PR #51, opened 2015-05-18
Add lint script, run in circle ci.

tomwilkie:


tomwilkie:
This is WIP; looks like static.go doesn't lint, so I deleted it, but now we need to get the UI building for this to be merged.

peterbourgon:
static.go will never lint; can we make an exception in the linter?

UI building we can do today with @davkal

tomwilkie:
I've added an exception, this is GTG

peterbourgon:
:dango:

Issue #26, opened 2015-05-12
Method for deduping reports


tomwilkie:
With #13, the possibility of an app connecting to a probe multiple times has been introduced.  This will allow an app to receive the same report more than once, and potentially double count.  We need to disallow this.

Options:
- Make report merge fully idempotent ie r1.merge(r1) == r1 and so on
- Make probes only accept one connection per app
- Include a random unique id in each report, and dedupe on the app

I like (1), as it has nice properties for the future; but it is also the hardest.

tomwilkie:
Short-term we'll only accept one connection per app, based on source ip.

peterbourgon:
Closed by #56

PR #56, opened 2015-05-19
One connection per app

peterbourgon:
Addresses #26.

### DIFF
--- a/xfer/publisher.go
+++ b/xfer/publisher.go
@@ -53,7 +53,7 @@ func (p *TCPPublisher) Publish(msg report.Report) {
 }
 
 func (p *TCPPublisher) loop(incoming <-chan net.Conn) {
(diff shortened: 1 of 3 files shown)


xfer/publisher.go
```diff
@@ -62,6 +62,27 @@ func (p *TCPPublisher) loop(incoming <-chan net.Conn) {
...
+					log.Printf("active connection: %s: %v (strange)", activeConn.RemoteAddr(), err)
+					continue
+				}
+				if host == activeHost {
+					log.Printf("duplicate connection: %s (dropped)", conn.RemoteAddr())
+					conn.Close()
+					continue outer
+				}
+			}
+
```
tomwilkie: This is a bit ugly; perhaps a new struct connection containing the encoder and the conn, and then key the above active map with the host from above?

tomwilkie:
LGTM

Issue #54, opened 2015-05-19
Make scope stop quicker


tomwilkie:
vagrant@ubuntu-14:~/src/github.com/weaveworks/scope$ time ./scope stop

    real     0m10.197s
    user     0m0.061s
    sys      0m0.076s

PR #59, opened 2015-05-19
Use runit inside the scope container

### LINKED ISSUE #32: Stop using supervisord in the container
Its appears to be the reason the container is so big.  There must be a more lightweight version.

tomwilkie:
Fixes #32 and #54


peterbourgon:
Lookin' real nice. :dango:

PR #62, opened 2015-05-19
A docker process mapper

### LINKED ISSUE #25: Docker process mapper
In the probe, extract metadata from the Docker runtime, and attach it to nodes as appropriate.

tomwilkie:
Fixes #25

### DIFF
--- a/probe/docker_process_mapper.go
+++ b/probe/docker_process_mapper.go
@@ -1,0 +1,128 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"sync"
+	"time"
+
+	docker "github.com/fsouza/go-dockerclient"
+)
+
+type dockerMapper struct {
+	sync.RWMutex
+	d        map[int]*docker.Container
+	procRoot string
+}
+
+func newDockerMapper(procRoot string, interval time.Duration) *dockerMapper {
+	m := dockerMapper{
+		procRoot: procRoot,
+		d:        map[int]*docker.Container{},
+	}
+	m.update()
+	go m.loop(interval)
+	return &m
+}
+
+func (m *dockerMapper) loop(d time.Duration) {
+	for range time.Tick(d) {
+		m.update()
+	}
+}
+
+// for mocking
+type dockerClient interface {
+	ListContainers(docker.ListContainersOptions) ([]docker.APIContainers, error)
+	InspectContainer(string) (*docker.Container, error)
+}
+
+func newRealDockerClient(endpoint string) (dockerClient, error) {
+	return docker.NewClient(endpoint)
+}
+
+var (
+	newDockerClient = newRealDockerClient
+	newPIDTreeStub  = newPIDTree
+)
+
+func (m *dockerMapper) update() {
+	pidTree, err := newPIDTreeStub(m.procRoot)
+	if err != nil {
+		log.Printf("docker mapper: %s", err)
+		return
+	}
+
+	endpoint := "unix:///var/run/docker.sock"
+	client, err := newDockerClient(endpoint)
+	if err != nil {
+		log.Printf("docker mapper: %s", err)
+		return
+	}
+
+	containers, err := client.ListContainers(docker.ListContainersOptions{All: true})
+	if err != nil {
+		log.Printf("docker mapper: %s", err)
+		return
+	}
+
+	pmap := map[int]*docker.Container{}
+	for _, container := range containers {
+		info, err := client.InspectContainer(container.ID)
+		if err != nil {
+			log.Printf("docker mapper: %s", err)
+			continue
+		}
+
+		if !info.State.Running {
+			continue
+		}
+
+		pids, err := pidTree.allChildren(info.State.Pid)
+		if err != nil {
+			log.Printf("docker mapper: %s", err)
+			continue
+		}
+		for _, pid := range pids {
+			pmap[pid] = info
+		}
+	}
+
+	m.Lock()
+	m.d = pmap
+	m.Unlock()
+}
+
+type dockerIDMapper struct {
+	*dockerMapper
+}
+
+func (m dockerIDMapper) Key() string { return "docker_id" }
+func (m dockerIDMapper) Map(pid uint) (string, error) {
+	m.RLock()
+	container, ok := m.d[int(pid)]
+	m.RUnlock()
+
+	if !ok {
+		return "", fmt.Errorf("no container found for PID %d", pid)
+	}
+
+	return container.ID, nil
+}
+
+type dockerNameMapper struct {
+	*dockerMapper
+}
+
+func (m dockerNameMapper) Key() string { return "docker_name" }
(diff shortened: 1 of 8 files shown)


tomwilkie:
This is WIP, but early feedback would be welcome.

probe/docker_process_mapper.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"fmt"
+	"log"
+	"sync"
+	"time"
+
+	docker "github.com/fsouza/go-dockerclient"
+)
+
+// dockerMapper is a cgroup task mapper.
```
tomwilkie: Oh no it isn't!

probe/main.go
```diff
@@ -28,6 +28,8 @@ func main() {
 		spyProcs           = flag.Bool("processes", true, "report processes (needs root)")
 		cgroupsRoot        = flag.String("cgroups.root", "", "if provided, enrich -processes with cgroup names from this root (e.g. /mnt/cgroups)")
 		cgroupsUpdate      = flag.Duration("cgroups.update", 10*time.Second, "how often to update cgroup names")
+		useDockerProcMap   = flag.Bool("docker", true, "collect docker-related attributes for processes.")
+		dockerUpdate       = flag.Duration("docker.update", 10*time.Second, "how often to update docker container info")
```
peterbourgon:
```go
dockerMapper   = flag.Bool("docker", true, "collect Docker-related attributes")
dockerInterval = flag.Duration("docker.interval", 10*time.Second, "Docker attribute poll interval")
```

probe/docker_process_mapper.go
```diff
@@ -0,0 +1,114 @@
...
+		return
+	}
+
+	containers, err := client.ListContainers(docker.ListContainersOptions{All: true})
+	if err != nil {
+		log.Printf("docker mapper: %s", err)
+		return
+	}
+
+	pmap := make(map[int]*docker.Container)
```
peterbourgon:
```go
pmap := map[int]*docker.Container{}
```

peterbourgon:
The architecture feels like the right way to do it. We should think if we shouldn't allow a single function invocation to annotate a PID with more-than-one piece of data, i.e. change `func Map(pid int) string` to `func Annotate(pid int) map[string]string`.

tomwilkie:
PTAL; they aren't exposed yet, but @peterbourgon is working on that.

probe/docker_process_mapper.go
```diff
@@ -0,0 +1,126 @@
...
+	ListContainers(docker.ListContainersOptions) ([]docker.APIContainers, error)
+	InspectContainer(string) (*docker.Container, error)
+}
+
+func newRealDockerClient(endpoint string) (dockerClient, error) {
+	return docker.NewClient(endpoint)
+}
+
+var newDockerClient = newRealDockerClient
+var newPidTreeStub = newPidTree
```
peterbourgon:
```go
var (
    newDockerClient = newRealDockerClient
    newPIDTree     = newRealPIDTree
)
```

probe/pidtree_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (f fileinfo) Name() string       { return f.name }
+func (f fileinfo) Size() int64        { return 0 }
+func (f fileinfo) Mode() os.FileMode  { return 0 }
+func (f fileinfo) ModTime() time.Time { return time.Now() }
+func (f fileinfo) IsDir() bool        { return true }
+func (f fileinfo) Sys() interface{}   { return nil }
+
+func TestPidTree(t *testing.T) {
+	oldReadDir, oldReadFile := ioutil.ReadDir, ioutil.ReadFile
```
peterbourgon:
```go
oldReadDir, oldReadFile := readDir, readFile
```

probe/docker_process_mapper_test.go
```diff
@@ -0,0 +1,81 @@
...
+				},
+			},
+		}, nil
+	}
+
+	dockerMapper := newDockerMapper("/proc", 10*time.Second)
+	dockerIDMapper := dockerIDMapper{dockerMapper}
+	dockerNameMapper := dockerNameMapper{dockerMapper}
+
+	fmt.Printf("%+v", dockerMapper)
```
peterbourgon: Never fmt.Print or println in tests; use t.Logf or (if it's for debugging the test) just remove it.

probe/docker_process_mapper_test.go
```diff
@@ -0,0 +1,81 @@
...
+	id, err = dockerIDMapper.Map(2)
+	if err != nil || id != "foo" {
+		t.Fatalf("err: %v, have '%s'", err, id)
+	}
+
+	name, err = dockerNameMapper.Map(2)
+	if err != nil || name != "bar" {
+		t.Fatalf("err: %v, have '%s'", err, name)
+	}
+}
```
peterbourgon: You can use a table-based test here.

```go
for pid, want := map[int]struct{id, name string}{
    1: {"foo", "bar"},
    2: {"foo", "bar"},
} {
    haveID, err := dockerIDMapper.Map(id)
    if err != nil || want.id != haveID {
        t.Errorf("%d: want %q, have %q (%v)", pid, want.id, haveID, err)
    }
    haveName, err := dockerNameMapper.Map(id)
    if err != nil || want.name != haveName {
        t.Errorf("%d: want %q, have %q (%v)", pid, want.name, haveName, err)
    }
}
```

probe/pidtree_test.go
```diff
@@ -0,0 +1,68 @@
...
+		if err != nil {
+			return nil, err
+		}
+		parent := pid - 1
+		return []byte(fmt.Sprintf("%d na R %d", pid, parent)), nil
+	}
+
+	pidtree, err := newPidTree("/proc")
+	if err != nil {
+		t.FailNow()
```
peterbourgon:
```go
t.Fatal(err)
```

probe/pidtree_test.go
```diff
@@ -0,0 +1,68 @@
...
+		fmt.Printf("wrong children: %v", children)
+		t.FailNow()
+	}
+
+	children, err = pidtree.allChildren(2)
+	if err != nil || !reflect.DeepEqual(children, []int{2, 3, 4}) {
+		fmt.Printf("wrong children: %v", children)
+		t.FailNow()
+	}
+}
```
peterbourgon: You can use a table-based test here.

```go
for pid, want := range map[int][]int{
    1: []int{1,2,3,4},
    2: []int{2,3,4},
} {
    have, err := pidtree.allChildren(pid)
    if err != nil || !reflect.DeepEqual(want, have) {
        t.Errorf("%d: want %#v, have %#v", pid, want, have)
    }
}
```

peterbourgon:
...and, in general, s/Pid/PID/g. Otherwise :dango:

probe/docker_process_mapper_test.go
tomwilkie: Whoops, that was old code that shouldn't have been committed.

PR #65, opened 2015-05-20
Fix grouping bar for topologies that don't support grouping

davkal:


peterbourgon:
Looks like a regression, when I click on nodes to get the detail pane, a request is made for `/<id>`, it should be `/api/topology/<topology>/<id>`.

peterbourgon:
Same regression exists for the grouped view of the Applications topology.

tomwilkie:
This LGTM

Issue #21, opened 2015-05-12
Populate real data in details struct


peterbourgon:
[Relevant](https://github.com/weaveworks/scope/blob/440b155cc529644dd47ed48dd007db90efb40e90/scope/app/api_topology.go#L104-132).

peterbourgon:
:ship:

PR #68, opened 2015-05-20
Enhance the details pane

peterbourgon:
This PR will enhance the details struct (and thereby the details pane) with details from the origin atom (process, IP), specifically stuff like cgroup names, Docker container info, etc.


peterbourgon:
Addresses #21

peterbourgon:
screen shot 2015-05-21 at 19 03 44

peterbourgon:
Steps to get a piece of data into the detail pane:

1. In the probe, add the data to the NodeMetadata, either [directly](https://sourcegraph.com/github.com/weaveworks/scope@c193d69b19918fcbb4540449627c0e3747b40dfc/.tree/probe/spy.go#startline=82&endline=82) or [with a ProcessMapper](https://sourcegraph.com/github.com/weaveworks/scope@c193d69b19918fcbb4540449627c0e3747b40dfc/.tree/probe/docker_process_mapper.go#startline=100&endline=111). Note which topology it's being added to; probably it will be the Process (Applications) topology.
2. In the app, [map the data to a table row](https://sourcegraph.com/github.com/weaveworks/scope@a1e8143f9b07b795c025ac233b55eebacaf3bde5/.tree/app/origin_node.go#startline=36&endline=38).

peterbourgon:
@tomwilkie ready for review, I think. This isn't a beautiful implementation but it'll do.

app/origin_node.go
```diff
@@ -0,0 +1,65 @@
...
+	return originNodeForNetworkHost(node), true
+}
+
+func originNodeForProcess(node report.NodeMetadata) OriginNode {
+	rows := []report.Row{
+		{Key: "Host", ValueMajor: node["domain"], ValueMinor: ""},
+		{Key: "PID", ValueMajor: node["pid"], ValueMinor: ""},
+		{Key: "Process name", ValueMajor: node["name"], ValueMinor: ""},
+	}
+	if val, ok := node["docker_id"]; ok {
```
tomwilkie: Would this be better as a table?  ie

```
for key, human := range map[string]string{
    "docker_id":  "Container ID",
    "docker_name": "Container Name",
    "cgroup": "cgroup",
} {
    if val, ok := node[key]; ok {
        rows = append(rows, report.Row{Key: human, ValueMajor: val, ValueMinor: ""})
    }
}
```

app/static.go
```diff
@@ -11,15 +11,15 @@ import (
 	"time"
```
tomwilkie: I'd leave this file out of this change.  Just push this separately without a review.

circle.yml
```diff
@@ -26,6 +26,8 @@ dependencies:
   post:
     - go clean -i net
     - go install -tags netgo std
+    - go get github.com/davecgh/go-spew/spew
+    - go get github.com/pmezard/go-difflib/difflib
```
tomwilkie: If PR #85 gets in first, these have moved to the Makefile.

tomwilkie:
Other than that one comment, LGTM.  Will definitely need a rebase and squash - we don't want merge commits, gofmt fixes, circle debugging etc.  I would squash into one.

app/origin_node.go
peterbourgon: Can do, but is it really better? I don't know. I kind of like it as-is, declaratively. But I'm a 2/10 on it, so if you feel more strongly, just say so and I change it.

tomwilkie: I think I'd prefer to push the human names into the node map and ultimately have the come from the process mappers.  But as is its slightly easier to follow (and extend) as a loop like this, I think.  I'm thinking we're going to be adding a lot more in here (docker image id and name, for instance).

circle.yml
tomwilkie: Actually do these need to be go gotten manually?  Wouldn't it be better to update the go get cmd in the makefile to fetch test dependancies, or was that not possible?

tomwilkie:
Problem with the table approach I suggested; ordering is not stable.

peterbourgon:
D'oh. Will fix.

Issue #74, opened 2015-05-21
Image on the Docker hub doesn't run


peterbourgon:
```
./run: exec: line 3: /home/weave/probe: not found
./run: exec: line 3: /home/weave/app: not found
./run: exec: line 3: /home/weave/probe: not found
./run: exec: line 3: /home/weave/app: not found
. . .
```

tomwilkie:
```
/home/weave # ldd probe
	/lib64/ld-linux-x86-64.so.2 (0x7f992268d000)
	libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7f992268d000)
	libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f992268d000)
```

tomwilkie:
Can confirm this is fixed.

peterbourgon:
Building from latest master, I get the same problems... [see gist](https://gist.github.com/peterbourgon/7d7c7262688144e0f719)

peterbourgon:
But I can confirm the image on Docker Hub is working. :(

tomwilkie:
I guess this is because you don't have docker squash - do your build docker image is just gibberish.

PR #78, opened 2015-05-21
Make the docker image built on circle runnable

tomwilkie:
- Use netgo tag to rebuild go std lib so resulting binaries are statically compiled.
- Couple of other tweaks to make it work.


tomwilkie:
Fixes #74

peterbourgon:
Oof. Well, if it works! :shipit:

PR #89, opened 2015-05-22
Embed the git commit id in a go file and print it on startup.

### LINKED ISSUE #88: Embed changeset id and log it on startup

tomwilkie:
Fixes #88

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -18,6 +18,9 @@ import (
 	"github.com/weaveworks/scope/xfer"
 )
 
+// Set during buildtime.
+var version = "unknown"
+
 func main() {
 	var (
 		defaultProbes = []string{fmt.Sprintf("localhost:%d", xfer.ProbePort), fmt.Sprintf("scope.weave.local:%d", xfer.ProbePort)}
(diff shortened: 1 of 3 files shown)


peterbourgon:
Oof. That's awkward. Let's have a [version variable](https://github.com/soundcloud/glimpse/blob/master/agent/main.go#L28) that's [set at build-time](https://github.com/soundcloud/glimpse/blob/master/agent/Makefile#L20).

tomwilkie:
Yeah thats much cleaner.  PTAL

app/main.go
```diff
@@ -63,7 +66,7 @@ func main() {
 		defer os.Remove(*pidfile)
 	}
 
-	log.Printf("starting")
+	log.Printf("app starting, commit id %s", version)
```
peterbourgon:
```go
log.Printf("starting, version %s", version)
```

probe/main.go
```diff
@@ -39,6 +42,8 @@ func main() {
 		os.Exit(1)
 	}
 
+	log.Printf("probe starting, commit id %s", version)
```
peterbourgon:
```go
log.Printf("starting, version %s", version)
```

peterbourgon:
Minor formatting nit aside, :dango:

app/main.go
tomwilkie: I put app and probe in their as their logs get mixed up when run under runit

PR #97, opened 2015-05-22
Don't set the hostname in the scope container when running with weavedns

### LINKED ISSUE #75: Hostname in scope container is "scope"

tomwilkie:
Fixes #75


peterbourgon:
:dango:

Issue #90, opened 2015-05-22
Wrapping of container id in details panel


tomwilkie:

tomwilkie:
![wrapping](https://cloud.githubusercontent.com/assets/444037/7770255/5d08836a-0084-11e5-991c-97df04bda6fe.png)

peterbourgon:
Can we just truncate that to the first 7 characters, in the short term?

peterbourgon:
And why is one Origin Process showing rows in the order (Container ID, Container name), where the other has the order (Container name, Container ID)? Origin Process tables should be deterministic and stable, in fact we rely on that to deduplicate. Is it from an earlier version where you maybe had implemented the rows as maps?

tomwilkie:
> Can we just truncate that to the first 7 characters, in the short term?

The fix has the browser do the truncate for you.

> And why is one Origin Process showing rows in the order...

The screenshot is from before you fixed that.

peterbourgon:
> The fix has the browser do the truncate for you.

:100:

> The screenshot is from before you fixed that.

Whew :relieved:

PR #98, opened 2015-05-24
Fix wrapping and scrolling in the details pane.

### LINKED ISSUE #80: Details pane doesn't grow or scroll when there's a lot of information in it
If there are multiple tables in the details pane, it exhibits `overflow: visible;` behavior. I guess we want `overflow: scroll;`.

tomwilkie:
Fixes #80 and #90


davkal:
LGTM

PR #106, opened 2015-05-26
Add Containers topology

### LINKED ISSUE #91: Add "container" topology
Group together processes in the same container.

peterbourgon:
- In the default view, nodes from the same container instance (docker ID) are merged
- In the grouped view, nodes from the same container image (docker image ID) are merged

Closes #91.

/cc @tomwilkie

### DIFF
--- a/report/mapping_functions.go
+++ b/report/mapping_functions.go
@@ -48,6 +48,37 @@ func ProcessPID(_ string, m NodeMetadata, grouped bool) (MappedNode, bool) {
 	}, show
 }
 
+// ProcessContainer maps Process topology nodes to the containers they run in.
+// We consider container and image IDs to be globally unique, and so don't
+// scope them further by e.g. host. If no container metadata is found, nodes
+// are grouped into the Uncontained node. If grouped is true, nodes with the
+// same container image ID are merged together.
+func ProcessContainer(_ string, m NodeMetadata, grouped bool) (MappedNode, bool) {
+	var (
+		containerID   = m["docker_id"]
+		containerName = m["docker_name"]
+		imageID       = m["docker_image_id"]
+		imageName     = m["docker_image_name"]
+		domain        = m["domain"]
+	)
+
+	var id, major, minor, rank string
+	if containerID == "" {
+		id, major, minor, rank = "uncontained", "Uncontained", "", "uncontained"
+	} else if grouped {
+		id, major, minor, rank = imageID, imageName, "", imageID
+	} else {
+		id, major, minor, rank = containerID, containerName, domain, imageID
+	}
+
+	return MappedNode{
+		ID:    id,
+		Major: major,
+		Minor: minor,
+		Rank:  rank,
+	}, true
+}
+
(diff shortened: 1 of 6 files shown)


report/mapping_functions.go
```diff
@@ -48,6 +48,37 @@ func ProcessPID(_ string, m NodeMetadata, grouped bool) (MappedNode, bool) {
 	}, show
 }
 
+// ProcessContainer maps Process topology nodes to the containers they run in.
+// We consider container and image IDs and names to be globally-scoped, and so
+// don't scope them further by e.g. host. If no container metadata is found,
+// nodes are grouped into the Uncontained node. If grouped is true, nodes are
+// grouped by container image.
```
tomwilkie: IDs are globally scoped; so are image names.  Container names are not; ie two processes on different hosts but in containers of the same name shouldn't be merged.  I don't think your code below does that though, so just need to update the comment.

tomwilkie:
Can you rebase and squash this into one please?  We don't need to introduce extra merge commits, or commits like "Works on my machine".  Also need to resolve why my two csets are appearing on your branch; probably go away if you rebase.

tomwilkie:
Ah I see you already got rid of the merge.

report/mapping_functions.go
peterbourgon: Ah OK, so — container names are only used in labels, never as IDs, so indeed we just need to make sure they're not ambiguous there. Will update...

tomwilkie: Yeah, its not quite that easy; container names are unique on a single machine, but not across a cluster.  The IDs should be unique (in that they are large and random), but there is no guarantee.  Images names and ids are enforced by the hub, so will be pretty unique, although you can make a local image with an arbitrary name, so I would only trust the IDs, which are content-addressable style hashes.

peterbourgon: Understood.

report/mapping_functions.go
```diff
@@ -48,6 +48,37 @@ func ProcessPID(_ string, m NodeMetadata, grouped bool) (MappedNode, bool) {
...
+		containerID   = m["docker_id"]
+		containerName = m["docker_name"]
+		imageID       = m["docker_image_id"]
+		imageName     = m["docker_image_name"]
+		domain        = m["domain"]
+	)
+
+	var id, major, minor, rank string
+	if containerID == "" {
+		id, major, minor, rank = "uncontained", "Uncontained", "", "uncontained"
```
tomwilkie: So you went with uncontained....  

I think this will potentially end up a lot like the pseudo nodes if we're not careful, so it might be best to just drop them at this stage?

peterbourgon: I'd really like to try it this way, keeping the data present in the view, and see how it feels. If we want to drop them, we can do it easily: just return the last parameter as false in this case.

tomwilkie: Okay we can try it this way for now.

tomwilkie:
LGTM

Issue #113, opened 2015-05-27
Pseudo nodes in containers view should be "uncontained", surely?


tomwilkie:
Its weird for the containers view to have three types of nodes - containers, everything not in a container, and hosts.

I think pseudo nodes in containers should either be hidden of uncontained.  I'm guessing the whole pseudo node thing is going to have to become a little more sophisticated - maybe a per-topology pseudo node factory (ie callback)?

peterbourgon:
We can definitely extend the [topology registry](https://github.com/weaveworks/scope/blob/8a735b4d8de8a94d4c67ea15a856800071867ca8/app/router.go#L43-52) to parameterize the treatment of pseudonodes. But I don't quite follow your goals. Can you explain a bit further what you want to accomplish?

tomwilkie:
(This is a bit of a thought-in-progress) I think psuedo nodes should be the same "kind" as the topology we're injecting them into,

- process topology - pseudo nodes should be our best approximation of processes (see #99)
- container topology - pseudo nodes should be hidden, or grouped into "uncontained"
- host topology - pseudo nodes should be hosts (we can do this very well)
- process name (process grouped) - maybe hidden? definitely not all grouped into one.
- container image (container grouped) - hidden; no way to know what image they're running

What do you think?

tomwilkie:
This is basically fixed.

PR #116, opened 2015-05-27
Group pseudo nodes by the node the referred to and their address in grouped topologies.

### LINKED ISSUE #105: All pseudo nodes are grouped together in the "grouped" view
This seems undesirable, implying relationships between unrelated services.

Whats more, they don't have a label in the UI.

Options to fix this include:
- just group by IP
- apply same grouping logic as in #99
- don't show them

tomwilkie:
Fixes #105


peterbourgon:
I don't think it's OK to have this behavior be default. Again, I _want_ to see pseudonodes in the general case. The case where Chrome is opening a bunch of short-lived connections is an outlier.

report/topology_test.go
```diff
@@ -231,12 +221,15 @@ func TestRenderByProcessPIDGrouped(t *testing.T) {
...
-			Adjacency:   NewIDList("curl", "localUnknown"),
+			ID:         "apache",
+			LabelMajor: "apache",
+			LabelMinor: "",
+			Rank:       "215",
+			Pseudo:     false,
+			Adjacency: NewIDList(
+				"curl",
+				"pseudo:;10.10.10.10;apache",
+				"pseudo:;10.10.10.11;apache"),
```
peterbourgon: We can leverage Go's tolerance of trailing commas to put the closing parens on a newline, making this much nicer.

```go
Adjacency: NewIDList(
	"curl",
	"pseudo:;10.10.10.10;apache",
	"pseudo:;10.10.10.11;apache",
),
```

tomwilkie:
This change is for grouped topologies, and stops all pseudo nodes being merged into one.  So it adds more pseudo nodes....

tomwilkie:
And yes, in general I'm factoring out the pseudo node handling to be per-topology, so different topologies can have different concepts for pseudo nodes.  So for the pid topology, we can show more, for containers, less, and hosts is pretty obvious.

peterbourgon:
Great! I noticed the formatLabel thing is basically dead code at this point, I'll remove that in another PR. :+1:

tomwilkie:
PTAL; I've factored out the pseudo node handling (hopefully with no semantic change)

report/mapping_functions.go
```diff
@@ -25,6 +26,17 @@ type MappedNode struct {
...
+// PseudoFunc is anything which can take an arbitrary node id and produce
+// a pseudo node for the given id, in the form of a node ID and a
+// human-readable major and minor labels.
+//
+// First input param is the src id (before mapping), second is the src rendered node,
+// third is the dst id, fourth is grouped.
+//
+// If the final output parameter is false, the node shall be omitted from the
+// rendered topology.
+type PseudoFunc func(string, RenderableNode, string, bool) (MappedNode, bool)
```
peterbourgon: How about this?

```go
// PseudoFunc maps arbitrary concrete nodes to renderable pseudonodes.
type PseudoFunc func(srcNodeID string, rn RenderableNode, dstNodeID string, grouped bool) (mn MappedNode, show bool)
```

report/topology.go
```diff
@@ -70,7 +69,7 @@ func NewTopology() Topology {
 //
 // RenderBy takes a a MapFunc, which defines how to group and label nodes. If
 // grouped is true, nodes that belong to the same "class" will be merged.
-func (t Topology) RenderBy(f MapFunc, grouped bool) map[string]RenderableNode {
+func (t Topology) RenderBy(f MapFunc, pseudo PseudoFunc, grouped bool) map[string]RenderableNode {
```
peterbourgon: For consistency...

```go
func (t Topology) RenderBy(map MapFunc, pseudo PseudoFunc, grouped bool) map[string]RenderableNode {
```

peterbourgon:
Yes, this is a brilliant change. :100:

report/topology.go
tomwilkie: map is not allowed as a variable name

report/mapping_functions.go
tomwilkie: It doesn't strictly map the renderable nodes to a pseudo node, but a node id to pseudo node (as there is no renderable node by definition).  I've made the role of srcNode more obvious.

tomwilkie:
Okay one more look pls.

tomwilkie:
Also fixes #113

report/mapping_functions.go
```diff
@@ -25,6 +25,12 @@ type MappedNode struct {
 // rendered topology.
 type MapFunc func(string, NodeMetadata, bool) (MappedNode, bool)
 
+// PseudoFunc creates MappedNode representing pseudo nodes given the dstNodeID.
+// The srcNode renderable nodes is essentially from MapFunc, representing one of
```
peterbourgon: "The srcNode renderable **node** is essentially from MapFunc..."

peterbourgon:
:+1:

PR #117, opened 2015-05-27
Remove formatLabel

peterbourgon:
Function has slowly become dead code; remove it.


tomwilkie:
I've consumed this into #116 as it was going to conflict.

PR #122, opened 2015-05-27
Don't do the hostproc dance, use --pid=host

### LINKED ISSUE #112: Invesitgate and use --pid=host
Turns out you can use the host's process namespace apparently.  This could mean no more /hostproc, and tracer might even run in a container.

tomwilkie:
Fixes #112


peterbourgon:
If we can rely on Docker 1.5+, LGTM.

tomwilkie:
Yeah we can, for now at very least.

Issue #120, opened 2015-05-27
Allow probes to attach metadata directly to entities


peterbourgon:
Currently the probes produce reports that look like

```go
type Report struct {
    Process Topology
    Network Topology
    HostMetadatas
}
```

We will change them to look like

```go
type Report struct {
    Endpoint  Topology // former Process
    Process   Topology // only nodes, no edges
    Container Topology // only nodes, no edges
    Address   Topology // former Network
    Host      Topology // former HostMetadatas; only nodes, no edges
}
```

Probes will attach metadata to the actual entities they scrape, rather than having to explode the metadata into the topologies that happen to be present. This will also make merging a little easier, as we're only dealing with a single type. (via #115)

As part of this work, we'll also add more process and container metadata. (via #114)

tomwilkie:
Could you elaborate on the key formats inside these topologies please?  

And I guess endpoints will need enough metadata on the nodes to link to processes and containers.

Whats in Address thats not in endpoint?  Ie could you just put host names as metadata on nodes in endpoint, and then reduce that to get Address?

peterbourgon:
> Could you elaborate on the key formats inside these topologies please?

Sure, I'll do this properly in the PR. I need to revisit it to get it straight in my head, to be honest :)

> And I guess endpoints will need enough metadata on the nodes to link to processes and containers.

Yes, we'll have foreign key relationships across topologies where we can.

> Whats in Address thats not in endpoint? 

Address is an IP address i.e. L3. Endpoint is the complete TCP n-tuple i.e. L4. Crucially, there is information we can get and associate with lower levels that won't translate to higher levels. Think ICMP: present in Address, invisible in Endpoint.

> Ie could you just put host names as metadata on nodes in endpoint, and then reduce that to get Address?

Yes, but the point of each topology is that it represents a different source of information, which doesn't map perfectly across OSI layer boundaries.

tomwilkie:
Yeah that makes sense. Am I right in thinking that address current has the
same info as endpoint (minus the port) ie they use the same source?

peterbourgon:
> Am I right in thinking that address current has the same info as endpoint (minus the port) ie they use the same source?

Yes, as an implementation detail, both Network and Process topologies [are currently populated](https://github.com/weaveworks/scope/blob/2743013c885f9b42e06b24cb3a5c512b36afee06/probe/spy.go#L29) by procspy.Connections. I want to change that ASAP :)

tomwilkie:
[...]

Issue #114, opened 2015-05-27
Pull data from docker stats


tomwilkie:
Its seems to offer memory and cpu on a per-container basis.

Perhaps pulling memory and cpu from proc on a per-process basis would be better?

peterbourgon:
We don't have a way to associate metadata to processes *or* containers at this stage, as they're both emergent entities from the underlying topologies. See my comments on #115. If we did what I proposed there, we could solve this issue, too. Shall I merge them and create a new issue?

peterbourgon:
Merged to #120

peterbourgon:
Re-opening as a supplementary ticket after for #123.

peterbourgon:
This is merged and done.

PR #123, opened 2015-05-27
Refactor Report for better per-entity metadata

peterbourgon:
Addresses #120.

- [x] Refactor Report and Topology structures
- [x] Make various ID generation consistent and understandable
- [x] Endpoint topology (was Process)
- [x] Address topology (was Network)
- [x] Process topology
- [x] Host topology (was OriginHost struct)

Follow-up tickets #114, #171, #172


peterbourgon:
@tomwilkie it's very possible I've forgotten some stuff in this PR, because I'm doing about six simultaneous things right now :) but I think this is good to go. Please give it a read, I suggest starting with **package report**. And also run it as a test, to make sure I didn't overlook any behaviors.

probe/tag/docker_tagger.go
```diff
@@ -0,0 +1,172 @@
+package tag
```
tomwilkie: You've lost all the docker event-driver stuff.

tomwilkie:
I've had a ~quick look (still took a few hours).  This change is un-reviewable.  At 7.6k lines, it would take me weeks to go through this code throughly.  It has to be broken down into smaller changes; I'm not going to rubber stamp such a big change.

Given that you're busy, I'll have a stab a teasing out smaller changes we can upstream quickly.  But I would fully expect this process to take the next week or so.

peterbourgon:
I can't really break it down. It's not a diff, it's a complete refactoring. Look at it on a per-package basis, not as a change but as a complete entity.

probe/tag/docker_tagger.go
peterbourgon: Right, because it was merged today. I'll allocate some time to bring it back tomorrow, if nothing more important comes up.

peterbourgon:
We're past the point of this being salvageable. Closing it out.

Issue #44, opened 2015-05-18
Minimize JS bundle


peterbourgon:
Using webpack maybe?

davkal:
Fixed via #132

PR #132, opened 2015-05-29
changed tooling from browserify to webpack

davkal:
* using browsersync instead of livereload
* `gulp serve` works as before
* `gulp sync` starts browserSync and serves UI on port 3000
* everything is built in `build/`
* build release with `gulp build --release`


davkal:
Fixes #44

tomwilkie:
LGTM

Issue #137, opened 2015-05-31
`basename: missing operand` when launching scope when weavedns is running


rade:
````
$ weave launch
476e00691ca1a5d9423f03377f0b18527478aff716e1ff867c70cec6724a8c17
$ weave launch-dns 10.2.0.1/24
e572552f539cd417d10e7c656b4854d65a752921498b27a733e83c65bcdff81a
$ ./scope launch
basename: missing operand
Try 'basename --help' for more information.
73f80066456faeb694ff119fa1de4dd3604f7b731d3fd8bb65a05e0a46cb96a1
````

This is with version 2e8df4a8e2521c9005a66dd93d0fd514b8ae6e4a, built according to the [instructions](https://github.com/weaveworks/scope#build).

tracing the exeuction with `sh -x` produces
````
...
+ dns_running
+ docker inspect --format={{.State.Running}} weavedns
+ status=true
+ [ true = true ]
+ return 0
+ + xargs -n1 basename
find /sys/class/net -name eth*
+ grep inet
+ + xargsgrep echo
 -oE [0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}
+ + awkxargs { print $2 } -n1 ip
 addrbasename: missing operand
Try 'basename --help' for more information.
 show
...
````

Looking at this and the code, I think the issue is that the script doesn't cope with my machine not having an `eth*` interface.

**Comment by peterbourgon** [comment]
Via my own educational archaeology, the [offending block](https://github.com/weaveworks/scope/blob/2e8df4a8e2521c9005a66dd93d0fd514b8ae6e4a/scope#L135-139)

```bash
		if dns_running; then
			IP_ADDRS=$(find /sys/class/net -name eth* | xargs -n1 basename | xargs -n1 ip addr show |
				grep inet | awk '{ print $2 }' | grep -oE "$IP_REGEXP" | xargs echo)
			tell_dns_fqdn PUT $CONTAINER $FQDN $IP_ADDRS
		fi
```

Compare against the [weave script's invocation](https://github.com/weaveworks/weave/blob/b86ead27374f2a421d1e4474266c8d68253acb21/weave#L977)

```bash
    dns-add)
        collect_ip_args "$@"
        shift $IP_COUNT
        [ $# -ge 1 ] || usage
        CONTAINER_ID="$1"
        if [ $# -eq 1 ] ; then
            tell_dns PUT $CONTAINER_ID $IP_ARGS
        else
[...]

PR #140, opened 2015-06-01
Don't just advertise addresses on eth devices; use blacklist on device name.

tomwilkie:
Helps towards fixing #137


tomwilkie:
Matthias asks why I don't just use hostname -I

rade:
That particular line of code looks-better-than-it-was-before.

PR #147, opened 2015-06-01
Make containers the default topology

### LINKED ISSUE #130: "Containers" should be the default view when you first navigate to the UI
We are all about containers, focus, being the right thing for containers etc.

tomwilkie:
Fixes #130


davkal:
LGTM

PR #152, opened 2015-06-02
Emit the urls at which the weave scope UI is reachable on scope launch.

### LINKED ISSUE #149: scope launch should at least try to print the url for the scope UI
This will be made harder by machines with multiple IPs.

tomwilkie:
Fixes #149 

The text is outputted on stderr, so $(scope launch) just gives you the uuid of the container.

### DIFF
--- a/scope
+++ b/scope
@@ -153,6 +153,11 @@ case "$COMMAND" in
 		fi
 
 		echo $CONTAINER
+
+		echo "Weave Scope is reachable at the following URL(s):" >&2
+		for ip in $(hostname -I); do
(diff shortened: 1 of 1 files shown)


scope
```diff
@@ -153,6 +153,11 @@ case "$COMMAND" in
 		fi
 
 		echo $CONTAINER
+
+		echo "Weave Scope is reachable at the following url(s):" >&2
```
peterbourgon: URL(s):

peterbourgon:
:+1:

PR #163, opened 2015-06-02
Use Docker events API instead of polling Docker daemon.

### LINKED ISSUE #84: Consider using docker events API instead of polling in docker process mapper.

tomwilkie:
Fixes #84

### DIFF
--- a/probe/docker_process_mapper.go
+++ b/probe/docker_process_mapper.go
@@ -10,25 +10,48 @@ import (
 	docker "github.com/fsouza/go-dockerclient"
 )
 
+const (
+	stop  = "stop"
+	start = "start"
+)
+
 type dockerMapper struct {
 	sync.RWMutex
-	containers map[int]*docker.Container
-	images     map[string]*docker.APIImages
-	procRoot   string
+
+	containers      map[string]*docker.Container
+	containersByPID map[int]*docker.Container
+	images          map[string]*docker.APIImages
+
+	procRoot string
+	pidTree  *pidTree
+
+	interval time.Duration
 }
 
-func newDockerMapper(procRoot string, interval time.Duration) *dockerMapper {
-	m := dockerMapper{
-		procRoot:   procRoot,
-		containers: map[int]*docker.Container{},
+func newDockerMapper(procRoot string, interval time.Duration) (*dockerMapper, error) {
+	pidTree, err := newPIDTreeStub(procRoot)
+	if err != nil {
+		return nil, err
 	}
-	m.update()
-	go m.loop(interval)
-	return &m
+
+	m := dockerMapper{
+		containers:      map[string]*docker.Container{},
+		containersByPID: map[int]*docker.Container{},
+		images:          map[string]*docker.APIImages{},
+
+		procRoot: procRoot,
+		pidTree:  pidTree,
+
+		interval: interval,
+	}
+
+	go m.loop()
+	return &m, nil
 }
 
-func (m *dockerMapper) loop(d time.Duration) {
-	for range time.Tick(d) {
+func (m *dockerMapper) loop() {
+	m.update()
+	for range time.Tick(m.interval) {
 		m.update()
 	}
 }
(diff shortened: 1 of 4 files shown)


probe/docker_process_mapper.go
```diff
@@ -50,63 +70,145 @@ var (
...
+		case <-otherUpdates:
+			if err := m.updatePIDTree(); err != nil {
+				log.Printf("docker mapper: %s", err)
+				return
+			}
+
+			if err := m.updateImages(client); err != nil {
+				log.Printf("docker mapper: %s", err)
+				return
+			}
```
peterbourgon: Do you really want to return in these error cases? Could you continue instead?

probe/docker_process_mapper.go
```diff
@@ -50,63 +70,145 @@ var (
...
-	pmap := map[int]*docker.Container{}
-	for _, container := range containers {
-		info, err := client.InspectContainer(container.ID)
+	if err := m.updateImages(client); err != nil {
+		log.Printf("docker mapper: %s", err)
+		return
+	}
+
+	otherUpdates := time.Tick(m.interval)
+	for true {
```
peterbourgon: `true` is unnecessary. Also, should this thing have a Stop mechanism?

probe/docker_process_mapper.go
```diff
@@ -50,63 +70,145 @@ var (
...
+		m.images[image.ID] = image
+	}
+	m.Unlock()
+
+	return nil
+}
+
+func (m *dockerMapper) handleEvent(event *docker.APIEvents, client dockerClient) {
+	switch event.Status {
+	case "stop":
```
peterbourgon: Can we use a constant here instead of a string literal?

probe/docker_process_mapper.go
```diff
@@ -50,63 +70,145 @@ var (
...
+		containerID := event.ID
+		m.Lock()
+		container, ok := m.containers[containerID]
+		if ok {
+			delete(m.containers, containerID)
+			delete(m.containersByPID, container.State.Pid)
+		} else {
+			log.Printf("docker mapper: container %s not found", containerID)
+		}
```
peterbourgon:
```go
if container, ok := m.containers[event.ID]; ok {
    delete(m.containers, event.ID)
    delete(m.containersByPID, container.State.Pid)
} else {
    log.Printf("...")
}
```

probe/docker_process_mapper_test.go
```diff
@@ -59,12 +67,14 @@ func TestDockerProcessMapper(t *testing.T) {
...
 	}
 
-	dockerMapper := newDockerMapper("/proc", 10*time.Second)
+	dockerMapper, _ := newDockerMapper("/proc", 10*time.Second)
 	dockerIDMapper := dockerMapper.idMapper()
 	dockerNameMapper := dockerMapper.nameMapper()
 	dockerImageIDMapper := dockerMapper.imageIDMapper()
 	dockerImageNameMapper := dockerMapper.imageNameMapper()
 
+	<-time.After(time.Millisecond)
```
peterbourgon:
```go
time.Sleep(time.Millisecond)
```

Or consider using

```go
runtime.Gosched()
```

peterbourgon:
>  coverage/coveralls — Coverage decreased (-1.79%) to 62.24%

probe/docker_process_mapper.go
tomwilkie: Yeah its probably not important, the whole thing is in a big while loop anyway.

probe/docker_process_mapper.go
tomwilkie: Yes we discussed this a while back, but I forgot to file a ticket.  This PR doesn't make the situation worse, so do you mind if we defer a Stop mechanism to another day?

tomwilkie: #166

probe/docker_process_mapper.go
tomwilkie: Done.

tomwilkie:
Thanks for feedback!  PTAL.  I'm not to fussed about small drops in coverage like that; majority is error handling.

probe/docker_process_mapper.go
peterbourgon: Sounds fine to me.

peterbourgon:
:+1:

PR #169, opened 2015-06-03
Allow users to specify app and probe arguments.

### LINKED ISSUE #153: Allow user to override app and probe default port

tomwilkie:
Fix #153

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,12 +1,16 @@
 #!/bin/sh
 
 usage() {
-	echo "$0 --dns <IP> --hostname <NAME> --searchpath <SEARCHPATH>"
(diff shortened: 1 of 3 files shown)


docker/entrypoint.sh
```diff
@@ -19,12 +23,30 @@ while true; do
...
+            ;;
         *)
             break
             ;;
     esac
 done
 
+mkdir -p /etc/weave
+echo "$APP_ARGS" >/etc/weave/app.conf
+echo "$PROBE_ARGS" >/etc/weave/probe.conf
```
peterbourgon: Can we use .args instead of .conf?

peterbourgon:
:+1:

PR #174, opened 2015-06-03
Refactor out the docker process mapper for use in tracer.

tomwilkie:


peterbourgon:
This will conflict annoyingly with #123 :\

tomwilkie:
Okay let me handle that; we'll get #123 in first.

We have to avoid doing such large PRs in the future.

tomwilkie:
This depends on #178

tomwilkie:
Now #178 is in, this change is much smaller.  @peterbourgon mind taking another look?

peterbourgon:
LGTM. We should start testing with the race detector.

PR #175, opened 2015-06-04
Rename Resolver to staticResolver

tomwilkie:
This is the first change teased from #123

### DIFF
--- a/app/resolver.go
+++ b/app/resolver.go
@@ -15,9 +15,7 @@ var (
 	lookupIP = net.LookupIP
 )
 
-// Resolver periodically tries to resolve the IP addresses for a given
-// set of hostnames.
-type Resolver struct {
+type staticResolver struct {
 	quit  chan struct{}
 	add   func(string)
 	peers []peer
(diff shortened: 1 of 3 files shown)


app/resolver.go
```diff
@@ -15,9 +15,11 @@ var (
 	lookupIP = net.LookupIP
 )
 
-// Resolver periodically tries to resolve the IP addresses for a given
-// set of hostnames.
-type Resolver struct {
+type resolver interface {
```
tomwilkie: What purpose does this interface serve?

peterbourgon: It maintains consistency with other, similar components e.g. the reporter, and it bounds the surface area of the component at its callsite. More generally, it's best if constructors return interface types rather than pointers to implementations, as it forces you to think about APIs.

tomwilkie: I'm all for abstraction and consistency, but I don't think this interface adds anything.  Its not as if the abstraction is used elsewhere to simplify something.  Rules like "constructors return interfaces" seem too general and remind me of all the horrors of Java.

peterbourgon:
LGTM

PR #176, opened 2015-06-04
Add interface to xfer/collector.go, and some more tests

tomwilkie:


tomwilkie:
This is another change broken out of #123

rade:
This does more than the description suggests. In particular note the type change for the `quit` field, and the introduction of `peek`.

peterbourgon:
- quit field change is to make it consistent with other Stoppable components in Scope
- peek is used only for tests, we could add a comment to that effect, but since it's not exported and only called in 1 place I believe it's pretty straightforward

xfer/collector.go
```diff
@@ -17,51 +17,61 @@ const (
 
 var (
 	// MaxBackoff is the maximum time between connect retries.
-	MaxBackoff = 2 * time.Minute // externally configurable.
+	// It's exported so it's externally configurable.
+	MaxBackoff = 2 * time.Minute
+
+	tick = time.Tick
```
rade: why have this? It's used exactly once, and makes the call site just that little bit more obscure.

peterbourgon:
LGTM

xfer/collector.go
tomwilkie: tick is extracted out for mocking in the tests.

rade: ah! comment, perhaps?

tomwilkie:
I've updated to account for comments.  Will merge when tests pass.

xfer/collector.go
tomwilkie: Added https://github.com/weaveworks/scope/pull/176/files#diff-2d26093b1b10de6776f9b567b831f282R23

PR #177, opened 2015-06-04
Docker tagger uses event API

peterbourgon:
This is a pull request into the new-report branch. It just brings the docker mapper event API changes into the new docker tagger. We don't necessarily have to merge it this way, it's just so you can see the diff.


tomwilkie:
I included this in #178

peterbourgon:
Cool, then I close this one in deference to that one.

PR #178, opened 2015-06-04
Refactor process mappers into taggers.

tomwilkie:
- Removed the cgroup mapper
- Adds topology tagger


probe/main.go
```diff
@@ -66,32 +63,14 @@ func main() {
...
-	}
-
-	if *dockerMapper {
-		docker, err := newDockerMapper(*procRoot, *dockerInterval)
+	taggers := []tag.Tagger{tag.NewTopologyTagger()}
+	if *dockerTagger {
+		t, err := tag.NewDockerTagger(*procRoot, *dockerInterval)
 		if err != nil {
-			log.Fatal(err)
+			log.Printf("warning: failed to start docker tagger: %v", err)
```
peterbourgon: This needs to fatal out, as `t` is nil...

report/mapping_functions.go
```diff
@@ -70,10 +83,10 @@ func ProcessName(_ string, m NodeMetadata) (MappedNode, bool) {
 // are grouped into the Uncontained node.
 func ProcessContainer(_ string, m NodeMetadata) (MappedNode, bool) {
 	var id, major, minor, rank string
-	if m["docker_id"] == "" {
+	if m["docker_container_id"] == "" {
 		id, major, minor, rank = "uncontained", "Uncontained", "", "uncontained"
 	} else {
-		id, major, minor, rank = m["docker_id"], m["docker_name"], m["domain"], m["docker_image_id"]
+		id, major, minor, rank = m["docker_container_id"], m["docker_container_name"], m["domain"], m["docker_image_id"]
```
tomwilkie: @peterbourgon just to bring your attention here; I think this change was missed out of the original PR

peterbourgon: Yep, great catch. I found a couple other places, too — 1794830.

peterbourgon:
LGTM

probe/tag/pidtree.go
```diff
@@ -24,7 +24,7 @@ var (
 	readFile = ioutil.ReadFile
 )
 
-func newPIDTree(procRoot string) (*pidTree, error) {
+func newRealPIDTree(procRoot string) (*pidTree, error) {
```
tomwilkie: I'm not a big fan of having real in all the names...  I'm guess this is to avoid a collision with the stub for testing?  Why not call that netpidtreestub?

probe/tag/tagger.go
```diff
@@ -0,0 +1,25 @@
+package tag
+
+import "github.com/weaveworks/scope/report"
+
+// Tagger tags nodes with value-add node metadata.
+type Tagger interface {
+	Tag(r report.Report, ts report.TopologySelector, id string) report.NodeMetadata
```
tomwilkie: @peterbourgon Whats the thoughts behind this interface?  Its strikes me as overly general and not particularly useful.  For one, we only have two implementations, and one call site - did you say we need three of something before we think about unification?

Also, the docker_tagger is going to become much more than just a tagger (its going to have to emit at least one topology itself).

Finally, the interface (report, selector, id) seems weird.  It seems a bit tortuous, and is only ever called in a loop by tagger. 

I suggest for topology tagger, we move the loop into the tagger.   For docker tagger, we make it take the process topology explicitly.  And we get rid of this interface.

probe/tag/pidtree.go
peterbourgon: - var newFoo = {newMockFoo, newRealFoo}
- var newFooStub = {newMockFoo, newFoo}

The first option is just more symmetric. Similarly I don't like invoking newFooStub in my actual code :)

tomwilkie: I would argue for the *Stub as it makes it obvious as to why you are taking a local reference to this function; see Mathias's comment earlier today.

probe/tag/tagger.go
peterbourgon: This interface is foundational to the new data model. A tagger takes a specific node ID, and some surrounding context to allow it to cross-reference keys and dig up more information, and produces arbitrary metadata that can be merged into that node. This is the primary mechanism to enrich nodes...

> the docker_tagger is going to become much more than just a tagger (its going to have to emit at least one topology itself)

We're talking about two distinct things here:

- The docker tagger is the thing that takes already existing nodes that have process IDs in their metadata, and enriches them with additional metadata taken from Docker
- The container topology producer will be the thing that produces the container topology with information from the Docker daemon

They both may be built on the same Docker event subscriber, but they're separate logical components, nothing really to do with each other in a component model of the probe.

> Finally, the interface (report, selector, id) seems weird. It seems a bit tortuous, and is only ever called in a loop by tagger.

It's a bit awkward only because we need a way to cross-reference metadata across different topologies. For that, we need access to both the node we're enriching, i.e.

```go
md := ts(t).NodeMetadata[id]
```

and a way for the thing that implements the interface to access other, arbitrary metadata from the report, e.g.

```
foreignKey, ok := md["some_key"]
if !ok {
    return NodeMetadata{} // nothing found
}
value := r.Container.NodeMetadatas[foreignKey]["other_key"]
return NodeMetadata{"some_container_data": value}
```

> I suggest for topology tagger, we move the loop into the tagger. For docker tagger, we make it take the process topology explicitly. And we get rid of this interface.

Arguably the whole point of #123 is to generalize the concept of a tagger to something that implements this interface (or one like it). So — if, after this explanation, you still want to do this, I guess we're missing some fundamental conceptual thing... let me know...

probe/tag/pidtree.go
peterbourgon: Creating a local reference to a standard function/constructor to swap out for tests is a common enough idiom, easily identifiable with some exposure. I've never seen Stub or Ref or anything in the name of the reference var, i.e.

```go
var (
    now = time.Now     // this
    nowStub = time.Now // not this
)
```

And I'd argue it's more awkward to parse in the place where it matters the most,

```go
func doSomething() {
    object := newFooStub() // stub? why? isn't this the real code?
    object := newFoo()     // makes more sense...
}
```

If you're still not convinced, I defer.

tomwilkie: I'm not convinced either way TBH - @rade what do you prefer?

rade: ditch the `Real`. That's just horrid.

probe/tag/tagger.go
tomwilkie: - do we ever tag anything other than a whole report?  
- is there any case where we might want to apply the same tagger to different topologies?

peterbourgon: > Do we ever tag anything other than a whole report?

No.

> Is there any case where we might want to apply the same tagger to different topologies?

Yes. In fact, we almost always want to apply the same tagger to multiple topologies. For example, the docker tagger should tag nodes in any topology that can be linked to a process, and that includes Endpoints, Addresses, Containers, Processes, and potentially more that we add in the future.

— – -

Maybe we can skip a little discussion — are you driving at something like

```go
type Tagger interface {
    Tag(Report)
}
```

?

tomwilkie: Maybe it would be helpful to document all the topologies you want to introduce (even if they are not in this PR)?  I get endpoint, container & process.  Whats address?

I have yet thought of what I'm driving at, but it does strike me at the very least that:

```
	myNodeMetadata, ok := ts(r).NodeMetadatas[id]
	if !ok {
		return report.NodeMetadata{}
	}
```

Appears at the top of ~every implementation, and we're calling these taggers on topologies where we expect them to fail - making it hard to distinguish programmer errors from normal operation.  I don't like how things fail silently.

I'm having a little play to see if I can come up with something nicer.

peterbourgon:
Address is an e.g. IP address-to-address topology. It's not new, it's a rename of the currently (badly) named Network.

I agree with your sentiment about the error handling... but remember that tagging is and will always be best effort — we can't assume that all Endpoint nodes will always have processes attached to them, for example.

We've uncovered something here, though. Take the docker tagger as an example. If we stick a docker_container_id key/val into every endpoint NodeMetadata, and then to merge multiple nodes together in report.Render, as it stands now we're going to lose information, the last writer to the merged metadata will win. (In the current version this works by having an appendable list of Origins, no information is lost there.) And I think this is what you were hinting at — what about this —

- Docker tagger becomes Container Topology generator (as you said)
- Every unique docker container ID, name, etc. is represented in just one node in that topology
- All cross-topology links are encoded as foreign keys, e.g. endpoint node has process_node_id FK, process node has container_node_id FK
- report.Render then resolves those FKs to produce appendable Origins in a RenderableNode, kind of as before

Make sense?

tomwilkie:
Yes this makes good sense.  I though this was the plan all along.  Yay code review!

I'm still not comfortable with this tagging interface, but I can't come up with anything nicer.  We'll probably have to come back to it at some later point.

peterbourgon:
Well if you agree with above we can rm the interface...

tomwilkie:
Lets not introduce the containers topology in the this PR; it will make the rest of #123 unmergable.

But I think what you're saying is that, given above, tagger will _not_ need to do foreign key lookups?  And therefore, we can massively simplify this interface?

peterbourgon:
Yes I think that's correct.

tomwilkie:
Do you want to have a crack at that (if you're not busy on marketing)?  I'm trying to break out some other changes right now.

peterbourgon:
If you can merge this and we do that other work in a separate PR — LGTM

peterbourgon:
**Motivation**

Right now, the component that reads from Docker attaches metadata to existing nodes in the topologies. For example, if one container is responsible for 10 connections, the same set of Docker metadata is duplicated across those 10 nodes in the process topology. And, if we try to merge those nodes, it's last-writer-wins on which information ends up in the merged node. Buggy.

What we want instead is for the Docker component to populate its own topology, with each node representing a Docker container, and the metadata existing in that single place. Then, using the same example, if one container is responsible for 10 connections, those 10 nodes in the process topology get a single container_node_id key/val, pointing to the relevant node in the container topology. When we merge the nodes, we'll perform special handling on foreign-key-style key/vals, merging them via set-union.

This will also require changes to report rendering, so that it understands how to follow those links.

**Gameplan**

This change will occur in 5 commits:

1. Introduce package spy, describing things that can spy on the host machine and produce topologies
2. Port current func spy to spy.Proc, producing process and address topologies
3. Add spy.Docker, producing the container topology
4. Update Render to perform foreign-key resolution
5. Remove tagging interface and components

peterbourgon:
Step 3 presents a speed bump. If Render will be able to perform foreign-key resolution, it will need to have access to all topologies. That means it needs to be defined on the full Report, rather than a single Topology. Starting that work, it became apparent that it would involve porting a significant portion of the work already done in #123, specifically implementing something roughly equivalent to its [Render function](https://github.com/weaveworks/scope/blob/new-report/report/render.go#L8) but with lots of small changes to keep the diff only to the immediately relevant components. Some of those are quite tricky, for example reverting the definition of the pseudo func, and therefore the behavior of pseudo node generation, to the previous broken/nondeterminstic mode ([previously described here](https://gist.github.com/peterbourgon/3c62f654621f187dd148)). And which would eventually be undone, anyway, as we continue this path.

This is enough of a speed bump that I'd like to ask for your opinion, @tomwilkie, on how to proceed. I'm afraid that it will be a waste of time. Perhaps there is a better strategy to proceed.

tomwilkie:
Some (friendly) comments:
- Stop renaming things!  At least, stop renaming them in the same PR as you introduce new features.  Its just bloats the PRs and makes them harder to review.
- Stop doing more than one thing per PR!  We don't need to introduce the container topology in this PR.  Lets get the taggers -> mappers refactoring in, then work on the next bit.

This change has already been open for too long; lets do the minimal thing to get it ready and then merge it.  I think that minimal thing is to simplify / remove the tagger interface.

peterbourgon:
The mapper creates new topologies. I don't understand how we can introduce this change without introducing it?

tomwilkie:
Don't make the mapper create new topologies?  Preserve the existing behaviour?

peterbourgon:
The end result of this PR is that Tagger operates on just a Report. @tomwilkie PTAL.

probe/tag/docker_tagger.go
```diff
@@ -217,88 +219,80 @@ func (m *dockerMapper) handleEvent(event *docker.APIEvents, client dockerClient)
...
-	}
-	m.RUnlock()
-
-	if err != nil {
-		return "", fmt.Errorf("no container found for PID %d", pid)
-	}
+func (t *dockerTagger) Tag(r report.Report) report.Report {
+	for _, topology := range []*report.Topology{
+		&(r.Process),
+		&(r.Network),
```
tomwilkie: As it stands, r.Network doesn't contain any Nodes with a pid, does it?

probe/tag/docker_tagger.go
```diff
@@ -217,88 +219,80 @@ func (m *dockerMapper) handleEvent(event *docker.APIEvents, client dockerClient)
...
-		if err != nil {
-			break
-		}
-	}
-	m.RUnlock()
-
-	if err != nil {
-		return "", fmt.Errorf("no container found for PID %d", pid)
-	}
+func (t *dockerTagger) Tag(r report.Report) report.Report {
```
tomwilkie: We seem to be mutating the passes in report here, although the intention seems to be for us not to do that.

If we are happy with the mutation, maybe make it a *Report and don't return Report.

probe/tag/tagger.go
```diff
@@ -0,0 +1,17 @@
+package tag
+
+import "github.com/weaveworks/scope/report"
+
+// Tagger tags nodes with value-add node metadata.
+type Tagger interface {
+	Tag(r report.Report) report.Report
+	Stop()
```
tomwilkie: Given not ever Tagger implements Stop, I suggest we remove it from the interface, return a concrete DockerMapper, and then defer docker.Stop().

probe/tag/docker_tagger.go
peterbourgon: That's correct. I wanted to keep this generic, but I can also remove it.

probe/tag/docker_tagger.go
peterbourgon: This does mutate, but it's a function of the fact that Reports are chock-full of reference-typed fields, not a deliberate decision. I'd prefer to stick to value-types and returning new versions, as that's how the e.g. append builtin works, and it's easier to test and reason about. Let's bookmark this for a later changeset.

probe/tag/tagger.go
peterbourgon: OK.

probe/tag/docker_tagger.go
tomwilkie: Yes please.

probe/tag/docker_tagger.go
tomwilkie: I prefer the value-types as well.  I think it would be as simple as taking a copy at the top of the function, mutating and returning that?

tomwilkie:
Please squash this down to one change and rebase on top of master.

tomwilkie:
\o/ LGTM \o/

PR #180, opened 2015-06-05
Fix up experimental, add a makefile and add it to circle to stop experimental getting broken again.

tomwilkie:
Could also remove all the lead makefiles in experimental?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -38,6 +38,7 @@ test:
     - cd $SRCDIR; make
     - cd $SRCDIR; ./bin/test
     - cd $SRCDIR; make client-test
+    - cd $SRCDIR/experimental; make
   post:
     - goveralls -repotoken $COVERALLS_REPO_TOKEN -coverprofile=$SRCDIR/profile.cov -service=circleci || true
     - cd $SRCDIR; cp coverage.html $CIRCLE_ARTIFACTS
--- a/experimental/Makefile
+++ b/experimental/Makefile
@@ -1,0 +1,16 @@
+.PHONY: all test clean
+
+DIRS=$(shell find . -maxdepth 2 -name *.go -printf "%h\n" | uniq)
+TARGETS=$(join $(patsubst %,%/,$(DIRS)),$(DIRS))
+
+all: $(TARGETS)
+
+$(TARGETS):
+	go get -tags netgo ./$(@D)
+	go build -ldflags "-extldflags \"-static\"" -tags netgo -o $@ ./$(@D)
+
+test:
+	go test ./...
+
+clean:
+	go clean ./...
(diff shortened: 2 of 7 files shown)


experimental/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+all: $(targets)
+
+define T
+$1/$1: $1/*.go
+	go get -tags netgo ./$$(@D)
+	go build -ldflags "-extldflags \"-static\"" -tags netgo -o $$@ ./$$(@D)
+endef
+
+$(foreach d,$(dirs),$(eval $(call T,$d)))
```
peterbourgon: Wow. Could we rather just have some some copy/pasted (dumb) targets?

experimental/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+test:
+	go test ./...
+
+clean:
+	go clean ./...
```
peterbourgon: Actually, can we not have a Makefile, please? Experimental stuff shouldn't be part of the regular build process.

circle.yml
```diff
@@ -38,6 +38,7 @@ test:
     - cd $SRCDIR; make
     - cd $SRCDIR; ./bin/test
     - cd $SRCDIR; make client-test
+    - cd $SRCDIR/experimental; make
```
peterbourgon: Similarly I'd vote to remove that

tomwilkie:
I added this as recent refactoring broken experimental, and a broken experimental makes automated refactoring tools (gorename) break.

experimental/graphviz/handle.go
```diff
@@ -14,7 +14,7 @@ import (
 func handleTXT(r Reporter) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
 		w.Header().Set("Content-Type", "text/plain")
-		dot(w, r.Report().Process.RenderBy(mapFunc(req), classView(req)))
+		//dot(w, r.Report().Process.RenderBy(mapFunc(req), nil, classView(req)))
```
peterbourgon: Can you fix this properly? If you don't want to bother, that's fine — in which case I think it's better to just temporarily remove graphviz from the Makefile or whatever.

experimental/Makefile
peterbourgon: I think this would be nicer if we just had individual, dumb, copy-pasted targets, removing dirs/targets/T/foreach

peterbourgon:
:+1:

experimental/Makefile
tomwilkie: I can make it a little clearer without loosing the concision:

```
.PHONY: all test clean

DIRS=$(shell find . -maxdepth 2 -name *.go -printf "%h\n" | uniq)
TARGETS=$(join $(patsubst %,%/,$(DIRS)),$(DIRS))

all: $(TARGETS)

$(TARGETS):
	go get -tags netgo ./$(@D)
	go build -ldflags "-extldflags \"-static\"" -tags netgo -o $@ ./$(@D)
```

But it does loose the dependancies.

PR #181, opened 2015-06-05
Rename NewReport -> MakeReport

tomwilkie:
Depends on #178 & #180


peterbourgon:
This PR includes both of those, correct?

tomwilkie:
Yes as it depends on them.  I won't submit this until they are submitted, so eventually it'll just have one.

Conversely, I could rebase this against master and get it in before #178 if that makes it easier for you?

peterbourgon:
If you can merge #178 and #180, I'll review this with the proper diff.

tomwilkie:
I've merged #180 already, so here is this off master; no dependancies.

peterbourgon:
LGTM

PR #182, opened 2015-06-05
probe: only allow Docker on Linux

### LINKED ISSUE #173: Suppress Docker /proc error messages on Darwin
```
sudo ./probe 
2015/06/03 17:13:31 probe starting, version unknown
2015/06/03 17:13:31 open /proc: no such file or directory
```

peterbourgon:
Fixes #173

### DIFF
--- a/probe/main.go
+++ b/probe/main.go
@@ -79,7 +79,7 @@ func main() {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


tomwilkie:
GLTM

PR #185, opened 2015-06-08
ID constructors

peterbourgon:
Use constructor functions to produce IDs in reports/topologies/tests. Remove manual, error-prone, and sometimes arbitrary string concatenation.

This is broken up into individual commits so you can see the progression, and if you have specific questions about a choice. I'll of course squash & rebase before merging.

### DIFF
--- a/report/id.go
+++ b/report/id.go
@@ -1,29 +1,117 @@
 package report
 
 import (
+	"fmt"
 	"net"
 	"strings"
 )
 
-// IDAddresser is used to get the IP address from an addressID. Or nil.
-type IDAddresser func(string) net.IP
+// TheInternet is used as a node ID to indicate a remote IP.
+const TheInternet = "theinternet"
 
-// AddressIPPort translates "scope;ip;port" to the IP address. These are used
-// by Process topologies.
-func AddressIPPort(id string) net.IP {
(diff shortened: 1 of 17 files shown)


report/id.go
```diff
@@ -1,29 +1,117 @@
...
+	if len(fields) != 2 {
+		//log.Printf("AddressIDAddresser: bad input %q", id)
+		return nil
 	}
-	return net.ParseIP(parts[1])
+	return net.ParseIP(fields[1])
+}
+
+// PanicIDAddresser will panic if it's ever called. It's used in topologies
+// where there are never and edges, and so it's nonsensical to try and extract
```
tomwilkie: s/and/any/

tomwilkie:
Might it be possible to use these ID functions instead of trySplitAddr in mapping_functions.go?

tomwilkie:
Otherwise, LGTM

peterbourgon:
> Might it be possible to use these ID functions instead of trySplitAddr in mapping_functions.go?

I agree it's a bit hacky. trySplitAddr is basically ParseArbitraryNodeID, since its callsites (pseudo funcs) just have opaque node IDs and don't know what topology they come from. So without changing how pseudo funcs work, we can't make it much smarter. I'll add a comment to this effect.

PR #187, opened 2015-06-08
Update IDList

peterbourgon:
- MakeIDList rather than NewIDList
- Iterate on test coverage

Simple.


report/id_list.go
```diff
@@ -1,19 +1,25 @@
...
-// NewIDList makes a new IDList.
-func NewIDList(ids ...string) IDList {
+// MakeIDList makes a new IDList.
+func MakeIDList(ids ...string) IDList {
 	sort.Strings(ids)
 	return IDList(ids)
 }
 
+// Copy returns a value copy, useful for tests.
+func (a IDList) Copy() IDList {
```
tomwilkie: Is the intention for this to be used by non-test code in the future?  If not, move it into the _test.go file

tomwilkie:
1 comment, otherwise LGTM

report/id_list.go
peterbourgon: It's primarily for tests, yes. But I can't move it there, because that's package report_test, and you can't define methods on types outside of their package. I could convert it to a free function. Is your concern simply pollution of the exported symbols?

peterbourgon: Actually, I'll just remove it for now, as it's apparently not used anywhere except its own unit test for the time being.

PR #190, opened 2015-06-09
Report fields

peterbourgon:
- Rename Process to Endpoint
- Rename Network to Address

Sets the foundation for future work.


report/report.go
```diff
@@ -6,21 +6,12 @@ import (
...
-// might make sense to make them more distinct in the user interface.
+// Report is the core data type. It's produced by probes, and consumed and
+// stored by apps. It's composed of multiple topologies, each representing
+// a different (related, but not equivalent) view of the network.
 type Report struct {
-	Process Topology
-	// Transport Topology
-	Network Topology
+	Endpoint Topology
+	Address  Topology
```
tomwilkie: I would love it if you could put a comment about what is in each topology - for instance, when their are nodes, when there are only edges, what the key type is etc.

tomwilkie:
Other than the one comment, LGTM

peterbourgon:
Merged with 1e0b83e

PR #192, opened 2015-06-09
Host topology

peterbourgon:
Continuation of piecemeal integration of #123. The HostMetadata struct is converted into a Host topology. This represents a significant change to the mechanics/implementation of reports, but yields the same net effect.

PR #194, opened 2015-06-09
OriginHosts and OriginNodes become Origins

peterbourgon:
Another baby step towards #123, this change follows from #192 and merges the two concepts of Origin in a renderable node. We also cut out a layer of abstraction, and add an OriginTable method to Report, which directly generates a table of info for the detail pane given any origin node ID.


report/detailed_node.go
```diff
@@ -0,0 +1,54 @@
...
+		if len(rows) > 0 {
+			tables = append(tables, Table{"Connections", true, rows})
+		}
+	}
+
+	// RenderableNode may be the result of merge operation(s), and so may have
+	// multiple origins. The ultimate goal here is to generate tables to view
+	// in the UI, so we skip the intermediate representations, but we could
+	// add them later.
+outer:
```
tomwilkie: I'm not a massive fan of labeled loop - they are a code smell to me, and usually point to something needing breaking out into a separate function.

report/report.go
```diff
@@ -111,3 +110,72 @@ func (r Report) LocalNetworks() []*net.IPNet {
 	}
 	return ipNets
 }
+
+// OriginTable produces a table (to be consumed directly by the UI) based on
+// an origin ID, which is (optimistically) a node ID in one of our topologies.
+func (r Report) OriginTable(originID string) (Table, bool) {
```
tomwilkie: It seems like this doesn't belong as a method on the Report struct, but rather should be associated with detailed_node.go?

report/report.go
```diff
@@ -111,3 +110,72 @@ func (r Report) LocalNetworks() []*net.IPNet {
...
+	for nodeID, nodeMetadata := range r.Endpoint.NodeMetadatas {
```
tomwilkie: It stikes me this could be done like this:
```
if nodeMetadata, ok := r.Endpoint.NodeMetadatas[nodeID]; ok {
  return endpointOriginTable(nodeMetadata)
}
```

report/detailed_node.go
```diff
@@ -0,0 +1,54 @@
...
+	for _, id := range n.Origins {
+		table, ok := r.OriginTable(id)
+		if !ok {
+			continue
+		}
+		// Naïve equivalence-based deduplication.
+		for _, existing := range tables {
+			if reflect.DeepEqual(existing, table) {
```
tomwilkie: Yuk!  Why can you not compare node ids?  Or even better, does n.Origins offer set-like semantics?

report/detailed_node.go
peterbourgon: Like goto, it's useful in extremely limited circumstances, and this is one of them.

report/report.go
peterbourgon: Ha, yes, that's much smarter.

report/report.go
peterbourgon: It has to be either this or

```go
func OriginTable(r Report, originID string) (Table, bool)
```

, right?

report/detailed_node.go
peterbourgon: Good eye. Yes, n.Origins is already a set. This code left over from a previous time, when we didn't have that guarantee. I believe we can remove the deduplication clause altogether.

report/report.go
tomwilkie: Looks like it.  Can then be moved into detailed_node.go

peterbourgon: I guess I can do it, but why is a free function that takes a report as the first parameter better than a method on report?

tomwilkie: I'm not fussed about free function or not, more that this function is only used from detailed_node.go (I think), so I guess it belongs there.  But I just realised they are in the same package so the receiver doesn't matter.

OTOH you could use the same argument about MakeDetailedNode; having both of these in the public API for a report seems weird.

report/detailed_node.go
peterbourgon: ...but it's irrelevant, as it goes away with one of the below changes, anyway.

peterbourgon:
How's that look?

app/detail_pane.go
```diff
@@ -1,94 +1 @@
 package main
```
tomwilkie: I'm confused; is there a reason you deleted everything except this one line?

tomwilkie:
One last comment; otherwise LGTM

Is coveralls being weird or did it really drop by 6%? That might be worth investigating.

app/detail_pane.go
peterbourgon: Oops! :)

tomwilkie:
Also squash pls!

peterbourgon:
> Is coveralls being weird or did it really drop by 6%? That might be worth investigating.

Coveralls has been being weird all day, but it wouldn't surprise me if coverage has dropped — there are a lot of tests that I'm waiting to port from #123, because they rely on features that haven't yet been moved over, and I'd rather do it all in one motion than waste time on getting them working in the intermediate steps and then blowing it away. 

At the end of all of this, I'll promise a significant increase in test coverage, probably double-digit.

tomwilkie:
SGTM!

PR #197, opened 2015-06-09
Don't make adjacency edges into 'the internet' if they refer to valid nodes

### LINKED ISSUE #196: Scope renders connections between processes/containers on the weave network as connections to "The Internet"
This is because scope uses the host's interface's IPs to decide whats the internet and what's not; but none of the IP addresses on a weave network appear in the host net namespace (the veth pairs don't have ips)

tomwilkie:
Fixes #196

### DIFF
--- a/report/mapping.go
+++ b/report/mapping.go
@@ -205,5 +205,8 @@ func trySplitAddr(addr string) (string, string) {
 	if len(fields) == 3 {
 		return fields[1], fields[2]
 	}
-	return fields[1], ""
+	if len(fields) == 2 {
+		return fields[1], ""
+	}
+	panic(addr)
 }
--- a/report/topology.go
+++ b/report/topology.go
@@ -184,12 +184,29 @@ func (t Topology) EdgeMetadata(mapFunc MapFunc, srcRenderableID, dstRenderableID
 
 // Squash squashes all non-local nodes in the topology to a super-node called
 // the Internet.
+// We rely on the values in the t.Adjacency lists being valid keys in
+// t.NodeMetadata (or t.Adjacency).
 func (t Topology) Squash(f IDAddresser, localNets []*net.IPNet) Topology {
-	isRemote := func(ip net.IP) bool { return !netsContain(localNets, ip) }
+	isRemote := func(id string) bool {
+		if _, ok := t.NodeMetadatas[id]; ok {
+			return false // it is a node, cannot possibly be remote
+		}
+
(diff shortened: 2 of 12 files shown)


report/topology.go
```diff
@@ -184,12 +184,30 @@ func (t Topology) EdgeMetadata(mapFunc MapFunc, srcRenderableID, dstRenderableID
...
+		// TODO make MakeAdjacencyID just take a single argument,
+		// and ParseAdjacencyID do the right thing.
+		_, ok = t.Adjacency[MakeAdjacencyID("", id)]
+		if ok {
+			return false
+		}
+
+		ip := f(id)
+		return ip == nil || !netsContain(localNets, ip)
+	}
```
peterbourgon:
```go
if _, ok := t.NodeMetadatas[id]; ok {
    return false // it's a node, cannot possibly be remote
}
if _, ok := t.Adjacency[MakeAdjacencyID("", id)]; ok {
    return false // it's in our adjacency list, cannot possibly be remote
}
if ip := f(id); ip != nil && !netsContain(localNets, ip) {
    return true // it's a valid IP, but not in our local nets, so it is remote
}
return false
```

report/topology.go
```diff
@@ -184,12 +184,30 @@ func (t Topology) EdgeMetadata(mapFunc MapFunc, srcRenderableID, dstRenderableID
...
 func (t Topology) Squash(f IDAddresser, localNets []*net.IPNet) Topology {
-	isRemote := func(ip net.IP) bool { return !netsContain(localNets, ip) }
+	isRemote := func(id string) bool {
+		_, ok := t.NodeMetadatas[id]
+		if ok {
+			return false
+		}
+
+		// TODO make MakeAdjacencyID just take a single argument,
+		// and ParseAdjacencyID do the right thing.
```
peterbourgon: Let's make this change now rather than leaving it as a TODO. I'm OK with changing the function definition in id.go as long as we leave the warning comments.

report/mapping.go
```diff
@@ -201,9 +202,13 @@ func InternetOnlyPseudoNode(_ string, _ RenderableNode, dst string) (MappedNode,
 //
 // TODO change how pseudofuncs work, and eliminate this helper.
 func trySplitAddr(addr string) (string, string) {
+	log.Println(addr)
```
peterbourgon: rm?

peterbourgon:
Apart from that one extra Println, LGTM!

PR #206, opened 2015-06-10
Prevent multiple connections for the same app.

### LINKED ISSUE #195: We seem to be allowing connections from the same app twice
```
2015/06/09 17:36:06 collector: got a report from 10.0.2.15:4030
2015/06/09 17:36:09 collector: got a report from 127.0.0.1:4030
```

tomwilkie:
- Apps generate a unique, random id on startup and send this to probes when connecting
- Probes dedupe connections based on this id

Fixes #195

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"log/syslog"
+	"math/rand"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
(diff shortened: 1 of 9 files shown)


tomwilkie:
Probably want to increase maxbackoff from 10s and quieten-down the logging?

peterbourgon:
Why is this better than simply using the remote addr as the ID, as we were doing previously? (I think I know the answer, I just want to get it on paper...)

> increase maxbackoff from 10s

Sure.

tomwilkie:
That bit was in the bug:

> Problem is connecting on 127.0.0.1 makes it look like the connection is coming from 127.0.0.1

peterbourgon:
Oh! Missed the bug link. LGTM!

PR #212, opened 2015-06-10
Process topology

peterbourgon:
—


probe/main.go
```diff
@@ -92,6 +94,9 @@ func main() {
 			case <-pubTick:
 				publishTicks.WithLabelValues().Add(1)
 				r.Host = hostTopology(hostID, hostName)
+				if dockerTagger != nil {
```
tomwilkie: I think this should be decoupled from dockerTagger; people will want this info even when not running on docker.

tomwilkie:
LGTM

PR #215, opened 2015-06-11
Fix back browsing after node was clicked

davkal:


peterbourgon:
Confirm it works! LGTM.

PR #216, opened 2015-06-11
Display a n/a header in details if nodes are gone

### LINKED ISSUE #193: If details window is empty, cannot close it
Can become empty if you remove the container you had selected, and refresh the page.

davkal:
fixes #193


tomwilkie:
It doesn't work when the node disappears whilst scope is loaded; it only works if you reload scope.

Try deleting a container whilst scope is running.

tomwilkie:
LGTM

PR #217, opened 2015-06-11
Subscribe to docker stats and place some in the Container topology.

tomwilkie:
Work towards #114

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -37,7 +37,7 @@ func makeTopologyList(rep Reporter) func(w http.ResponseWriter, r *http.Request)
 					subTopologies = append(subTopologies, APITopologyDesc{
 						Name:  subDef.human,
 						URL:   "/api/topology/" + subName,
-						Stats: stats(render(rpt, subDef.maps)),
+						Stats: stats(subDef.renderer.Render(rpt)),
 					})
 				}
 			}
@@ -45,7 +45,7 @@ func makeTopologyList(rep Reporter) func(w http.ResponseWriter, r *http.Request)
 				Name:          def.human,
 				URL:           "/api/topology/" + name,
 				SubTopologies: subTopologies,
-				Stats:         stats(render(rpt, def.maps)),
+				Stats:         stats(def.renderer.Render(rpt)),
 			})
 		}
 		respondWith(w, http.StatusOK, topologies)
--- a/app/api_topology.go
+++ b/app/api_topology.go
@@ -30,19 +30,10 @@ type APIEdge struct {
 	Metadata report.AggregateMetadata `json:"metadata"`
 }
 
-func render(rpt report.Report, maps []topologyMapper) report.RenderableNodes {
-	result := report.RenderableNodes{}
-	for _, m := range maps {
-		rns := m.selector(rpt).RenderBy(m.mapper, m.pseudo)
-		result.Merge(rns)
-	}
-	return result
-}
-
 // Full topology.
 func handleTopology(rep Reporter, t topologyView, w http.ResponseWriter, r *http.Request) {
 	respondWith(w, http.StatusOK, APITopology{
-		Nodes: render(rep.Report(), t.maps),
+		Nodes: t.renderer.Render(rep.Report()),
 	})
 }
 
@@ -69,7 +60,7 @@ func handleNode(rep Reporter, t topologyView, w http.ResponseWriter, r *http.Req
 		vars     = mux.Vars(r)
 		nodeID   = vars["id"]
 		rpt      = rep.Report()
-		node, ok = render(rpt, t.maps)[nodeID]
+		node, ok = t.renderer.Render(rep.Report())[nodeID]
 	)
 	if !ok {
 		http.NotFound(w, r)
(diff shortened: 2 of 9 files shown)


probe/main.go
```diff
@@ -144,8 +144,18 @@ func hostTopology(hostID, hostName string) report.Topology {
...
+	for {
+		switch <-sigs {
+		case syscall.SIGINT:
+			log.Printf("=== received SIGINT ===\n*** exiting\n")
+			return
+		case syscall.SIGQUIT:
+			stacklen := runtime.Stack(buf, true)
+			log.Printf("=== received SIGQUIT ===\n*** goroutine dump...\n%s\n*** end\n", buf[:stacklen])
+		}
+	}
```
peterbourgon: If you simply don't trap SIGQUIT, you get this behavior by default in the runtime.

tomwilkie:
Note this introduces a startup lag (as it connects all the stats go routines on the main thread).  This should be moved off to a background thread (and parallelised?)

probe/tag/docker_tagger.go
```diff
@@ -157,15 +179,15 @@ func (t *DockerTagger) update() bool {
...
 
-func (t *DockerTagger) updateContainers(client dockerClient) error {
-	apiContainers, err := client.ListContainers(docker.ListContainersOptions{All: true})
+func (t *DockerTagger) updateContainers() error {
+	apiContainers, err := t.client.ListContainers(docker.ListContainersOptions{All: true})
 	if err != nil {
 		return err
 	}
 
 	containers := []*docker.Container{}
```
peterbourgon: t.update never returns unless there's an error during set up, or the user issues a quit. But you call it repeatedly in loop, and always on a ticker. Why?

tomwilkie: To make it retry if it fails; if a connection is dropped for instance, or Docker is restarted.

probe/main.go
tomwilkie: D'oh.  Weave has this everywhere, so I assumed not.  Might be worth filing a bug with weave too.

peterbourgon:
There's a lot of manual mutex locking going on. Since you already have a goroutine handling Docker events, why not thread all user requests through it, too? Then you can eliminate all the locks.

tomwilkie:
Yes the locking makes me nervous; going to an actor model seems appropriate here.

I was always considering having our own Container struct, and moving all the stats & updating stuff into that.  Then docker mapper is responsible for creating / removing containers in response to events.

We should probably also break out the mapper & topology logic to be a client of this type.

probe/tag/docker_tagger.go
```diff
@@ -249,6 +295,79 @@ func (t *DockerTagger) updatePIDTree() error {
...
+	}
+
+	conn.Close()
+	delete(t.statsConns, containerID)
+	delete(t.latestStats, containerID)
+	return
+}
+
 // Containers returns the Containers the DockerTagger knows about.
 func (t *DockerTagger) Containers() []*docker.Container {
```
peterbourgon: I think this is dead code?

probe/tag/docker_tagger.go
```diff
@@ -336,8 +455,42 @@ func (t *DockerTagger) ContainerTopology(scope string) report.Topology {
...
 		result.NodeMetadatas[nodeID] = nmd
 	}
 	return result
 }
+
+// called whilst holding t.RLock()
+func (t *DockerTagger) getStats(containerID string) map[string]string {
+	stats, ok := t.latestStats[containerID]
+	if !ok {
+		return nil
```
peterbourgon:
```go
return map[string]string{}
```

is safer :)

probe/tag/docker_tagger.go
tomwilkie: Used by tracer.

peterbourgon:
I've been reading this for awhile and am having trouble building and keeping a mental model of what's happening. There's a lot of action at a distance, mutating the DockerTagger struct in-place. The preponderance of mutex-locking is an artifact of that. I think it would benefit from a single loop() method handling user requests, events from the Docker daemon, and events from individual /container/foo/stats streams.

tomwilkie:
@peterbourgon I didn't go full actor on this, but I've rationalised the flow and locking.

Let me know what you think.  There is a lot of potential for refactoring here, but I'd prefer to save anything major for another PR.

probe/tag/docker_tagger.go
```diff
@@ -249,13 +217,52 @@ func (t *DockerTagger) updatePIDTree() error {
...
 }
 
+func (t *DockerTagger) addContainer(containerID string) error {
+	container, err := t.client.InspectContainer(containerID)
+	if err != nil {
+		// Don't spam the logs if the container was short lived
+		if _, ok := err.(*docker.NoSuchContainer); !ok {
+			return err
+		}
+		return nil
```
peterbourgon: Hmm, should this be inverted?

```go
if _, ok := err.(*docker.NoSuchContainer); ok {
    return nil
}
return err
```

peterbourgon:
Much, much easier to understand. Thanks a bunch. LGTM.

tomwilkie:
Going to hold off on this until post 0.3

PR #219, opened 2015-06-11
WeaveTagger and the Overlay topology 

peterbourgon:
PTAL! No changes to the Weave /status-json yet, obviously.

### DIFF
--- a/probe/main.go
+++ b/probe/main.go
@@ -33,6 +33,7 @@ func main() {
 		spyProcs           = flag.Bool("processes", true, "report processes (needs root)")
 		dockerEnabled      = flag.Bool("docker", true, "collect Docker-related attributes for processes")
 		dockerInterval     = flag.Duration("docker.interval", 10*time.Second, "how often to update Docker attributes")
+		weaveRouterAddr    = flag.String("weave.router.addr", "", "IP address or FQDN of the Weave router")
 		procRoot           = flag.String("proc.root", "/proc", "location of the proc filesystem")
 	)
 	flag.Parse()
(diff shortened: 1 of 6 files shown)


probe/main.go
```diff
@@ -66,8 +67,12 @@ func main() {
 	}
 	defer publisher.Close()
 
+	var (
+		dockerTagger *tag.DockerTagger
+		weaveTagger  *tag.WeaveTagger
```
tomwilkie: Is tagger the right name?  It produces a topology, doesn't tag an existing one.

tomwilkie:
1 comment, otherwise LGTM.  Hold off until post v0.3.0?

probe/main.go
peterbourgon: It is consistent with DockerTagger, which both produces a topology and tags existing topologies. I agree the naming of probe components is currently schizophrenic; that's a consequence of the organic growth of the package, which has resulted in poor separation of concerns and awkward names. I intend to address those things holistically in a refactor...

tomwilkie: SGTM, we've got a ticket for reorganising all this.

probe/tag/weave_tagger.go
```diff
@@ -0,0 +1,102 @@
...
+// OverlayTopology produces an overlay topology from the Weave router.
+func (t WeaveTagger) OverlayTopology() report.Topology {
+	topology := report.NewTopology()
+
+	resp, err := http.Get(t.url)
+	if err != nil {
+		log.Printf("Weave Tagger: %v", err)
+		return topology
+	}
+	defer resp.Body.Close()
```
tomwilkie: Note Body.close will block until it reads an EOF.

probe/tag/weave_tagger.go
```diff
@@ -0,0 +1,102 @@
...
+		topology.NodeMetadatas[report.MakeOverlayNodeID(peer.Name)] = report.NodeMetadata{
+			WeavePeerName:     peer.Name,
+			WeavePeerNickName: peer.NickName,
+		}
+	}
+
+	return topology
+}
+
+func sanitize(scheme string, port int, path string) func(string) (string, error) {
```
tomwilkie: Why return a function?  As far as I can tell its used immediately.

tomwilkie:
LGTM.  Pls rebase & squash.

probe/tag/weave_tagger.go
peterbourgon: It's just currying the sanitization parameters. Better separation of concerns than passing params + target string as equal parameters, lighter-weight than defining a method on a POD struct.

Issue #131, opened 2015-05-29
Render pseudonodes differently


peterbourgon:
Pseudonodes are less important than regular nodes. We should render them in a way that makes that clear. Maybe with less saturation and smaller?

davkal:
Fixed via #221

PR #221, opened 2015-06-12
Styled pseudonodes pale and remove interaction

davkal:
Solution for #131


peterbourgon:
Confirmed working! Thanks. LGTM.

PR #225, opened 2015-06-15
Use crypto/rand to generate app id on startup.

### LINKED ISSUE #224: App always starts with an ID of 4d65822107fcfd52
Not very random, causes scope clustering to fail...

tomwilkie:
Fixes #224

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -1,11 +1,13 @@
 package main
 
 import (
+	"crypto/rand"
 	"flag"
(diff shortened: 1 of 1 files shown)


peterbourgon:
To fix #224 you just need to invoke `rand.Seed(time.Now().UnixNano())` at the top of func main. No need to import crypto/rand or any of this, AFAICT.

PR #226, opened 2015-06-15
Seed random so different apps have different ids.

### LINKED ISSUE #224: App always starts with an ID of 4d65822107fcfd52

tomwilkie:
Fixes #224


peterbourgon:
LGTM! :)

PR #229, opened 2015-06-15
Fix node scoping rules

### LINKED ISSUE #227: Cross-host connects not showing in UI when connection traverses docker port mapping
conn

peterbourgon:
We only want to scope (i.e. prefix with hostID) those addresses that are deemed loopback, to disambiguate them. Otherwise, we want to leave addresses in unscoped form, so they can be matched, and links between communicating nodes properly made.

So, we make the isLoopback check in MakeAddressID, and omit hostID if the address isn't loopback. So far so good.

But this breaks topology rendering, as we were relying on extracting hostID from adjacency node IDs, to populate origin hosts in the rendered node output. So we need another way to get origin host from an arbitrary node.

A survey revealed no reliable way to get that information from IDs in their new form. However, we have access to node metadata, so this changeset introduces the OriginHostTagger, which tags each node with its origin host, via the foreign-key semantics we'll use going forward.

Fixes #227


probe/tag/origin_host_tagger.go
```diff
@@ -0,0 +1,29 @@
...
+
+type originHostTagger struct{ hostNodeID string }
+
+// NewOriginHostTagger tags each node with a foreign key linking it to its
+// origin host in the host topology.
+func NewOriginHostTagger(hostID string) Tagger {
+	return &originHostTagger{hostNodeID: report.MakeHostNodeID(hostID)}
+}
+
+func (t originHostTagger) Tag(r report.Report) report.Report {
```
tomwilkie: Could slightly adapt and use report.Topologies here?

probe/tag/topology_tagger.go
```diff
@@ -13,8 +13,11 @@ func NewTopologyTagger() Tagger {
 
 func (topologyTagger) Tag(r report.Report) report.Report {
 	for val, topology := range map[string]*report.Topology{
```
tomwilkie: Similarly here.

tomwilkie:
Otherwise LGTM

PR #231, opened 2015-06-15
Squash merged reports as they come out of the LIFO, not individual reports as they come in.

### LINKED ISSUE #227: Cross-host connects not showing in UI when connection traverses docker port mapping

tomwilkie:
Helps #227

### DIFF
--- a/app/report_lifo.go
+++ b/app/report_lifo.go
@@ -41,7 +41,6 @@ func NewReportLIFO(r reporter, maxAge time.Duration) *ReportLIFO {
 			select {
 			case report := <-r.Reports():
 				// Incoming report from the collecter.
-				report = report.Squash() // TODO?: make this a CLI argument.
 				tr := timedReport{
 					Timestamp: time.Now(),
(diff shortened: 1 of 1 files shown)


peterbourgon:
What's the net effect?

tomwilkie:
There is no guarantee that the merged reports coming out of the collector are the result of merging reports from multiple probes.  The way timeouts are configured right now makes it very unlikely (collector interval is 1s, probe is 2s).  This change makes it way more likely, and therefore makes the cross-host linking stuff behave as it should.

tomwilkie:
Note that we do the pseudo node munging on the LIFO merged report, so this also brings the squashing in line with that.

tomwilkie:
And we have to put this in 0.3.0

peterbourgon:
Got it. Let's add a report_lifo_test.go with a test that ensures 2 sequential reports with different local_networks squash to the right thing.

peterbourgon:
LGTM.

PR #235, opened 2015-06-15
Prevent leftover nodes after scope restart

### LINKED ISSUE #162: Restarting scope results in leftover "probe" nodes in the graph
Not really a problem, just seems a bit weird. Nodes are cleared when going to a different view and back.

```bash
sudo scope launch
open http://localhost:4040
sudo scope stop
sleep 5
sudo scope launch
```

After a few restarts:
screen shot 2015-06-02 at 16 23 30

davkal:
* clear nodes cache when websocket closes connection
* also show reconnection status

fixes #162

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -13,3 +13,17 @@ To see a topology, `../app/app` needs to be running, as well as a probe.
 This directory has a `.eslintrc`, make sure your editor supports linter hints.
 To run a linter, you also run `gulp lint`.
 
+## Logging
+
+The Scope UI uses [debug](https://www.npmjs.com/package/debug) for logging, e.g.,:
+
+```
+const debug = require('debug')('scope:app-store');
+debug('Store log message');
+```
+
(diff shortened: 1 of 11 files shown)


peterbourgon:
@davkal confirmed it doesn't duplicate nodes anymore, but when I `scope stop`, the UI immediately blanks out — good — but doesn't have a little status badge telling me it's disconnected — bad?

screen shot 2015-06-15 at 20 58 38


If that's expected then LGTM.

davkal:
Thinking about it, the immediate wiping on error makes it look broken. I'll make sure that the nodes stay and the reconnect message appears.

I'm wondering why your reconnect message did not appear. `scope stop` stops the web server, right?

peterbourgon:
Latest version works as I expect. After I `scope stop` I get the `◻︎ Trying to reconnect...` notice, and the nodes stay in place. When scope comes back, all is normal again. LGTM!

Issue #228, opened 2015-06-15
Produce containers topology from endpoints via processes...


tomwilkie:
Mainly just refactoring:
- Allow the Renderers to recurse
- Add detailednode -> detailed node mappers

tomwilkie:
This is the aim:

```
render.Reduce{
    render.LeafMap{Selector: report.SelectContainer, Mapper: report.MapContainerIdentity, Pseudo: report.InternetOnlyPseudoNode},
    render.Map{
        Mapper: report.MapProcess2Container,
        Renderer: render.Reduce{
            render.LeafMap{Selector: report.SelectProcess, Mapper: report.MapProcessIdentity, Pseudo: report.InternetOnlyPseudoNode},
            render.Map{
                Mapper: report.MapEndpoint2Process,
                Renderer: render.LeafMap{Selector: report.SelectEndpoint, Mapper: report.MapEndpointIdentity, Pseudo: report.InternetOnlyPseudoNode},
            }
        }
    }
}
```

As part of this change, I'll be moving all the rendering code from report to render/

PR #240, opened 2015-06-16
Move mapping functions and main render function in render package.

tomwilkie:
This is a continuation of #228, moving all the report -> renderablenodes code into the render/ package.

Also moves the topodiff code into render/.

Should have no semantic change.


peterbourgon:
Seems fine to me.

PR #242, opened 2015-06-16
Move RenderableNode and DetailedNode into render/

tomwilkie:
A continuation of #228

### DIFF
--- a/render/detailed_node_test.go
+++ b/render/detailed_node_test.go
@@ -1,9 +1,10 @@
-package report_test
+package render_test
 
 import (
 	"reflect"
 	"testing"
 
+	"github.com/weaveworks/scope/render"
 	"github.com/weaveworks/scope/report"
 )
 
@@ -12,26 +13,30 @@ func TestMakeDetailedNode(t *testing.T) {
 }
 
 func TestOriginTable(t *testing.T) {
-	if _, ok := report.OriginTable(reportFixture, "not-found"); ok {
+	if _, ok := render.OriginTable(rpt, "not-found"); ok {
 		t.Errorf("unknown origin ID gave unexpected success")
 	}
-	for originID, want := range map[string]report.Table{
-		client54001EndpointNodeID: {
+	for originID, want := range map[string]render.Table{
+		client54001NodeID: {
 			Title:   "Origin Endpoint",
 			Numeric: false,
-			Rows:    []report.Row{{"Host name", clientHostName, ""}},
+			Rows: []render.Row{
+				{"Host name", clientHostName, ""},
+				{"PID", "10001", ""},
+				{"Process name", "curl", ""},
+			},
 		},
 		clientAddressNodeID: {
 			Title:   "Origin Address",
 			Numeric: false,
-			Rows: []report.Row{
+			Rows: []render.Row{
 				{"Host name", clientHostName, ""},
 			},
 		},
 		report.MakeProcessNodeID(clientHostID, "4242"): {
 			Title:   "Origin Process",
 			Numeric: false,
-			Rows: []report.Row{
+			Rows: []render.Row{
 				{"Name (comm)", "curl", ""},
 				{"PID", "4242", ""},
 			},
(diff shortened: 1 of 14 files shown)


render/detailed_node_test.go
```diff
@@ -39,14 +43,14 @@ func TestOriginTable(t *testing.T) {
 		serverHostNodeID: {
 			Title:   "Origin Host",
 			Numeric: false,
-			Rows: []report.Row{
-				{"Host name", serverHostName, ""},
+			Rows: []Row{
+				{"Host name", serverHostID, ""},
```
peterbourgon: This isn't semantically correct. We currently set host ID == hostname in the probe, but they're not the same concept.

render/render_test.go
```diff
@@ -94,6 +94,7 @@ var (
 					"domain":          "client-54001-domain",
 					"pid":             "10001",
 					report.HostNodeID: clientHostNodeID,
+					"host_name":       clientHostID,
```
peterbourgon: Similarly, this isn't correct.

render/render_test.go
```diff
@@ -186,11 +209,23 @@ var (
 				},
 			},
 		},
+		Host: report.Topology{
+			Adjacency: report.Adjacency{},
+			NodeMetadatas: report.NodeMetadatas{
+				serverHostNodeID: report.NodeMetadata{
+					"host_name":      serverHostID,
```
peterbourgon: Similarly, this isn't correct.

render/detailed_node_test.go
```diff
@@ -12,26 +12,30 @@ func TestMakeDetailedNode(t *testing.T) {
...
+			},
 		},
-		clientAddressNodeID: {
+		clientIP: {
 			Title:   "Origin Address",
 			Numeric: false,
-			Rows: []report.Row{
-				{"Host name", clientHostName, ""},
+			Rows: []Row{
+				{"Host name", "client.hostname.com", ""},
```
peterbourgon: Why did this get literalized?

peterbourgon:
Gimme another poke when those are resolved...

tomwilkie:
Sorry about that; the origin tests were originally using the report fixture from report/report_fixture_test.go, but I changed them to use the one from render_test.go.

I'll canonicalise a report fixture for the render module in a future changeset.

render/detailed_node_test.go
```diff
@@ -1,4 +1,4 @@
-package report_test
+package render
```
peterbourgon: Please keep this in the _test package. The only reason _test.go files should be in the original package namespace is if they're explicitly testing package/component internals, and even then it's a bad smell unless the package/component is really complex.

render/detailed_node_test.go
```diff
@@ -12,26 +12,30 @@ func TestMakeDetailedNode(t *testing.T) {
...
 
 func TestOriginTable(t *testing.T) {
-	if _, ok := report.OriginTable(reportFixture, "not-found"); ok {
+	if _, ok := OriginTable(rpt, "not-found"); ok {
 		t.Errorf("unknown origin ID gave unexpected success")
 	}
-	for originID, want := range map[string]report.Table{
-		client54001EndpointNodeID: {
+	for originID, want := range map[string]Table{
+		client54001: {
```
peterbourgon: client54001 is never an origin node ID, the keys here must take the form xxxNodeID.

render/detailed_node_test.go
```diff
@@ -12,26 +12,30 @@ func TestMakeDetailedNode(t *testing.T) {
...
 			Numeric: false,
-			Rows:    []report.Row{{"Host name", clientHostName, ""}},
+			Rows: []Row{
+				{"Host name", clientHostName, ""},
+				{"PID", "10001", ""},
+				{"Process name", "curl", ""},
+			},
 		},
-		clientAddressNodeID: {
+		clientIP: {
```
peterbourgon: Likewise above

render/detailed_node_test.go
tomwilkie: Right now the tests reference renderTopology, so need to be in the same namespace.

In a future PR I've merged renderTopology into Map.Render, so the tests can be moved out.  I was trying to keep changes to a minimum in this PR.

peterbourgon: Can we do that in this PR instead? I'm happy with larger PRs if it means we avoid creating future work or technical debt.

peterbourgon:
Thanks for making it more comprehensive. LGTM!

PR #243, opened 2015-06-16
Include roboto in build

davkal:
Included so fonts work also w/o internet connection


tomwilkie:
Pls squash away the first commit, otherwise LGTM

PR #248, opened 2015-06-16
Produce containers topology from endpoints via processes

### LINKED ISSUE #228: Produce containers topology from endpoints via processes...

tomwilkie:
With this change, the process topology is used to produce the containers topology - fixes #228.

Changes:
- render.Map -> render.LeafMap (Leaf to indicate that its a tree of renderers, and these ones have to be at the leaf of that tree)
- a new render.Map, which allows for mapping from NMD -> NMDs.  Tests for this.
- add and accumulate some NodeMetadata in the RenderableNode struct, to enable the above remapping
- Some WIP mapping function for identity mappers and mappers to up-convert (ie from endpoints to processes)

The meat of the review is the new render.Map type, the mapping functions and render/topologies.go (naming suggestions?).

Left TODO:
- [x] remove some annotations on nodes from the probe
- [x] preserve pseudo node behaviour with the new mappers
- [x] filter out empty nodes from the process topology
- [x] moar testing

Note I won't be going the container image topology in this PR, so "the old way" of producing renderablenodes will be around for a little longer.


render/mapping.go
```diff
@@ -49,6 +32,88 @@ type MapFunc func(report.NodeMetadata) (RenderableNode, bool)
...
+// MapEndpointIdentity maps a endpoint topology node to endpoint RenderableNode node.
+// As it is only ever run on endpoint topology nodes, we can safely assume the
+// presences of certain keys.
+func MapEndpointIdentity(m report.NodeMetadata) (RenderableNode, bool) {
+	var (
+		id    = fmt.Sprintf("endpoint:%s:%s:%s", m[report.HostNodeID], m["addr"], m["port"])
+		major = fmt.Sprintf("%s:%s", m["addr"], m["port"])
+		minor = fmt.Sprintf("%s (%s)", m[report.HostNodeID], m["pid"])
+		rank  = m["pid"]
+	)
```
peterbourgon: You cannot safely assume the presence of "pid".

render/mapping.go
```diff
@@ -49,6 +32,88 @@ type MapFunc func(report.NodeMetadata) (RenderableNode, bool)
...
+	}
+
+	id := fmt.Sprintf("pid:%s:%s", n.NodeMetadata[report.HostNodeID], n.NodeMetadata["pid"])
+	return newDerivedNode(id, n), true
+}
+
+// MapProcess2Container maps process RenderableNodes to container RenderableNodes.
+//
+// If this function is given a node without a docker_container_id (including other
+// psueod nodes), it will produce an "Uncontained" pseudo node.
```
peterbourgon: Please fix typos (psueod) and wrap at 80 columns.

render/mapping.go
```diff
@@ -49,6 +32,88 @@ type MapFunc func(report.NodeMetadata) (RenderableNode, bool)
 // node IDs prior to mapping.
 type PseudoFunc func(srcNodeID string, srcNode RenderableNode, dstNodeID string) (RenderableNode, bool)
 
+// MapFunc is anything which can take an arbitrary RenderableNode and
+// return another RenderableNode.
+type MapFunc func(RenderableNode) (RenderableNode, bool)
```
peterbourgon: What does the second output parameter mean?

render/mapping.go
tomwilkie: Ah but you can, as this mapper is only every run on the endpoints topology, which always contains a pid.

render/mapping.go
tomwilkie: As with LeafMapFunc, allows you to not emit node (when false).  I'll extend the comment.

render/mapping.go
```diff
@@ -67,34 +132,7 @@ func ProcessPID(m report.NodeMetadata) (RenderableNode, bool) {
...
-func MapContainerIdentity(m report.NodeMetadata) (RenderableNode, bool) {
-	var id, major, minor, rank string
-	if m["docker_container_id"] == "" {
-		id, major, minor, rank = "uncontained", "Uncontained", "", "uncontained"
-	} else {
-		id, major, minor, rank = m["docker_container_id"], m["docker_container_name"], m["domain"], m["docker_image_id"]
-	}
-
-	return newRenderableNode(id, major, minor, rank), true
+	return NewRenderableNode(m["name"], m["name"], "", m["name"], m), show
```
peterbourgon: I think it's time to issue a ban on all "name" keys in NodeMetadatas, and always qualify them. In this case I think it should be "host_name", no?

peterbourgon: I think it's time to issue a ban on all "name" keys in NodeMetadatas, and always qualify them. In this case I think it should be "host_name"—is that right?

render/mapping.go
peterbourgon: It will only emit a PID in the endpoint topology if procspy is configured to do so and the host OS has the info.

render/renderable_node.go
```diff
@@ -8,14 +8,15 @@ import (
...
-	Metadata   report.AggregateMetadata `json:"metadata"`              // Numeric sums
+	ID           string                   `json:"id"`                    //
+	LabelMajor   string                   `json:"label_major"`           // e.g. "process", human-readable
+	LabelMinor   string                   `json:"label_minor,omitempty"` // e.g. "hostname", human-readable, optional
+	Rank         string                   `json:"rank"`                  // to help the layout engine
+	Pseudo       bool                     `json:"pseudo,omitempty"`      // sort-of a placeholder node, for rendering purposes
+	Adjacency    report.IDList            `json:"adjacency,omitempty"`   // Node IDs (in the same topology domain)
+	Origins      report.IDList            `json:"origins,omitempty"`     // Core node IDs that contributed information
+	Metadata     report.AggregateMetadata `json:"metadata"`              // Numeric sums
+	NodeMetadata report.NodeMetadata      `json:"-"`                     // merged NodeMetadata of the nodes used to build this
```
peterbourgon: I think it'd be good to disambiguate the two metadatas by e.g.

```go
type RenderableNode struct {
    // ...
    AggregateMetadata report.AggregateMetadata `json:"metadata"`
    NodeMetadata     report.NodeMetadata `json:"-"`
}
```

or, perhaps even better,

```
type RenderableNode struct {
    // ...
    report.AggregateMetadata `json:"metadata"`
    report.NodeMetadata      `json:"-"`
}
```

tomwilkie: That works for me.  I think AggregateMetadata is only used in render now, so its a candidate for moving.  Not in this PR.

render/mapping.go
tomwilkie: Yes!  This function is going away though, the new ones use comm.

peterbourgon:
From discussion:

- Every Leaf mapping MUST append the origin node ID to the list of OriginNodes in the destination RenderableNode. This is done [in LeafMap.Render](https://github.com/tomwilkie/scope/blob/0643507057e854f33f1343b83144a7d26208250f/render/render.go#L172)
- Identity LeafMapFuncs (mapping into the same domain, e.g. process to process) SHALL copy their NodeMetadata into the destination RenderableNode. This is done [in NewRenderableNode](https://github.com/tomwilkie/scope/blob/0643507057e854f33f1343b83144a7d26208250f/render/renderable_node.go#L72), which is always invoked by the identity LeafMapFuncs.
- Non-identity LeafMapFuncs (mapping from one domain into another, e.g. endpoint to process) MUST NOT copy their NodeMetadata into the destination RenderableNode. That information isn't necessary for building the topology, and can be retrieved later by following the OriginNode foreign keys. That's currently the case, as non-identity LeafMapFuncs always invoke [newDerivedNode](https://github.com/tomwilkie/scope/blob/0643507057e854f33f1343b83144a7d26208250f/render/renderable_node.go#L72), which doesn't touch NodeMetadata.

tomwilkie:
@peterbourgon I think this is ready to go now.  You were right about not-propagating-NMDs for derived nodes.

peterbourgon:
OK let me do another parse...

render/mapping.go
```diff
@@ -49,52 +33,132 @@ type MapFunc func(report.NodeMetadata) (RenderableNode, bool)
...
 type PseudoFunc func(srcNodeID string, srcNode RenderableNode, dstNodeID string) (RenderableNode, bool)
 
-// ProcessPID takes a node NodeMetadata from topology, and returns a
-// representation with the ID based on the process PID and the labels based on
-// the process name.
-func ProcessPID(m report.NodeMetadata) (RenderableNode, bool) {
+// MapFunc is anything which can take an arbitrary RenderableNode and
+// return another RenderableNode.
+//
+// As with LeadMapFunc, if the final output parameter is false, the node
```
peterbourgon: LeafMapFunc

render/mapping.go
```diff
@@ -49,52 +33,132 @@ type MapFunc func(report.NodeMetadata) (RenderableNode, bool)
...
+// shall be omitted from the rendered topology.
+type MapFunc func(RenderableNode) (RenderableNode, bool)
+
+// MapEndpointIdentity maps a endpoint topology node to endpoint RenderableNode node.
+// As it is only ever run on endpoint topology nodes, we can safely assume the
+// presences of certain keys.
```
peterbourgon: "presence" + wrap at 80col please

render/mapping.go
```diff
@@ -49,52 +33,132 @@ type MapFunc func(report.NodeMetadata) (RenderableNode, bool)
...
+	node := newDerivedNode(name, n)
+	node.LabelMajor = name
+	node.Rank = name
+	return node, true
+}
+
+func getHostname(m report.NodeMetadata) string {
+	hostname, _, _ := report.ParseNodeID(m[report.HostNodeID])
+	return hostname
 }
```
peterbourgon: 1. If you want to parse a host node ID, better to write report.ParseHostNodeID
2. This function is called getHostname but it's actually getHostID, they're not the same
3. If you want the hostname, you need to extract it from the host node metadata...

render/mapping.go
```diff
@@ -49,52 +33,132 @@ type MapFunc func(report.NodeMetadata) (RenderableNode, bool)
...
-// grouped into the Uncontained node.
-func MapEndpoint2Container(m report.NodeMetadata) (RenderableNode, bool) {
-	var id, major, minor, rank string
-	if m["docker_container_id"] == "" {
-		id, major, minor, rank = "uncontained", "Uncontained", "", "uncontained"
-	} else {
-		id, major, minor, rank = m["docker_container_id"], "", m["domain"], ""
+// MapContainerIdentity maps a container topology node to container
+// RenderableNode node. As it is only ever run on container topology
+// nodes, we can safely assume the presences of certain keys.
```
peterbourgon: s/the presences of/the presence of/g

render/mapping.go
```diff
@@ -49,52 +33,132 @@ type MapFunc func(report.NodeMetadata) (RenderableNode, bool)
...
-// scope them further by e.g. host. If no container metadata is found, nodes are
-// grouped into the Uncontained node.
-func MapEndpoint2Container(m report.NodeMetadata) (RenderableNode, bool) {
-	var id, major, minor, rank string
-	if m["docker_container_id"] == "" {
-		id, major, minor, rank = "uncontained", "Uncontained", "", "uncontained"
-	} else {
-		id, major, minor, rank = m["docker_container_id"], "", m["domain"], ""
+// MapContainerIdentity maps a container topology node to container
+// RenderableNode node. As it is only ever run on container topology
```
peterbourgon: "to container RenderableNode."

render/render.go
```diff
@@ -16,6 +16,31 @@ type Renderer interface {
 // other renderers
 type Reduce []Renderer
 
+// Map is a Renderer which produces a set of RendererNodes from the set of
```
peterbourgon: a set of RenderableNodes

render/render.go
```diff
@@ -16,6 +16,31 @@ type Renderer interface {
...
+// RendererNodes produces by another Renderer
```
peterbourgon: Render**able**Nodes (2 places)

render/render.go
```diff
@@ -16,6 +16,31 @@ type Renderer interface {
...
+type Map struct {
+	MapFunc
+	Renderer
+}
+
+// LeafMap is a Renderer which produces a set of RendererNodes from a report.Topology
```
peterbourgon: Render**able**Nodes

render/render.go
```diff
@@ -16,6 +16,31 @@ type Renderer interface {
...
+// by using a map functions and topology selector.
```
peterbourgon: by using a map function (singlular)

render/render.go
```diff
@@ -34,37 +59,101 @@ func (r Reduce) AggregateMetadata(rpt report.Report, localID, remoteID string) r
...
 }
 
-// Map is a Renderer which produces a set of RendererNodes by using a
-// Mapper functions and topology selector.
-type Map struct {
-	Selector report.TopologySelector
-	Mapper   MapFunc
-	Pseudo   PseudoFunc
+// Render transforms a set of RendererNodes produces by another Renderer
+// using a map function
```
peterbourgon: - Render**able**Nodes produce**d**
- End punctuation

render/render.go
```diff
@@ -34,37 +59,101 @@ func (r Reduce) AggregateMetadata(rpt report.Report, localID, remoteID string) r
...
+		}
+
+		output[outRenderable.ID] = outRenderable
+		mapped[inRenderable.ID] = outRenderable.ID
+		adjacencies[outRenderable.ID] = adjacencies[outRenderable.ID].Add(inRenderable.Adjacency...)
+	}
+
+	// Rewrite Adjacency for new node IDs.
+	// NB we don't do pseudo nodes here; we assumer the input graph
+	// we properly-connected, and if the map func dropped a node,
```
peterbourgon: **was** properly connected?

render/render.go
```diff
@@ -34,37 +59,101 @@ func (r Reduce) AggregateMetadata(rpt report.Report, localID, remoteID string) r
...
+	// we drop links to it.
+	for outNodeID, inAdjacency := range adjacencies {
+		outAdjacency := report.MakeIDList()
+		for _, inAdjacent := range inAdjacency {
+			outAdjacent, ok := mapped[inAdjacent]
+			if ok {
+				outAdjacency = outAdjacency.Add(outAdjacent)
+			}
```
peterbourgon:
```go
if outAdjacent, ok := mapped[inAdjacent]; ok {
    outAdjacency = outAdjacency.Add(outAdjacent)
}
```

render/render.go
```diff
@@ -142,16 +227,33 @@ func edgeMetadata(t report.Topology, mapFunc MapFunc, srcRenderableID, dstRender
...
+			continue
+		}
+
+		output[id] = node
+		for _, id := range node.Adjacency {
+			output[id] = input[id]
+		}
+	}
+	return output
 }
```
peterbourgon: I'm not *totally* sure what the second clause in your loop does (walking each node.Adjacency) but I think this is equivalent:

```go
output := f.Renderer.Render(rpt)
for id, node := range output {
    if len(node.Adjacency) == 0 {
        delete(output, id)
    }
}
return output
```

render/render_test.go
```diff
@@ -55,360 +48,123 @@ func TestReduceEdge(t *testing.T) {
...
-	randomHostNodeID = report.MakeHostNodeID(randomHostID)
-
-	client54001NodeID = report.MakeEndpointNodeID(clientHostID, "10.10.10.20", "54001") // curl (1)
-	client54002NodeID = report.MakeEndpointNodeID(clientHostID, "10.10.10.20", "54002") // curl (2)
-	unknownClient1    = report.MakeEndpointNodeID(serverHostID, "10.10.10.10", "54010") // we want to ensure two unknown clients, connnected
-	unknownClient2    = report.MakeEndpointNodeID(serverHostID, "10.10.10.10", "54020") // to the same server, are deduped.
-	unknownClient3    = report.MakeEndpointNodeID(serverHostID, "10.10.10.11", "54020") // Check this one isn't deduped
-	server80          = report.MakeEndpointNodeID(serverHostID, "192.168.1.1", "80")    // apache
+func TestMapRender(t *testing.T) {
+	// 1. Check when we return false, the node gets filtered out
```
peterbourgon: Since there's no shared functionality, should these be different tests?

peterbourgon:
Okay, I follow everything. After those edits, LGTM.

render/render.go
tomwilkie: This was me being defensive; AFAICT there is nothing stopping symmetric edges (ie A->B without B->A).  In that case I wanted to be sure not to drop B.

If we want to ensure symmetric edges, I'll add a check to report.Validate.

tomwilkie:
Address feedback; thanks!

there are 2 problems with the PR as it stands:
(1) containers by image is broken; I have another PR for that.  Will "fix it up" in this branch
(2) we get this weird periodic "shuffling" in the UI - I think its something to do with the process topology getting out of sync with the endpoint topology

PR #252, opened 2015-06-17
Don't calculate coverage metrics in bin/test by default.

### LINKED ISSUE #208: bin/test has got awfully slow
I think its the -race and the errcheck; if true, I propose we has a --slow that is run on circle.

tomwilkie:
Also:
- Don't go get -t by default
- Don't try and test git history

Fixes #208

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -2,31 +2,42 @@
 
 set -eu
 
-GO_TEST_ARGS="-covermode=count"
-RUN_ERRCHECK=""
+GO_TEST_ARGS=""
+SLOW=""
 if [ $# -eq 1 ] && [ "$1" = "-slow" ]; then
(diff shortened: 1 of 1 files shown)


peterbourgon:
> Don't go get -t by default

It looks like you still run `go test` by default, so `go get -t` is necessary for correctness...

tomwilkie:
We run bin/test -slow on circle, so circle will always for go get -t.

These optimisations are for users running in manually.  It will now fail and behave exactly like running go test ./... if you don't have the right test dependancies; I think thats a reasonable compromise.  If it does this, the user is expected to run bin/test -slow then continue running bin/test.

peterbourgon:
Alright. LGTM.

PR #254, opened 2015-06-17
Add ContainerImage topology and use it to build the Container By Image graph.

### LINKED ISSUE #230: container details are missing from details panel on containers by images view.
Should be fixed by #228

tomwilkie:
This makes container image details show the containers (and processes) correctly.

Fixes #230 

Also:
- introduces a 'test' package, moved Diff function there.
- adds some tests for this new rendered view.

### DIFF
--- a/render/detailed_node_test.go
+++ b/render/detailed_node_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/weaveworks/scope/render"
+	"github.com/weaveworks/scope/test"
 )
 
 func TestOriginTable(t *testing.T) {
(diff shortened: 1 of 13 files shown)


render/detailed_node_test.go
```diff
@@ -109,6 +111,6 @@ func TestMakeDetailedNode(t *testing.T) {
 		},
 	}
 	if !reflect.DeepEqual(want, have) {
-		t.Errorf("%s", diff(want, have))
+		t.Errorf("%s", test.Diff(want, have))
```
paulbellamy: Personally, not a big fan of `test` as a package name. Maybe, `diff.New(want, have)`??

tomwilkie: Yeah me neither, but a whole package for a 20 line file seems overkill.  At least with the name test, it might one day contain other things.

paulbellamy: Ah, it's only used in tests, so test is fine.

paulbellamy:
LGTM

PR #255, opened 2015-06-17
Move AggregateMetadata into render package.

tomwilkie:


render/metadata_test.go
```diff
@@ -1,61 +1,64 @@
-package report
+package render_test
```
peterbourgon: Yay

peterbourgon:
Yep! LGTM.

PR #262, opened 2015-06-19
Expose some more information on containers.

tomwilkie:
And stop using strings to index into dicts (at least, use constant strings).


tomwilkie:
I'm going to go out on a limb and not have this reviewed...

PR #264, opened 2015-06-19
Emit per-host Uncontained pseudo nodes.

### LINKED ISSUE #127: In Container view, render uncontained processes as distinct pseudonodes

tomwilkie:
Fixes #127

### DIFF
--- a/render/mapping.go
+++ b/render/mapping.go
@@ -155,7 +155,6 @@ func MapEndpoint2Process(n RenderableNode) (RenderableNode, bool) {
 
 	pid, ok := n.NodeMetadata["pid"]
 	if !ok {
-		// TODO: Propogate a pseudo node instead of dropping this?
 		return RenderableNode{}, false
 	}
 
@@ -180,11 +179,22 @@ func MapProcess2Container(n RenderableNode) (RenderableNode, bool) {
 		return n, true
 	}
 
+	// Don't propogate non-internet pseudo nodes
+	if n.Pseudo {
+		return n, false
+	}
+
 	// Otherwise, if the process is not in a container, group it
-	// into an "Uncontained" node
+	// into an per-host "Uncontained" node.  If for whatever reason
+	// this node doesn't have a host id in their nodemetadata, it'll
+	// all get grouped into a single uncontained node.
 	id, ok := n.NodeMetadata[docker.ContainerID]
-	if !ok || n.Pseudo {
-		return newDerivedPseudoNode(UncontainedID, UncontainedMajor, n), true
+	if !ok {
+		hostID := report.ExtractHostID(n.NodeMetadata)
+		id = fmt.Sprintf("%s:%s", UncontainedID, hostID)
+		node := newDerivedPseudoNode(id, UncontainedMajor, n)
+		node.LabelMinor = hostID
+		return node, true
 	}
 
 	return newDerivedNode(id, n), true
@@ -203,7 +213,6 @@ func MapProcess2Name(n RenderableNode) (RenderableNode, bool) {
 
 	name, ok := n.NodeMetadata["comm"]
 	if !ok {
-		// TODO: Propogate a pseudo node instead of dropping this?
 		return RenderableNode{}, false
 	}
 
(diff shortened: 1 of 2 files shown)


render/topologies_test.go
```diff
@@ -412,17 +412,17 @@ func TestContainerRenderer(t *testing.T) {
...
 			Origins:    report.MakeIDList(serverContainerNodeID, server80NodeID, serverProcessNodeID, serverHostNodeID),
 			AggregateMetadata: render.AggregateMetadata{
 				render.KeyBytesIngress: 150,
 				render.KeyBytesEgress:  1500,
 			},
 		},
-		render.UncontainedID: {
-			ID:                render.UncontainedID,
+		fmt.Sprintf("%s:%s", render.UncontainedID, serverHostName): {
+			ID:                fmt.Sprintf("%s:%s", render.UncontainedID, serverHostName),
```
peterbourgon: Please use an ID constructor.

render/topologies_test.go
```diff
@@ -458,17 +458,17 @@ func TestContainerImageRenderer(t *testing.T) {
...
 			Origins:    report.MakeIDList(serverContainerImageNodeID, serverContainerNodeID, server80NodeID, serverProcessNodeID, serverHostNodeID),
 			AggregateMetadata: render.AggregateMetadata{
 				render.KeyBytesIngress: 150,
 				render.KeyBytesEgress:  1500,
 			},
 		},
-		render.UncontainedID: {
-			ID:                render.UncontainedID,
+		fmt.Sprintf("%s:%s", render.UncontainedID, serverHostName): {
+			ID:                fmt.Sprintf("%s:%s", render.UncontainedID, serverHostName),
```
peterbourgon: Please use an ID constructor.

render/mapping.go
```diff
@@ -180,11 +179,20 @@ func MapProcess2Container(n RenderableNode) (RenderableNode, bool) {
...
+
 	// Otherwise, if the process is not in a container, group it
-	// into an "Uncontained" node
+	// into an per-host "Uncontained" node
 	id, ok := n.NodeMetadata[docker.ContainerID]
-	if !ok || n.Pseudo {
-		return newDerivedPseudoNode(UncontainedID, UncontainedMajor, n), true
+	if !ok {
+		hostID := report.ExtractHostID(n.NodeMetadata)
+		id = fmt.Sprintf("%s:%s", UncontainedID, hostID)
```
peterbourgon: Please use an ID constructor.

render/mapping.go
```diff
@@ -180,11 +179,20 @@ func MapProcess2Container(n RenderableNode) (RenderableNode, bool) {
...
+	}
+
 	// Otherwise, if the process is not in a container, group it
-	// into an "Uncontained" node
+	// into an per-host "Uncontained" node
 	id, ok := n.NodeMetadata[docker.ContainerID]
-	if !ok || n.Pseudo {
-		return newDerivedPseudoNode(UncontainedID, UncontainedMajor, n), true
+	if !ok {
+		hostID := report.ExtractHostID(n.NodeMetadata)
```
peterbourgon: This can fail i.e. return an empty string. What will happen in that case?

peterbourgon:
As far as I understand, this will split up the current singleton Uncontained pseudonode in both of the UI container topologies into per-host Uncontained pseudonodes. Is that correct?

tomwilkie:
That's the plan, yes

render/topologies_test.go
tomwilkie: These aren't report-level node ids; these are for rendered nodes. None of
the functions in this file use constructors for id. I agree it would be a
good thing, but I think it's beyond the scope of this PR.

What do you think? Do it this PR or the next?

render/mapping.go
tomwilkie: They'll all get grouped into a single uncontained node, if they don't have a host id in their nodemetadata.

render/topologies_test.go
peterbourgon: Ugh. I see. There is some prior art in fmt.Sprintfing renderable node IDs. Gross. OK, continue as-is, we'll address later :(

render/mapping.go
peterbourgon: OK, can you leave a comment to that effect?

peterbourgon:
LGTM after that one added comment.

PR #270, opened 2015-06-22
Duplicate endpoints in the endpoint topology to account for NAT mapping.

### LINKED ISSUE #245: Capture connections that traverse Docker port mapping
Follow on from #227, the work here is to capture the fact that a process in a container can listen to e.g. port 80, but be accessible to other processes as e.g. port 23456.

tomwilkie:
Fixes #245


probe/endpoint/nat.go
```diff
@@ -0,0 +1,131 @@
...
+
+		output = append(output, conn)
+	}
+
+	return output, nil
+}
+
+// NatIt duplicates some nodes in report.Endpoints to account for docker
+// NAT port mapping.
+func NatIt(scope string, rpt report.Report) {
```
peterbourgon: What's this function name meant to be? "NAT it"? Please no :)

peterbourgon:
`package endpoint` is about everything related to the endpoint topology? If yes, let's move other things in there too, so we don't have this in-between state.

probe/spy.go
```diff
@@ -34,6 +35,8 @@ func spy(
 		addConnection(&r, conn, hostID, hostName)
 	}
 
+	endpoint.NatIt(hostID, r)
```
peterbourgon: We should guard this invocation with a check for the `conntrack` binary in the path.

```go
var nat = func(string, report.Report) {}
if endpoint.NATAvailable() {
    nat = endpoint.NAT
}

// ...

nat(hostID, r)
```

tomwilkie:
All good points; should have put a WIP tag on this.  I was considering adapting typetypetype/conntrack for this but it looks like too much work; we can do it in the future if necessary.

probe/endpoint/nat.go
paulbellamy: +1, or even `func (r *Report) ApplyNAT(scope string)`

probe/spy.go
paulbellamy: Or at least expose the error from endpoint.natTable, and check that.

probe/endpoint/nat.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func natTable() ([]connection, error) {
+	var conntrack conntrack
+	cmd := exec.Command("conntrack", "-L", "--any-nat", "-o", "xml")
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+	if err := cmd.Start(); err != nil {
```
paulbellamy: Run, instead of Start/Wait?

probe/endpoint/nat.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil, err
+	}
+	if err := cmd.Wait(); err != nil {
+		return nil, err
+	}
+
+	output := []connection{}
+	for _, flow := range conntrack.Flows {
+		original, reply := meta{}, meta{}
+		for _, meta := range flow.Metas {
```
paulbellamy: Comment, please

probe/endpoint/nat.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+		}
+
+		output = append(output, conn)
+	}
+
+	return output, nil
+}
+
+func applyNAT(scope string, rpt report.Report) {
```
paulbellamy: I'd prefer report as the first argument (since it gets modified), but not critical.

paulbellamy:
LGTM

PR #273, opened 2015-06-23
First pass at integration tests

### LINKED ISSUE #79: Integration tests
- ./scope launch the container, do a http get
- ./scope launch on two containers with weave installed, check they can talk to each other

we could use the bash "micro" framework weave uses, and all the gce scripts to bring up vms.

paulbellamy:
Will fix #79.

Just adds a basic test for now. If people like this approach, I'll add another couple tests to this PR.

Depends on [these changes](https://github.com/weaveworks/weave/compare/master...paulbellamy:scope-tests) being merged first, into the weave repo.

The `setup.sh` script in here will `go get github.com/weaveworks/weave`, for our working copy, and uses a few of the test helpers from there, to avoid duplicating. We considered having a separate repo for the integration test helpers, but didn't want to manage another repo.

Having separate vagrant files in here is a bit messy, potentially, so am open to suggestions on that, also.

Also, need to set these up to run on circle.


tomwilkie:
- after discussion we decided we don't need the vagrant file as all prs are run through circle.

tomwilkie:
this will break scheduling as its not namespaced per repo.  Could either disable it, namespace it, or add prefix to all the tests.

integration/setup.sh
```diff
@@ -0,0 +1,16 @@
...
+. ./config.sh
+
+echo Copying scope images and scripts to hosts
+for HOST in $HOSTS; do
+    docker_on $HOST load -i ../scope.tar
+    upload_executable $HOST ../scope
+done
+
+echo Building weave
+(cd "$WEAVE_ROOT"      && make)
```
tomwilkie: lets use a release version of weave.

paulbellamy:
Won't running these on PRs have the same issue as on the weave repo (exposing gce credentials)?

tomwilkie:
It won't expose the keys if you put them in a variable in the circle ui (as opposed to checked in).  This is how we do it on weave.

And Peter & David push to branched on weaveworks/scope, so the tests will run fine there.

integration/README.md
```diff
@@ -0,0 +1,52 @@
...
+The `Vagrantfile` in this directory constructs two such VMs. To meet
+the aforementioned ssh requirement you may want to
+
+    cp ~/.vagrant.d/insecure_private_key .
+
+## Running tests
+
+    ./setup.sh
+
+Downloads a copy of weave (if not already downloaded) and calls it's
```
peterbourgon: of Scope

paulbellamy:
Just has one basic test for now. Needs more added...

Will skip integration tests on forked PRs (i.e. if there is no $SECRET_PASSWORD), but will run them on main-repo PRs.

paulbellamy:
@tomwilkie could you please review the [weave side of things](https://github.com/weaveworks/weave/pull/999), while doing this. It will need to be in a release before this will work. We could point the integration tests at weave/master until that point.

paulbellamy:
These will be broken until weave does a release.

circle.yml
```diff
@@ -2,6 +2,7 @@ general:
   branches:
     ignore:
       - gh-pages
+      - /pull\/.*/
```
tomwilkie: Thinking about it, can we have this run on pull requests please.  And some how stop the integration tests running for them.

tomwilkie:
1 comment otherwise LGTM

PR #276, opened 2015-06-23
Fix order of tables in details panel to go from most to least granular.

### LINKED ISSUE #261: Order of tables in details panel depends on details on node id formatting
We should order from least granular to most, ie docker image -> docker container -> process -> endpoints

tomwilkie:
Fixes #261

### DIFF
--- a/render/detailed_node.go
+++ b/render/detailed_node.go
@@ -3,6 +3,7 @@ package render
 import (
 	"fmt"
 	"reflect"
+	"sort"
 	"strconv"
 
 	"github.com/weaveworks/scope/probe/docker"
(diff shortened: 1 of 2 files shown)


paulbellamy:
It would be nice if the rankings were concentrated in one section of the file (or one datastructure), so it is easier to tell how they relate.

paulbellamy:
There is code to gather bits and pieces, and generate the Tables, then there is code to organise them for display. They both feel very entwined together at the moment. But possibly separating them is a bigger refactor.

paulbellamy:
But, other than that, it looks fine.

tomwilkie:
Maybe make the numbers into constants?

paulbellamy:
I was thinking something like:
```Go
var tableDefinitions := []struct{
    name string
    generator
    rank int
  }{
  {"table1", table1Func, 0},
}

for _, table :=  range tableDefinitions {
//...
}
```

But constants would be a smaller refactor, so that sounds fine.

paulbellamy:
lgtm

Issue #281, opened 2015-06-24
Details on process-by-name grouping aren't showing both processes


tomwilkie:

tomwilkie:
This was causes by newline (\n) making it into rendered node ids.  I've trimmed the newline off the process name to prevent this.

I believe that was another UI issue @davkal; looks like the UI doesn't escape newlines in generate URLs correctly.  Will leave this ticket open to discuss that.

davkal:
Why are there newlines in the first place?

tomwilkie:
Oh the file we were reading to get the process name had a newline at its end.  Thats fixed now.  But I think it unearthed a bug in the UI, so probably worth you investigating.

PR #283, opened 2015-06-24
Trim whitespace from process names

tomwilkie:
This causes detailed node lookups for the grouped-by-process-name view to fail.  Also, add a test for process walker trimmming whitespace, and a test the process-by-name view gives the right result.

Works around #281, but I believe there is an underlying UI bug where URLs aren't escaped properly.

### DIFF
--- a/app/api_topology_test.go
+++ b/app/api_topology_test.go
@@ -43,14 +43,14 @@ func TestAPITopologyApplications(t *testing.T) {
 		// Let's not unit-test the specific content of the detail tables
 	}
 	{
-		body := getRawJSON(t, ts, fmt.Sprintf("/api/topology/applications/%s/%s", expected.ClientProcessID, expected.ServerProcessID))
+		body := getRawJSON(t, ts, fmt.Sprintf("/api/topology/applications/%s/%s", expected.ClientProcess1ID, expected.ServerProcessID))
 		var edge APIEdge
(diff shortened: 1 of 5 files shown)


tomwilkie:
Paul's on holiday now.

app/api_topology_test.go
```diff
@@ -43,14 +43,14 @@ func TestAPITopologyApplications(t *testing.T) {
...
+		body := getRawJSON(t, ts, fmt.Sprintf("/api/topology/applications/%s/%s", expected.ClientProcess1ID, expected.ServerProcessID))
 		var edge APIEdge
 		if err := json.Unmarshal(body, &edge); err != nil {
 			t.Fatalf("JSON parse error: %s", err)
 		}
 		want := render.AggregateMetadata{
-			"egress_bytes":  30,
-			"ingress_bytes": 300,
+			"egress_bytes":  10,
+			"ingress_bytes": 100,
```
peterbourgon: Maybe it's still there and I'm not seeing it, but if not, can we add another test for something that would return the aggregate 30/300 bytes? Just want to make sure that functionality isn't overlooked.

peterbourgon:
Otherwise LGTM.

app/api_topology_test.go
tomwilkie: Yeah a bunch of tests in render/topologies_test.go check for that on the the process by name, container and container images topologies.

Issue #284, opened 2015-06-24
Probe is using 40% cpu


tomwilkie:
```
top - 14:48:42 up  3:50,  4 users,  load average: 1.08, 1.21, 0.93
Tasks: 120 total,   2 running, 118 sleeping,   0 stopped,   0 zombie
%Cpu(s):  7.0 us, 22.2 sy,  0.0 ni, 70.6 id,  0.0 wa,  0.2 hi,  0.0 si,  0.0 st
KiB Mem:   1017796 total,   565664 used,   452132 free,    11924 buffers
KiB Swap:  3063804 total,   176724 used,  2887080 free.   138760 cached Mem

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                                                                                                         
 6850 root      20   0   59044  54084   2800 S  38.0  5.3  64:07.44 probe                                                                                                           
 1217 root      20   0  375036   6956   1748 S   4.0  0.7   7:23.28 dragent                                                                                                         
 1262 root      20   0  748996  29240   4464 S   4.0  2.9   3:54.85 ruxitagent                                                                                                      
 1218 root      20   0 1707540  65120   7056 S   2.6  6.4   7:28.29 java                                                                                                            
  655 root      20   0 1200600  13084   3144 S   2.0  1.3  12:48.40 docker                                                                                                          
 2250 root      20   0  256652  37376  20596 S   1.7  3.7   3:57.39 weaver                                                                                                          
 2485 root      20   0   13328   2840   1540 S   1.3  0.3   2:42.29 weavedns                                                                                                        
 5644 ruxitusr  20   0  215408  70100  64848 S   1.0  6.9   0:37.53 ruxitagentnetwo                                                                                                 
    7 root      20   0       0      0      0 S   0.7  0.0   1:34.30 rcu_sched             
```

tomwilkie:
I think the cpu usage stems from walking the proc tree 3 times - once for procspy, onces for process topology, and once for docker tagger.

tomwilkie:
![prof copy](https://cloud.githubusercontent.com/assets/444037/8383186/28f56fbc-1c30-11e5-9f2a-2e6c6d9c6f89.jpg)

tomwilkie:
70% of our CPU activity is syscalls (read, listdir, stat etc)

We could reduce this by two thirds if we walk the proc tree once, gathering the required info, then let prospy, process reporter and docker tag loose on it.

[...]

PR #287, opened 2015-06-26
Cache the walk of the process tree, reusing it in docker tagger and process reporter.

tomwilkie:
- Update the cache every spy tick.
- This change make CPU usage of scope on my box go from ~40% to ~17%.

Helps towards #284

### DIFF
--- a/probe/main.go
+++ b/probe/main.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"log"
 	"net/http"
+	_ "net/http/pprof"
 	"os"
 	"os/signal"
 	"runtime"
@@ -57,7 +58,10 @@ func main() {
 			log.Printf("exposing Prometheus endpoint at %s%s", *httpListen, *prometheusEndpoint)
 			http.Handle(*prometheusEndpoint, makePrometheusHandler())
 		}
-		go func(err error) { log.Print(err) }(http.ListenAndServe(*httpListen, nil))
+		go func() {
+			err := http.ListenAndServe(*httpListen, nil)
+			log.Print(err)
+		}()
 	}
 
 	if *spyProcs && os.Getegid() != 0 {
@@ -76,7 +80,8 @@ func main() {
 	)
 
 	var (
-		weaveTagger *tag.WeaveTagger
+		weaveTagger  *tag.WeaveTagger
+		processCache *process.CachingWalker
 	)
 
 	taggers := []tag.Tagger{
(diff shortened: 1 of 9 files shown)


tomwilkie:
Profile after this change:

prof

probe/main.go
```diff
@@ -57,7 +58,10 @@ func main() {
 			log.Printf("exposing Prometheus endpoint at %s%s", *httpListen, *prometheusEndpoint)
 			http.Handle(*prometheusEndpoint, makePrometheusHandler())
 		}
-		go func(err error) { log.Print(err) }(http.ListenAndServe(*httpListen, nil))
+		go func() {
+			err := http.ListenAndServe(*httpListen, nil)
+			log.Print(err)
+		}()
```
peterbourgon: Why?

probe/process/walker.go
```diff
@@ -76,3 +78,34 @@ var Walk = func(procRoot string, f func(*Process)) error {
 
 	return nil
 }
+
+var (
+	cache     []*Process
+	cacheLock sync.RWMutex
+)
```
peterbourgon: Is there a reason these are package globals vs. e.g. `type Cache struct`?

probe/main.go
tomwilkie: "The function value and parameters are evaluated as usual in the calling goroutine" - https://golang.org/ref/spec#Go_statements

ListenAndServe blocks, so go f(ListenAndServe()) also blocks.

probe/process/walker.go
tomwilkie: No; it would make the change a little bigger (would need a Walker interface etc) and I wasn't sure it was worth it.  But I can make that change if you like.

peterbourgon: Yeah, I'm very wary of package globals.. if you don't mind, I'd appreciate it.

probe/main.go
peterbourgon: Whoa. Obvious in hindsight. How did that ever work?

tomwilkie: It hasn't since we added errcheck (and therefore logged the error returned by this).  We were also missing the pprof import.

tomwilkie:
@peterbourgon PTAL

probe/main.go
```diff
@@ -89,19 +94,25 @@ func main() {
 		endpoint.NewReporter(hostID, hostName, *spyProcs),
 	}
 
-	if *dockerEnabled && runtime.GOOS == linux {
-		if err = report.AddLocalBridge(*dockerBridge); err != nil {
-			log.Fatalf("failed to get docker bridge address: %v", err)
-		}
+	// TODO provide an alternate implementation for Darwin.
```
peterbourgon: Does this mean the current implementation won't populate PIDs and process names on Darwin?

tomwilkie: My understanding is that Darwin has no /proc FS.  So yes, this won't produce Process or Docker topologies on Darwin.  Note this PR doesn't change that behaviour - this comment isn't new.

peterbourgon: OK, I follow. We should be careful to not introduce any regressions here, in the future.

probe/process/walker.go
```diff
@@ -76,3 +91,43 @@ var Walk = func(procRoot string, f func(*Process)) error {
 
 	return nil
 }
+
+// CachingWalker is a walker than caches a copy of the output from another
+// Walker, and then allows other cocurrent readed to Walk that copy.
```
peterbourgon: other concurrent readers

peterbourgon:
That one comment aside, LGTM. Thanks!

PR #289, opened 2015-06-29
Fix host on darwin

peterbourgon:
Let's see if Circle will work...

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -37,7 +37,8 @@ test:
     - cd $SRCDIR; ./bin/lint .
     - cd $SRCDIR; make client-test
     - cd $SRCDIR; make static
-    - cd $SRCDIR; make
+    - cd $SRCDIR; rm -f app/app probe/probe; GOOS=darwin make
+    - cd $SRCDIR; rm -f app/app probe/probe; GOOS=linux make
     - cd $SRCDIR; ./bin/test -slow
     - cd $SRCDIR/experimental; make
   post:
--- a/probe/host/system_test.go
+++ b/probe/host/system_test.go
@@ -1,0 +1,38 @@
+package host_test
+
+import (
+	"strings"
+	"testing"
+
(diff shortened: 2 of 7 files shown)


circle.yml
```diff
@@ -37,7 +37,8 @@ test:
     - cd $SRCDIR; ./bin/lint .
     - cd $SRCDIR; make client-test
     - cd $SRCDIR; make static
-    - cd $SRCDIR; make
+    - cd $SRCDIR; rm -f app/app probe/probe; GOOS=linux make
+    - cd $SRCDIR; rm -f app/app probe/probe; GOOS=darwin make
```
tomwilkie: Is this going to build a darwin executable (and stick it in the docker container, and upload it to the docker hub)?

peterbourgon: Ha, good catch.

probe/host/system_test.go
```diff
@@ -0,0 +1,38 @@
+package host_test
```
tomwilkie: These tests are a bit weak; why did you remove the original unit tests (that mock syscall.Uname and Readfile?)

peterbourgon: - I didn't remove any tests, only added some
- You can't mock syscall.Uname, as it's only defined for GOOS=linux
- You can't really mock ReadFile, as the files you're reading (/proc) don't exist on Darwin, so the implementation would fail...

tomwilkie: You can have platform specific tests...

peterbourgon: Of course, but then you're just mocking and testing for a single platform. It seems like it makes more sense to move the mocking boundary "up" one level, and mock the functions that actually return cross-platform-usable data. Incidentally, that's why I added system_test.go — to make sure the underlying implementations return sane data.

tomwilkie:
LGTM with squash

Issue #286, opened 2015-06-25
Rename app & probe


tomwilkie:
Maybe something like weaveapp and weaveprobe?

App is too generic in the process view, and might collide with users apps.  Lets make it more obvious.

peterbourgon:
Agree. scope-app and scope-probe, maybe.

peterbourgon:
Done.

PR #293, opened 2015-06-29
In Docker, {app,probe} -> {scope-app,scope-probe}

peterbourgon:
A "more correct" way of doing this would be renaming the app and probe directories to scope-app and scope-probe, but that makes importing the packages under probe very ugly. I think this strikes a reasonable balance: when doing the blessed build via `make`, we produce special scope-app and scope-probe binaries, and reference them in the various Docker scripts.

LMKWYT.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,8 @@
 SUDO=sudo
 DOCKER_SQUASH=$(shell which docker-squash)
 DOCKERHUB_USER=weaveworks
-APP_EXE=app/app
-PROBE_EXE=probe/probe
+APP_EXE=app/scope-app
+PROBE_EXE=probe/scope-probe
 FIXPROBE_EXE=experimental/fixprobe/fixprobe
 SCOPE_IMAGE=$(DOCKERHUB_USER)/scope
 SCOPE_EXPORT=scope.tar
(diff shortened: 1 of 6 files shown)


peterbourgon:
Oh, yeah: addresses #286

Makefile
```diff
@@ -17,7 +17,8 @@ all: $(SCOPE_EXPORT)
 
 $(SCOPE_EXPORT): $(APP_EXE) $(PROBE_EXE) docker/*
 	@if [ -z '$(DOCKER_SQUASH)' ]; then echo "Please install docker-squash by running 'make deps'." && exit 1; fi
-	cp $(APP_EXE) $(PROBE_EXE) docker/
+	cp $(APP_EXE) docker/scope-app
+	cp $(PROBE_EXE) docker/scope-probe
```
tomwilkie: why was this bit necessary?

tomwilkie:
Is it important this goes in before release?

Makefile
peterbourgon: Oops! Artifact from an earlier revision. Fixing.

peterbourgon:
No.

tomwilkie:
LGTM; fine with approach.  I'd hold off until after release, although up to you.

peterbourgon:
Happy to hold off.

peterbourgon:
We're post-release now.

PR #294, opened 2015-06-30
Probe re-org

### LINKED ISSUE #238: Reorg probe code
Ideas:
- directory per data source (docker, process, endpoints)
- tagger interface & topology generating interface.
- break up the docker tagger into different responsibilities.

peterbourgon:
- tag/weave* -> package overlay
- tag/origin_host* -> package host
- tag/topology* -> package main

Closes #238


probe/endpoint/reporter.go
```diff
@@ -33,10 +33,11 @@ var SpyDuration = prometheus.NewSummaryVec(
...
 // NewReporter creates a new Reporter that invokes procspy.Connections to
 // generate a report.Report that contains every discovered (spied) connection
-// on the host machine, at the granularity of host and port. It optionally
-// enriches that topology with process (PID) information.
-func NewReporter(hostID, hostName string, includeProcesses bool) tag.Reporter {
-	return &reporter{
+// on the host machine, at the granularity of host and port. That information
+// is stored in the Endpoint topology. It optionally enriches that topology
+// with process (PID) information.
+func NewReporter(hostID, hostName string, includeProcesses bool) *Reporter {
```
tomwilkie: Bit of a shame not to return the interface here, but I guess you had to do it to stop circular imports?

probe/main.go
```diff
@@ -80,16 +80,16 @@ func main() {
 	)
 
 	var (
-		weaveTagger  *tag.WeaveTagger
+		weaveTagger  *overlay.WeaveTagger
```
tomwilkie: Seems like a missed opportunity to have weaveTagger implement reporter, appended onto the reporters list and the special cases removed?

tomwilkie:
Do you know why coverage has gone down so much?

probe/endpoint/reporter.go
peterbourgon: We could define Reporter and Tagger in another package, but it seems needless, since every package {host,endpoint,process,...} would have 1 Reporter and Tagger struct.

tomwilkie: Yeah I agree not worth it.

probe/main.go
peterbourgon: Good catch. Updated.

peterbourgon:
I don't! I didn't remove any tests... I can add some more tests to bump the metric back up, if you think it's important. Or we can file another issue to work on coverage specifically.

probe/main.go
```diff
@@ -80,16 +80,16 @@ func main() {
 	)
 
 	var (
-		weaveTagger  *tag.WeaveTagger
+		weave        *overlay.Weave
```
tomwilkie: This variable isn't needed outside the ```if *weaveRouterAddr != "" {``` block now, is it?

tomwilkie:
Nah the coverage isn't worth it for the sake of it, although we had just got it over 70%...

probe/main.go
peterbourgon: Well spotted! Updated.

probe/endpoint/nat.go
```diff
@@ -149,7 +149,7 @@ func applyNAT(rpt report.Report, scope string) error {
 func conntrackModulePresent() bool {
 	f, err := os.Open(modules)
 	if err != nil {
-		log.Printf("conntrack error: %v", err)
+		//log.Printf("conntrack error: %v", err) // /proc doesn't exist on Darwin
```
tomwilkie: Lets just delete this line (as opposed to commenting it out)

tomwilkie:
Looks much nicer now.  LGTM with that one comment.

Issue #154, opened 2015-06-02
Allow users to inspect connections


tomwilkie:
Some confusion has arisen from the number of active connections we should in the details panel.

We could potentially allow users to click this number, and see a table of (src, dst, time observed) connections.

tomwilkie:
I think it would be nice to, instead of showing a bunch of endpoint tables in the details panel, we should a single table of all the connections:

```
host:port -> host':port', timestamp
```

We could try and show the amount of time this connection has been open for.

paulbellamy:
afaict, on linux you only get "local" and "remote" addresses, not the directionality of the connection. So, procspy will have to check which ports/interfaces each container is listening on, and use that to infer which end is the initiator.

tomwilkie:
@paulbellamy whats left here?

paulbellamy:
Nothing! :+1:

PR #298, opened 2015-06-30
Replace "Origin Endpoint" tables with "Connection Details" table

paulbellamy:
Aggregates all connections into a single table. Only shows Local and Remote endpoints. Connection age to be added later.

Helps #154, but needs timestamps to fully fix it.

screen shot 2015-06-30 at 13 22 31

### DIFF
--- a/render/detailed_node.go
+++ b/render/detailed_node.go
@@ -2,7 +2,6 @@ package render
 
 import (
 	"fmt"
-	"reflect"
 	"sort"
 	"strconv"
 
@@ -80,15 +79,8 @@ func MakeDetailedNode(r report.Report, n RenderableNode) DetailedNode {
 	// in the UI, so we skip the intermediate representations, but we could
 	// add them later.
 	connections := []Row{}
-outer:
 	for _, id := range n.Origins {
 		if table, ok := OriginTable(r, id); ok {
-			// TODO there's a bug that yields duplicate tables. Quick fix.
-			for _, existing := range tables {
-				if reflect.DeepEqual(existing, table) {
-					continue outer
-				}
-			}
 			tables = append(tables, table)
 		} else if nmd, ok := r.Endpoint.NodeMetadatas[id]; ok {
 			connections = append(connections, connectionDetailsRows(r.Endpoint, id, nmd)...)
--- a/render/detailed_node.go
+++ b/render/detailed_node.go
@@ -79,6 +79,7 @@ func MakeDetailedNode(r report.Report, n RenderableNode) DetailedNode {
 	// multiple origins. The ultimate goal here is to generate tables to view
 	// in the UI, so we skip the intermediate representations, but we could
 	// add them later.
+	connections := []Row{}
 outer:
 	for _, id := range n.Origins {
 		if table, ok := OriginTable(r, id); ok {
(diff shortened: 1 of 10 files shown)


render/detailed_node.go
```diff
@@ -128,22 +132,31 @@ func OriginTable(r report.Report, originID string) (Table, bool) {
...
-		{"addr", "Endpoint"},
-		{"port", "Port"},
-	} {
-		if val, ok := nmd[tuple.key]; ok {
-			rows = append(rows, Row{Key: tuple.human, ValueMajor: val, ValueMinor: ""})
+	local := fmt.Sprintf("%s:%s", nmd["addr"], nmd["port"])
+	adjacencies := endpointTopology.Adjacency[report.MakeAdjacencyID(originID)]
+	sort.Strings(adjacencies)
+	for _, adj := range adjacencies {
+		segments := strings.SplitN(adj, report.ScopeDelim, 3)
```
tomwilkie: do we have a parsing function for this?  if not, make one in report/ids.go

render/detailed_node.go
```diff
@@ -79,6 +79,7 @@ func MakeDetailedNode(r report.Report, n RenderableNode) DetailedNode {
 	// multiple origins. The ultimate goal here is to generate tables to view
 	// in the UI, so we skip the intermediate representations, but we could
 	// add them later.
+	connections := []Row{}
 outer:
 	for _, id := range n.Origins {
 		if table, ok := OriginTable(r, id); ok {
```
tomwilkie: Should be able to get rid of the duplicates detection now (the only thing causing duplicates was origin endpoints)

render/detailed_node.go
```diff
@@ -128,22 +123,29 @@ func OriginTable(r report.Report, originID string) (Table, bool) {
...
-func endpointOriginTable(nmd report.NodeMetadata) (Table, bool) {
+func connectionDetailsRows(endpointTopology report.Topology, originID string, nmd report.NodeMetadata) []Row {
 	rows := []Row{}
-	for _, tuple := range []struct{ key, human string }{
-		{"addr", "Endpoint"},
-		{"port", "Port"},
-	} {
-		if val, ok := nmd[tuple.key]; ok {
-			rows = append(rows, Row{Key: tuple.human, ValueMajor: val, ValueMinor: ""})
+	local := fmt.Sprintf("%s:%s", nmd["addr"], nmd["port"])
```
peterbourgon: In keeping with the tradition of const-izing the metadata keys — is this an opportunity to do the same here?

render/detailed_node.go
```diff
@@ -128,22 +123,29 @@ func OriginTable(r report.Report, originID string) (Table, bool) {
...
+	return rows
+}
+
+func connectionDetailsTable(connectionRows []Row) Table {
 	return Table{
-		Title:   "Origin Endpoint",
+		Title:   "Connection Details",
 		Numeric: false,
-		Rows:    rows,
+		Rows:    append([]Row{{Key: "Local", ValueMajor: "Remote"}}, connectionRows...),
```
peterbourgon: Oof. I guess this will do for now, but let's file an issue to modify the Table data type to more properly support this kind of thing. (Maybe we can have it be an interface with a RenderToHTML method, something like this.)

report/id.go
```diff
@@ -103,6 +103,16 @@ func ParseNodeID(nodeID string) (hostID string, remainder string, ok bool) {
 	return fields[0], fields[1], true
 }
 
+// ParseEndpointNodeID produces the host ID, address, and port and remainder (typically an address) from
+// an endpoint node ID. Note that hostID may be blank.
```
peterbourgon: Please wrap comments at 80 col.

peterbourgon:
:+1:

render/detailed_node.go
paulbellamy: Yeah. RenderToHTML interface would be nice.

PR #299, opened 2015-07-01
Use container name (minus version) as id in containers-by-image view.

### LINKED ISSUE #291: Containers-by-image groups by image id, doesn't group images of different version together
Whilst this is strictly correct (two different versions of with the same image name are different images), I don't think this distinction is useful.  We should probably just group by image name - this grouping is intended to reduce the cardinality, and currently sometimes doesn't.

tomwilkie:
Also, build a custom URL matcher to cope with container image names having a encoded forward slash in them.

Fixes #291 and #281


app/router.go
```diff
@@ -2,12 +2,45 @@ package main
...
 	"github.com/gorilla/mux"
 
 	"github.com/weaveworks/scope/render"
 )
 
+// customMatcher uses request.RequestURI (the raw, unparsed request) to attempt
+// to match pattern.  It does this as go's URL.Parse method is broken, and
+// mistakenly unescapes the Path before parsing it.  This breaks %2F (encoded
+// forward slashes) in the paths.
+func customMatcher(pattern string) mux.MatcherFunc {
```
paulbellamy: Let's pick a more informative name for this.

app/router.go
```diff
@@ -2,12 +2,45 @@ package main
...
+
+		rm.Vars = map[string]string{}
+		for i, part := range parts {
+			unescaped, err := url.QueryUnescape(part)
+			if err != nil {
+				return false
+			}
+			match := matchParts[i]
+			if strings.HasPrefix(match, "{") && strings.HasSuffix(match, "}") {
+				rm.Vars[strings.Trim(match, "{}")] = unescaped
```
paulbellamy: The distinction between `{variable}` and `{{variable}}}}`, is rarely important.

app/router.go
tomwilkie: URLMatcher? CorrectlyEscapingURLMatcher?  The information is in the comment I guess.

paulbellamy: Is there any reason we *wouldn't* want to use this matcher for a route? Could it be HandleFunc(mux, path, handler)?

paulbellamy:
update test name, then LGTM

PR #302, opened 2015-07-02
weave expose prior to scope launch

peterbourgon:
Seems to work. Whether or not it's a good idea... maybe we want to put it on a switch? `scope launch --no-expose`?

### DIFF
--- a/scope
+++ b/scope
@@ -27,6 +27,7 @@ HOSTNAME=scope
 DOMAINNAME=weave.local
 FQDN=$HOSTNAME.$DOMAINNAME
 DOCKER_BRIDGE=${DOCKER_BRIDGE:-docker0}
+WEAVE=$(which weave)
 IP_REGEXP="[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}"
 IP_ADDR_CMD="find /sys/class/net -type l | xargs -n1 basename | grep -vE 'docker|veth|lo' | \
     xargs -n1 ip addr show | grep inet | awk '{ print \$2 }' | grep -oE '$IP_REGEXP'"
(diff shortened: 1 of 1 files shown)


scope
```diff
@@ -70,6 +71,20 @@ check_not_running() {
     esac
 }
 
+# Run `weave expose` if it's not already exposed.
+weave_expose() {
+    weave=$(which weave)
```
tomwilkie: We could update all the dns stuff to call ```weave dns-add``` etc...

scope
```diff
@@ -137,21 +152,28 @@ container_ip() {
 case "$COMMAND" in
 
 	launch)
-		check_not_running $CONTAINER_NAME $SCOPE_IMAGE
-
-		# If WeaveDNS is running, we want to automatically tell the scope
-		# image to use weave dns.  We can't use --dns with --net=host, so we have to hack it.
-		if dns_running; then
+		check_not_running $SCOPE_CONTAINER_NAME $SCOPE_IMAGE
```
tomwilkie: whitespace

peterbourgon: Ack, tabs!

scope
tomwilkie: The more I think about this, the more I think its a good idea, because (1) consistency and (2) dns is moving inside the router, with the weave script being update for that - so if we just depend on the weave script, we don't need to change anything.

peterbourgon:
@tomwilkie Give me a +1 and I'll update the DNS stuff in this PR as well.

tomwilkie:
Yes please!

peterbourgon:
@tomwilkie PTAL?

scope
```diff
@@ -70,31 +72,40 @@ check_not_running() {
...
         return
     fi
-    # get the long form of the container ID
-    CONTAINER=$(docker inspect --format='{{.Id}}' $CONTAINER_ID 2>/dev/null)
-    MORE_ARGS="--data-urlencode fqdn=$CONTAINER_FQDN"
-    for ADDR; do
-        http_call $DNS_CONTAINER_NAME $DNS_HTTP_PORT $METHOD /name/$CONTAINER/${ADDR%/*} $MORE_ARGS || true
+
+    for ip in $*; do
+        $WEAVE dns-add $ip $CONTAINER_ID -h $CONTAINER_FQDN
```
tomwilkie: What the behaviour of weave dns-add if weavedns is not running?

scope
```diff
@@ -136,48 +147,53 @@ container_ip() {
...
+        check_not_running $SCOPE_CONTAINER_NAME $SCOPE_IMAGE
+
+        # If Weave is running, we want to expose a Weave IP to the host
+        # network namespace, so Scope can use it.
+        weave_expose
+
+        # If WeaveDNS is running, we want to automatically tell the scope
+        # image to use weave dns.  We can't use --dns with --net=host, so we
+        # have to hack it.
+        if is_running $DNS_CONTAINER_NAME; then
```
tomwilkie: This is a bit fragile - the weave dns container is going away.  Is there a better way to test if weavedns is running?

peterbourgon: `weave status | grep 'weave DNS'`?

tomwilkie: I'm trying to discourage people from relying on the output of weave status.

The new weavedns code will be always-on, so I think we want to check the weave version output for >1.1, or the weavedns container being present.

peterbourgon: Then I'd propose to leave it as-is until the new weave DNS code is shipped.

tomwilkie: I'm just worried that scope launch will stop working when we ship 1.1  OTOH we probably shouldn't make assumptions about what 1.1 will look like.  And we can release a new scope version to deal with 1.1 super quick.

So yeah, not doing anything now might be best idea.

PR #307, opened 2015-07-06
Add -cpu 4 to tests.

### LINKED ISSUE #303: Add -cpu 4 flag to bin/test

tomwilkie:
Fixes #303

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -2,10 +2,10 @@
 
 set -eu
 
-GO_TEST_ARGS=""
+GO_TEST_ARGS="-cpu 4 -timeout 10s -tags netgo"
 SLOW=""
 if [ $# -eq 1 ] && [ "$1" = "-slow" ]; then
-    GO_TEST_ARGS="-race -covermode=atomic"
+    GO_TEST_ARGS="$GO_TEST_ARGS -race -covermode=atomic"
     SLOW="yes"
 fi
 
@@ -26,7 +26,7 @@ for dir in $(find . -type f -name '*_test.go' | grep -v '^./.git/'  | grep -v '^
     GO_TEST_ARGS_RUN="$GO_TEST_ARGS"
   fi
 
-  if ! go test $GO_TEST_ARGS_RUN -timeout 10s -tags netgo $dir ; then
+  if ! go test $GO_TEST_ARGS_RUN $dir ; then
     fail=1
   fi
 
--- a/probe/docker/container_test.go
+++ b/probe/docker/container_test.go
@@ -8,9 +8,12 @@ import (
 	"net/http"
 	"runtime"
 	"testing"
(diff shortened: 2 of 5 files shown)


bin/test
```diff
@@ -26,7 +26,7 @@ for dir in $(find . -type f -name '*_test.go' | grep -v '^./.git/'  | grep -v '^
     GO_TEST_ARGS_RUN="$GO_TEST_ARGS"
```
2opremio: I wonder why we have both `GO_TEST_ARGS` and `GO_TEST_ARGS_RUN`. Not directly related to your changes though.

tomwilkie: GO_TEST_ARGS_RUN contains a per-run temporary filename.

test/poll.go
```diff
@@ -0,0 +1,20 @@
...
+func Poll(t *testing.T, d time.Duration, condition func() bool, msg string) {
+	deadline := time.Now().Add(d)
+	for {
+		if time.Now().After(deadline) {
+			t.Fatal(msg)
+		}
+		if condition() {
+			return
+		}
+		time.Sleep(d / 10)
```
2opremio: Why  `d / 10`? Wouldn't it make sense to make it configurable (or a fixed amount)?

probe/docker/container_test.go
```diff
@@ -56,11 +59,10 @@ func TestContainer(t *testing.T) {
...
 		t.Error(err)
 	}
-	runtime.Gosched() // wait for StartGatheringStats goroutine to receive the stats
 
 	// Now see if we go them
-	nmd := c.GetNodeMetadata()
-	if nmd[docker.MemoryUsage] != "12345" {
-		t.Errorf("want 12345, got %s", nmd[docker.MemoryUsage])
-	}
+	test.Poll(t, 10*time.Millisecond, func() bool {
```
2opremio: What if there's a logical failure? (i.e. always != 12345). Then, the test will take 10s no matter what. I guess we don't want to optimize for failures but this can make things unpleasant when doing big refactorings.

test/poll.go
tomwilkie: I've copied existing code here; and its only a test.

probe/docker/container_test.go
tomwilkie: Again, not changed the existing code here, just move some things about.

And its 10ms, not 10s.

test/poll.go
2opremio: Oh, I just saw you simply moved the function, sorry. I think my comment still holds though.

probe/docker/container_test.go
2opremio: You are right.

2opremio:
LGTM

Issue #266, opened 2015-06-19
Scope uses more memory than expected.


tomwilkie:
~50MB.  Might be worth looking into.

peterbourgon:
Why is that more than expected?

tomwilkie:
I was expecting something around 10mb.

peterbourgon:
Why would you expect 10MB? Even a do-nothing daemon is ~5MB. (Honest questions, I promise.)

tomwilkie:
No real reasoning, I guess we should set an expectation. Looking at it
today we seem to have gone up to 150MB so I wonder if there's a leak.

Weaver and DNS both use about 10MB

tomwilkie:
Even worse:
```
Created
24 Jun 15 10:14 UTC
Command
/home/weave/entrypoint.sh --dns 172.17.42.1 --searchpath weave.local
Memory Usage (MB):
413.07
```

bboreham:
I did some diagnosis on this, although I did not have the opportunity to observe a process at ~50MB.

I started scope, built from master.  I only have a few containers running on my machine, and they are not connected to any other machines.  Linux sees the size as 18MB:

````
$ grep -i VM /proc/25014/status
VmPeak:	   18676 kB
VmSize:	   18676 kB
````

After half an hour, the size has grown very slightly:

````
$ grep -i VM /proc/25014/status
VmPeak:	   18932 kB
VmSize:	   18932 kB
````

Since scope is linked with pprof, we can view its internal stats at http://127.0.0.1:4040/debug/pprof/heap?debug=1 (scroll to the bottom):

Soon after startup:
````
# runtime.MemStats
# Alloc = 4,178,904
# TotalAlloc = 301,409,080
...
# HeapInuse = 5,193,728
# HeapReleased = 0
...
# NextGC = 5,059,840
````

(I inserted the commas to make the numbers easier to read)

`Alloc` is how much memory is allocated now; `TotalAlloc` is how much memory has been allocated since the program started; the difference between these numbers is the amount of garbage collected.

When `Alloc` hits the level of `NextGC`, the next garbage-collect will run.

After half an hour we see:

````
# Alloc = 4,984,624
# TotalAlloc = 2,563,262,288
...
# HeapInuse = 6,356,992
# HeapReleased = 802,816
...
# NextGC = 5,633,072
````

[...]

PR #309, opened 2015-07-08
Reduce garbage generation with smarter IDList merge

bboreham:
Arising from #266 (but not claiming to fix it).

The previous implementation, when adding M elements to N, would do O(M * NlogN) compare operations, O(M * N) copies to move the data around and M*2 heap allocations.

This merge implementation should do O(M+N) compares, M+N copies and 1 allocation.

A couple of ancillary things were required to keep tests passing: the use of `MakeIDList` in [`TestMergeAdjacency`](https://github.com/weaveworks/scope/blob/master/report/merge_test.go#L85) created an `IDList` that broke its own invariant, so I made `MakeIDList` enforce it.  It is not clear whether this feature is required for the real code or just an artefact of the test.  Also the early-out on empty `IDList` - otherwise `DeepEquals` complains that an empty slice is not the same as a nil slice.

The third commit, 35f9dc6, is very low-impact compared to the above, but provides minor reduction in allocation.

### DIFF
--- a/render/render.go
+++ b/render/render.go
@@ -85,7 +85,7 @@ func (m Map) render(rpt report.Report) (RenderableNodes, map[string]string) {
 
 		output[outRenderable.ID] = outRenderable
 		mapped[inRenderable.ID] = outRenderable.ID
-		adjacencies[outRenderable.ID] = adjacencies[outRenderable.ID].Add(inRenderable.Adjacency...)
+		adjacencies[outRenderable.ID] = adjacencies[outRenderable.ID].Merge(inRenderable.Adjacency)
 	}
 
 	// Rewrite Adjacency for new node IDs.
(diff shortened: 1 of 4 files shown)


tomwilkie:
One small comment, other a very nice change.  Thanks!

bboreham:
You've reminded me that that MakeIDList implementation is rather inefficient if the supplied list has a lot of duplicates.  It appears that, outside of tests, it is only called with 0 or 1 argument.

bboreham:
Modified `MakeIDList` as suggested, rebased and `push -f`'d

bboreham:
Oops, sorry, my push had actually stalled.  Shall I do another PR?

tomwilkie:
Whoops.  My bad.  Yes please.

PR #317, opened 2015-07-14
gopacket sniffer

peterbourgon:
This PR adds packet sniffing to the probe's bag of tricks. Here's how it works:

- We sample via a duty cycle, e.g. 1s on, 5s off.
- When off, we still count (but don't decode) packets, this is necessary to compute a proper sample rate.
- Currently gopacket/pcap only, could think about adding e.g. pfring.

Misc notes from the implementation:

- This removes the AggregateMetadata type, as (it turns out) we can merge EdgeMetadatas directly just fine.
- ~~There is no real concept of bytes ingress/egress on an edge, as an edge is directed. So those separate values have been removed. If there will be a concept of ingress/egress, it should be derived as part of the rendering process.~~
- We can't cross-compile anymore :( so I had to remove the Darwin test from Circle.

To review, start with [the Sniffer](https://github.com/weaveworks/scope/pull/317/files#diff-12b45ef81ad4adaef7515ba523884796) and work your way out, being sure to look at how it's wired up [in probe.main](https://github.com/weaveworks/scope/pull/317/files#diff-3b8e458fa59e2dc53bde306a3e21b6e9). (It's modeled as a Reporter.)

We still need to wire this into the (revised) details pane, but I wanted to get this work reviewed and merged first.


peterbourgon:
@tomwilkie PTAL and let me know what you think! Maybe also interesting to @paulbellamy

probe/endpoint/reporter.go
```diff
@@ -71,34 +71,34 @@ func (r *Reporter) Report() (report.Report, error) {
 
 func (r *Reporter) addConnection(rpt *report.Report, c *procspy.Connection) {
 	var (
-		scopedLocal  = report.MakeAddressNodeID(r.hostID, c.LocalAddress.String())
-		scopedRemote = report.MakeAddressNodeID(r.hostID, c.RemoteAddress.String())
-		key          = report.MakeAdjacencyID(scopedLocal)
-		edgeKey      = report.MakeEdgeID(scopedLocal, scopedRemote)
+		localAddressNodeID  = report.MakeAddressNodeID(r.hostID, c.LocalAddress.String())
+		remoteAddressNodeID = report.MakeAddressNodeID(r.hostID, c.RemoteAddress.String())
+		adjecencyID         = report.MakeAdjacencyID(localAddressNodeID)
```
paulbellamy: s/adjecencyID/adjacencyID

paulbellamy:
I'm not quite seeing why EdgeMetadata.* all have to be pointers? Is it so they can be nil, or for mutability?

peterbourgon:
They need maybe semantics, because the zero value may be meaningful.

peterbourgon:
I'm seeing very weird behavior when testing the details pane. Still tracking down some bugs.

probe/sniff/sniffer.go
```diff
@@ -0,0 +1,247 @@
...
+
+	go func() {
+		s.read(src, packets, &process, &total, &count)
+		close(done)
+	}()
+
+	for {
+		select {
+		case p := <-packets:
+			s.Merge(p, rpt)
```
paulbellamy: If this is slow it'll block packet flow, correct?

probe/sniff/sniffer.go
```diff
@@ -0,0 +1,247 @@
...
+		case <-done:
+			return
+		}
+	}
+}
+
+// applySampling compensates for sampling by artifically inflating counts
+// throughout the report. It should be run once, by the probe, before emitting
+// the report upstream.
+func applySampling(r report.Report) {
```
paulbellamy: maybe "interpolateSamples" is more descriptive of the intent?

probe/sniff/sniffer.go
peterbourgon: In the on state, yes. Good catch. I think it'd be better to do a nonblocking send to the packets chan, what do you think?

probe/sniff/sniffer.go
peterbourgon: +1

probe/sniff/sniffer.go
paulbellamy: Sounds good.

Makefile
```diff
@@ -26,7 +26,7 @@ $(APP_EXE): app/*.go render/*.go report/*.go xfer/*.go
 $(PROBE_EXE): probe/*.go probe/docker/*.go probe/endpoint/*.go probe/host/*.go probe/process/*.go probe/overlay/*.go report/*.go xfer/*.go
 
 $(APP_EXE) $(PROBE_EXE):
-	go get -tags netgo ./$(@D)
+	go get -d -tags netgo ./$(@D)
```
tomwilkie: -d ?

experimental/demoprobe/generate.go
```diff
@@ -123,3 +121,5 @@ func DemoReport(nodeCount int) report.Report {
 
 	return r
 }
+
+func newu64(value uint64) *uint64 { return &value }
```
tomwilkie: Not particularly keep on this - len(newu64) > len(&)

probe/endpoint/reporter.go
```diff
@@ -71,34 +71,34 @@ func (r *Reporter) Report() (report.Report, error) {
 
 func (r *Reporter) addConnection(rpt *report.Report, c *procspy.Connection) {
 	var (
-		scopedLocal  = report.MakeAddressNodeID(r.hostID, c.LocalAddress.String())
-		scopedRemote = report.MakeAddressNodeID(r.hostID, c.RemoteAddress.String())
-		key          = report.MakeAdjacencyID(scopedLocal)
-		edgeKey      = report.MakeEdgeID(scopedLocal, scopedRemote)
+		localAddressNodeID  = report.MakeAddressNodeID(r.hostID, c.LocalAddress.String())
```
tomwilkie: Looks like these renamings are orthogonal to the meat of this change and could be put in separately.

probe/main.go
```diff
@@ -153,3 +176,16 @@ func interrupt() chan os.Signal {
 	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
 	return c
 }
+
+func interfaces() string {
```
tomwilkie: I'm not sure capturing on all interfaces is a good idea.  Will this capture on the docker bridge and eth0?  Will packets get captured twice?

probe/process/walker_test.go
```diff
@@ -8,6 +8,16 @@ import (
 	"github.com/weaveworks/scope/test"
 )
 
+func TestBasicWalk(t *testing.T) {
```
tomwilkie: Seems unrelated to this PR...

experimental/demoprobe/generate.go
peterbourgon: This is unfortunately a function of necessity: there's no way to take the address of a primitive type.

probe/main.go
peterbourgon: I think it is a sensible default, or at least I can't think of a way to produce a better default. Would you suggest something like only take `{en,eth}[0-9]*` as a default? Note that for the Docker entrypoint, we can change it via commandline flag...

tomwilkie: just taking eth* will missing inter-container traffic on this host.  Your probably better of just taking docker0 and weave bridges.  I'd talk to someone who knows more though.

probe/sniff/sniffer.go
```diff
@@ -0,0 +1,247 @@
...
+		data, _, err = src.ZeroCopyReadPacketData()
+		if err == io.EOF {
+			return // done
+		}
+		if err != nil {
+			log.Printf("sniffer: read: %v", err)
+			continue
+		}
+		atomic.AddUint64(total, 1)
+		if atomic.LoadUint64(process) == 0 {
```
tomwilkie: If I read this correctly, we're always capturing all packets, and only decoding them if we're 'active'?

probe/sniff/sniffer.go
```diff
@@ -0,0 +1,247 @@
...
+
+			case layers.LayerTypeTCP:
+				p.SrcPort = strconv.Itoa(int(s.tcp.SrcPort))
+				p.DstPort = strconv.Itoa(int(s.tcp.DstPort))
+				p.Transport += len(s.tcp.Payload)
+
+			case layers.LayerTypeUDP:
+				p.SrcPort = strconv.Itoa(int(s.udp.SrcPort))
+				p.DstPort = strconv.Itoa(int(s.udp.DstPort))
+				p.Transport += len(s.udp.Payload)
```
tomwilkie: For performance you might want to consider not converting ips & ports to strings until you have to produce the report - ie keep them as []byte and ints, merger based on that, and convert at the last step.

experimental/demoprobe/generate.go
tomwilkie: :-(

tomwilkie:
Whats the performance impact of this change?  Its not enabled by default - will we ever be able to do this? 

I've voiced concerns previously that we should get this kind of information (bandwidth) from the router / FDP and not through scope packet capture.

probe/sniff/sniffer.go
peterbourgon: That's correct.

probe/sniff/sniffer.go
peterbourgon: I'll bookmark that as one thing we can try if we hit a performance wall.

probe/sniff/sniffer.go
peterbourgon: MakeReport has to block packet flow, as the loop is a synchronization point for mutations on the report. I'm also comfortable with throwing away any buffered packets on teardown.

Makefile
peterbourgon: Please correct me if I'm wrong, but as I understand it, the `go get` step here has the purpose of fetching the dependencies for the component. But without -d, it also builds and installs the component itself to $GOPATH/bin — needless work that we duplicate on the line below.

peterbourgon:
@tomwilkie In my performance tests, adding packet capture at 100% sample rate (-capture=true -capture.off=0) takes weighted moving average CPU usage from ~9% to ~11%. I don't think this is significant, but I'm probably not stressing the probe with enough traffic. It would be great to get more data here. Would you be willing to try it on your instances?

Sadly, decreasing the sample rate through the duty-cycle seems to have no effect, or, in the worst case, _increase_ CPU usage. With -capture.on=3s -capture.off=3s (50% effective sample rate) I see CPU usage at ~15%. With -capture.on=1s -capture.off=5s I see CPU usage at ~12%. Two strategies to make this better:

1. Abandon the duty-cycle approach, make a decision with each packet
1. Try pfring and/or netfilter, both provided in gopacket

The first would make the code cleaner (and sample rates more accurate), but I think only the second stands a real chance of improving performance. That said, I don't think it's a pressing concern.

paulbellamy:
Looks fine, but @tomwilkie knows more about the performance characteristics of this.
In general, any reason to use pcap instead of pf_ring? pcap is lossless, but that doesn't really bother us here...

tomwilkie:
Sorry for delay, been beavering away on SAAS.

re: performance; I don't think customer will be too interested in those figures, I'm afraid.  They'll want an answer the the question "If I run this on my machine, how much does it slow down my application?"

To measure this you could setup a pair of GCE VMs, run iperf between them, measure bandwidth, and then repeat with this enabled.  Right now I imagine the performance impact isn't huge (as weave is already doing pcap, so the top line figure is low).  I would measure without weave to get a better idea.  I worry once we introduce FDP, running scope with packet capture will be like reverting to weave legacy performance (We're taking 10's MB/s vs 100's MB/s).

Given the above, I can't see how this approach will every be satisfactorily turned on by default.  If we want bandwidth info, I'd rather bake the stats gathering into FDP, where it can be done in kernel with minimal overhead.

peterbourgon:
To be clear the performance overhead of turning on packet capture in this branch is 1%. Is that what you're calling unsatisfactory?

tomwilkie:
How are you measuring the that? 1% of what?

peterbourgon:
> Adding packet capture at 100% sample rate … takes … CPU usage from ~9% to ~11%

tomwilkie:
"re: performance; I don't think customer will be too interested in those figures, I'm afraid. They'll want an answer the the question "If I run this on my machine, how much does it slow down my application?""

I don't think users care how much CPU scope uses; they care what the effect is on their application.  Claiming 1% performance overhead could be misleading - they might thing an app serving 1k QPS will only lose 10 qps, whereas I suspect the reality is very different.

peterbourgon:
I'm not sure how to respond to that. It's a lot of hypotheticals. Are you :+1: to merge this PR, do you prefer I close it without merging, or is there a third option?

peterbourgon:
Alright, I'll merge this in the present disabled state, just so it doesn't rot. We'll leave it on ice until we decide what to do with it. We can rip it out later if it turns out we're not going to follow through with packet capture.

Issue #315, opened 2015-07-13
Add controls


tomwilkie:
Want to have the ability for the UI to send actions all the way back to the probe.  Example is the "quarantine" action.

Heres a rough design / plan:

In the report data types, do the follow:
- [x] <s>Change NodeMetadata to a struct, existing metadata will live in ndm.metadata field.</s> (Done in #318)
- [x] Add "controls" field to this struct; plan for just a []string of control identifiers
- [x] Add a "controls" field to top-level Topology struct, containing the i18n & parameter info

In the App / Client:
- [x] Render all this out into the detailed node struct, using the i18n data from the topologies
- [x] Have the UI stick this in the details panel (any maybe a right click menu?)
- [x] Add new POST `/api/topology/{topology}/{id}` handler, which will use the origin host node metadata info to route the action back to the right probe
- [x] Add a new websocket handler at `/api/controls`, which the probe will connect a websocket to, to allow the app to send and receive async messages to/from the probe

In the Probe:
- [x] Connect to the app, handle control requests and responses
- [x] In probe, have a map of control id -> function, have various modules register in this map, and when we get a action just lookup and run said function
- <s>Have actions travel the other way down the app<->probe connection; reuse existing connection.</s>  (now that we post the reports to the app, this makes little sense)

Future
- <s>Have WeaveTagger add quarantine control to the Container topology</s> (this will have to be in some eventual weave run integration)
- [x] Have Docker Reporter add stop / restart controls to the Container topology

Open questions:
- Do we want to support actions on nodes that are merges from multiple probes (eg container images?)  I think no, for now
- Do we want actions to have responses? yes.
- Do we want actions to have parameters? yes

peterbourgon:
> Add new /api/topology/{topology}/{id}/{action}

The semantics of this are interesting to me... it would need to be a POST, and support parameters, but I'm not sure the best way to do that — JSON body? Arbitrary URL params? Also, I think it needs to be synchronous, waiting for a response from the probe, which has interesting consequences on package xfer...

> Do we want to support actions on nodes that are merges from multiple probes (eg container images?) I think no, for now

I don't see the problem, you just need to select which origin host you want the action to apply to, in the detail pane. Maybe that's not UI we care to build out in a first cut.

> Do we want actions to have responses? I think yes.

Agree.

> Do we want actions to have parameters? I think no for now.

I don't think we can do anything useful without parameters. What did you have in mind?

tomwilkie:
[...]

PR #318, opened 2015-07-15
Make NodeMetadata a struct, move existing map[string]string to NodeMetadata.Metadata

tomwilkie:
Part 1 of #315

### DIFF
--- a/report/topology.go
+++ b/report/topology.go
@@ -41,28 +41,28 @@ type EdgeMetadata struct {
 }
 
 // NodeMetadata describes a superset of the metadata that probes can collect
-// about a given node in a given topology. Right now it's a weakly-typed map,
-// which should probably change (see comment on type MapFunc).
-type NodeMetadata map[string]string
+// about a given node in a given topology.
+type NodeMetadata struct {
+	Metadata
+}
+
+// Metadata is shorthand for a string:string map.
+type Metadata map[string]string
+
+// NewNodeMetadata creates a new NodeMetadata with the supplied Metadata.
+func NewNodeMetadata(m Metadata) NodeMetadata {
+	return NodeMetadata{
+		Metadata: m,
+	}
+}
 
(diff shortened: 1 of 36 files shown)


report/topology.go
```diff
@@ -41,28 +41,28 @@ type EdgeMetadata struct {
...
+	Metadata
+}
 
-// Copy returns a value copy, useful for tests.
-func (nm NodeMetadata) Copy() NodeMetadata {
-	cp := make(NodeMetadata, len(nm))
-	for k, v := range nm {
-		cp[k] = v
+// Metadata is shorthand for a string:string map.
+type Metadata map[string]string
```
peterbourgon: You really only want to define a new type to implement some value-add behavior via a method. I'd prefer to remove this definition and use map[string]string instead.

peterbourgon:
Can we continue adding commits to this PR until it represents the complete feature we're driving towards, and then merge it?

report/topology.go
tomwilkie: I originally did that but having map[string]string everywhere was messy.  Metadata is much easier to type.

tomwilkie:
I'd rather not; I think the change will end up being too big.  As it stands it represents a nice, self contained and self consistent change.

report/topology.go
tomwilkie: report.Metadata is longer than map[string]string OTOH

peterbourgon: > having map[string]string everywhere was messy

It requires one less mental dereference, which is something we should (and Go does) aim to optimize for.

tomwilkie: Yeah I've come round to changing them all to map[string]string now.

peterbourgon:
As per discussion, I would rather PRs encapsulate issues (optimize for completeness) rather than checkpoints (optimize for size). But as a checkpoint, LGTM.

PR #320, opened 2015-07-16
Naïve process walker for Darwin

peterbourgon:
This fixes the regression where process names weren't appearing for Darwin probes. Makes testing easier.

### DIFF
--- a/probe/main.go
+++ b/probe/main.go
@@ -7,7 +7,6 @@ import (
 	_ "net/http/pprof"
 	"os"
 	"os/signal"
-	"runtime"
 	"strconv"
 	"syscall"
 	"time"
@@ -24,8 +23,6 @@ import (
 
 var version = "dev" // set at build time
 
-const linux = "linux" // runtime.GOOS
-
 func main() {
 	var (
 		httpListen         = flag.String("http.listen", "", "listen address for HTTP profiling and instrumentation server")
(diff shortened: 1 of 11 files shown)


probe/main.go
```diff
@@ -83,24 +83,22 @@ func main() {
 	)
 
 	// TODO provide an alternate implementation for Darwin.
```
tomwilkie: Can this todo go away now?

probe/main.go
```diff
@@ -83,24 +83,22 @@ func main() {
...
-			if err = report.AddLocalBridge(*dockerBridge); err != nil {
-				log.Fatalf("failed to get docker bridge address: %v", err)
-			}
-
-			dockerRegistry, err := docker.NewRegistry(*dockerInterval)
-			if err != nil {
-				log.Fatalf("failed to start docker registry: %v", err)
-			}
-			defer dockerRegistry.Stop()
+	if runtime.GOOS == linux && *dockerEnabled {
```
tomwilkie: Alternatively, we could default dockerEnabled to false, and set it in the docker init scripts.  Then we could remove this if.

probe/main.go
```diff
@@ -83,24 +83,22 @@ func main() {
...
+		if err = report.AddLocalBridge(*dockerBridge); err != nil {
```
tomwilkie: should probably be :=

probe/main.go
tomwilkie: Also, the if processCache != nil line below can go too (I can't seem to comment on that line, sorry)

probe/process/walker_darwin.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (walker) Walk(f func(*Process)) error {
+	// Copied from procspy spy_darwin
+	out, err := exec.Command(
+		lsofBinary,
+		"-i",       // only Internet files
+		"-n", "-P", // no number resolving
+		"-w",             // no warnings
+		"-F", lsofFields, // \n based output of only the fields we want.
+	).CombinedOutput()
```
tomwilkie: :-( thats a bit icky.  Only other way I could find involves cgo:

https://github.com/jondot/gosigar/blob/master/sigar_darwin.go#L195

probe/main.go
peterbourgon: I don't understand what you mean by "docker init scripts"?

peterbourgon:
OK, I think I addressed your feedback...

probe/main.go
tomwilkie: Sorry wasn't clear.  Things like entrypoint.sh or even our scope script.  I'm assuming you don't use them when launching on Darwin.

peterbourgon: Ah, yeah, but we'd still need to check the flag here, no?

tomwilkie: Yeah.  Not a big deal.  Just seems weird that we wouldn't run the docker integration and we won't error even if explicitly run with --docker=true

peterbourgon: Yep, I agree with you. Updated.

peterbourgon:
Good?

tomwilkie:
Yeah LGTM.  Please squash, and hold until David's done the release.

peterbourgon:
Of course.

PR #325, opened 2015-07-16
Detect WeaveDNS by an open port 53 on the docker bridge && the presence of the weave script.

### LINKED ISSUE #321: Scope doesn't work with master builds of weave anymore
Due to gossip dns being merged into the router.

tomwilkie:
We can't rely on the presence of the WeaveDNS container.

Fixes #321

### DIFF
--- a/scope
+++ b/scope
@@ -21,8 +21,6 @@ IMAGE_VERSION=${VERSION:-$IMAGE_VERSION}
 SCOPE_IMAGE=weaveworks/scope:$IMAGE_VERSION
 SCOPE_CONTAINER_NAME=weavescope
 WEAVE_CONTAINER_NAME=weave
-DNS_CONTAINER_NAME=weavedns
-DNS_HTTP_PORT=6785
 HOSTNAME=scope
 DOMAINNAME=weave.local
 FQDN=$HOSTNAME.$DOMAINNAME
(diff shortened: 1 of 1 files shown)


peterbourgon:
The code looks fine, but is this methodology really sound?

tomwilkie:
Best I could think of.  I'm open to other suggestions if you have some.  @rade what do you think?

rade:
These days one can't really turn off weavedns, can one? So perhaps just check that weave is running?

peterbourgon:
If Weave is running, is it possible, without playing tricks, to have something other than then Weave DNS listening on the docker bridge IP port 53?

tomwilkie:
Its very easy in Weave 1.0 to have a non functional weavedns (perhaps you didn't launch it, or it failed to get an address from ipam).  I'd like scope to work in these situations, so in my head this rules out just relying on weave being there.

And I'd like scope to work with master, so this rules out checking for the weavedns container.

I could have scope look at the weave version - 1.0 or less looks for the weavedns container, master or >1.0 just assumes its running.  But this seems easier.

@peterbourgon yes other people could put things on port 52 of the weave bridge; its relatively straight forward.

scope
```diff
@@ -163,7 +165,7 @@ case "$COMMAND" in
             $WEAVESCOPE_DOCKER_ARGS $SCOPE_IMAGE $WEAVESCOPE_DNS_ARGS "$@")
 
         IP_ADDRS=$(docker run --net=host gliderlabs/alpine /bin/sh -c "$IP_ADDR_CMD")
-        if is_running $DNS_CONTAINER_NAME && is_running $WEAVE_CONTAINER_NAME && command_exists weave; then
+        if command_exists weave && weave_dns_present; then
```
peterbourgon: Re: conversation in the issue, how about

```bash
if command_exists weave && is_running $WEAVE_CONTAINER_NAME && weave_dns_present; then
```

PR #333, opened 2015-07-30
Fix some variable names

peterbourgon:
Also, don't use docker.Addr as a synonym for 'addr' string literal. 

This PR extracted from the sniff branch.

### DIFF
--- a/probe/docker/tagger.go
+++ b/probe/docker/tagger.go
@@ -7,15 +7,11 @@ import (
 	"github.com/weaveworks/scope/report"
 )
 
-// These constants are keys used in node metadata
-// TODO: use these constants in report/{mapping.go, detailed_node.go} - pending some circular references
+// Node metadata keys.
 const (
-	Addr        = "addr"
 	ContainerID = "docker_container_id"
-	Domain      = "domain"
-	Name        = "name"
-	PID         = "pid"
-	Port        = "port"
+	Domain      = "domain" // TODO this is ambiguous, be more specific
+	Name        = "name"   // TODO this is ambiguous, be more specific
 )
 
 // These vars are exported for testing.
@@ -50,7 +46,7 @@ func (t *Tagger) Tag(r report.Report) (report.Report, error) {
 
 func (t *Tagger) tag(tree process.Tree, topology *report.Topology) {
 	for nodeID, nodeMetadata := range topology.NodeMetadatas {
-		pidStr, ok := nodeMetadata.Metadata["pid"]
+		pidStr, ok := nodeMetadata.Metadata[process.PID]
 		if !ok {
 			continue
 		}
--- a/probe/endpoint/reporter.go
+++ b/probe/endpoint/reporter.go
@@ -8,8 +8,14 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/weaveworks/procspy"
-	"github.com/weaveworks/scope/probe/docker"
+	"github.com/weaveworks/scope/probe/process"
 	"github.com/weaveworks/scope/report"
+)
+
+// Node metadata keys.
+const (
+	Addr = "addr" // typically IPv4
+	Port = "port"
 )
 
 // Reporter generates Reports containing the Endpoint topology.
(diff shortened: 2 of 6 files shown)


probe/endpoint/reporter.go
```diff
@@ -71,34 +70,34 @@ func (r *Reporter) Report() (report.Report, error) {
...
+		edgeID              = report.MakeEdgeID(localAddressNodeID, remoteAddressNodeID)
 	)
 
-	rpt.Address.Adjacency[key] = rpt.Address.Adjacency[key].Add(scopedRemote)
+	rpt.Address.Adjacency[adjacencyID] = rpt.Address.Adjacency[adjacencyID].Add(remoteAddressNodeID)
 
-	if _, ok := rpt.Address.NodeMetadatas[scopedLocal]; !ok {
-		rpt.Address.NodeMetadatas[scopedLocal] = report.NewNodeMetadata(map[string]string{
-			docker.Name: r.hostName,
-			docker.Addr: c.LocalAddress.String(),
```
tomwilkie: Can we at least stick a const for Name and Addr at the top of the file?  And use these in detail_node.go:addressOriginTable()?

tomwilkie: And connectionDetailsRows

peterbourgon: It actually doesn't make sense to define a constant for the "addr" literal, because we've been imprecise in how we [ab]use it throughout the code. We use it as a key to read, without knowing which component actually used it to write.

```fish
ugh ~/src/github.com/weaveworks/scope (consistent-variable-naming) ack '"addr"'
experimental/example/searchapp/app.go
22:             addr = flag.String("addr", ":8080", "HTTP listen address")

experimental/example/shout/shout.go
22:             addr = flag.String("addr", ":8090", "HTTP listen address")

probe/docker/tagger.go
13:     Addr        = "addr"

probe/endpoint/reporter.go
84:                     "addr": c.LocalAddress.String(),
103:                            "addr": c.LocalAddress.String(),

render/detailed_node.go
128:    local := fmt.Sprintf("%s:%s", nmd.Metadata["addr"], nmd.Metadata["port"])
153:    if val, ok := nmd.Metadata["addr"]; ok {

render/mapping.go
52:             id      = MakeEndpointID(report.ExtractHostID(m), m.Metadata["addr"], m.Metadata["port"])
53:             major   = fmt.Sprintf("%s:%s", m.Metadata["addr"], m.Metadata["port"])
112:            id    = MakeAddressID(report.ExtractHostID(m), m.Metadata["addr"])
113:            major = m.Metadata["addr"]

test/report_fixture.go
89:                                     "addr":            ClientIP,
95:                                     "addr":            ClientIP,
101:                                    "addr":            ServerIP,
213:                                    "addr":            ClientIP,
217:                                    "addr":            ServerIP,
```

tomwilkie: Ignoring flags and tests, all those instance of "addr" are for either the Address or Endpoint topology.  So I think it makes sense to put a constant there.

peterbourgon: OK, then I guess it makes the most sense to put

```go
const Addr = "addr"
```

in package address, and 

```go
const Port = "port"
```

in package endpoint. Will make that change now.

peterbourgon: Ah, there is no package address, because of how we've overloaded the duties of the endpoint reporter and tagger. Both constants in endpoint, then.

peterbourgon:
@tomwilkie How about that?

probe/docker/tagger.go
```diff
@@ -50,7 +46,7 @@ func (t *Tagger) Tag(r report.Report) (report.Report, error) {
 
 func (t *Tagger) tag(tree process.Tree, topology *report.Topology) {
 	for nodeID, nodeMetadata := range topology.NodeMetadatas {
-		pidStr, ok := nodeMetadata.Metadata["pid"]
+		pidStr, ok := nodeMetadata.Metadata[process.PID]
```
tomwilkie: :-)

tomwilkie:
Much nicer.  LGTM

PR #335, opened 2015-07-30
Fix new node metadata

peterbourgon:
Fix a couple of dangling issues with the NewNodeMetadata constructor pattern from a ~week ago.

Extracted from an upcoming branch.

### DIFF
--- a/render/expected/expected.go
+++ b/render/expected/expected.go
@@ -17,18 +17,21 @@ var (
 		ID:                unknownPseudoNode1ID,
 		LabelMajor:        "10.10.10.10",
 		Pseudo:            true,
+		NodeMetadata:      report.MakeNodeMetadata(),
 		AggregateMetadata: render.AggregateMetadata{},
 	}
 	unknownPseudoNode2 = render.RenderableNode{
 		ID:                unknownPseudoNode2ID,
 		LabelMajor:        "10.10.10.11",
 		Pseudo:            true,
+		NodeMetadata:      report.MakeNodeMetadata(),
 		AggregateMetadata: render.AggregateMetadata{},
 	}
 	theInternetNode = render.RenderableNode{
 		ID:                render.TheInternetID,
 		LabelMajor:        render.TheInternetMajor,
 		Pseudo:            true,
+		NodeMetadata:      report.MakeNodeMetadata(),
 		AggregateMetadata: render.AggregateMetadata{},
 	}
 
@@ -50,6 +53,7 @@ var (
 				test.ClientProcess1NodeID,
 				test.ClientHostNodeID,
 			),
+			NodeMetadata: report.MakeNodeMetadata(),
 			AggregateMetadata: render.AggregateMetadata{
 				render.KeyBytesIngress: 100,
 				render.KeyBytesEgress:  10,
@@ -67,6 +71,7 @@ var (
 				test.ClientProcess2NodeID,
 				test.ClientHostNodeID,
 			),
+			NodeMetadata: report.MakeNodeMetadata(),
 			AggregateMetadata: render.AggregateMetadata{
 				render.KeyBytesIngress: 200,
 				render.KeyBytesEgress:  20,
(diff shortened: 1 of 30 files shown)


render/expected/expected.go
```diff
@@ -17,18 +17,21 @@ var (
 		ID:                unknownPseudoNode1ID,
 		LabelMajor:        "10.10.10.10",
 		Pseudo:            true,
+		NodeMetadata:      report.MakeNodeMetadata(map[string]string{}),
```
tomwilkie: Why are these now needed?  The fields are niled out in the tests.

peterbourgon: It is not valid for a NodeMetadata instance to exist with a nil map anywhere.

tomwilkie: But what problem did it cause?

peterbourgon: The problem is that the expression

```go
report.NodeMetadata{}
```

used to produce a usable object, and now it doesn't. So I updated all instances of that to

```go
report.MakeNodeMetadata(map[string]string{})
```

which in turn caused test failures, as some expected NodeMetadata fields changed from having a nil Metadata member to a constructed-but-empty Metadata member. Does it make sense?

peterbourgon: Actually I would prefer to split

```go
func MakeNodeMetadata(m map[string]string)
```

to

```go
func MakeNodeMetadata()
func MakeNodeMetadataWith(m map[string]string)
```

to make the more-numerous empty initialisms a little nicer to read. What do you think?

tomwilkie: Which tests failed?

tomwilkie: Yeah that change SGTM

peterbourgon: If I revert my change like this

```diff
diff --git a/render/expected/expected.go b/render/expected/expected.go
index 752d19d..5b7211e 100644
--- a/render/expected/expected.go
+++ b/render/expected/expected.go
@@ -17,21 +17,18 @@ var (
                ID:                unknownPseudoNode1ID,
                LabelMajor:        "10.10.10.10",
                Pseudo:            true,
-               NodeMetadata:      report.MakeNodeMetadata(map[string]string{}),
                AggregateMetadata: render.AggregateMetadata{},
        }
        unknownPseudoNode2 = render.RenderableNode{
                ID:                unknownPseudoNode2ID,
                LabelMajor:        "10.10.10.11",
                Pseudo:            true,
-               NodeMetadata:      report.MakeNodeMetadata(map[string]string{}),
                AggregateMetadata: render.AggregateMetadata{},
        }
        theInternetNode = render.RenderableNode{
                ID:                render.TheInternetID,
                LabelMajor:        render.TheInternetMajor,
                Pseudo:            true,
-               NodeMetadata:      report.MakeNodeMetadata(map[string]string{}),
                AggregateMetadata: render.AggregateMetadata{},
        }
 ...
```

then I get failures like this

```
--- FAIL: TestProcessRenderer (0.00s)
        topologies_test.go:26:
 ...
                --- want
                +++ have
                @@ -18,7 +18,8 @@
                    (string) (len=13) "ingress_bytes": (int) 100
                   },
                   NodeMetadata: (report.NodeMetadata) {
                -   Metadata: (map[string]string) <nil>
                +   Metadata: (map[string]string) {
                +   }
                   }
                  },
 ...
```

tomwilkie: I think it you revert your change to trimNodeMetadata those errors will go away.

peterbourgon: Well, sure, the tests will pass, but the output of trimNodeMetadata will leave the report in an invalid state. A instance of NodeMetadata with a nil Metadata member is invalid.

peterbourgon: Ugh. This breaks down in the app tests, when we unmarshal to a RenderableNode, which has an invisible (`json:"-"`) NodeMetadata field. We'll have to nil-out the expected NodeMetadatas in that instance. I hate reference semantics.

peterbourgon:
@tomwilkie squashed and rebased, ready for your last pass.

tomwilkie:
LGTM

PR #339, opened 2015-07-31
Fix a subtle bug in render mapping

peterbourgon:
During rendering, RenderableNodes are created by Map funcs, which take only NodeMetadata. Previously, it was simply assumed that the relevant keys for a given Map func would be present in the metadata. If that implicit invariant failed, the returned RenderableNode would be invalid, with e.g. an ID of
"hostid::". That, in turn, would trigger undefined behavior later on in the rendering workflow.

This bug was detected by creating a partial node metadata for a non-local endpoint node. That node was detected during the first phase of rendering, and given an invalid renderable node ID of "myhostname::", which prevented it from attaching to TheInternet pseudonode. It eventually got removed from the  set
of valid nodes, which meant nodes that were adjacent to it suddenly became orphans, and got filtered out by the FilterUnconnected step of the rendering pipeline.

With this change, every map func checks for the presence of mandatory fields, i.e. the fields that compose the resulting renderable node's ID.

### DIFF
--- a/render/mapping.go
+++ b/render/mapping.go
@@ -46,33 +46,47 @@ type PseudoFunc func(srcNodeID string, srcNode RenderableNode, dstNodeID string,
 // shall be omitted from the rendered topology.
 type MapFunc func(RenderableNode) (RenderableNode, bool)
 
-// MapEndpointIdentity maps a endpoint topology node to endpoint RenderableNode
-// node. As it is only ever run on endpoint topology nodes, we can safely
-// assume the presence of certain keys.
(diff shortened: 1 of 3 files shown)


paulbellamy:
Worth checking if `m.Metadata == nil`?

Also, worth a unit test for the map fns?

peterbourgon:
> Worth checking if m.Metadata == nil?

A NodeMetadata with a nil Metadata field is always invalid. I've added that constraint to the topology validator.

> Also, worth a unit test for the map fns?

:+1:

PR #340, opened 2015-08-03
Cola layout

peterbourgon:
@davkal any reason we shouldn't merge this?

### DIFF
--- a/client/app/scripts/charts/node.js
+++ b/client/app/scripts/charts/node.js
@@ -30,7 +30,7 @@ const Node = React.createClass({
     const label = this.ellipsis(props.label, 14, scale(4 * scaleFactor));
     const subLabel = this.ellipsis(props.subLabel, 12, scale(4 * scaleFactor));
 
-    if (this.props.highlighted) {
+    if (props.highlighted) {
       classNames.push('highlighted');
     }
     if (this.props.blurred) {
--- a/client/app/scripts/charts/nodes-chart.js
+++ b/client/app/scripts/charts/nodes-chart.js
@@ -417,7 +417,9 @@ const NodesChart = React.createClass({
     const normalizedNodeSize = nodeSize / Math.sqrt(n); // assuming rectangular layout
     const nodeScale = this.state.nodeScale.range([0, normalizedNodeSize]);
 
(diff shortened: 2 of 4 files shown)


davkal:
@peterbourgon breaks chrome with this report by @tomwilkie : https://gist.github.com/davkal/5e40f7455d87b86dbc87

davkal:
I need to debug this.

davkal:
Cola produces some very clean layouts, but jumps all the time

screen shot 2015-08-24 at 19 50 58

This is based on Tom's setup.

tomwilkie:
Oh yes that does look better.  Anyway we can stop it jumping?  Have you tried the same trick?

davkal:
Will try the reuse trick tomorrow

davkal:
The reuse trick doesnt even though I cache node positions and then feed them back :( maybe some randomizer in the constraints engine works against us, I'll dig a bit through the code.

davkal:
Just verified, that the topology is always the same, it's still jumpy.

davkal:
I'm migrating the node management to ImmutableJS objects, so that it will be easier to compare updated topologies and see if they need re-rendering.

davkal:
Fixing tests...

davkal:
Internet is now on top, and they are nicer spaced. Switching back and forth between topos is still a bit unpredictable

screen shot 2015-08-27 at 17 45 20

tomwilkie:
I'm not too worries about switching-back-and-forth, its more when staying on the same topology, and the graph changes, do it completely re-layout the page?

tomwilkie:
Also that layout really highlights the need to do better with the internet node: all but one of those connections if going out to the internet, and one is coming in.  Confusing...

davkal:
Same topo, but a node get added/removed, then the layout similarity between renders depends how connected the node was. But we can smooth this out be adding an animation so nodes travel instead of jump.

davkal:
Agreed with the directionality, it's weird now with hovering over nodes, because some nodes seem connected, but dont highlight their peer because it's an ingress link.

davkal:
Closing this, dagre wins out.

PR #342, opened 2015-08-06
Probes POST to apps

peterbourgon:
Change the flow of the streams.

image


app/router.go
```diff
@@ -41,10 +43,13 @@ func URLMatcher(pattern string) mux.MatcherFunc {
...
-// Router gives of the HTTP dispatcher. It will always use the embedded HTML
-// resources.
-func Router(c Reporter) *mux.Router {
+// Router returns the HTTP dispatcher, managing API and UI requests, and
+// accepting reports from probes.. It will always use the embedded HTML
+// resources for the UI.
+func Router(c collector) *mux.Router {
 	router := mux.NewRouter()
+	post := router.Methods("POST").Subrouter()
+	post.HandleFunc("/", handleReport(c))
```
tomwilkie: What do you think about putting this at a different path to /?  How about /api/report?

app/router.go
```diff
@@ -58,7 +63,28 @@ func Router(c Reporter) *mux.Router {
 	return router
 }
 
-func captureTopology(rep Reporter, f func(Reporter, topologyView, http.ResponseWriter, *http.Request)) func(http.ResponseWriter, *http.Request) {
+type collector interface {
```
tomwilkie: Could this interface be moved to xfer?

tomwilkie:
For the OSS to work, doesn't the probe need to use the resolver to publish to multiple apps?

probe/main.go
```diff
@@ -25,10 +25,11 @@ var version = "dev" // set at build time
 
 func main() {
 	var (
+		publish            = flag.String("publish", fmt.Sprintf("http://localhost:%d", xfer.AppPort), "URL to publish reports")
```
tomwilkie: Shouldn't this also include scope.weave.works?

xfer/collector.go
```diff
@@ -1,233 +1,59 @@
...
+func (c *Collector) Add(rpt report.Report) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	c.reports = append(c.reports, timestampReport{now(), rpt})
+}
 
-		case ip := <-c.remove:
-			remove(ip)
+// Report returns a merged report over all added reports.
+func (c *Collector) Report() report.Report {
```
tomwilkie: Is something periodically calling Report, or is it only called when a UI is attached?

If so, won't c.reports get arbitrarily long?

peterbourgon: Yep, we should clean on both ingress and egress.

app/router.go
peterbourgon: It was there originally, but I think it makes more sense here, as the concept of an abstract collector is only useful to the app. More generally, because interfaces define expected behavior, they're actually tools for consumers, rather than producers, of APIs. Ben Johnson gets at it in an article on testing:

> [the] caller should create the interface instead of the callee providing an interface. This makes sense because the caller can declare exactly what it needs.

—https://medium.com/@benbjohnson/structuring-tests-in-go-46ddee7a25c

probe/main.go
peterbourgon: Yep, this and the resolver is coming back in another commit, I just wanted a clean foundation to build up on.

app/router.go
peterbourgon: Yep, sounds good!

peterbourgon:
@tomwilkie @paulbellamy Ready for review!

xfer/publisher.go
```diff
@@ -1,132 +1,125 @@
...
+	return nil
 }
 
-// Close stops a TCPPublisher and closes the socket.
-func (p *TCPPublisher) Close() {
-	close(p.msg)
-	p.closer.Close()
+// AuthorizationHeader returns a value suitable for an HTTP Authorization
+// header, based on the passed token string.
+func AuthorizationHeader(token string) string {
```
paulbellamy: Why not use this above?

paulbellamy:
Overall, LGTM

xfer/publisher.go
peterbourgon: Whoops!

PR #347, opened 2015-08-11
Support --probe.foo=bar style in entrypoint.sh

tomwilkie:


paulbellamy:
Am pleasantly surprised it handles `--app.one="two=three"`.

PR #351, opened 2015-08-12
Add X-Scope-Probe-ID header to POSTs

peterbourgon:
/cc @2opremio @paulbellamy

### DIFF
--- a/xfer/publisher.go
+++ b/xfer/publisher.go
@@ -22,10 +22,17 @@ type Publisher interface {
 type HTTPPublisher struct {
 	url   string
 	token string
+	id    string
 }
 
+// ScopeProbeIDHeader is the header we use to carry the probe's unique ID. The
+// ID is currently set to the probe's hostname. It's designed to deduplicate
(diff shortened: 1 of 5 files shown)


paulbellamy:
Would something like the hostname be more meaningful to the user?

peterbourgon:
Hmm, I can't think of a normal scenario where the hostname wouldn't work... will make the change.

xfer/publisher.go
```diff
@@ -22,10 +22,18 @@ type Publisher interface {
 type HTTPPublisher struct {
 	url   string
 	token string
+	id    string
 }
 
+// ScopeProbeIDHeader is the header we use to carry the probe's unique ID. The
+// ID is generated at runtime and has the same lifetime as the probe process.
```
paulbellamy: This isn't true anymore.

paulbellamy:
Fix comment, please, then LGTM

peterbourgon:
Test failure seems to be GCE related.

PR #354, opened 2015-08-13
Use various scripts from tools.git

tomwilkie:
This reduces duplication as these scripts are shared with service.git, and will also be shared with weave.git in due course.


tomwilkie:
Drop in coverage is due to the new test script calculating coverage more accurately (it include coverage of modules you use, and merges the results properly)

tomwilkie:
Intermittent test failure: #363

Makefile
```diff
@@ -61,7 +61,7 @@ client-sync:
 		-v $(shell pwd)/client/build:/home/weave/build \
 		$(SCOPE_UI_BUILD_IMAGE) gulp sync
 
-$(SCOPE_UI_BUILD_EXPORT): client/Dockerfile client/gulpfile.js client/package.json
+$(SCOPE_UI_BUILD_EXPORT): client/Dockerfile client/gulpfile.js client/package.json  client/webpack.config.js client/.eslintrc
```
paulbellamy: control your whitespace

client/Dockerfile
```diff
@@ -1,19 +1,5 @@
-FROM debian:latest
+FROM node:0.10
```
peterbourgon: I know David did FROM debian instead of FROM node deliberately. So I guess it would be best if he gave a thumbs-up on this change...

tomwilkie: That was alpine-node; the node image is something we only recently learned about.  Its what we use in service, and updating to use it here makes the builds significantly faster.  Whats more, the node image is just debian + node: https://github.com/joyent/docker-node/blob/787d816f90dc1bf933b6503b876bdeb752b24f96/0.10/Dockerfile

davkal:
Make adding package and npm install a separate step because it takes the longest to compile.
```diff
diff --git a/client/Dockerfile b/client/Dockerfile
index 47cd0e2..741e765 100644
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -1,5 +1,6 @@
 FROM node:0.10
 WORKDIR /home/weave
-COPY package.json gulpfile.js webpack.config.js .eslintrc /home/weave/
 RUN npm install -g gulp
+COPY package.json
 RUN npm install
+COPY gulpfile.js webpack.config.js .eslintrc /home/weave/
```

client/Dockerfile
davkal: Yeah, the node base image is good because it has build dependencies baked in.

tomwilkie:
Updated the Docker file.  How does it look now?

davkal:
LGTM :dango:

PR #355, opened 2015-08-13
Make edge direction flow from client->server.

tomwilkie:
So it turns out the layout engine we use is already trying to layout the graph from top-to-bottom: https://github.com/cpettitt/dagre/wiki

This change uses a simple heuristic (client port is the higher port) to orient the edges in the graph such that they flow from client to server.

It somewhat improves the graph layout.  Before:

screen shot 2015-08-13 at 15 51 00

After:

screen shot 2015-08-13 at 15 49 45

### DIFF
--- a/probe/endpoint/reporter.go
+++ b/probe/endpoint/reporter.go
@@ -80,18 +80,18 @@ func (r *Reporter) addConnection(rpt *report.Report, c *procspy.Connection) {
 		localIsClient       = int(c.LocalPort) > int(c.RemotePort)
 		localAddressNodeID  = report.MakeAddressNodeID(r.hostID, c.LocalAddress.String())
 		remoteAddressNodeID = report.MakeAddressNodeID(r.hostID, c.RemoteAddress.String())
-		adjecencyID         = ""
+		adjacencyID         = ""
 		edgeID              = ""
 	)
 
 	if localIsClient {
-		adjecencyID = report.MakeAdjacencyID(localAddressNodeID)
-		rpt.Address.Adjacency[adjecencyID] = rpt.Address.Adjacency[adjecencyID].Add(remoteAddressNodeID)
+		adjacencyID = report.MakeAdjacencyID(localAddressNodeID)
+		rpt.Address.Adjacency[adjacencyID] = rpt.Address.Adjacency[adjacencyID].Add(remoteAddressNodeID)
 
 		edgeID = report.MakeEdgeID(localAddressNodeID, remoteAddressNodeID)
 	} else {
-		adjecencyID = report.MakeAdjacencyID(remoteAddressNodeID)
-		rpt.Address.Adjacency[adjecencyID] = rpt.Address.Adjacency[adjecencyID].Add(localAddressNodeID)
+		adjacencyID = report.MakeAdjacencyID(remoteAddressNodeID)
+		rpt.Address.Adjacency[adjacencyID] = rpt.Address.Adjacency[adjacencyID].Add(localAddressNodeID)
 
 		edgeID = report.MakeEdgeID(remoteAddressNodeID, localAddressNodeID)
 	}
@@ -109,18 +109,18 @@ func (r *Reporter) addConnection(rpt *report.Report, c *procspy.Connection) {
 		var (
 			localEndpointNodeID  = report.MakeEndpointNodeID(r.hostID, c.LocalAddress.String(), strconv.Itoa(int(c.LocalPort)))
 			remoteEndpointNodeID = report.MakeEndpointNodeID(r.hostID, c.RemoteAddress.String(), strconv.Itoa(int(c.RemotePort)))
-			adjecencyID          = ""
+			adjacencyID          = ""
 			edgeID               = ""
 		)
 
 		if localIsClient {
-			adjecencyID = report.MakeAdjacencyID(localEndpointNodeID)
-			rpt.Endpoint.Adjacency[adjecencyID] = rpt.Endpoint.Adjacency[adjecencyID].Add(remoteEndpointNodeID)
+			adjacencyID = report.MakeAdjacencyID(localEndpointNodeID)
+			rpt.Endpoint.Adjacency[adjacencyID] = rpt.Endpoint.Adjacency[adjacencyID].Add(remoteEndpointNodeID)
 
 			edgeID = report.MakeEdgeID(localEndpointNodeID, remoteEndpointNodeID)
 		} else {
-			adjecencyID = report.MakeAdjacencyID(remoteEndpointNodeID)
-			rpt.Endpoint.Adjacency[adjecencyID] = rpt.Endpoint.Adjacency[adjecencyID].Add(localEndpointNodeID)
+			adjacencyID = report.MakeAdjacencyID(remoteEndpointNodeID)
+			rpt.Endpoint.Adjacency[adjacencyID] = rpt.Endpoint.Adjacency[adjacencyID].Add(localEndpointNodeID)
 
 			edgeID = report.MakeEdgeID(remoteEndpointNodeID, localEndpointNodeID)
 		}
--- a/probe/endpoint/reporter.go
+++ b/probe/endpoint/reporter.go
@@ -77,13 +77,24 @@ func (r *Reporter) Report() (report.Report, error) {
 
 func (r *Reporter) addConnection(rpt *report.Report, c *procspy.Connection) {
 	var (
+		localIsClient       = int(c.LocalPort) > int(c.RemotePort)
 		localAddressNodeID  = report.MakeAddressNodeID(r.hostID, c.LocalAddress.String())
 		remoteAddressNodeID = report.MakeAddressNodeID(r.hostID, c.RemoteAddress.String())
-		adjecencyID         = report.MakeAdjacencyID(localAddressNodeID)
-		edgeID              = report.MakeEdgeID(localAddressNodeID, remoteAddressNodeID)
+		adjecencyID         = ""
+		edgeID              = ""
 	)
 
-	rpt.Address.Adjacency[adjecencyID] = rpt.Address.Adjacency[adjecencyID].Add(remoteAddressNodeID)
+	if localIsClient {
+		adjecencyID = report.MakeAdjacencyID(localAddressNodeID)
+		rpt.Address.Adjacency[adjecencyID] = rpt.Address.Adjacency[adjecencyID].Add(remoteAddressNodeID)
+
+		edgeID = report.MakeEdgeID(localAddressNodeID, remoteAddressNodeID)
+	} else {
+		adjecencyID = report.MakeAdjacencyID(remoteAddressNodeID)
+		rpt.Address.Adjacency[adjecencyID] = rpt.Address.Adjacency[adjecencyID].Add(localAddressNodeID)
+
+		edgeID = report.MakeEdgeID(remoteAddressNodeID, localAddressNodeID)
+	}
 
 	if _, ok := rpt.Address.NodeMetadatas[localAddressNodeID]; !ok {
 		rpt.Address.NodeMetadatas[localAddressNodeID] = report.MakeNodeMetadataWith(map[string]string{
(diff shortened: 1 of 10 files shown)


peterbourgon:
I don't think you can just swap which end of the connection is used as the key in the adjacency map. [We assume that every key in the NodeMetadata map has representation in the adjacency map, too](https://github.com/weaveworks/scope/blob/82e59bfe2065aa887625dfe62561e17eb86dc23a/render/render.go#L184). I'm surprised this doesn't break rendering.

What about just using [this change](https://github.com/weaveworks/scope/pull/355/files#diff-9898e724b96b9b4456dc3674787aeb26R35) in isolation?

tomwilkie:
That change you refer to is a noop (its the default value). I didn't realise I'd committed it.

tomwilkie:
This next change reverses the assumption that keys in the Adjacency map always exist in the NodeMetadata map.  It makes the pseudo node generator take a direction flag, and updates the tests to reflect reality.

One side effect of this PR is previously, when you have a probe running on two different machines, communicating processes on those machines would have two edges in the UI.

A negative side effect is that now, in the details panel, one only sees outgoing connections from the selected node.  I intend to address this in this PR, next.

render/detailed_node.go
```diff
@@ -156,16 +174,26 @@ func OriginTable(r report.Report, originID string) (Table, bool) {
...
+			if clientNodeID, ok := report.ParseAdjacencyID(clientAdjID); ok {
+				if _, address, port, ok := report.ParseEndpointNodeID(clientNodeID); ok {
+					rows = append(rows, Row{
+						Key:        fmt.Sprintf("%s:%s", address, port),
+						ValueMajor: local,
+					})
+				}
+			}
+		}
+	}
```
tomwilkie: @paulbellamy could you take a look at this bit please?

probe/endpoint/reporter.go
```diff
@@ -77,13 +77,24 @@ func (r *Reporter) Report() (report.Report, error) {
 
 func (r *Reporter) addConnection(rpt *report.Report, c *procspy.Connection) {
 	var (
+		localIsClient       = int(c.LocalPort) > int(c.RemotePort)
 		localAddressNodeID  = report.MakeAddressNodeID(r.hostID, c.LocalAddress.String())
 		remoteAddressNodeID = report.MakeAddressNodeID(r.hostID, c.RemoteAddress.String())
-		adjecencyID         = report.MakeAdjacencyID(localAddressNodeID)
-		edgeID              = report.MakeEdgeID(localAddressNodeID, remoteAddressNodeID)
+		adjecencyID         = ""
```
peterbourgon: adjacency

render/detailed_node.go
peterbourgon: Proposal?

```go
for clientAdjID, serverNodeIDs := range endpointTopology.Adjacency {
	if !serverNodeIDs.Contains(originID) {
		continue
	}
	clientNodeID, ok := report.ParseAdjacencyID(clientAdjID)
	if !ok {
		continue
	}
	_, addr, port, ok := report.ParseEndpointNodeID(clientNodeID)
	if !ok {
		continue
	}
	rows = append(rows, Row{
		Key:        fmt.Sprintf("%s:%s", addr, port),
		ValueMajor: local,
	})
}
```

tomwilkie: Yeah thats what I originally had.  I forget why I rewrote it; yours looks much better.  Will change.

render/mapping.go
```diff
@@ -33,11 +33,12 @@ const (
...
-// PseudoFunc creates RenderableNode representing pseudo nodes given the dstNodeID.
-// The srcNode renderable node is essentially from MapFunc, representing one of
-// the rendered nodes this pseudo node refers to. srcNodeID and dstNodeID are
-// node IDs prior to mapping.
-type PseudoFunc func(srcNodeID string, srcNode RenderableNode, dstNodeID string, local report.Networks) (RenderableNode, bool)
+// PseudoFunc creates RenderableNode representing pseudo nodes given the nodeID.
+// dstNodeID is the node id of one of the nodes this node is attached to.
+// nodeID and dstNodeID are node IDs prior to mapping.  isClient indicated the direction
+// of the edge to dstNodeID - true indicates nodeID is the client, false indicates
+// nodeID is the server.
```
peterbourgon: Cleanup...

```go
// PseudoFunc creates RenderableNode representing pseudo nodes given the
// nodeID. dstNodeID is the node ID of one of the nodes this node is attached
// to. nodeID and dstNodeID are node IDs prior to mapping. isClient indicates
// the direction of the edge to dstNodeID: true indicates nodeID is the
// client, false indicates nodeID is the server.
```

render/mapping.go
```diff
@@ -323,29 +324,36 @@ func MapAddress2Host(n RenderableNode) (RenderableNode, bool) {
...
-		if !local.Contains(dstNodeAddr) {
+		if !local.Contains(nodeAddr) {
 			return newPseudoNode(TheInternetID, TheInternetMajor, ""), true
 		}
 
-		// Otherwise, the rule for non-internet psuedo nodes; emit 1 new node for each
-		// dstNodeAddr, srcNodeAddr, srcNodePort.
-		srcNodeAddr, srcNodePort := trySplitAddr(src)
+		if isClient {
+			// If the client node is missing, generate a single pseudo node for every (client ip, server ip, server por)
```
peterbourgon: port

render/mapping.go
```diff
@@ -323,29 +324,36 @@ func MapAddress2Host(n RenderableNode) (RenderableNode, bool) {
 // the report's local networks.  Otherwise, the returned function will
 // produce a single pseudo node per (dst address, src address, src port).
 func GenericPseudoNode(addresser func(id string) net.IP) PseudoFunc {
-	return func(src string, srcMapped RenderableNode, dst string, local report.Networks) (RenderableNode, bool) {
-		// Use the addresser to extract the destination IP
-		dstNodeAddr := addresser(dst)
-
+	return func(nodeID, dstNodeId string, isClient bool, local report.Networks) (RenderableNode, bool) {
```
peterbourgon: dstNodeID

render/mapping.go
```diff
@@ -33,11 +33,12 @@ const (
...
-// The srcNode renderable node is essentially from MapFunc, representing one of
-// the rendered nodes this pseudo node refers to. srcNodeID and dstNodeID are
-// node IDs prior to mapping.
-type PseudoFunc func(srcNodeID string, srcNode RenderableNode, dstNodeID string, local report.Networks) (RenderableNode, bool)
+// PseudoFunc creates RenderableNode representing pseudo nodes given the nodeID.
+// dstNodeID is the node id of one of the nodes this node is attached to.
+// nodeID and dstNodeID are node IDs prior to mapping.  isClient indicated the direction
+// of the edge to dstNodeID - true indicates nodeID is the client, false indicates
+// nodeID is the server.
+type PseudoFunc func(nodeID string, dstNodeID string, isClient bool, local report.Networks) (RenderableNode, bool)
```
peterbourgon: isClient is a bit ambiguous. How about one of these?

```go
type PseudoFunc(srcNodeID, dstNodeID string, srcIsClient bool, local report.Networks) (RenderableNode, bool)

type PseudoFunc(nodeID, dstNodeID string, client2server bool, local report.Networks) (RenderableNode, bool)
```

render/mapping.go
```diff
@@ -323,29 +324,36 @@ func MapAddress2Host(n RenderableNode) (RenderableNode, bool) {
...
+			serverIP, serverPort := trySplitAddr(dstNodeId)
+			outputID := MakePseudoNodeID(nodeAddr.String(), serverIP, serverPort)
+			major := nodeAddr.String()
+			return newPseudoNode(outputID, major, ""), true
+		}
 
-		outputID := MakePseudoNodeID(dstNodeAddr.String(), srcNodeAddr, srcNodePort)
-		major := dstNodeAddr.String()
-		return newPseudoNode(outputID, major, ""), true
+		// Othereise (the server node is missing), generate a pseudo node for every (server ip, server port)
```
peterbourgon: Otherwise

render/render.go
```diff
@@ -177,37 +177,69 @@ func (m LeafMap) Render(rpt report.Report) RenderableNodes {
 		source2mapped[nodeID] = mapped.ID
 	}
 
+	mkPseudoNode := func(srcID, dstId string, srcIsClient bool) (string, bool) {
```
peterbourgon: Please use consistent variable names to the PseudoNode type definition

render/detailed_node.go
```diff
@@ -156,16 +174,31 @@ func OriginTable(r report.Report, originID string) (Table, bool) {
...
+		}
+		_, address, port, ok := report.ParseEndpointNodeID(clientNodeID)
+		if !ok {
+			continue
+		}
+		rows = append(rows, Row{
+			Key:        fmt.Sprintf("%s:%s", address, port),
+			ValueMajor: local,
+		})
+	}
```
paulbellamy: This function might be more readable as:
```Go
rows := []Row{}
rows = append(rows, inboundConnectionDetailRows(endpointTopology, originID, local)...)
rows = append(rows, outboundConnectionDetailRows(endpointTopology, originID, local)...)
return rows
```

Or just a comment by each for loop, whichever.

paulbellamy: Won't the second loop overlap with the first, and include outbound connections, backwards?

tomwilkie: It shouldn't - connection are directional, so nodeA -> nodeB will exist but nodeB -> nodeA won't.

tomwilkie: But as discussed, this could duplicate loops; in reality this won't happen - edges only exist in the endpoint topology, and the node id in that topology also includes ports, and well, its weird if you connect from port A to port A.

peterbourgon:
> edges only exist in the endpoint topology

This isn't a constraint of the data model. It's just an accident of the implementation. You can't treat it as an invariant and program against it.

But I was thinking — rather than restructuring the adjacency list, why don't you just indicate the client-to-server directionality as a boolean field in the EdgeMetadata?

tomwilkie:
> This isn't a constraint of the data model. It's just an accident of the implementation. You can't treat it as an invariant and program against it.

I'm not; I was pointing out that whilst Paul had spotted a bug, the reality was it was never going to be hit.  I fixed it anyway.

> But I was thinking — rather than restructuring the adjacency list, why don't you just indicate the client-to-server directionality as a boolean field in the EdgeMetadata?

I didn't consider this at the time; what would it buy us?

There is an inherent directionality in the adjacency map; previously it was from local->remote.  This approach changed the assumption without changing the datastructure.  With this approach there is no duplication of information (previously two edges would have existed between communicating nodes, now only one).  Also no UI changes are required, as the UI already assumes the edges are directional, although it was probably just a coincidence.

render/detailed_node.go
tomwilkie: This has been addressed in the changeset below

peterbourgon:
> I'm not; I was pointing out that whilst Paul had spotted a bug, the reality was it was never going to be hit. I fixed it anyway.

Okeedokey.

> I didn't consider this at the time; what would it buy us?

We would be able to keep the invariant that NodeMetadatas[id] —implies→ Adjacency[id], which I think is a nice property. But...

> There is an inherent directionality in the adjacency map; previously it was from local->remote. This approach changed the assumption without changing the datastructure. With this approach there is no duplication of information (previously two edges would have existed between communicating nodes, now only one). Also no UI changes are required, as the UI already assumes the edges are directional, although it was probably just a coincidence.

I agree with everything here. I'm thinking out loud what "directionality" actually means for us. The edges have [both ingress and egress counters](https://github.com/weaveworks/scope/blob/9be81c520c55a8dd583c21197a0ffaff47337ee2/report/topology.go#L34-L37) so they actually represent both directions of data flow. Therefore, we don't actually have a directed graph, we just have a graph. And the directionality encoded by the adjacency list is kind of meaningless. Well, it has the meaning we assign to it: previously that was (reliably) local-to-remote, now it is (heuristically) client-to-server.

It seems to me that we are trading the property I listed above for more code in rendering and a potentially nicer graph layout. Is that right?

tomwilkie:
The graph is still directed; previous I would say it was undirected, as in most situations having a (src, dst) edge meant you also had a (dst, src) edge.  Now you'll only get a single (highest port number, lowest port number) edge - which looks much nicer in the UI, and is factually more correct.  The edge represents a tcp connection, not a dataflow.  The directionality is super meaningful, but the edgemetadata needs to be updated to deal with it (ingress and egress swapped when appropriate).

I'd phrase this change more as adding more semantics to the topology, which in turns makes the rendering slightly nicer.  I don't see how the property of each node metadata key being a valid key in adjacencies was particularly useful - we pretty much only relied on it in one place (the leaf map code).  And we were mistakenly relying on both directions of an edge to exist in detailed node rendering, so this change improves that.

tomwilkie:
@peterbourgon I'd like to get this merged today; whats left to do?

peterbourgon:
Nothing left to do, I was just giving it some brain time to see if there were potentially other problems, but I couldn't come up with any. :+1:

tomwilkie:
Cool, no problem.  I'll file a follow up ticket to ensure we handle the directionality correcting in EdgeMetadata; currently the only populated field is tcp connection count, which isn't directional, so its fine.  But when we get the bandwidth stuff enabled by default, it will bite us.

PR #366, opened 2015-08-18
Add container names to process minor label if possible.

### LINKED ISSUE #331: Show container name instead of process name in "Applications"
It's really hard to tell difference when you have multiple node/ruby/java/etc applications running. Could it be possible to label these with container name (if process is inside a container)?

tomwilkie:
Fixes #331

### DIFF
--- a/app/api_topology_test.go
+++ b/app/api_topology_test.go
@@ -62,10 +62,8 @@ func TestAll(t *testing.T) {
 	}
 }
 
-func TestAPITopologyApplications(t *testing.T) {
+func TestAPITopologyContainers(t *testing.T) {
 	ts := httptest.NewServer(Router(StaticReport{}))
-	defer ts.Close()
-	is404(t, ts, "/api/topology/applications/foobar")
 	{
 		body := getRawJSON(t, ts, "/api/topology/containers")
 		var topo APITopology
@@ -77,6 +75,12 @@ func TestAPITopologyApplications(t *testing.T) {
 			t.Error(test.Diff(want, have))
 		}
 	}
+}
+
+func TestAPITopologyApplications(t *testing.T) {
+	ts := httptest.NewServer(Router(StaticReport{}))
+	defer ts.Close()
+	is404(t, ts, "/api/topology/applications/foobar")
 	{
 		body := getRawJSON(t, ts, "/api/topology/applications/"+expected.ServerProcessID)
 		var node APINode
--- a/app/api_topology_test.go
+++ b/app/api_topology_test.go
@@ -67,13 +67,13 @@ func TestAPITopologyApplications(t *testing.T) {
 	defer ts.Close()
 	is404(t, ts, "/api/topology/applications/foobar")
 	{
-		body := getRawJSON(t, ts, "/api/topology/applications")
+		body := getRawJSON(t, ts, "/api/topology/containers")
 		var topo APITopology
 		if err := json.Unmarshal(body, &topo); err != nil {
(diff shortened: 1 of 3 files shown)


peterbourgon:
Can you provide some context here? When I'm in the applications view, I generally don't want to know about containers. Did someone ask for this feature?

app/api_topology_test.go
```diff
@@ -67,13 +67,13 @@ func TestAPITopologyApplications(t *testing.T) {
 	defer ts.Close()
 	is404(t, ts, "/api/topology/applications/foobar")
 	{
-		body := getRawJSON(t, ts, "/api/topology/applications")
+		body := getRawJSON(t, ts, "/api/topology/containers")
```
peterbourgon: Why is TestAPITopologyApplications making a GET on /api/topology/containers? Can you move this to TestAPITopologyContainers?

app/router.go
```diff
@@ -104,7 +104,7 @@ var topologyRegistry = map[string]topologyView{
 	"applications": {
 		human:    "Applications",
 		parent:   "",
-		renderer: render.FilterUnconnected{Renderer: render.ProcessRenderer},
+		renderer: render.FilterUnconnected{Renderer: render.ProcessWithContainerNameRenderer{}},
```
peterbourgon: Is there a reason we wouldn't just have this be the default behavior of ProcessRenderer?

render/topologies.go
```diff
@@ -25,6 +29,45 @@ var ProcessRenderer = MakeReduce(
...
 
+// ProcessWithContainerNameRenderer is a Renderer which produces a process
+// graph enriched with container names where appropriate
+type ProcessWithContainerNameRenderer struct{}
+
+// Render produces a process graph where the minor labels contain the
+// container name, if found.
+func (r ProcessWithContainerNameRenderer) Render(rpt report.Report) RenderableNodes {
+	var processes = ProcessRenderer.Render(rpt)
+	var containers = LeafMap{
```
peterbourgon: [Always prefer type-inferred assignment with :=](https://docs.google.com/document/d/18UTcmy8VoecEaYNb0DrI8NZrI1uJbq-XYDEko0l5Ny4/edit) unless you're doing a var block.

render/topologies.go
```diff
@@ -25,6 +29,45 @@ var ProcessRenderer = MakeReduce(
...
+		}
+		containerID, ok := p.NodeMetadata.Metadata[docker.ContainerID]
+		if !ok {
+			continue
+		}
+		container, ok := containers[containerID]
+		if !ok {
+			continue
+		}
+		p.LabelMinor = fmt.Sprintf("%s (%s:%s)", report.ExtractHostID(p.NodeMetadata), container.LabelMajor, pid)
```
peterbourgon: Just to be clear, this is overwriting the existing minor label, which is of the format "hostname (pid)" — right?

app/router.go
tomwilkie: The process renderer is used by the container render; there very little point in going to all the effort of labelling up the processes if we're just going to blow it away in the container renderer.

render/topologies.go
tomwilkie: Yes

tomwilkie:
@peterbourgon thanks for the feedback; PTAL

app/router.go
peterbourgon: Ah, right, gotcha.

peterbourgon:
:+1:

PR #367, opened 2015-08-18
Condense origin address entries into a single table.

### LINKED ISSUE #323: Lots of "Origin Address" entries in the details panels for hosts
This seems like we should probably condense it into one table, like with connections:

screen shot 2015-07-16 at 11 15 34

tomwilkie:
Fixes #323

### DIFF
--- a/render/detailed_node.go
+++ b/render/detailed_node.go
@@ -6,7 +6,6 @@ import (
 	"strconv"
 
 	"github.com/weaveworks/scope/probe/docker"
-	"github.com/weaveworks/scope/probe/endpoint"
 	"github.com/weaveworks/scope/probe/host"
 	"github.com/weaveworks/scope/probe/process"
 	"github.com/weaveworks/scope/report"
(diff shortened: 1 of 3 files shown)


render/detailed_node.go
```diff
@@ -153,18 +151,27 @@ func OriginTable(r report.Report, originID string) (Table, bool) {
...
+			return addr
 		}
+		return ""
+	}
+
+	local := labeler(originID)
+	adjacencies := topology.Adjacency[report.MakeAdjacencyID(originID)]
+	sort.Strings(adjacencies)
+	for _, nodeID := range adjacencies {
+		rows = append(rows, Row{
```
2opremio: `labeler` can return an empty string. We shouldn't add a row when this happens. And, if it *can't* happen then I think we should make it explicit (e.g. by logging it).

2opremio:
Minor comment aside, LGTM.

PR #368, opened 2015-08-18
make should auto-build app/static.go if needed

paulbellamy:
@davkal Does this still fit your workflow?

I replaced `make static` with `make app/static.go`, which the main `make` depends on. It means that you can out-of-the-box do `make`, and have it build everything, which is nice for me.


tomwilkie:
The reasoning against this was that the ui build image takes a long time to build and pulls down a lot of packages.  We didn't want to burden non-ui people with this, so we checked in static.go.

If you're keen on this, then you'll want to delete static.go in this PR.

paulbellamy:
Just floating it as a suggestion.

I'd personally rather wait a bit and have a complete build. It'll only rebuild the UI when it changes, anyway. But if other non-ui people prefer it the way it is I'm happy with that also.

PR #369, opened 2015-08-18
Show how many containers have been grouped together under an image

### LINKED ISSUE #336: Show how many containers have been grouped together under an image
And similarly for process-by-name view.

tomwilkie:
Fixes #336

### DIFF
--- a/render/mapping.go
+++ b/render/mapping.go
@@ -19,6 +19,9 @@ const (
 
 	TheInternetID    = "theinternet"
 	TheInternetMajor = "The Internet"
+
+	containersPrefix = "container-"
+	processesPrefix  = "process-"
 )
 
 // LeafMapFunc is anything which can take an arbitrary NodeMetadata, which is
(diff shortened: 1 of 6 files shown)


render/mapping.go
```diff
@@ -276,7 +307,10 @@ func MapContainer2ContainerImage(n RenderableNode) (RenderableNode, bool) {
 		return n, false
 	}
 
-	return newDerivedNode(id, n), true
+	// Add container-<id> key to NMD, which will later be counted to produce the minor label
+	result := newDerivedNode(id, n)
+	result.NodeMetadata.Metadata[containersPrefix+n.ID] = ""
+	return result, true
```
peterbourgon: What do you think about a "container_count" key, with the merge semantics defined as interpreting the value as a number and summing?

peterbourgon:
Apart from fixing the keys-with-empty-values thing, LGTM. Quite like it.

render/mapping.go
tomwilkie: This is 'the mapreduce way'.  Emit unique keys from the mapper and then count them in the reducer.

Are you suggesting the container_count be a field on NodeMetadata?   Or a key in the NodeMetadata.Metadata map?  That sounds awfully special case.

A more generic solution would be to add a map[string]int field to the RenderableNode struct called counters, which was not json serialised, and define the merge semantics on that.  But that seems like massive overkill for these two uses.

peterbourgon: > This is 'the mapreduce way'. Emit unique keys from the mapper and then count them in the reducer.

It's the wordcount way. But we're not reducing text files, we're semantically merging dicts. A key with no value feels like a yellow flag.

> Are you suggesting the container_count be a field on NodeMetadata? Or a key in the NodeMetadata.Metadata map?

Either one.

> A more generic solution would be to add a map[string]int field to the RenderableNode struct called counters, which was not json serialised, and define the merge semantics on that. But that seems like massive overkill for these two uses.

I agree, that would be premature generalization.

tomwilkie: I'm not keen on special casing keys in the ndm metadata struct like that.  Thats a yellow flag to me.

As a half-way step between a counters field on renderable node, how about any keys in NMD.metadata with the prefix "-counter" gets add semantics?

peterbourgon: For the record: we decided offline to pursue the separate counters struct in the NodeMetadata.

PR #376, opened 2015-08-19
Don't report empty fields from process

### LINKED ISSUE #370: Empty fields in details pane look broken
The command field does not seem to have a value when probe is run in OSX. The field in the details pane has no value, which looks broken. Maybe display `(n/a)` if a field has no value?

screen shot 2015-08-18 at 23 10 54

2opremio:
Fixes #370

### DIFF
--- a/probe/process/reporter.go
+++ b/probe/process/reporter.go
@@ -45,12 +45,17 @@ func (r *Reporter) processTopology() (report.Topology, error) {
 	err := r.walker.Walk(func(p Process) {
 		pidstr := strconv.Itoa(p.PID)
 		nodeID := report.MakeProcessNodeID(r.scope, pidstr)
-		t.NodeMetadatas[nodeID] = report.MakeNodeMetadataWith(map[string]string{
-			PID:     pidstr,
-			Comm:    p.Comm,
-			Cmdline: p.Cmdline,
-			Threads: strconv.Itoa(p.Threads),
-		})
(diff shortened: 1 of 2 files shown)


probe/process/reporter.go
```diff
@@ -45,12 +45,19 @@ func (r *Reporter) processTopology() (report.Topology, error) {
...
 			Threads: strconv.Itoa(p.Threads),
-		})
+		}
+		// Filter out empty fields
+		for key, value := range metadata {
+			if value == "" {
+				delete(metadata, key)
+			}
+		}
+		t.NodeMetadatas[nodeID] = report.MakeNodeMetadataWith(metadata)
```
tomwilkie: This would be better if it were more similar to detailed_node rendering:
```
	for _, tuple := range []struct{ key, value string }{
		{Comm, p.Comm},
		{PID, pidstr},
		{Cmdline, p.Cmdline},
		{Threads, strconv.Itoa(p.Threads)},
	} {
		if tuple.value != "" {
			rt.NodeMetadatas[nodeID][key] = value
		}
	}
```
Or similar...

2opremio: Uhm, I actually wrote it like that initially, but then I thought it was easier to read the *complete* metadata first. I will change it for consistency. Thanks.

tomwilkie:
LGTM

PR #379, opened 2015-08-19
Stop the rendered graph from dancing.

tomwilkie:
I'm not sure this actually works, but I wanted to try it out.

The idea is that instead of doing a fresh layout every time we get a graph update, we cache the layout from last time and mutate that.

Some basic testing implies this works, but I'll try harder to break it tomorrow.

Let me know what you think, @davkal!


davkal:
There was some provision for reusing old locations [in the code](https://github.com/weaveworks/scope/blob/95448e73ae0ea0dea0aec4a8e01be9ea94e3e875/client/app/scripts/charts/nodes-chart.js#L134) but that never informed the layout engine. So yeah, looks promising!

davkal:
Tried it out. Layout is less jumpy with coming/going nodes in the same topology. Switching back and forth between topos still makes different layouts every time.
We can try an approach where we cache all node positions, and then feed those back into the engine.

tomwilkie:
@davkal can you run this with?  My JS is not so good.

davkal:
@tomwilkie sure thing!

davkal:
Hmm, caching the layout positions was a non-starter, they are not being passed back to the internal node objects that the graph engine uses, they are not on [the whitelist](https://github.com/cpettitt/dagre/blob/master/lib/layout.js#L100), and even if I patch dagre, they dont seem to inform the layout.

I'll try updating to the latest version, and then go for the one-graph-engine-per-topology approach.

davkal:
Looks better with one engine per topo.

tomwilkie:
LGTM.  Will need to regen static.go.

PR #380, opened 2015-08-20
Improved error message when docker-squash is not detected

inercia:
Ditto

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 # If you can use Docker without being root, you can `make SUDO= <target>`
 SUDO=sudo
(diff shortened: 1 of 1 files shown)


tomwilkie:
Hi Alvaro; can you rebase and fix the issues with this?  In general I'm not in favour of assuming you have docker squash installed in a specific localtion; assuming its in the PATH is more flexible, as the user has control over the path.  We could always update the error message to specify that it needs to be in the path?

inercia:
@tomwilkie ok, I'll update the message. The problem was that the message is misleading: if `docker-squash` is not in the `PATH`, you get an error that suggests to do a  `make dep`, you do that and it still fails...

PR #384, opened 2015-08-20
gzip compression

### LINKED ISSUE #371: Analyze the amount of traffic generated by Scope
We should analyze the amount of traffic generated by Scope when sending data to `scope.weave.works`, and study possible scalability problems. In my tests, 3 machines with 7 containers/machine generate about 100KiB/s. Maybe this could be reduced by applying some compression...

inercia:
Fixes #371

### DIFF
--- a/app/router.go
+++ b/app/router.go
@@ -1,11 +1,14 @@
 package main
 
 import (
+	"compress/gzip"
 	"encoding/gob"
+	"io"
 	"net/http"
 	"net/url"
 	"strings"
 
+	"github.com/PuerkitoBio/ghost/handlers"
 	"github.com/gorilla/mux"
 
 	"github.com/weaveworks/scope/render"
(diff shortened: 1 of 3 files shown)


inercia:
@tomwilkie do you think it is worth to apply compression to some other HTTP endpoints (eg, `/api/topology`)?

tomwilkie:
Looks good.  Can we make the server side a little more generic (like https://gist.github.com/bryfry/09a650eb8aac0fb76c24) so it can be used for other endpoints (as you suggest)?

Making it more generic will also make it unit-testable.

inercia:
@tomwilkie I thought about adding a more generic solution, but I didn't want to clutter the code until I could confirm that _a)_ we get significant amounts of data from other endpoints, and _b)_ this does not disturb the UI (well, I guess some of these endpoints are used from the UI...).

I'll add compression for all endpoints and I'll see what happens...

inercia:
I have added a [gorilla handler](https://github.com/gorilla/handlers)) at the root mux that performs all the compression/decompression transparently on the app side.

app/router.go
```diff
@@ -71,7 +72,19 @@ func Router(c collector) *mux.Router {
 func makeReportPostHandler(a xfer.Adder) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		var rpt report.Report
-		if err := gob.NewDecoder(r.Body).Decode(&rpt); err != nil {
+
+		reader := r.Body
```
inercia: It seems that the content in a POST is not automatically decompressed, even when the `Content-Type` is `gzip`... :worried:

xfer/publisher_test.go
```diff
@@ -1,16 +1,19 @@
...
 	"reflect"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/weaveworks/scope/report"
 	"github.com/weaveworks/scope/test"
-	"github.com/weaveworks/scope/xfer"
+
+	"github.com/gorilla/handlers"
```
tomwilkie: pls put this import above the scope ones.

xfer/publisher_test.go
```diff
@@ -1,16 +1,19 @@
-package xfer_test
+package xfer
```
tomwilkie: Pls don't change the package; @peterbourgon is very keen on having tests in separate packages, even if it means exposing slightly more of an API.

tomwilkie:
gorilla handler: nice!

xfer/publisher_test.go
inercia: Done.

xfer/publisher_test.go
inercia: Done.

xfer/publisher.go
```diff
@@ -52,16 +53,24 @@ func NewHTTPPublisher(target, token, id string) (*HTTPPublisher, error) {
 
 // Publish publishes the report to the URL.
 func (p HTTPPublisher) Publish(rpt report.Report) error {
-	var buf bytes.Buffer
-	if err := gob.NewEncoder(&buf).Encode(rpt); err != nil {
+	gzbuf := new(bytes.Buffer)
```
tomwilkie: we prefer bytes.Buffer{} as opposed to new(...)

https://docs.google.com/document/d/18UTcmy8VoecEaYNb0DrI8NZrI1uJbq-XYDEko0l5Ny4/edit

xfer/publisher.go
```diff
@@ -52,16 +53,24 @@ func NewHTTPPublisher(target, token, id string) (*HTTPPublisher, error) {
...
+	gzwriter := gzip.NewWriter(gzbuf)
```
tomwilkie: Also, I think we could use a io.Pipe in a way such that we don't buffer the whole encoded report in memory.  Give one end to http.NewRequest, and the other to the gob.NewEncoder.  Not 100% sure, but can you give it a try please?

tomwilkie:
Looking really good; just the one comment really.

xfer/publisher.go
inercia: I tried that, but there was a problem with data not being flushed. I guess someone is missing a `Close()` call...

xfer/publisher.go
inercia: Done.

tomwilkie:
LGTM; have you had a chance to measure the impact this has on cpu & network?

inercia:
There is a problem wth _gorilla handlers_: they close the connection after processing the request. I'm looking at how to fix this...

inercia:
@tomwilkie I couldn't do it with Gorilla handlers, but [this library](https://github.com/PuerkitoBio/ghost) has done the job. Data compression reduces data usage by 10 (a report with compression is 13 Kbytes, while it was about 130 Kbytes without compression). I cannot get an accurate number for the CPU: it is basically negligible...

app/main.go
```diff
@@ -14,6 +14,8 @@ import (
 	"time"
 
 	"github.com/weaveworks/scope/xfer"
+
+	"github.com/PuerkitoBio/ghost/handlers"
```
tomwilkie: Please put third part imports above weaveworks/scope imports, and below standard library imports.

xfer/publisher.go
```diff
@@ -52,21 +53,30 @@ func NewHTTPPublisher(target, token, id string) (*HTTPPublisher, error) {
 
 // Publish publishes the report to the URL.
 func (p HTTPPublisher) Publish(rpt report.Report) error {
-	var buf bytes.Buffer
-	if err := gob.NewEncoder(&buf).Encode(rpt); err != nil {
+	gzbuf := &bytes.Buffer{}
```
tomwilkie: Minor comment: I prefer gzbuf := bytes.Buffer{} and then using the & operator when passing it to gzip.NewWriter and http.NewRequest

tomwilkie:
2 minor comments, LGTM otherwise.

app/main.go
inercia: Fixed.

xfer/publisher.go
inercia: Fixed.

inercia:
@tomwilkie I've had to move the GZip handler to the router handlers in order to skip the compression for WebSockets (it was breaking the Web client)...

PR #385, opened 2015-08-20
Update the JS tooling

davkal:
* get rid of gulp
* use webpack-dev-server to serve app.js
* express serves static files
* 2 webpack configs: local and production
* for development with backend, websocket URL is injected
* remove unneeded static files


tomwilkie:
Also want to update this line https://github.com/weaveworks/scope/blob/js-tooling/circle.yml#L38

tomwilkie:
LGTM, merge away

PR #386, opened 2015-08-21
Use conntrack to detect short lived connections.

### LINKED ISSUE #356: We need a way to capture short lived connections.
Conntrack will do it, but we can't map it back to a pid (at least on the client side; for servers we could look at the output of lsof)

tomwilkie:
Fixes #356

Status: Good to go.  Review feedback welcome.

Steps are:
1. Refactor the rendering pipeline such that mappers can emit multiple nodes
2. Use conntrack to populate the endpoint table (in addition to procspy).  NB nodes from conntrack _will not_ have a pid
3. Enhance the container render with an extra pipeline which joins containers to endpoints via their IP addresses

### DIFF
--- a/probe/endpoint/conntrack.go
+++ b/probe/endpoint/conntrack.go
@@ -9,7 +9,7 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/weaveworks/scope/test/exec"
+	"github.com/weaveworks/scope/common/exec"
 )
 
 // Constants exported for testing
--- a/probe/endpoint/conntrack.go
+++ b/probe/endpoint/conntrack.go
@@ -1,0 +1,225 @@
+package endpoint
+
+import (
+	"bufio"
+	"encoding/xml"
+	"fmt"
+	"log"
+	"os"
+	"strings"
+	"sync"
+
+	"github.com/weaveworks/scope/test/exec"
+)
+
+// Constants exported for testing
+const (
+	modules          = "/proc/modules"
+	conntrackModule  = "nf_conntrack"
+	XMLHeader        = "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n"
+	ConntrackOpenTag = "<conntrack>\n"
+	TimeWait         = "TIME_WAIT"
+	TCP              = "tcp"
+	New              = "new"
+	Update           = "update"
+	Destroy          = "destroy"
+)
+
+// Layer3 - these structs are for the parsed conntrack output
+type Layer3 struct {
+	XMLName xml.Name `xml:"layer3"`
+	SrcIP   string   `xml:"src"`
+	DstIP   string   `xml:"dst"`
+}
+
+// Layer4 - these structs are for the parsed conntrack output
+type Layer4 struct {
+	XMLName xml.Name `xml:"layer4"`
+	SrcPort int      `xml:"sport"`
+	DstPort int      `xml:"dport"`
+	Proto   string   `xml:"protoname,attr"`
+}
+
+// Meta - these structs are for the parsed conntrack output
+type Meta struct {
+	XMLName   xml.Name `xml:"meta"`
+	Direction string   `xml:"direction,attr"`
+	Layer3    Layer3   `xml:"layer3"`
+	Layer4    Layer4   `xml:"layer4"`
+	ID        int64    `xml:"id"`
+	State     string   `xml:"state"`
+}
+
+// Flow - these structs are for the parsed conntrack output
+type Flow struct {
+	XMLName xml.Name `xml:"flow"`
+	Metas   []Meta   `xml:"meta"`
+	Type    string   `xml:"type,attr"`
+
+	Original, Reply, Independent *Meta `xml:"-"`
+}
+
+// Conntracker uses the conntrack command to track network connections
+type Conntracker struct {
+	sync.Mutex
+	cmd           exec.Cmd
+	activeFlows   map[int64]Flow // active flows in state != TIME_WAIT
+	bufferedFlows []Flow         // flows coming out of activeFlows spend 1 walk cycle here
+}
+
+// NewConntracker creates and starts a new Conntracter
+func NewConntracker(args ...string) (*Conntracker, error) {
+	if !ConntrackModulePresent() {
+		return nil, fmt.Errorf("No conntrack module")
+	}
+	result := &Conntracker{
+		activeFlows: map[int64]Flow{},
+	}
+	go result.run(args...)
+	return result, nil
+}
+
+// ConntrackModulePresent returns true if the kernel has the conntrack module
+// present.  It is made public for mocking.
+var ConntrackModulePresent = func() bool {
+	f, err := os.Open(modules)
+	if err != nil {
+		return false
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, conntrackModule) {
+			return true
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		log.Printf("conntrack error: %v", err)
+	}
+
+	log.Printf("conntrack: failed to find module %s", conntrackModule)
+	return false
+}
+
+// NB this is not re-entrant!
+func (c *Conntracker) run(args ...string) {
+	args = append([]string{"-E", "-o", "xml"}, args...)
+	cmd := exec.Command("conntrack", args...)
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		log.Printf("conntrack error: %v", err)
+		return
+	}
+	if err := cmd.Start(); err != nil {
+		log.Printf("conntrack error: %v", err)
+		return
+	}
+
+	c.Lock()
+	c.cmd = cmd
+	c.Unlock()
+	defer func() {
+		if err := cmd.Wait(); err != nil {
+			log.Printf("conntrack error: %v", err)
+		}
+	}()
+
+	// Swallow the first two lines
+	reader := bufio.NewReader(stdout)
+	if line, err := reader.ReadString('\n'); err != nil {
+		log.Printf("conntrack error: %v", err)
+		return
+	} else if line != XMLHeader {
+		log.Printf("conntrack invalid output: '%s'", line)
+		return
+	}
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -12,12 +12,23 @@ SCOPE_EXPORT=scope.tar
...
+DOCKER_DISTRIB=docker/docker-$(DOCKER_VERSION).tgz
+DOCKER_DISTRIB_URL=https://get.docker.com/builds/Linux/x86_64/docker-$(DOCKER_VERSION).tgz
 
 all: $(SCOPE_EXPORT)
 
-$(SCOPE_EXPORT): $(APP_EXE) $(PROBE_EXE) docker/*
+$(DOCKER_DISTRIB):
+	curl -o $(DOCKER_DISTRIB) $(DOCKER_DISTRIB_URL)
+
+docker/weave:
```
2opremio: I thought it was a product decision to keep scope decoupled from Weave.

tomwilkie: This integration is completely optional.

peterbourgon:
To be clear: short-lived connections are visible, for processes that are running in containers, and talking to other containers?

Makefile
peterbourgon: Well, with this change, it's required to build Scope?

probe/endpoint/reporter.go
```diff
@@ -1,11 +1,12 @@
 package endpoint
 
 import (
-	"fmt"
+	"log"
 	"strconv"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/typetypetype/conntrack"
```
peterbourgon: We should fork this.

tomwilkie: Yes next up is adding live NAT mappings; that will require a fork.

Makefile
tomwilkie: This is just downloading the latest stable weave script to put it in the scope container; if the weave integration is enabled (automatically detected by the scope script), this script is used to get the weave ip addresses of the containers.

tomwilkie: ie weave doesn't need to be present or running to build or run scope.

tomwilkie:
> To be clear: short-lived connections are visible, for processes that are running in containers, and talking to other containers?

Short lived connections will only be visible for containers with a unique IP address in the containers and container image view.  This rules out some of the kubernetes use cases as containers share a network namespace and IP address (although we can do the same thing for pods in the k8n views, so its not too bad), and also rules out --net=host containers (ie scope itself).

These short lived connections will not be visible in the process view (nothing to join them on).  We could back-propagate the links for singleton process in containers (again, not scope).

Container-to-container links are covered, as are/will be pseudo nodes (both random ips and the internet node)

probe/endpoint/conntrack.go
```diff
@@ -0,0 +1,220 @@
...
+		log.Printf("conntrack error: %v", err)
+		return
+	}
+	if line, err := reader.ReadString('\n'); line != ConntrackOpenTag {
+		log.Printf("conntrack invalid output: '%s'", line)
+		return
+	} else if err != nil {
+		log.Printf("conntrack error: %v", err)
+		return
+	}
```
peterbourgon: Should probably switch the order of evaluation for both of these blocks, e.g.

```go
if line, err := reader.ReadString('\n'); err != nil {
    // handle error
} else if line != XMLHeader {
    // handle bad line
}
```

Of course, since your behavior is identical in both cases... not such a big deal.

probe/endpoint/conntrack.go
```diff
@@ -0,0 +1,220 @@
...
+
+	// Now loop on the output stream
+	decoder := xml.NewDecoder(reader)
+	for {
+		var f Flow
```
peterbourgon: You can reduce GC pressure by leaving this out of the for loop.

probe/endpoint/conntrack.go
```diff
@@ -0,0 +1,220 @@
...
+		case "reply":
+			f.Reply = meta
+		case "independent":
+			f.Independent = meta
+		}
+	}
+
+	if f.Original.Layer4.Proto != TCP {
+		return
+	}
```
peterbourgon: Can we have a small comment to why this is here?

probe/endpoint/conntrack_test.go
```diff
@@ -0,0 +1,138 @@
+package endpoint_test
+
+import (
+	"bufio"
+	"encoding/xml"
+	"io"
+	"testing"
+	"time"
+
+	. "github.com/weaveworks/scope/probe/endpoint"
```
peterbourgon: Please don't :)

probe/endpoint/conntrack_test.go
```diff
@@ -0,0 +1,138 @@
...
+	defer func() { exec.Command = oldExecCmd }()
+
+	reader, writer := io.Pipe()
+	exec.Command = func(name string, args ...string) exec.Cmd {
+		return exec.NewMockCmd(reader)
+	}
+
+	conntracker, err := NewConntracker()
+	if err != nil {
+		t.Fatal(err)
```
peterbourgon:
```
ugh ~/src/github.com/weaveworks/scope/probe/endpoint (356-conntrack) go test -v -test.run=TestConntracker
=== RUN   TestConntracker
--- FAIL: TestConntracker (0.00s)
        conntrack_test.go:76: No conntrack module
FAIL
exit status 1
```

Please put this test behind an OS flag somehow. Maybe call the file conntracker_linux_test.go?

probe/endpoint/conntrack_test.go
```diff
@@ -0,0 +1,138 @@
...
+		result := []Flow{}
+		conntracker.WalkFlows(func(f Flow) {
+			f.Original = nil
+			f.Reply = nil
+			f.Independent = nil
+			result = append(result, f)
+		})
+		return result
+	}
+	ts := 1000 * time.Millisecond
```
peterbourgon: Just flagging this for later, I can't test it now due to lack of access to a Linux box, but does the unit test take 1s to complete?

probe/endpoint/reporter.go
```diff
@@ -65,50 +96,68 @@ func (r *Reporter) Report() (report.Report, error) {
 	}
 
 	for conn := conns.Next(); conn != nil; conn = conns.Next() {
-		r.addConnection(&rpt, conn)
+		r.addConnection(&rpt, conn.LocalAddress.String(), conn.RemoteAddress.String(),
+			conn.LocalPort, conn.RemotePort, &conn.Proc)
```
peterbourgon: Maybe there's a nicer way to split this?

probe/endpoint/conntrack_test.go
tomwilkie: Yeah I think this is the one acceptable place for a dot import. In a separate PR I'll update the other tests to use this style. 

https://github.com/golang/go/wiki/CodeReviewComments#import-dot

probe/endpoint/conntrack_test.go
tomwilkie: Ah I meant to mock out access to that file. His should work on a Mac.

probe/endpoint/conntrack_test.go
tomwilkie: Nah this is just a timeout for the polls. It could be much lower (100ms). Will update.

probe/endpoint/reporter.go
tomwilkie: Will do.

probe/endpoint/conntrack_test.go
peterbourgon: Ah, that page gives good guidance: only use it when there is a circular dependency issue in foo_test packages. Do we have a circular dependency here?

tomwilkie: "The import . form can be useful in tests that, due to circular dependencies, cannot be made part of the package being tested:"

The go-way is to have tests part of the package; but as we put the tests in separate packages using the dot import still applied.

probe/endpoint/conntrack_test.go
tomwilkie: Okay thats mocked now, should work on mac.

probe/endpoint/conntrack.go
tomwilkie: the xml parser has some weird behaviour if you don't pass it a fresh instance each time - it ends up appended to the list of metas.

tomwilkie:
I solved #403 by ensure each edge is only included once, by construction.

probe/endpoint/conntrack.go
peterbourgon: Ha. Awesome.

render/mapping.go
```diff
@@ -48,20 +45,20 @@ type PseudoFunc func(srcNodeID, dstNodeID string, srcIsClient bool, local report
 //
 // As with LeafMapFunc, if the final output parameter is false, the node
 // shall be omitted from the rendered topology.
-type MapFunc func(RenderableNode) (RenderableNode, bool)
+type MapFunc func(RenderableNode) RenderableNodes
```
peterbourgon: Update comment, please.

peterbourgon:
After reading this a little bit, very eager to go ahead and merge RenderableNode/NodeMetadata/.... as we discussed...

tomwilkie:
Yeah I've got a WIP branch to get there - https://github.com/weaveworks/scope/tree/357-adjacencies-to-nodemetadata

Will do it tomorrow.

probe/endpoint/conntrack.go
```diff
@@ -0,0 +1,225 @@
...
+import (
+	"bufio"
+	"encoding/xml"
+	"fmt"
+	"log"
+	"os"
+	"strings"
+	"sync"
+
+	"github.com/weaveworks/scope/test/exec"
```
peterbourgon: If exec will be used by non-test packages, I guess it shouldn't be nested under test...

peterbourgon:
Alright, those comments aside, this LGTM-ish. It's a shame that we have to exploit IP-per-container to get this info, but now that I understand it, I think it's reasonable enough.

tomwilkie:
Thanks!  Any suggestions on where to put exec?  Maybe a lib package?
Theres more to follow; we've got bits and pieces of a mock filesystem that
should probably be brought together.

peterbourgon:
> Any suggestions on where to put exec?

common/{exec,foo,bar}?

PR #387, opened 2015-08-21
Improve titles and context of tables in the details pane

### LINKED ISSUE #296: On groupings which traverse hosts (process-by-name or container-by-image), you can't tell which hosts the details come from
In the details panel you see multiple "origin process" tables, and you don't know which is from which host.

I suggest we change the name of the table to be the same as the major (minor) of the node, ie they should be something like "Process Apache (341 - host1)" or "Conatiner Weave (host1)" or something

2opremio:
Fixes #296

### DIFF
--- a/app/router.go
+++ b/app/router.go
@@ -107,10 +107,9 @@ var topologyRegistry = map[string]topologyView{
 		renderer: render.FilterUnconnected{Renderer: render.ProcessWithContainerNameRenderer{}},
 	},
 	"applications-by-name": {
-		human:       "by name",
-		parent:      "applications",
-		renderer:    render.FilterUnconnected{Renderer: render.ProcessNameRenderer},
-		isMultiHost: true,
+		human:    "by name",
+		parent:   "applications",
+		renderer: render.FilterUnconnected{Renderer: render.ProcessNameRenderer},
 	},
 	"containers": {
 		human:    "Containers",
@@ -118,10 +117,9 @@ var topologyRegistry = map[string]topologyView{
 		renderer: render.ContainerRenderer,
 	},
 	"containers-by-image": {
-		human:       "by image",
-		parent:      "containers",
-		renderer:    render.ContainerImageRenderer,
-		isMultiHost: true,
+		human:    "by image",
+		parent:   "containers",
+		renderer: render.ContainerImageRenderer,
 	},
 	"hosts": {
 		human:    "Hosts",
@@ -131,8 +129,7 @@ var topologyRegistry = map[string]topologyView{
 }
 
 type topologyView struct {
-	human       string
-	parent      string
-	renderer    render.Renderer
-	isMultiHost bool // does the view involve information from multiple hosts?
+	human    string
+	parent   string
+	renderer render.Renderer
 }
--- a/app/router.go
+++ b/app/router.go
@@ -107,9 +107,10 @@ var topologyRegistry = map[string]topologyView{
 		renderer: render.FilterUnconnected{Renderer: render.ProcessWithContainerNameRenderer{}},
 	},
 	"applications-by-name": {
-		human:    "by name",
-		parent:   "applications",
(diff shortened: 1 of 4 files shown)


app/router.go
```diff
@@ -129,7 +131,8 @@ var topologyRegistry = map[string]topologyView{
 }
 
 type topologyView struct {
-	human    string
-	parent   string
-	renderer render.Renderer
+	human       string
+	parent      string
+	renderer    render.Renderer
+	isMultiHost bool // does the view involve information from multiple hosts?
```
tomwilkie: Instead of this being set explicitly per view, can this be inferred from the node you've been asked about?  Ie examine all the origin ids for a given renderable node, fetch the node id from their node metadata, and see if you have more than one?

render/detailed_node.go
```diff
@@ -64,23 +64,23 @@ func (r sortableRows) Less(i, j int) bool {
 	}
 }
 
-type tables []Table
+type sortableTables []Table
```
tomwilkie: Why was this rename necessary?

render/detailed_node.go
```diff
@@ -237,20 +238,33 @@ func processOriginTable(nmd report.NodeMetadata) (Table, bool) {
...
+	var (
+		commFound, pidFound bool
+		name, pid           string
+	)
+	if name, commFound = nmd.Metadata[process.Comm]; commFound {
+		title += ` "` + name + `"`
+	}
+	if pid, pidFound = nmd.Metadata[process.PID]; pidFound {
+		title += " (" + pid + ")"
+	}
```
tomwilkie: I'm not a fan of this kind of string construction; I would do it like this:
```
var (
	comm, commFound = nmd.Metadata[process.Comm]
	pid, pidFound = nmd.Metadata[process.PID]
	title string
)
switch {
case commFound && !pidFound:
	title = fmt.Sprintf(`Process "%s"`, comm)
case !commFound && pidFound:
	title = fmt.Sprintf(`Process (%s)`, pid)
case commFound && pidFound:
	title = fmt.Sprintf(`Process "%s" (%s)`, comm, pid)
default:
 	title = "Process"
}
```

render/detailed_node.go
2opremio: It wasn't strictly necessary, I did it for consistency with the `rows` -> `sortableRows`renaming.

app/router.go
2opremio: Sure, I can do that, I avoided it to save a pass on `RendarableNode.Origins` but I guess that won't buy us much.

render/detailed_node.go
2opremio: I like the declaration block, but I am not in favor of systematically covering every case when you can do it incrementally. For me it's more compact and easier to read.

2opremio:
@tomwilkie I've realized that providing Host context is not enough. In the Container Image view, when showing processes from containers, I should also be providing what container they belong to since they can belong to different containers.

I will add that later today.

tomwilkie:
Oh yes good idea.  Take a look at https://github.com/weaveworks/scope/pull/366 to see how I did something similar for process labels.

render/detailed_node.go
```diff
@@ -64,23 +64,40 @@ func (r sortableRows) Less(i, j int) bool {
...
+			if nmd, ok := topology.NodeMetadatas[id]; ok {
+				originHosts[report.ExtractHostID(nmd)] = struct{}{}
+				if id, ok := nmd.Metadata[docker.ContainerID]; ok {
+					originContainers[id] = struct{}{}
+				}
+			}
+		}
+	}
+	multiHost := len(originHosts) > 1
+	multiContainer := len(originContainers) > 1
```
tomwilkie: You could move this check into the loop and break out early.

I'd also put this logic in a separate function.

tomwilkie:
1 comment, otherwise LGTM

PR #389, opened 2015-08-24
Update for new weave status interface; 'intelligently' enable by default

tomwilkie:
Weave 1.1 has a new status json interface, and the old one has been removed.

I'm not worries about backwards compatibility, as the old code was never enabled.

### DIFF
--- a/probe/overlay/weave.go
+++ b/probe/overlay/weave.go
@@ -3,7 +3,6 @@ package overlay
 import (
 	"encoding/json"
 	"fmt"
-	"log"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


probe/overlay/weave.go
```diff
@@ -30,14 +29,45 @@ type Weave struct {
...
 	if err != nil {
 		return nil, err
 	}
-	return &Weave{s}, nil
+	return &Weave{
+		url: s,
+	}, nil
+}
+
+func (w Weave) update() (weaveStatus, error) {
```
tomwilkie: I moved this out of Report() as I was reusing it in Tag(); but then I decided to save that for another change.

probe/overlay/weave.go
```diff
@@ -30,14 +29,45 @@ type Weave struct {
...
+		return result, err
+	}
+	req.Header.Add("Accept", "application/json")
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return result, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
```
peterbourgon: http.StatusOK

probe/overlay/weave.go
```diff
@@ -30,14 +29,45 @@ type Weave struct {
...
+		return result, fmt.Errorf("Weave Tagger: got %d", resp.StatusCode)
+	}
+
+	return result, json.NewDecoder(resp.Body).Decode(&result)
```
peterbourgon: Haha! Weird.

probe/overlay/weave_test.go
```diff
@@ -42,15 +41,18 @@ func TestWeaveTaggerOverlayTopology(t *testing.T) {
...
 	mockWeavePeerName     = "winnebago"
 	mockWeavePeerNickName = "winny"
+	mockResponse = `{
+		"router": {
+			"peers": [{
+				"name": "winnebago",
+				"nickname": "winny"
+			}]
+		}
+	}`
```
peterbourgon: Maybe remove the constants mockWeavePeer[Nick]Name if you're not using them in the response?

peterbourgon:
:+1:

probe/overlay/weave_test.go
tomwilkie: Their used in the checking code above; I've made them also be used in the the response.

PR #396, opened 2015-08-25
Tag containers with Weave and Docker IPs

### LINKED ISSUE #326: Get output of weave ps in scope
So we can tag containers with their weave addresses.

tomwilkie:
Fixes #326

This change involved periodically invoking ```weave ps```, which requires the weave script to be installed in our container, which requires the Docker CLI to be installed in out container.

### DIFF
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -244,3 +244,8 @@ func (c *container) GetNodeMetadata() report.NodeMetadata {
 	}))
 	return result
 }
+
+// ExtractContainerIPs returns the list of container IPs given a NodeMetadata from the Container topology.
+func ExtractContainerIPs(nmd report.NodeMetadata) []string {
+	return strings.Fields(nmd.Metadata[ContainerIPs])
+}
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -26,6 +26,7 @@ const (
 	ContainerCommand = "docker_container_command"
 	ContainerPorts   = "docker_container_ports"
 	ContainerCreated = "docker_container_created"
+	ContainerIPs     = "docker_container_ips"
 
 	NetworkRxDropped = "network_rx_dropped"
 	NetworkRxBytes   = "network_rx_bytes"
@@ -212,6 +213,8 @@ func (c *container) GetNodeMetadata() report.NodeMetadata {
 		ContainerCreated: c.container.Created.Format(time.RFC822),
 		ContainerCommand: c.container.Path + " " + strings.Join(c.container.Args, " "),
 		ImageID:          c.container.Image,
+		ContainerIPs: strings.Join(append(c.container.NetworkSettings.SecondaryIPAddresses,
+			c.container.NetworkSettings.IPAddress), " "),
 	})
 
 	if c.latestStats == nil {
--- a/probe/overlay/weave.go
+++ b/probe/overlay/weave.go
@@ -1,13 +1,19 @@
 package overlay
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
+	"io"
+	"log"
 	"net"
 	"net/http"
 	"net/url"
+	"os/exec"
+	"regexp"
 	"strings"
 
+	"github.com/weaveworks/scope/probe/docker"
 	"github.com/weaveworks/scope/report"
 )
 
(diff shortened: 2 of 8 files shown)


probe/docker/container.go
```diff
@@ -212,6 +213,7 @@ func (c *container) GetNodeMetadata() report.NodeMetadata {
 		ContainerCreated: c.container.Created.Format(time.RFC822),
 		ContainerCommand: c.container.Path + " " + strings.Join(c.container.Args, " "),
 		ImageID:          c.container.Image,
+		ContainerIPs:     c.container.NetworkSettings.IPAddress,
```
2opremio: Two comments:
1. Don't you want to include the `SecondaryIPAddresses` too?
2. I think you should provide a mechanism to easily parse the addresses in case there's more than one. For the docker labels I am encoding them in json.

tomwilkie: 1. yes
2. strings.Fields is more than good enough.

tomwilkie:
@2opremio PTAL

probe/overlay/weave.go
```diff
@@ -83,6 +106,59 @@ func (w Weave) update() (weaveStatus, error) {
...
+
+func (w Weave) tagContainer(r report.Report, containerIDPrefix, macAddress string, ips []string) {
+	for nodeid, nmd := range r.Container.NodeMetadatas {
+		idPrefix := nmd.Metadata[docker.ContainerID][:12]
+		if idPrefix != containerIDPrefix {
+			continue
+		}
+
+		existingIPs := report.MakeIDList(strings.Fields(nmd.Metadata[docker.ContainerIPs])...)
+		existingIPs = existingIPs.Add(ips...)
```
2opremio: I would put `strings.Fields(nmd.Metadata[docker.ContainerIPs])` in a function, say `func ExtractContainerIPs (nmd report.NodeMetadata) []string`, in `docker/container.go`

probe/overlay/weave.go
```diff
@@ -21,8 +27,25 @@ const (
 
 	// WeaveDNSHostname is the ket for the WeaveDNS hostname
 	WeaveDNSHostname = "weave_dns_hostname"
+
+	// WeaveMACAddress is the key for the mac address of the container on the
+	// weave network, to be found in container node metadata
+	WeaveMACAddress = "weave_mac_address"
 )
 
+var weavePsMatch = regexp.MustCompile(`^([0-9a-f]{12}) ((?:[0-9a-f][0-9a-f]\:){5}(?:[0-9a-f][0-9a-f]))(.*)$`)
```
2opremio: This is rather fragile (i.e. the output of `weave ps` change in the future). Have you considered replicating the behaviour from `weave ps` in using the docker API?

render/detailed_node.go
```diff
@@ -297,13 +297,15 @@ func containerOriginTable(nmd report.NodeMetadata, addHostTag bool) (Table, bool
 	rows := []Row{}
 	for _, tuple := range []struct{ key, human string }{
 		{docker.ContainerID, "ID"},
-		{overlay.WeaveDNSHostname, "Weave DNS Hostname"},
 		{docker.ImageID, "Image ID"},
 		{docker.ContainerPorts, "Ports"},
 		{docker.ContainerCreated, "Created"},
 		{docker.ContainerCommand, "Command"},
+		{docker.ContainerIPs, "IP Addresses"},
```
2opremio: I would print one IP per row. One row can only fit a could of them

probe/overlay/weave.go
tomwilkie: This is only done in one place (here), so its not clear breaking this out into a function adds anything.

probe/overlay/weave.go
tomwilkie: ```weave ps``` is a supported interface.  See #1350 for further discussion.

probe/overlay/weave.go
2opremio: It *hides* how you encode the IPs. Plus you would be using it in two places if you render one IP per row in the details pane to make sure they fit.

tomwilkie:
@2opremio PTAL

probe/overlay/weave.go
2opremio: Oh, I didn't know it was supported, great then.

2opremio:
LGTM

PR #399, opened 2015-08-26
Moar functional: Topology Merge methods don't mutate the receiver

peterbourgon:
This PR changes all Merge methods in package report to return a copy, rather than mutate the receiver. The immediate impetus is that it simplifies the rendering pipeline for the DSL. But it's part of a broader effort to make all of our transformation code more functional, and therefore more coherent.

As a follow-up, we should change the RenderableNode Merge methods to have the same semantics. But  those types may not be around for much longer, so I didn't include it in this PR.

/cc @paulbellamy @tomwilkie @2opremio


report/topology.go
```diff
@@ -16,18 +14,105 @@ type Topology struct {
 	NodeMetadatas
 }
 
-// Adjacency is an adjacency-list encoding of the topology. Keys are node IDs,
-// as produced by the relevant MappingFunc for the topology.
+// MakeTopology gives you a Topology.
+func MakeTopology() Topology {
```
paulbellamy: Why rename this?

report/report.go
```diff
@@ -58,18 +58,49 @@ type Report struct {
...
+		ContainerImage: MakeTopology(),
+		Host:           MakeTopology(),
+		Overlay:        MakeTopology(),
 		Sampling:       Sampling{},
 		Window:         0,
 	}
 }
 
+// Copy returns a value copy of the report.
+func (r Report) Copy() Report {
```
paulbellamy: Why do we need all the `Copy()` fns? `r` is already a copy, no?

report/topology.go
peterbourgon: MakeFoo returns Foo, NewFoo returns *Foo.

report/report.go
peterbourgon: r itself is a copy, but r's fields have reference semantics and need to be manually copied.

paulbellamy: Oh, bleh. Because they're all maps.

tomwilkie:
@peterbourgon I get a blank UI when I load this up; looks like the test failure is indicative of something.

Will investigate further.

tomwilkie:
```/api/report``` is reporting an empty report with this change.

**Comment by tomwilkie** [comment]
I'd use this opportunity to tidy up the taggers too - the interface implies it shouldn't mutate the report, but return a fresh one; however they all seem to mutate the report.

We could consider using some third party immutable lists and maps, if only go had generics...

**Comment by tomwilkie** [comment]
This seems to fix it:
```
diff --git a/probe/main.go b/probe/main.go
index d636b38..9a27d69 100644
--- a/probe/main.go
+++ b/probe/main.go
@@ -183,7 +183,7 @@ func main() {
                                        if err != nil {
                                                log.Printf("error generating report: %v", err)
                                        }
-                                       r.Merge(newReport)
+                                       r = r.Merge(newReport)
                                }
                                r = Apply(r, taggers)
```

Issue #372, opened 2015-08-19
More intuitive detail pane


2opremio:
* Include Docker labels
* Fix bugs (origin container, etc.)
* More intelligent data structure, better rendering; backend produce HTML

peterbourgon:
Other tickets related to the detail pane...

- [On groupings which traverse hosts . . . you can't tell which hosts the details come from](https://github.com/weaveworks/scope/issues/296)
- [Connections and Connection Details seem redundant](https://github.com/weaveworks/scope/issues/322)
- [Table type should be more flexible](https://github.com/weaveworks/scope/issues/301)
- [Empty fields in details pane look broken](https://github.com/weaveworks/scope/issues/370)

2opremio:
Closing this ticket since the only remaining one is #301 which we are tracking separately

2opremio:
Actually ... reopening since there's no ticket for including docker labels.

PR #400, opened 2015-08-26
Report docker container labels and render them in the details pane

2opremio:
Addresses part of #372


tomwilkie:
Failure is due to 
```
./probe/docker/container.go:206:1: exported function ExtractContainerLabels should have comment or be unexported
```

probe/docker/container.go
```diff
@@ -215,6 +223,7 @@ func (c *container) GetNodeMetadata() report.NodeMetadata {
 		ImageID:          c.container.Image,
 		ContainerIPs: strings.Join(append(c.container.NetworkSettings.SecondaryIPAddresses,
 			c.container.NetworkSettings.IPAddress), " "),
+		ContainerLabels: string(labels),
```
tomwilkie: After discussion with @peterbourgon, we'd prefer if this was encoded as multiple entries in the nodemetadata map - where the key in the map is of the form docker_container_label_<label key>.

tomwilkie: (This will make it possible to filter on them in the DSL)

tomwilkie: (And make them merge ~correctly)

tomwilkie:
Also, docker images can have labels, as can the docker daemon.  We should probably fetch these and attach them to the docker image and host topology.  In which case, make the prefix ```docker_label_```

probe/docker/container.go
```diff
@@ -206,6 +207,7 @@ func (c *container) GetNodeMetadata() report.NodeMetadata {
 	c.RLock()
 	defer c.RUnlock()
 
+	labels, _ := json.Marshal(c.container.Config.Labels)
```
peterbourgon: Don't ignore this error.

probe/docker/container.go
```diff
@@ -245,6 +248,13 @@ func (c *container) GetNodeMetadata() report.NodeMetadata {
 	return result
 }
 
+// ExtractContainerLabels returns the list of Docker container labels given a NodeMetadata from the Container topology.
+func ExtractContainerLabels(nmd report.NodeMetadata) map[string]string {
+	result := make(map[string]string)
```
peterbourgon: [Avoid make and new](https://docs.google.com/document/d/18UTcmy8VoecEaYNb0DrI8NZrI1uJbq-XYDEko0l5Ny4/edit), preferring struct-literal initialization when possible, i.e.

```go
result := map[string]string{}
```

peterbourgon: (This will go away with Tom's suggestion anyway, but just for the record...)

probe/docker/container.go
peterbourgon: Never ignore errors.

(This will go away with Tom's suggestion, but just for the record...)

render/detailed_node.go
```diff
@@ -313,6 +313,18 @@ func containerOriginTable(nmd report.NodeMetadata, addHostTag bool) (Table, bool
 		rows = append(rows, Row{Key: "IP Address", ValueMajor: ip, ValueMinor: ""})
 	}
 
+	// Add labels in alphabetical order
+	labels := docker.ExtractContainerLabels(nmd)
+	labelKeys := make([]string, 0, len(labels))
+	for k := range labels {
+		labelKeys = append(labelKeys, k)
+	}
```
peterbourgon: You can do this more directly, e.g.

```go
labelKeys := make([]string, len(labels))
for i, k := range labels {
    labelKeys[i] = k
}
```

(This will go away with Tom's suggestion anyway, but just for the record...)

probe/docker/container.go
2opremio: Uhm, even if they can't have *by construction* and you are checking that in unit tests?

peterbourgon: Yep.

render/detailed_node.go
2opremio: `i` is a string, I don't think you can do that.

peterbourgon: Oh! I'm so sorry, I misread `labels` type. You're totally right. Disregard my comment.

probe/docker/container.go
2opremio: OK, this seems more reasonable :)

2opremio:
@tomwilkie Unfortunately the fsouza client doesn't support Docker daemon labels yet: see [the message from my last commit](https://github.com/weaveworks/scope/commit/a8c72974424300b50c4177e8efa0337493a93ae2)

probe/docker/container.go
```diff
@@ -22,11 +22,12 @@ import (
...
-	ContainerCommand = "docker_container_command"
-	ContainerPorts   = "docker_container_ports"
-	ContainerCreated = "docker_container_created"
-	ContainerIPs     = "docker_container_ips"
+	ContainerName        = "docker_container_name"
+	ContainerCommand     = "docker_container_command"
+	ContainerPorts       = "docker_container_ports"
+	ContainerCreated     = "docker_container_created"
+	ContainerIPs         = "docker_container_ips"
+	ContainerLabelPrefix = "docker_container_label_"
```
tomwilkie: Can we drop container from the name?  its used for images too.

probe/docker/container.go
```diff
@@ -242,6 +244,7 @@ func (c *container) GetNodeMetadata() report.NodeMetadata {
 		CPUUsageInKernelmode: strconv.FormatUint(c.latestStats.CPUStats.CPUUsage.UsageInKernelmode, 10),
 		CPUSystemCPUUsage:    strconv.FormatUint(c.latestStats.CPUStats.SystemCPUUsage, 10),
 	}))
+
```
tomwilkie: Unnecessary

tomwilkie:
Assign it back to me when the tests pass.

probe/docker/container.go
2opremio: Oh, thanks, that shouldn't even be there anymore, there's a new `label.go` file for this.

tomwilkie: Which I think you forgot to add?

2opremio: True! Just fixed it, sorry.

tomwilkie:
LGTM

Issue #373, opened 2015-08-19
Ensure we handle the directionality correctly in EdgeMetadata


tomwilkie:
Follow up from  #355

Issue #249, opened 2015-06-16
Pseudo nodes are unclickable


tomwilkie:
Is this intentional?

davkal:
> we don't have any NodeMetadata for them, so nothing to render in the side panel

This was also my reasoning. Nothing to see there.
Under which circumstances is there any meta data?

tomwilkie:
Pseudo nodes in the containers view have a ton of metadata.

peterbourgon:
By definition pseudonodes are nodes we have to invent because we have them as a destination in an edge, but don't have any other data (metadata) for them. If a node has metadata, by definition it is not a pseudonode.

tomwilkie:
Are you suggesting uncontained should not be a pseudo node?

Note that even under you definition, we still have useful info we should
show the user - for instance, the Internet pseudo node has a mass be list
of ip addresses we could show, and other pseudo nodes has port numbers we
could show.

peterbourgon:
Agh, uncontained. Right. For the record, we're abusing the pseudo flag for that one.

But, point taken. There's always going to be *some* information we can display, actually.

davkal:
Which nodes in which topologies do not have any data?

I'm asking because we could show an actionable message like *install a probe* in the details panel in that case.

tomwilkie:
Not much in the process topology to show.  And "the internet" never has any info.  Really its only the uncontained node that has meaningful info I think.  OTOH paul's recent connections stuff might work for all pseudo nodes...

PR #401, opened 2015-08-26
Only group processes not in containers AND doing network IO into uncontained nodes.

tomwilkie:
With the re-enablement of the selection of pseudo nodes (#249), it turns out our 'Uncontained' nodes in the containers view contains all processes.  This isn't particularly useful; we only want to show the 'Uncontained' node if there is a process on the machine with network connections with isn't in a container.

The changes to the bandwidth figures are due to #373, which is not a big deal right now as the bandwidth figures aren't enabled by default, but I'll fix that next.

### DIFF
--- a/render/render.go
+++ b/render/render.go
@@ -269,27 +269,79 @@ func (m LeafMap) EdgeMetadata(rpt report.Report, srcRenderableID, dstRenderableI
 	return metadata
 }
 
-// FilterUnconnected is a Renderer which filters out unconnected nodes.
-type FilterUnconnected struct {
+// CustomRenderer allow for mapping functions that recived the entire topology
+// in one call - useful for functions that need to consider the entire graph
+type CustomRenderer struct {
+	RenderFunc func(RenderableNodes) RenderableNodes
 	Renderer
 }
 
-// Render produces a set of RenderableNodes given a Report
-func (f FilterUnconnected) Render(rpt report.Report) RenderableNodes {
-	return OnlyConnected(f.Renderer.Render(rpt))
+// Render implements Renderer
+func (c CustomRenderer) Render(rpt report.Report) RenderableNodes {
+	return c.RenderFunc(c.Renderer.Render(rpt))
 }
+
+// IsConnected is the key added to NodeMetadata by ColorConnected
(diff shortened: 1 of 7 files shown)


render/render.go
```diff
@@ -294,3 +300,14 @@ func OnlyConnected(input RenderableNodes) RenderableNodes {
...
+// Render implements Renderer
+func (f Filter) Render(rpt report.Report) RenderableNodes {
+	output := RenderableNodes{}
+	for id, node := range f.Renderer.Render(rpt) {
+		if f.f(node) {
+			output[id] = node
+		}
+	}
+	return output
+}
```
peterbourgon: I made this change in #399, but can you keep types and their methods grouped?

render/topologies.go
```diff
@@ -82,8 +82,19 @@ var ProcessNameRenderer = Map{
 // graph by merging the process graph and the container topology.
 var ContainerRenderer = MakeReduce(
 	Map{
-		MapFunc:  MapProcess2Container,
-		Renderer: ProcessRenderer,
+		MapFunc: MapProcess2Container,
+		Renderer: MakeReduce( // We only want processes in container _or_ processes with network connections
+			FilterUnconnected{
+				Renderer: ProcessRenderer,
+			},
```
peterbourgon: This appears to be a change in the behavior, previously we weren't filtering unconnected, now we are?

tomwilkie: Indeed.  We filter unconnected process and we filter processes not in containers and merge the result, to get the set of processes either in a container or connected to something.

render/render.go
tomwilkie: Done

peterbourgon:
Simple enough. Looks like the new Filter is unit tested by the detailed_node_test? If so, :+1:

tomwilkie:
This now includes a fix for #403 - by construction, we only include edges from the endpoint topology once, guaranteeing we don't over count the edgemetadata.

peterbourgon:
It's difficult to follow the rendering pipeline. I wish for something more procedural. That said, I can understand what's happening here. LGTM.

PR #402, opened 2015-08-26
Fail if want != have!

tomwilkie:
git blame says it was me...


tomwilkie:
@paulbellamy says LGTM

Issue #364, opened 2015-08-17
Consider doing async reverse dns lookups in probe, breaking out the internet node into something more interesting.


tomwilkie:
Esp when running apps in the cloud, would be nice to show which components are calling out to which services (s3, ses, etc)

peterbourgon:
Ooh, I like it: we could just tag remote nodes with a variety of whois-y lookup data...

```json
"123.45.6.78;80": {
    "country": "UK",
    "city": "Birmingham",
    "owner": "HP Foods LLC"
}
```

awh:
Other metadata types that could be interesting - geoIP, [autonomous system](https://en.wikipedia.org/wiki/Autonomous_system_%28Internet%29)

tomwilkie:
Design considerations:
- we don't want to hammer any backend (dns) service, so we'll want some caching
- we don't want the lookups in dns to block the probes main thread

peterbourgon:
For geolocation stuff, it may be better to ship a GeoIP database with every Scope release, rather than pinging some remote service.

inercia:
@peterbourgon there are some GeoIP databases [publicly available](http://dev.maxmind.com/geoip/legacy/geolite/). Maybe Scope could download it on the first run and keep it until a new version is available...

peterbourgon:
The probes would need to download it, and you may have lots of probes in your infrastructure. If I were an ops guy, I would be very unhappy with a monitoring agent reaching out to the internet and downloading some megabytes from a server when it first boots. I would be very unhappy with it reaching out to the internet at all, actually, unless I explicitly turned some feature on.

inercia:
@peterbourgon you are right: we should include the GeoIP database in the Scope image. Maybe we  could add an option for overriding the embeded database in the future...

inercia:
And what about adding this kind of information not in the probe but in the server? The result of geolocating an IP would be the same, and it would be easier to keep an up-to-date database in the server...

tomwilkie:
The app server might not be running in the same context as the probes, and might not have access to the same dns (for instance).

Lets worry less about geoip for now, and more about reverse dns lookups.

awh:
> Lets worry less about geoip for now, and more about reverse dns lookups.

Yes sorry about that - I didn't mean to pollute this PR, just capture the conversation we had around it the other day...

inercia:
Some implementation questions:

[...]

PR #404, opened 2015-08-27
reverse resolutions for remote side of connections.

inercia:
Async reverse dns lookups in probe that creates nodes for the remote side of connections.

It's part of the solution for #364

TODO:

- [x] unit test
- [x] update detailed_node.go to display the remote name


probe/endpoint/resolver.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Get the reverse resolution for a
+func (r *reverseResolver) Get(ip net.IP) (string, error) {
+	address := ip.String()
+	val, err := r.cache.Get(address)
+	if err != nil {
+		if err == gcache.NotFoundKeyError {
+			// we trigger a asynchronous reverse resolution when not cached, but we return
+			// immediately so the client does not block...
+			r.addresses <- address
```
tomwilkie: pls make this a non blocking channel put 

https://gobyexample.com/non-blocking-channel-operations

inercia: I didn't know Go had this non-bocking put... Fixed.

tomwilkie:
#411 is merged, so you should now be able to put the reverse dns lookups on pseudo nodes and plumb this all the way through to the UI.

probe/endpoint/reporter.go
```diff
@@ -98,11 +101,21 @@ func (r *Reporter) addConnection(rpt *report.Report, c *procspy.Connection) {
...
 
+	// Add the remote address iff we have a reverse resolution available
+	if _, found := rpt.Address.NodeMetadatas[remoteAddressNodeID]; !found {
+		if raddr, err := r.revResolver.Get(c.RemoteAddress); err == nil {
+			rpt.Address.NodeMetadatas[remoteAddressNodeID] = report.MakeNodeMetadataWith(map[string]string{
+				Name: raddr,
+				Addr: c.RemoteAddress.String(),
+			})
+		}
+	}
```
tomwilkie: We're a fan of less nesting, ie:

```
if _, found := rpt.Address.NodeMetadatas[remoteAddressNodeID]; !found {
    raddr, err := r.revResolver.Get(c.RemoteAddress);
    if err != nil {
        continue
    }
	rpt.Address.NodeMetadatas[remoteAddressNodeID] = report.MakeNodeMetadataWith(map[string]string{
		Name: raddr,
		Addr: c.RemoteAddress.String(),
	})
}
```

tomwilkie: Also, please use the new topology.WithNode, as in this case you might be accidentally overwriting an existing node in the topology.  WithNode will make sure its merge.

probe/endpoint/reporter.go
```diff
@@ -14,6 +14,7 @@ import (
 
 // Node metadata keys.
 const (
+	Name = "name"
```
tomwilkie: I seem to remember @peterbourgon didn't want this for some reason.  Peter?

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,71 @@
...
+
+// NewReverseResolver starts a new reverse resolver that
+// performs reverse resolutions and caches the result.
+func NewReverseResolver(cacheLen int) *ReverseResolver {
+	r := ReverseResolver{
+		addresses: make(chan string, raddrBacklog),
+		cache:     gcache.New(cacheLen).LRU().Expiration(rAddrCacheExpiration).Build(),
+	}
+
+	go func() {
```
tomwilkie: Please pull this out into a separate function, like in https://github.com/weaveworks/scope/blob/master/probe/endpoint/conntrack.go#L78

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,71 @@
...
+		rate := time.Second / 10
+		throttle := time.Tick(rate)
+		for address := range r.addresses {
+			<-throttle // rate limit our DNS resolutions
```
tomwilkie: Nice :-)

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,71 @@
...
+	"time"
+
+	"github.com/bluele/gcache"
+)
+
+// Default cache length
+const RAddrCacheLen = 500
+
+const raddrBacklog = 1000
+const rAddrCacheExpiration = 30 * time.Minute
```
tomwilkie:
```
const (
...
)
```

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,71 @@
...
+			select {
+			case r.addresses <- address:
+			default:
+			}
+		} else {
+			log.Printf("could not GET %s from cache: %v", address, err)
+		}
+		return "", err
+	}
+	return val.(string), nil
```
tomwilkie: Prefer less nesting:

```
if err == nil {
    return val.(string), nil
}
    
if err == gcache.NotFoundKeyError {
	// we trigger a asynchronous reverse resolution when not cached
	select {
	case r.addresses <- address:
	default:
	}
} else {
	log.Printf("could not GET %s from cache: %v", address, err)
}
```

tomwilkie: Also consider not logging this, as it could log on every connection, and there can be lots!

probe/main.go
```diff
@@ -40,6 +40,7 @@ func main() {
 		dockerInterval     = flag.Duration("docker.interval", 10*time.Second, "how often to update Docker attributes")
 		dockerBridge       = flag.String("docker.bridge", "docker0", "the docker bridge name")
 		weaveRouterAddr    = flag.String("weave.router.addr", "", "IP address or FQDN of the Weave router")
+		revResolvLen       = flag.Int("raddr.len", endpoint.RAddrCacheLen, "reverse resolution cache length")
```
tomwilkie: We prefer less configurable on scope.  I think its fine for this to be a const and not user controllable.

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,71 @@
...
+// performs reverse resolutions and caches the result.
+func NewReverseResolver(cacheLen int) *ReverseResolver {
+	r := ReverseResolver{
+		addresses: make(chan string, raddrBacklog),
+		cache:     gcache.New(cacheLen).LRU().Expiration(rAddrCacheExpiration).Build(),
+	}
+
+	go func() {
+		rate := time.Second / 10
+		throttle := time.Tick(rate)
```
tomwilkie: 1 line:
```
throttle := time.Tick(time.Second / 10)
```

tomwilkie:
We'll also need unit tests for the resolver.

probe/endpoint/reporter.go
peterbourgon: Yes, I understand this change is just elevating a string literal to a const, but "name" is too generic to be reified in this way. What kind of name is it? Can we also change the string literal, or do we rely on that ambiguity in other places (I recall that we might)? If the answers to these questions are too difficult, as I suspect the might be, better to leave it as a string literal.

probe/main.go
peterbourgon: +1. And if you decide to leave it, please make the var name reflect the flag name, e.g. `raddrLen`.

probe/endpoint/reporter.go
inercia: I understand that having a _Name_ constant is not very nice, but I don't think having a structure like this:

```
			remoteNode = report.MakeNodeMetadataWith(map[string]string{
				"name": revRemoteAddr,
				Addr:   remoteAddr,
			})
```

where "name" is the only literal, does not make sense either, especially when some other parts of the code depend on that constant. Maybe we should standardise these names with something like:

 ```
			remoteNode = report.MakeNodeMetadataWith(map[string]string{
				NodeMetadataName: revRemoteAddr,
				NodeMetadataAddr: remoteAddr,
			})
```

File: probe/endpoint/resolver.go (hunk 4 above)
Comment (inercia) [inline probe/endpoint/resolver.go]: Done.

File: probe/endpoint/reporter.go (hunk 2 above)
Comment (inercia) [inline probe/endpoint/reporter.go]: Done.

File: probe/endpoint/resolver.go (hunk 6 above)
Comment (inercia) [inline probe/endpoint/resolver.go]: Done.

File: probe/main.go (hunk 8 above)
Comment (inercia) [inline probe/main.go]: I'll remove the config option...

File: probe/endpoint/resolver.go (hunk 9 above)
Comment (inercia) [inline probe/endpoint/resolver.go]: Done.

File: probe/endpoint/reporter.go (hunk 3 above)
Comment (peterbourgon) [inline probe/endpoint/reporter.go]: Yes, if you can detangle the web of dependency and use a consistent constant like NodeMetadataName, please do so. 

But it may be (and I fear that) the "name" literal is e.g. written as if it were NodeMetadataName, but then later read as if it were SomeOtherKindOfName. And things just kind of work, but accidentally. If that's true, I think it's much better to leave it as a string literal. At least that way we know it's bad, and we're not tricking ourselves into a false sense of safety.

**Comment by tomwilkie** [comment]
Test failure due to:

```
./probe/endpoint/resolver.go:11:2: exported const RAddrCacheLen should have comment (or a comment on this block) or be unexported
```

File: probe/main.go [hunk 10]
```diff
@@ -104,7 +104,8 @@ func main() {
 	}
 
 	var (
-		endpointReporter = endpoint.NewReporter(hostID, hostName, *spyProcs, *useConntrack)
+		revResolver      = endpoint.NewReverseResolver(endpoint.RAddrCacheLen)
+		endpointReporter = endpoint.NewReporter(hostID, hostName, *spyProcs, *useConntrack, revResolver)
```
tomwilkie: Given cache len is now not configurable, lets roll the initialisation of this in to NewReporter and make RAddrCacheLen 'private'

tomwilkie:
Looking good; I'll leave the final comment to Peter but once resolved, LGTM

probe/endpoint/reporter.go
```diff
@@ -132,6 +135,11 @@ func (r *Reporter) addConnection(rpt *report.Report, localAddr, remoteAddr strin
 
 	// Update address topology
 	{
+		revRemoteAddr := ""
+		if r, err := r.revResolver.Get(remoteAddr); err != nil {
+			revRemoteAddr = r
+		}
+
```
tomwilkie: Scratch that. It's bad practice to put empty values in the nose metadata, so move this section down and make it:

```
name, err := r...get(...)
if err != nill {
    Remotenode = remotenode.
}
```

probe/main.go
inercia: Yes, I'll do that...

peterbourgon:
Just waiting for the WIP prefix to go away, and those two checkboxes to be checked :)

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,69 @@
...
+			continue
+		}
+		if len(names) > 0 {
+			r.cache.Set(address, names[0])
+		}
+	}
+}
+
+// Stop the async reverse resolver
+func (r *ReverseResolver) Stop() {
```
tomwilkie: Is this called anywhere?

inercia: No, but we should... I think we should implement a proper, ordered shutdown in the probe. In some cases it has not even responded to `Ctrl-C`s, so maybe we should invoke `Stop()`s in `main.go` after the `<-interrupt()`...

tomwilkie: endpoint/reporter.go has a Stop method; this should be called from there.

File a ticket to ensure the report.Stop() is called by the probes main.

inercia: Done.

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+	}
+	return "", err
+}
+
+func (r *ReverseResolver) run() {
+	throttle := time.Tick(time.Second / 10)
+	for address := range r.addresses {
+		<-throttle // rate limit our DNS resolutions
+		names, err := net.LookupAddr(address)
```
tomwilkie: At this point, we should lookup the address in the cache, as there is a race between the goroutine calling Get, not finding it, the putting it on the queue and this goroutine.  Imagine if a bunch of "foo.bar.com" got passed to Get in quick succession.

inercia: I think the cache we are using is supposed to be thread-safe, so there shouldn't be any problem with Getting from one goroutine while Putting from another one... Getting 1000 times should not have any effect...

tomwilkie: Thats not what I'm referring to; I'm refer to the fact we enqueue a request on the channel after checking the cache, then some time later do the request.  In the time between the two we might have already put the entry in the cache (from a previous request).

Therefore we need to check the cache a second time before actually doing the request.

inercia: Oh, I understand. I'll check if the name is name is already in the cache before doing the reverse resolution.

inercia:
@tomwilkie what should I do with the reverse address in `detailed_node.go`? should I replace IPs by names when available? or maybe should I show those names in some other way?

tomwilkie:
@inercia yes I'd try and replace IP addresses in the connections table, in connectionDetailsRows()

https://github.com/weaveworks/scope/blob/master/render/detailed_node.go#L206

inercia:
@tomwilkie the problem with doing a `s/IP/name/` is that some names are rather long (eg, `db3msgr5011705.gateway.messenger.live.com.`, `c2-54-163-163-128.compute-1.amazonaws.com.`) so I'm concerned about the result in the UI...

tomwilkie:
The UI will truncate them and offer tooltips (recent change just went in, you'll need to rebase).

I wouldn't be too worries about it.

inercia:
This is how it looks like now:

addresses1

There are scroll bars in the details window now... :confused:

tomwilkie:
@davkal do you know why this is happening?

Álvaro, once you've rebased you may need to do a make static.

davkal:
@tomwilkie never seen this. @inercia which browser/OS is this?

inercia:
@tomwilkie @davkal chrome://version/
```
Google Chrome	45.0.2454.85 (Official Build) (64-bit)
Revision	4321fbd965a0cfb327829005e6054c07669f81d5-refs/branch-heads/2454@{#439}
OS	Linux 
Blink	537.36 (@201276)
JavaScript	V8 4.5.103.29
Flash	18.0.0.233
User Agent	Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/45.0.2454.85 Safari/537.36
```

I'll try again in a few secs after a `make static`...

inercia:
I rebased again and did a `make clean scope_ui_build.tar static all`, and I'm still getting the same result...

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+const (
+	rAddrCacheLen        = 500 // Default cache length
+	rAddrBacklog         = 1000
+	rAddrCacheExpiration = 30 * time.Minute
+)
+
+// ReverseResolver is a caching, reverse resolver
+type ReverseResolver struct {
```
peterbourgon: Since this is only ever used within the endpoint package, can it be unexported, i.e. `type reverseResolver struct`? (That'll mean you need a resolver_internal_test.go for unit tests, but that's fine.)

probe/endpoint/reporter.go
```diff
@@ -64,12 +65,16 @@ func NewReporter(hostID, hostName string, includeProcesses bool, useConntrack bo
 			log.Printf("Failed to start natMapper: %v", err)
 		}
 	}
+
+	revRes := NewReverseResolver(rAddrCacheLen)
```
peterbourgon: Why have a parameter if you just pass a const default?

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,72 @@
...
+
+// NewReverseResolver starts a new reverse resolver that
+// performs reverse resolutions and caches the result.
+func NewReverseResolver(cacheLen int) *ReverseResolver {
+	r := ReverseResolver{
+		addresses: make(chan string, rAddrBacklog),
+		cache:     gcache.New(cacheLen).LRU().Expiration(rAddrCacheExpiration).Build(),
+	}
+
+	go r.run()
```
peterbourgon: Please call this `loop` for consistency.

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,72 @@
...
+	return &r
+}
+
+// Get the reverse resolution for an IP address if already in the cache, gcache.NotFoundKeyError otherwise
+// Note: it returns one of the possible names that can be obtained for that IP
```
peterbourgon: Please wrap all comments at 80col.

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return "", err
+}
+
+func (r *ReverseResolver) run() {
+	throttle := time.Tick(time.Second / 10)
+	for address := range r.addresses {
+		<-throttle                     // rate limit our DNS resolutions
+		_, err := r.cache.Get(address) // and check if the answer is already in the cache
+		if err == nil {
```
peterbourgon:
```go
if _, err := r.cache.Get(address); err == nil {
```

probe/endpoint/resolver_test.go
```diff
@@ -0,0 +1,40 @@
+package endpoint_test
+
+import (
+	"testing"
+
+	"github.com/weaveworks/scope/probe/endpoint"
+	"time"
```
peterbourgon: Please group stdlib imports together.

probe/endpoint/resolver_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	// first time: no names are returned for our reverse resolutions
+	for _, v := range tests {
+		if name, err := revRes.Get(v.ip); name != "" || err == nil {
+			t.Errorf("we didn't get an error, or the cache was not empty, hen trying to resolve '%s'", v.ip)
+		}
+	}
+
+	// after some time, the async resolver should have the names in the cache
+	time.Sleep(500 * time.Millisecond)
```
peterbourgon: Hmm, can we make this deterministic, to avoid unit tests sleeping for half a second? That may require extra hooks in the reverse resolver...

probe/endpoint/resolver_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	// so, if we check again these IPs, we should have the names now
+	for _, v := range tests {
+		name, err := revRes.Get(v.ip)
+		if err != nil {
+			t.Errorf("we got an error when trying to resolve '%s': %s", v.ip, err)
```
peterbourgon: This can be more terse.

```go
t.Errorf("%s: %v", v.ip, err)
```

probe/endpoint/resolver_test.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+		if name != v.name {
+			t.Errorf("got %s for '%s', but expected '%s'", name, v.ip, v.name)
```
peterbourgon: Likewise we prefer a more standard form:

```go
if want, have := v.name, name; want != have {
    t.Errorf("want %q, have %q", want, have)
}
```

render/detailed_node.go
```diff
@@ -205,8 +205,11 @@ func OriginTable(r report.Report, originID string, addHostTags bool, addContaine
 
 func connectionDetailsRows(topology report.Topology, originID string) []Row {
 	rows := []Row{}
-	labeler := func(nodeID string) (string, bool) {
+	labeler := func(nodeID string, meta map[string]string) (string, bool) {
 		if _, addr, port, ok := report.ParseEndpointNodeID(nodeID); ok {
+			if name, found := meta["name"]; found {
```
peterbourgon:
```go
if name, ok := meta["name"]; ok {
```

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,72 @@
...
+			continue
+		}
+		names, err := net.LookupAddr(address)
+		if err != nil {
+			continue
+		}
+		if len(names) > 0 {
+			r.cache.Set(address, names[0])
```
peterbourgon: Please trim the trailing `.`, if it exists...

probe/endpoint/resolver_test.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func TestReverseResolver(t *testing.T) {
+	revRes := endpoint.NewReverseResolver(100)
+	tests := []struct {
+		ip   string
+		name string
+	}{
+		{"8.8.8.8", "google-public-dns-a.google.com."},
+		{"8.8.4.4", "google-public-dns-b.google.com."},
```
peterbourgon: This unit test will fail if the host isn't internet-connected, or if Google changes their DNS infrastructure.

I suppose a good way to fix this would be to mock out net.LookupAddr, and swap it at the start of the test.

probe/endpoint/resolver_test.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	// first time: no names are returned for our reverse resolutions
+	for _, v := range tests {
+		if name, err := revRes.Get(v.ip); name != "" || err == nil {
+			t.Errorf("we didn't get an error, or the cache was not empty, hen trying to resolve '%s'", v.ip)
```
peterbourgon: `'%s'` in a format string is better spelled `%q`.

probe/endpoint/reporter.go
peterbourgon: Also, name consistency with conntracker and natmapper would lead me to suggest revresolver.

inercia: I left the configurable cache length because an important part of the reverse resolver is the cache, and that is configurable with that parameter, so I was reluctant to hardcode that value with a constant... but it is not a big deal, I will use a constant.

probe/endpoint/resolver.go
inercia: Done.

probe/endpoint/resolver.go
inercia: Done.

probe/endpoint/resolver.go
inercia: Done.

probe/endpoint/resolver_test.go
inercia: This is the result of automatic imports in IntelliJ... fixed.

probe/endpoint/resolver_test.go
inercia: Too terse maybe, but it's ok...

probe/endpoint/resolver_test.go
inercia: Done.

render/detailed_node.go
inercia: Done.

probe/endpoint/resolver.go
inercia: Done.

probe/endpoint/resolver_test.go
inercia: Done.

probe/endpoint/resolver_test.go
inercia: Done.

probe/endpoint/resolver.go
```diff
@@ -0,0 +1,85 @@
...
+	val, err := r.cache.Get(address)
+	if err == nil {
+		return val.(string), nil
+	}
+	if err == gcache.NotFoundKeyError {
+		request := revResRequest{address: address, done: make(chan struct{})}
+		// we trigger a asynchronous reverse resolution when not cached
+		select {
+		case r.addresses <- request:
+			if wait {
```
tomwilkie: wait is just for testing right?

probe/endpoint/resolver_internal_test.go
```diff
@@ -0,0 +1,41 @@
...
+	// first time: no names are returned for our reverse resolutions
+	for ip := range tests {
+		if have, err := revRes.Get(ip, true); have != "" || err == nil {
+			t.Errorf("we didn't get an error, or the cache was not empty, when trying to resolve '%q'", ip)
+		}
+	}
+
+	// so, if we check again these IPs, we should have the names now
+	for ip, want := range tests {
+		have, err := revRes.Get(ip, true)
```
tomwilkie: Instead of wait, we could use test.Poll (like our other tests).  This would keep the API simpler.

probe/endpoint/resolver_internal_test.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"errors"
+	"testing"
+)
+
+func TestReverseResolver(t *testing.T) {
+	tests := map[string]string{
+		"8.8.8.8": "google-public-dns-a.google.com",
+		"8.8.4.4": "google-public-dns-b.google.com",
+	}
```
tomwilkie: We need to mock out the actual dns lookup - do you mind if I have a stab at that?

tomwilkie: Ignore me.

tomwilkie:
I just put in a few small changes to address some of my own review feedback...

probe/endpoint/resolver_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"time"
+
+	. "github.com/weaveworks/scope/probe/endpoint"
+	"github.com/weaveworks/scope/test"
+)
+
+func TestReverseResolver(t *testing.T) {
+	tests := map[string]string{
+		"1.2.3.4": "test.domain.name",
+		"4.3.2.1": "im.a.little.tea.pot",
```
peterbourgon: :+1:

peterbourgon:
LGTM!

PR #353, opened 2015-08-13
Makes scope works with boot2docker (Mac OSX)

barravi:
In order for scope to work with boot2docker, the variable DOCKER_BRIDGE_IP can now be preset to the ip returned by the command 'boot2docker ip'.

DOCKER_BRIDGE_IP=$(boot2docker ip) scope launch <- now works with boot2docker)
scope launch <- usual behavior.

PR #409, opened 2015-08-27
Works with boot2docker.

paulbellamy:
The DOCKER_BRIDGE_IP is set on the remote host, so it will be found
correctly on boot2docker.

When a container dies, the logs show:
```
2015/08/27 15:38:51 docker container: error reading event, did container stop? read unix /var/run/docker.sock: use of closed network connection
2015/08/27 15:38:51 docker container: stopped collecting stats for 44514f936c6631ac4fd95fe434f47a94faa04c94cecbcd540c7387c1648a24fc
```
Which is a bit messy, but not an issue.

Closes #353

### DIFF
--- a/scope
+++ b/scope
@@ -99,7 +99,7 @@ weave_dns_present() {
 }
 
 set_docker_bridge_ip() {
-    DOCKER_BRIDGE_IP=$(ip -4 addr show dev $DOCKER_BRIDGE | grep -m1 -o 'inet [.0-9]*')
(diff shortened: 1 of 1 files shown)


scope
```diff
@@ -99,7 +99,7 @@ weave_dns_present() {
 }
 
 set_docker_bridge_ip() {
-    DOCKER_BRIDGE_IP=$(ip -4 addr show dev $DOCKER_BRIDGE | grep -m1 -o 'inet [.0-9]*')
+    DOCKER_BRIDGE_IP=$(docker run --rm --net=host gliderlabs/alpine ip -f inet address show dev $DOCKER_BRIDGE | grep -m1 -o 'inet \([.0-9]\)*')
```
tomwilkie: Can you use the weavescope image with --entrypoint=/bin/sh please?

That way we'll automatically be labelled with works.weave.role=system and filtered when that happens.

tomwilkie:
1 minor comment, otherwise LGTM

PR #410, opened 2015-08-27
Use immutable objects for storing nodes in browser

davkal:
* makes checking for topology changes easier, which triggers fewer re-renders
* fix JS store tests with Immutables
* only update graph if topology changed

This was cherry-picked and patched from the cola-layout branch.

### DIFF
--- a/client/app/scripts/charts/nodes-chart.js
+++ b/client/app/scripts/charts/nodes-chart.js
@@ -43,14 +43,13 @@ const NodesChart = React.createClass({
   },
 
   componentWillReceiveProps: function(nextProps) {
-    if (this.getTopologyFingerprint(nextProps.nodes) !== this.getTopologyFingerprint(this.props.nodes)) {
+    if (nextProps.nodes !== this.props.nodes) {
       this.setState({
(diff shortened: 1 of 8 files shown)


tomwilkie:
Seems to work for me.  Didn't buy us much though?  LGTM

tomwilkie:
Actually hold on; details panel doesn't seem to work anymore.

Just confirming its this change

davkal:
This brings down browser CPU use considerably since it skips layout calls after topology data changes that wont have an effect on layout anyways. I wanted to add it now and use it as a base for other layout experiments.

tomwilkie:
Yeah I can confirm the details panel is broken with this change - it always say node missing.

davkal:
Will fix and add a test for this.

tomwilkie:
@davkal do you want someone to review this again?

davkal:
@tomwilkie yeah, that be great. I found other places where I want to use immutables in new branches.

tomwilkie:
Okay can confirm details panel is fixed.  LGTM!

PR #411, opened 2015-08-28
Move Adjacency info inside the NodeMetadata struct

tomwilkie:
Also moves pseudo node generation to the probes


probe/sniff/sniffer.go
```diff
@@ -250,21 +250,30 @@ func (s *Sniffer) Merge(p Packet, rpt report.Report) {
...
+	addAdjacency := func(t report.Topology, srcNodeID, dstNodeID string) {
+		srcNode, ok := t.NodeMetadatas[srcNodeID]
+		if !ok {
+			srcNode = report.MakeNodeMetadata()
+		}
+		srcNode.Adjacency = srcNode.Adjacency.Add(dstNodeID)
+		t.NodeMetadatas[srcNodeID] = srcNode
+
+		if _, ok := t.NodeMetadatas[dstNodeID]; !ok {
+			t.NodeMetadatas[dstNodeID] = report.MakeNodeMetadata()
```
peterbourgon: So here is where we do pseudonode generation in the probe, correct?

probe/sniff/sniffer_test.go
```diff
@@ -77,7 +72,16 @@ func TestMerge(t *testing.T) {
...
+			srcEndpointNodeID: {
+				Metadata:  map[string]string{},
+				Counters:  map[string]int{},
+				Adjacency: report.MakeIDList(dstEndpointNodeID),
+			},
+			dstEndpointNodeID: report.NodeMetadata{
+				Metadata:  map[string]string{},
+				Counters:  map[string]int{},
+				Adjacency: report.MakeIDList(),
+			},
```
peterbourgon: Please add constructors for this initialization style. If we leave it as-is, and later add a field to NodeMetadata that needs initialization, it's very easy to forget to declare it in this callsite, and get very subtle runtime breakage.

I'm not at all happy with this pattern, but I'm not aware of a better way to do it. Suggestions welcome.

probe/sniff/sniffer_test.go
```diff
@@ -100,7 +99,16 @@ func TestMerge(t *testing.T) {
...
+			srcAddressNodeID: report.NodeMetadata{
+				Metadata:  map[string]string{},
+				Counters:  map[string]int{},
+				Adjacency: report.MakeIDList(dstAddressNodeID),
+			},
+			dstAddressNodeID: report.NodeMetadata{
+				Metadata:  map[string]string{},
+				Counters:  map[string]int{},
+				Adjacency: report.MakeIDList(),
+			},
```
peterbourgon: Likewise above.

report/id.go
```diff
@@ -21,19 +21,6 @@ const (
...
-	return ">" + srcNodeID
-}
-
-// ParseAdjacencyID produces a node ID from an adjancency ID.
-func ParseAdjacencyID(adjacencyID string) (string, bool) {
-	if !strings.HasPrefix(adjacencyID, ">") {
-		return "", false
-	}
-	return adjacencyID[1:], true
-}
```
peterbourgon: Yay :)

report/id_list.go
```diff
@@ -46,6 +37,13 @@ func (a IDList) Add(ids ...string) IDList {
 	return a
 }
 
+// Copy returns a copy of the IDList.
+func (a IDList) Copy() IDList {
+	result := make(IDList, len(a))
+	copy(result, a)
```
peterbourgon: ...oh, yes :sheep:

peterbourgon:
I didn't look deeply at the tests, but this looks good. Once the constructor thing is taken care of (or an alternate suggestion made) :+1:

probe/sniff/sniffer.go
tomwilkie: Correct.  From the probes PoV, its just node generation, and now it generates nodes at both 'ends' of the adjacency.  The local node is always tagged with the origin host id, whereas the remote node remains untagger.  The rendering pipeline line uses this info as a clue that it should treat remote ends as pseudo nodes.  NB for local<->local connections, we get the same connection twice, once for each pid, and end up putting host id in each node.

probe/sniff/sniffer_test.go
tomwilkie: We could do the method cascading AKA fluent API:
```
MakeNodeMetadata().SetMetadata(...).SetCounters(...).SetAdjacency(...)
```
I'm not a huge fan of this, but it beats the N^2 constructors we'd need otherwise.

peterbourgon: I've never been a fan of that builder style of constructor, but it may be the least-bad option in this case...

```go
md := report.MakeNodeMetadata().SetMetadata(map[string]string{
    "foo": "bar",
    "baz": "quux",
}).SetAdjacency(report.MakeIDList(
    dstEndpointNodeID,
))
```

okay

PR #420, opened 2015-09-01
Add filtering options to views; add options for filtering out system containers & unconnected containers.

peterbourgon:
Me and @davkal


render/render.go
```diff
@@ -354,3 +356,62 @@ func (f Filter) Render(rpt report.Report) RenderableNodes {
 	}
 	return output
 }
+
+// FilterSystem is a Renderer which filters out system nodes.
+func FilterSystem(r Renderer) Renderer {
```
tomwilkie: Could use a Filter{} for this; save a few lines.

render/render.go
```diff
@@ -354,3 +356,62 @@ func (f Filter) Render(rpt report.Report) RenderableNodes {
...
+	"weavescope": struct{}{},
+	"weavedns":   struct{}{},
+	"weave":      struct{}{},
+}
+
+var systemImageNames = map[string]struct{}{
+	"weaveworks/scope":    struct{}{},
+	"weaveworks/weavedns": struct{}{},
+	"weaveworks/weave":    struct{}{},
+}
```
tomwilkie: hmmm... instead of hardcoding a bunch of images / container names, I thought we were going to use labels?

Also, you forgot weaveexec and weaveproxy.

tomwilkie:
Not using the DSL for this?

render/render.go
peterbourgon: Only weavescope has a label at the moment. Did the weave PRs get merged and I'm just behind the times?

peterbourgon:
> Not using the DSL for this?

It turned out mapping between report.Topology (what the DSL expects) and RenderableNodes (what the rendering pipeline outputs, and what the UI needs) was quite a bit of work, some of which was subtle. In the interest of pushing the feature, I went with this approach instead. 

It will be simpler to have a full-expression renderer, i.e. only dumping to RenderableNodes at the end. That's what I'm targeting for the Kubernetes integration.

render/render.go
peterbourgon: Nope, looks like weave containers still don't have labels...

tomwilkie:
```
vagrant@ubuntu-14:~/src/github.com/weaveworks/scope$ docker inspect weave
...
        "Labels": {
            "works.weave.role": "system"
        }
...
```

See: https://github.com/weaveworks/weave/pull/1290

NB this is only in master, so will end up in 1.1

app/router.go
```diff
@@ -155,4 +179,16 @@ type topologyView struct {
 	human    string
 	parent   string
 	renderer render.Renderer
+	options  optionParams
+}
+
+type optionParams map[string][]optionValue // param: values
+
+type optionValue struct {
```
tomwilkie: Is there a way to not have APITopologyOption and this struct?  Some cleve combinations of struct tags might be able to give the same result.

render/render.go
tomwilkie: @peterbourgon ping?

tomwilkie:
@peterbourgon what happened to using labels instead of hardcoding the container names?

peterbourgon:
@tomwilkie we are using labels ([see here](https://github.com/weaveworks/scope/pull/420/files#diff-38a809c428b3d9b68a706c5267759007R411)) just not exclusively, as current weave master still doesn't put system labels on weave components.

app/router.go
peterbourgon: PRs welcome :)

tomwilkie:
> as current weave master still doesn't put system labels on weave components.

It should do (I made that change!) - what version are you running?

client/app/scripts/stores/app-store.js
```diff
@@ -156,6 +172,14 @@ const AppStore = assign({}, EventEmitter.prototype, {
 AppStore.registeredCallback = function(payload) {
   switch (payload.type) {
 
+    case ActionTypes.CHANGE_TOPOLOGY_OPTION:
+      if (activeTopologyOptions[payload.option] !== payload.value) {
+        nodes = {};
```
davkal: Should be `nodes = nodes.clear()`

tomwilkie: Pls update; when I rebase I'm sure there are a few cases of this.

davkal:
Not getting any errors anymore.

tomwilkie:
Thanks; I'm just adding some tests but this is feeling really close now.

tomwilkie:
Okay I'm done; its looking good.

One issue:  I don't think the UI is selecting the default filter - I've changed the default to filter system containers, but I'm still seeing them in the UI.  David, can you fix?

peterbourgon:
LGTM!

PR #421, opened 2015-09-01
Move EdgeMetadata into the NodeMetadata struct.

tomwilkie:
Part of #357

Issue #357, opened 2015-08-14
Consider gently converging NodeMetadata struct and RenderableNode


tomwilkie:
- move adjacency information in a topology into the NMDs
- have probes make psuedo nodes

inercia:
How are these _pseudo-nodes_ going to be generated? I assume that this will break the _"Internet"_ node in some other _pseudo-nodes_, but we probably do not want to generate a _pseudo-node_ for each remote end of a connection, so is there any rule we will follow for generating these nodes?

tomwilkie:
Currently the probes only create nodes for stuff (endpoints, processes) it actually has locally - but it creates edges to things that might (or might not exist).  This means the graph is incomplete.  In the rendering stage, we introduce these pseudo nodes if they don't exist.

The idea here would be to have the probes create nodes for both ends of every edge in a way that merges nicely if both ends of the connection are seen by two different probes.

inercia:
Cool, but wouldn't this mean that probes would be sending lots of data for nodes that would never be rendered?

tomwilkie:
> Cool, but wouldn't this mean that probes would be sending lots of data 

Not "lots" - a constant factor more than we already send.  And we have compression.

> for nodes that would never be rendered?

Pseudo nodes are rendered.

tomwilkie:
Part of this is addressed in #411

Suggested next steps:
* Move EdgeMetadatas into NodeMetadata.EdgeMetadatas, define appropriate merge and collapse semantics.  Merge is as-is, collapse would merge together all the edgemetadatas for a single node into the edgemetadata in RenderableNode.  Could also solve the edge double-counting issue.
* Rename NodeMetadata (and associated types & functions) to Node
* Topology just has a single field map[string]Node - don't want to collapse, as we'll be adding capability descriptors there in the future.

Then options for converging RenderableNode and Node.  We shouldn't change the API to the UI, so we still need to serve RenderableNodes (or things that look like RenderableNodes):
* Embed Node struct in RenderableNode, provide a function to generate a basic RenderableNode from a Node, perhaps with heuristics for filling out Names.  Leave the current rendering pipeline in tact.
* Extend Node to have LabelMajor, LableMinor, Rank, etc etc fields.  Have the probe attempt to fill them in appropriately.  Define some merge semantics for how these fields are merged...

Other ideas?

tomwilkie:
Independently, we can now merge LeafMap and Map in the rendering pipeline, which I'll do next.

[...]

PR #425, opened 2015-09-02
Remove LeafMap - have a generic function which converts a Topology to RenderableNodes

tomwilkie:
Includes #421, those changes will go away when thats merged.

Part of #357 

Doesn't actually buy us as much as I'd hoped.

### DIFF
--- a/render/expected/expected.go
+++ b/render/expected/expected.go
@@ -15,8 +15,8 @@ func Sterilize(r render.RenderableNodes) render.RenderableNodes {
 	// state not sent to the client.  So in the tests we ignore
 	// this state.
 	for id, n := range r {
-		n.NodeMetadata.Metadata = map[string]string{}
-		n.NodeMetadata.Counters = map[string]int{}
+		n.NodeMetadata.Metadata = report.Metadata{}
+		n.NodeMetadata.Counters = report.Counters{}
 		n.NodeMetadata.Edges = report.EdgeMetadatas{}
 		r[id] = n
 	}
--- a/render/expected/expected.go
+++ b/render/expected/expected.go
@@ -10,22 +10,14 @@ import (
 
 // Sterilize cleans up RenderableNodes test fixtures
 func Sterilize(r render.RenderableNodes) render.RenderableNodes {
-	// Since introducing new map fields to the report.NodeMetadata type, its
-	// zero value is •not valid• -- every time you need one, you need to use
-	// the report.MakeNodeMetadata constructor. (Similarly, but not exactly
-	// the same, is that a zero-value Adjacency is not the same as a created
-	// but empty Adjacency.)
-	//
-	// But we're not doing this in tests.  We also don't particularly care
-	// about the output of NodeMetadata for the rendering pipeline, as this
-	// is never serialised to json.  So this function sterilizes invalid
-	// RenderableNodes by setting an empty NodeMetadata from the proper
-	// constructor.
+	// RenderableNodes contain NodeMetadatas, but generally we
+	// only care about the Adjacency field - the rest is internal
+	// state not sent to the client.  So in the tests we ignore
+	// this state.
 	for id, n := range r {
-		if n.Adjacency == nil {
-			n.Adjacency = report.MakeIDList()
-		}
-		n.NodeMetadata = report.MakeNodeMetadata()
+		n.NodeMetadata.Metadata = map[string]string{}
+		n.NodeMetadata.Counters = map[string]int{}
+		n.NodeMetadata.Edges = report.EdgeMetadatas{}
 		r[id] = n
 	}
 	return r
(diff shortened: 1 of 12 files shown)


render/renderable_node.go
```diff
@@ -107,6 +70,73 @@ func newDerivedPseudoNode(id, major string, node RenderableNode) RenderableNode
 		Pseudo:       true,
 		Origins:      node.Origins.Copy(),
 		EdgeMetadata: node.EdgeMetadata.Copy(),
-		NodeMetadata: report.MakeNodeMetadata(),
+		NodeMetadata: node.NodeMetadata.Copy(),
+	}
+}
+
+// WithNode creates a new RenderableNode bases on rn, with n
+func (rn RenderableNode) WithNode(n report.NodeMetadata) RenderableNode {
```
peterbourgon: It's not WithNode, but rather WithNodeMetadata...

peterbourgon:
I appreciate the change. Removing the distinction of a LeafMapRenderer is a big win for understandability.

render/expected/expected.go
```diff
@@ -10,22 +10,14 @@ import (
...
+	// state not sent to the client.  So in the tests we ignore
+	// this state.
 	for id, n := range r {
-		if n.Adjacency == nil {
-			n.Adjacency = report.MakeIDList()
-		}
-		n.NodeMetadata = report.MakeNodeMetadata()
+		n.NodeMetadata.Metadata = map[string]string{}
+		n.NodeMetadata.Counters = map[string]int{}
+		n.NodeMetadata.Edges = report.EdgeMetadatas{}
```
peterbourgon: I guess use the constructors here...

render/renderable_node.go
tomwilkie: Sorry go ahead of myself; next change it to rename that struct.  Will fix.

tomwilkie:
@peterbourgon PTAL

render/mapping.go
```diff
@@ -25,26 +25,17 @@ const (
...
-// ID to a human-readable major and minor labels.
-//
-// A single NodeMetadata can yield arbitrary many representations, including
-// representations that reduce (or even increase) the cardinality of the set of nodes.
-type LeafMapFunc func(report.NodeMetadata, report.Networks) RenderableNodes
-
 // MapFunc is anything which can take an arbitrary RenderableNode and
 // return a set of other RenderableNodes.
 //
 // As with LeafMapFunc, if the output is empty, the node
```
paulbellamy: "As with LeafMapFunc"

render/renderable_node.go
```diff
@@ -13,67 +13,42 @@ type RenderableNode struct {
 	LabelMinor string        `json:"label_minor,omitempty"` // e.g. "hostname", human-readable, optional
 	Rank       string        `json:"rank"`                  // to help the layout engine
 	Pseudo     bool          `json:"pseudo,omitempty"`      // sort-of a placeholder node, for rendering purposes
-	Adjacency  report.IDList `json:"adjacency,omitempty"`   // Node IDs (in the same topology domain)
 	Origins    report.IDList `json:"origins,omitempty"`     // Core node IDs that contributed information
 
 	report.EdgeMetadata `json:"metadata"` // Numeric sums
-	report.NodeMetadata `json:"-"`        // merged NodeMetadata of the nodes used to build this
+	report.NodeMetadata
```
peterbourgon: The non-omitted fields of NodeMetadata will get flattened and rendered into the RenderableNode JSON, i.e. you'll get

```json
{
    "label_minor": "whatever",
    "origins": ["a", "b", "c"],
    "adjacency": ["x", "y", "z"]
}
```

Is that your intent?

render/renderable_node.go
```diff
@@ -107,6 +70,73 @@ func newDerivedPseudoNode(id, major string, node RenderableNode) RenderableNode
...
 		EdgeMetadata: node.EdgeMetadata.Copy(),
-		NodeMetadata: report.MakeNodeMetadata(),
+		NodeMetadata: node.NodeMetadata.Copy(),
+	}
+}
+
+// WithNodeMetadata creates a new RenderableNode based on rn, with n
+func (rn RenderableNode) WithNodeMetadata(n report.NodeMetadata) RenderableNode {
+	result := rn.Copy()
+	result.NodeMetadata = n
```
peterbourgon: Should this be

```go
result.NodeMetadata = result.NodeMetadata.Merge(n)
```

?

render/renderable_node.go
tomwilkie: Yes thats my intention; NodeMetadata only exposes adjacency on json rendering.

render/renderable_node.go
tomwilkie: Yes why not.  Its only used in constructors, so it should affect anything, but this would be nicer.

peterbourgon:
:+1: otherwise.

Semi-related, I'm noticing that most churn in recent PRs is in the tests. I feel this myself, when I make a simple change in e.g. package render, I find I have to change a lot of "distant" tests in other components. I'm wondering if there's not a better testing strategy, relying a lot less or even even not at all on coarse-grained transforms of reports (i.e. package expected), and a lot more on fine-grained unit testing of individual functions and packages. What do you think?

tomwilkie:
On one hand I agree, the tests are pretty coarse.  But at the end of the day they are a good way to catch unintended side effects of small changes that more focused test might miss, and they do provide excellent coverage.

An alternative would be to capture a json report and the json output of the renders and plumb them together; that might help with some of the tedious struct munging.   Or just make it tedious json munging instead.

PR #428, opened 2015-09-02
Show container image details for containers.

### LINKED ISSUE #398: In container details panel I want to see container image info.

tomwilkie:
Fixes #398

Includes (and requires) #425 

Implementation note: #425 moves the responsibility for propagating origin ids to the mappers, which in turn rely on the RenderableNode constructors to do it properly.  This opens up the opportunity for the mappers to inject extra origin node ids, which we take advantage of here.

### DIFF
--- a/render/mapping.go
+++ b/render/mapping.go
@@ -131,8 +131,8 @@ func MapContainerIdentity(m RenderableNode, _ report.Networks) RenderableNodes {
 
 	node := NewRenderableNodeWith(id, major, minor, rank, m)
 	if imageID, ok := m.Metadata[docker.ImageID]; ok {
-		scope, _, _ := report.ParseContainerNodeID(m.ID)
-		node.Origins = node.Origins.Add(report.MakeContainerNodeID(scope, imageID))
+		hostID, _, _ := report.ParseContainerNodeID(m.ID)
+		node.Origins = node.Origins.Add(report.MakeContainerNodeID(hostID, imageID))
 	}
 	return RenderableNodes{id: node}
 }
--- a/report/id.go
+++ b/report/id.go
@@ -121,10 +121,10 @@ func ParseEndpointNodeID(endpointNodeID string) (hostID, address, port string, o
 	return fields[0], fields[1], fields[2], true
 }
 
-// ParseContainerNodeID produces the host ID, and container id from an container
+// ParseContainerNodeID produces the host and container id from an container
 // node ID.
 func ParseContainerNodeID(containerNodeID string) (hostID, containerID string, ok bool) {
-	fields := strings.SplitN(containerNodeID, ScopeDelim, 3)
+	fields := strings.SplitN(containerNodeID, ScopeDelim, 2)
 	if len(fields) != 2 {
 		return "", "", false
 	}


render/expected/expected.go
```diff
@@ -10,22 +10,14 @@ import (
...
+	// state not sent to the client.  So in the tests we ignore
+	// this state.
 	for id, n := range r {
-		if n.Adjacency == nil {
-			n.Adjacency = report.MakeIDList()
-		}
-		n.NodeMetadata = report.MakeNodeMetadata()
+		n.NodeMetadata.Metadata = map[string]string{}
+		n.NodeMetadata.Counters = map[string]int{}
+		n.NodeMetadata.Edges = report.EdgeMetadatas{}
```
peterbourgon: I guess use the constructors here...

peterbourgon:
If it's OK, I'll wait for #425 to merge and this gets rebased...

tomwilkie:
@peterbourgon try now

render/mapping.go
```diff
@@ -121,7 +121,12 @@ func MapContainerIdentity(m RenderableNode, _ report.Networks) RenderableNodes {
 		rank  = m.Metadata[docker.ImageID]
 	)
 
-	return RenderableNodes{id: NewRenderableNodeWith(id, major, minor, rank, m)}
+	node := NewRenderableNodeWith(id, major, minor, rank, m)
+	if imageID, ok := m.Metadata[docker.ImageID]; ok {
+		scope, _, _ := report.ParseContainerNodeID(m.ID)
+		node.Origins = node.Origins.Add(report.MakeContainerNodeID(scope, imageID))
```
peterbourgon:
```go
hostID, _, _ := report.ParseContainerNodeID(m.ID)
node.Origins = node.Origins.Add(report.MakeContainerNodeID(hostID, imageID))
```

report/id.go
```diff
@@ -87,6 +87,16 @@ func ParseEndpointNodeID(endpointNodeID string) (hostID, address, port string, o
 	return fields[0], fields[1], fields[2], true
 }
 
+// ParseContainerNodeID produces the host ID, and container id from an container
```
peterbourgon: container ID

report/id.go
```diff
@@ -87,6 +87,16 @@ func ParseEndpointNodeID(endpointNodeID string) (hostID, address, port string, o
...
+// node ID.
+func ParseContainerNodeID(containerNodeID string) (hostID, containerID string, ok bool) {
+	fields := strings.SplitN(containerNodeID, ScopeDelim, 3)
+	if len(fields) != 2 {
```
peterbourgon: 3 vs. 2?

peterbourgon:
I don't quite understand the motivation. The goal here is to display additional container image details in container node detail panes, right? Why can't we get the same outcome by doing a normal foreign-key-style join across the Container and ContainerImage topologies during rendering?

tomwilkie:
> I don't quite understand the motivation. The goal here is to display additional container image details in container node detail panes, right? 

Thats right

> Why can't we get the same outcome by doing a normal foreign-key-style join across the Container and ContainerImage topologies during rendering?

A foreign key join would reduce the number of nodes, and going backwards would be tricky; also we only need the container image id, which is already in the container node as the foreign key, so it'd be wasteful.

peterbourgon:
Could you instead have something in the rendering pipeline which did the join, but just copied the relevant metadata from ContainerImage to Container topologies, and didn't reduce the number of nodes?

I'm concerned, because this PR changes the semantics of origins. Previously, origins were topology nodes which had been merged, fully, into the rendered node. This PR loosens that definition to... well, nothing, really—just any topology node which could have some kind of abstract "parent" relationship to the rendered node. Feels wrong...

tomwilkie:
> Could you instead have something in the rendering pipeline which did the join, but just copied the relevant metadata from ContainerImage to Container topologies, and didn't reduce the number of nodes?

I don't see how - the container image topology doesn't have container ids in the nodes, so any join would need to reduce container ids to image ids, which makes it hard to go back.  Also the only relevant metadata is the container image id, which is already in the container topology nodes.

Maybe I'm missing the point - could you sketch it out in pseudo code?

peterbourgon:
I'll comment inline, one second...

render/mapping.go
```diff
@@ -121,7 +121,12 @@ func MapContainerIdentity(m RenderableNode, _ report.Networks) RenderableNodes {
 		rank  = m.Metadata[docker.ImageID]
 	)
 
-	return RenderableNodes{id: NewRenderableNodeWith(id, major, minor, rank, m)}
+	node := NewRenderableNodeWith(id, major, minor, rank, m)
+	if imageID, ok := m.Metadata[docker.ImageID]; ok {
+		hostID, _, _ := report.ParseContainerNodeID(m.ID)
+		node.Origins = node.Origins.Add(report.MakeContainerNodeID(hostID, imageID))
```
peterbourgon: 1. A renderable node's origins are defined to be the topology nodes that have been merged to produce it. You're not merging the container image node into this one, so it's not appropriate to use origins.
2. Minor point compared to the above, but: you're overloading MakeContainerNodeID, i.e. "make a node ID for the container topology", and using it to mean "make a node ID for the container image topology".

If you want to display the image ID in the detail pane, and you know the `docker.ImageID` is present in the metadata for that RenderableNode, then you can put this logic in the thing which generates the detail pane, and leave origins untouched, I think?

tomwilkie: (1) is not strictly true; we include the host id in the origins list so that the host info appears in the details panel.

I would describe origins as the list of node ids for which you want info to appear in the details panel...

peterbourgon: Hmm. This definition

> I would describe origins as the list of node ids for which you want info to appear in the details panel...

is too loose, but re: host ID, I'll agree that my definition is too strict. How about this: origins are nodes which have a strict parent relationship to the renderable node. Accordingly, how do you feel about renaming the field Parents?

tomwilkie: @peterbourgon Whats the definition of a 'strict parent relationship'?

tomwilkie:
Looks like this accidentally got merged as part of #434 - this PR is just the review feedback now.

peterbourgon:
Not cool :(

> Whats the definition of a 'strict parent relationship'?

A parent is a node which conceptually comes before/above or envelops another node. A container image node `weaveworks/foo:1.0.0` is strictly the parent of a container node `a0b1c2d3e4` which runs that image. An address node `10.11.12.13` is strictly the parent of an endpoint node `10.11.12.13:80`.

tomwilkie:
Hmm that doesn't make sense then; you suggested we rename Origins to Parents, but Origins for a endpoint doesn't contain the address node id, and before this change, origins for a container didn't contain the contained image node id.  If anything, Origins is closer to some notion of 'Children', except it also contains the host node id.

I think the easiest definition is to define origins as 'the list of node ids for which you want info to appear in the details panel'.  If you like I could rename it to 'Details' - although it shouldn't really be exposed in the API anyway.

tomwilkie:
We settled this on hipchat.

PR #429, opened 2015-09-02
Trim long container names and labels in details

### LINKED ISSUE #418: Trim Columns keys and Table titles in details pane
screen shot 2015-09-01 at 1 51 23 pm

Also, both for the keys and values, provide a way to view the full value when trimmed (i.e. hover-on) so that the user can view the full text.

davkal:
fixes #418

### DIFF
--- a/client/app/scripts/components/node-details-table.js
+++ b/client/app/scripts/components/node-details-table.js
@@ -14,11 +14,15 @@ const NodeDetailsTable = React.createClass({
         {this.props.rows.map(function(row) {
           return (
             <div className="node-details-table-row">
-              <div className="node-details-table-row-key">{row.key}</div>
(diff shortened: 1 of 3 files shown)


tomwilkie:
Can confirm row labels are truncated as expected.

However, table headers are still wrapping:

screen shot 2015-09-02 at 17 04 24

davkal:
I thought that was feature ;)

2opremio:
@davkal Maybe this should be addressesed somewhere else, but #418 also mentions there should be a way to display the full names (I suggested to do it on hover-on).

davkal:
@fons that works already.

davkal:
all right, lets keep this only for #418

tomwilkie:
Can confirm this seems to all work as expected.  LGTM

Issue #413, opened 2015-08-28
Trim long container names in UI


2opremio:
Here's an example of how the UI looks like with long container names (mangled by ECS):

<img width="1193" alt="screen shot 2015-08-28 at 6 29 35 pm" src="https://cloud.githubusercontent.com/assets/2362916/9553850/b6c5f964-4db9-11e5-806e-7348bef97e8b.png">

tomwilkie:
We should definitely trim the names.

Also perhaps a feature for th DSL would be to allow users to apply regexs
to the names, like we do in the proxy for weave DNS? @peterbourgon?

peterbourgon:
@tomwilkie Easily done. +1 to the idea.

tomwilkie:
Moving the regex discussion to another ticket

PR #430, opened 2015-09-02
Trim node titles in graph

davkal:
Trim node titles in graph via JS

fix for #413

Sadly, the title attribute to show something on hover does not work in
SVG. So we might need another solution for that.


tomwilkie:
Not too worried about no hover on the graph, as they can get the full name from the details panel.

tomwilkie:
Can confirm this works; LGTM

PR #435, opened 2015-09-04
Color nodes by rank

davkal:
used to be by label_major.


tomwilkie:
We should also ensure the rank is the container image name, not the image id (which changes between versions)

PR #440, opened 2015-09-04
Make scope compile under go 1.5

tomwilkie:


paulbellamy:
Should probably update the gce demo as well.

tomwilkie:
Nah I want to keep building everything with 1.4 for now.  This is just for the brave people...

PR #441, opened 2015-09-04
Basic Kubernetes Integration

paulbellamy:
@tomwilkie, @peterbourgon any feedback would be great.

I'm not super-happy with the result, as there are still some shortcomings:

* We can't yet correlate connections going through the kube-proxy into pseudo service nodes, which means that all connections go through the "Uncontained" pseudo-node. This is because I haven't been able to figure out how to do it.
* We can't ever get full resolution on connections going through a service, as it might include a userspace hop (depending on configuration).
* There is a bug in k8s where [some pods are not listed in the API](https://github.com/kubernetes/kubernetes/issues/14738). I've raised an issue, and added a workaround.
* Need to pass a `--probe.kuberetes true` flag on the master node to enable it (for now).
* There may be some issues rendering incoming internet connections from external load-balancers, depending on how they do that routing, but none beyond what scope would already have.
* Kubernetes uses a lot of ephemeral (high) ports, which breaks the assumption about a higher port number being the client, and causes the graph to be laid out strangely.

In the example, there are:
* 3 php (frontend) pods, accepting incoming http requests, and making requests to redis
* 1 redis-master pod, handling writes
* 2 redis-slave pods (one per minion), handling reads. The redis-slave pods maintain an open socket to redis-master for replication.

Screenshots of it:
Pods view. I'm not sure if we should show the "2 containers" under each node, or if we should skip that.
screen shot 2015-09-29 at 16 45 40

Pods-by-Service view, where pods are grouped when they belong to a common service
screen shot 2015-09-29 at 16 45 53

As you can see, all the connections routing through "Uncontained" makes the output mostly useless. However, we *can* filter out the loads of kube-system pods, which clutter up the view considerably.

### DIFF
--- a/app/api_topologies_test.go
+++ b/app/api_topologies_test.go
@@ -1,9 +1,17 @@
 package main
 
 import (
+	"bytes"
+	"encoding/gob"
 	"encoding/json"
 	"net/http/httptest"
 	"testing"
+
+	"k8s.io/kubernetes/pkg/api"
+
+	"github.com/weaveworks/scope/probe/kubernetes"
+	"github.com/weaveworks/scope/report"
+	"github.com/weaveworks/scope/test"
 )
 
 func TestAPITopology(t *testing.T) {
@@ -36,5 +44,58 @@ func TestAPITopology(t *testing.T) {
 		if have := topology.Stats.NonpseudoNodeCount; have <= 0 {
 			t.Errorf("NonpseudoNodeCount isn't positive for %s: %d", topology.Name, have)
 		}
+	}
+}
+
+func TestAPITopologyAddsKubernetes(t *testing.T) {
+	ts := httptest.NewServer(Router(StaticReport{}))
+	defer ts.Close()
+
+	body := getRawJSON(t, ts, "/api/topology")
+
+	var topologies []APITopologyDesc
+	if err := json.Unmarshal(body, &topologies); err != nil {
+		t.Fatalf("JSON parse error: %s", err)
+	}
+	equals(t, 3, len(topologies))
+
+	// Enable the kubernetes topologies
+	rpt := report.MakeReport()
+	rpt.Pod = report.MakeTopology()
+	rpt.Pod.Nodes[test.ClientPodNodeID] = kubernetes.NewPod(&api.Pod{
+		ObjectMeta: api.ObjectMeta{
+			Name:      "pong-a",
+			Namespace: "ping",
+			Labels:    map[string]string{"ponger": "true"},
+		},
+		Status: api.PodStatus{
+			HostIP: "1.2.3.4",
+			ContainerStatuses: []api.ContainerStatus{
+				{ContainerID: "container1"},
+				{ContainerID: "container2"},
+			},
+		},
+	}).GetNode()
+	buf := &bytes.Buffer{}
+	if err := gob.NewEncoder(buf).Encode(rpt); err != nil {
+		t.Fatalf("GOB encoding error: %s", err)
+	}
+	checkRequest(t, ts, "POST", "/api/report", buf.Bytes())
+
+	body = getRawJSON(t, ts, "/api/topology")
+	if err := json.Unmarshal(body, &topologies); err != nil {
+		t.Fatalf("JSON parse error: %s", err)
+	}
+	equals(t, 4, len(topologies))
+
+	found := false
+	for _, topology := range topologies {
+		if topology.Name == "Pods" {
+			found = true
+			break
+		}
+	}
+	if !found {
+		t.Error("Could not find pods topology")
 	}
 }
--- a/app/router.go
+++ b/app/router.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 
 	"github.com/PuerkitoBio/ghost/handlers"
 	"github.com/gorilla/mux"
@@ -96,6 +97,9 @@ func makeReportPostHandler(a xfer.Adder) http.HandlerFunc {
 			return
 		}
 		a.Add(rpt)
+		if len(rpt.Pod.Nodes) > 0 {
+			enableKubernetesTopologies()
+		}
 		w.WriteHeader(http.StatusOK)
 	}
 }
@@ -113,7 +117,7 @@ func decorateTopologyForRequest(r *http.Request, topology *topologyView) {
 
 func captureTopology(rep xfer.Reporter, f func(xfer.Reporter, topologyView, http.ResponseWriter, *http.Request)) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		topology, ok := topologyRegistry[mux.Vars(r)["topology"]]
+		topology, ok := topologyRegistry.get(mux.Vars(r)["topology"])
 		if !ok {
 			http.NotFound(w, r)
 			return
@@ -133,50 +137,98 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 	respondWith(w, http.StatusOK, APIDetails{ID: uniqueID, Version: version})
 }
 
+// registry is a threadsafe store of the available topologies
+type registry struct {
+	sync.RWMutex
+	items map[string]topologyView
+}
+
+func (r *registry) add(ts map[string]topologyView) {
+	r.Lock()
+	defer r.Unlock()
+	result := map[string]topologyView{}
+	for name, t := range r.items {
+		result[name] = t
+	}
+	for name, t := range ts {
+		result[name] = t
+	}
+	r.items = result
+}
+
+func (r *registry) get(name string) (topologyView, bool) {
+	r.RLock()
+	defer r.RUnlock()
+	t, ok := r.items[name]
+	return t, ok
+}
+
+func (r *registry) walk(f func(string, topologyView, map[string]topologyView)) {
+	r.RLock()
+	defer r.RUnlock()
+	for name, def := range r.items {
+		if def.parent != "" {
+			continue
+		}
+
+		subDefs := map[string]topologyView{}
+		for subName, subDef := range r.items {
+			if subDef.parent != name {
+				continue
+			}
+			subDefs[subName] = subDef
+		}
+
+		f(name, def, subDefs)
+	}
+}
+
 // Topology option labels should tell the current state. The first item must
 // be the verb to get to that state
-var topologyRegistry = map[string]topologyView{
-	"applications": {
-		human:    "Applications",
-		parent:   "",
-		renderer: render.FilterUnconnected(render.ProcessWithContainerNameRenderer),
-		options: optionParams{"unconnected": {
-			// Show the user why there are filtered nodes in this view.
-			// Don't give them the option to show those nodes.
-			{"hide", "Unconnected nodes hidden", true, nop},
-		}},
-	},
-	"applications-by-name": {
-		human:    "by name",
-		parent:   "applications",
-		renderer: render.FilterUnconnected(render.ProcessNameRenderer),
-		options: optionParams{"unconnected": {
-			// Ditto above.
-			{"hide", "Unconnected nodes hidden", true, nop},
-		}},
-	},
-	"containers": {
-		human:    "Containers",
-		parent:   "",
-		renderer: render.ContainerWithImageNameRenderer,
-		options: optionParams{"system": {
-			{"show", "System containers shown", false, nop},
-			{"hide", "System containers hidden", true, render.FilterSystem},
-		}},
-	},
-	"containers-by-image": {
-		human:    "by image",
-		parent:   "containers",
-		renderer: render.ContainerImageRenderer,
-		options: optionParams{"system": {
-			{"show", "System containers shown", false, nop},
-			{"hide", "System containers hidden", true, render.FilterSystem},
-		}},
-	},
-	"hosts": {
-		human:    "Hosts",
-		parent:   "",
-		renderer: render.HostRenderer,
+var topologyRegistry = &registry{
+	items: map[string]topologyView{
+		"applications": {
+			human:    "Applications",
+			parent:   "",
+			renderer: render.FilterUnconnected(render.ProcessWithContainerNameRenderer),
+			options: optionParams{"unconnected": {
+				// Show the user why there are filtered nodes in this view.
+				// Don't give them the option to show those nodes.
+				{"hide", "Unconnected nodes hidden", true, nop},
+			}},
+		},
+		"applications-by-name": {
+			human:    "by name",
+			parent:   "applications",
+			renderer: render.FilterUnconnected(render.ProcessNameRenderer),
+			options: optionParams{"unconnected": {
+				// Ditto above.
+				{"hide", "Unconnected nodes hidden", true, nop},
+			}},
+		},
+		"containers": {
+			human:    "Containers",
+			parent:   "",
+			renderer: render.ContainerWithImageNameRenderer,
+			options: optionParams{"system": {
+				{"show", "System containers shown", false, nop},
+				{"hide", "System containers hidden", true, render.FilterSystem},
+			}},
+		},
+		"containers-by-image": {
+			human:    "by image",
+			parent:   "containers",
+			renderer: render.ContainerImageRenderer,
+			options: optionParams{"system": {
+				{"show", "System containers shown", false, nop},
+				{"hide", "System containers hidden", true, render.FilterSystem},
+			}},
+		},
+		"hosts": {
+			human:    "Hosts",
+			parent:   "",
+			renderer: render.HostRenderer,
+		},
 	},
 }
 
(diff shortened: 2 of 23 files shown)


probe/docker/container.go
```diff
@@ -16,7 +16,6 @@ import (
 	"time"
 
 	docker "github.com/fsouza/go-dockerclient"
-
```
tomwilkie: pls don't.

probe/kubernetes/client.go
```diff
@@ -0,0 +1,109 @@
...
+	k.listeners[c] = l
+	go l.Listen(c)
+
+	return nil
+}
+
+func (k *client) RemoveEventListener(c chan watch.Event) error {
+	k.lock.Lock()
+	defer k.lock.Unlock()
+	return k.removeEventListener(c)
```
peterbourgon: Good idiom.

probe/kubernetes/client.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+type listener struct {
+	pods, services, nodes watch.Interface
+}
+
+func (l listener) Listen(c chan<- watch.Event) {
+	podsChan := l.pods.ResultChan()
+	servicesChan := l.services.ResultChan()
+	nodesChan := l.nodes.ResultChan()
```
peterbourgon: Maybe var-block this, and can they just be called pods/services/chan? ([via](https://docs.google.com/document/d/18UTcmy8VoecEaYNb0DrI8NZrI1uJbq-XYDEko0l5Ny4/edit#heading=h.74h5axa7f66g))

probe/kubernetes/pod.go
```diff
@@ -0,0 +1,68 @@
...
+func (p *pod) Name() string {
+	return p.ObjectMeta.Name
+}
+
+func (p *pod) Namespace() string {
+	return p.ObjectMeta.Namespace
+}
+
+func (p *pod) Created() string {
+	return p.ObjectMeta.CreationTimestamp.Format(time.RFC822)
```
peterbourgon: Is this method for debug or something? (Not represented in interface..)

probe/kubernetes/registry.go
```diff
@@ -0,0 +1,192 @@
...
+	pods     map[string]Pod
+	services map[string]Service
+	nodes    map[string]*api.Node
+}
+
+// Client interface for mocking.
+type Client interface {
+	AddEventListener(chan<- watch.Event) error
+	RemoveEventListener(chan watch.Event) error
+}
```
peterbourgon: Maybe Subscribe/Unsubscribe is more succinct?

probe/kubernetes/reporter.go
```diff
@@ -0,0 +1,41 @@
+package kubernetes
+
+import "github.com/weaveworks/scope/report"
+
+// Reporter generate Reports containing Container and ContainerImage topologies
```
peterbourgon: I remember discussing why we chose this route vs. a new topology, but I don't remember the finer details :) Maybe you can add a bit more commentary?

render/id.go
```diff
@@ -25,6 +25,11 @@ func MakeHostID(hostID string) string {
 	return fmt.Sprintf("host:%s", hostID)
 }
 
+// MakePodID makes an pod node ID for rendered nodes.
```
peterbourgon: *a

probe/kubernetes/reporter.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (r *Reporter) serviceTopology() (report.Topology, []Service, error) {
+	var (
+		result   = report.MakeTopology()
+		services = []Service{}
+	)
+	err := r.client.WalkServices(func(s Service) error {
+		nodeID := report.MakeServiceNodeID(s.Namespace(), s.Name())
+		result.Nodes[nodeID] = s.GetNode()
```
tomwilkie: Use AddNode pls.

probe/kubernetes/reporter.go
```diff
@@ -0,0 +1,62 @@
...
+func (r *Reporter) podTopology(services []Service) (report.Topology, error) {
+	result := report.MakeTopology()
+	err := r.client.WalkPods(func(p Pod) error {
+		for _, service := range services {
+			if service.Selector().Matches(p.Labels()) {
+				p.AddServiceID(service.ID())
+			}
+		}
+		nodeID := report.MakePodNodeID(p.Namespace(), p.Name())
+		result.Nodes[nodeID] = p.GetNode()
```
tomwilkie: AddNode

app/router.go
```diff
@@ -188,3 +195,31 @@ type optionValue struct {
 }
 
 func nop(r render.Renderer) render.Renderer { return r }
+
+func addKubernetesTopologies() {
```
tomwilkie: I wonder weather instead of adding / removing the topology, it would be better to have a flag like "hide if empty" that the UI could honor?

probe/main.go
```diff
@@ -134,6 +138,16 @@ func main() {
 		defer dockerRegistry.Stop()
 	}
 
+	if *kubernetesEnabled {
+		client, err := kubernetes.NewClient(*kubernetesAPI, *kubernetesInterval)
+		if err != nil {
+			log.Fatalf("Kubernetes: failed to start client: %v", err)
```
tomwilkie: We've stop using fatal here I think.

probe/main.go
```diff
@@ -134,6 +138,16 @@ func main() {
...
+		}
+		defer client.Stop()
```
tomwilkie: Client will be nil on err.

render/mapping.go
```diff
@@ -463,6 +559,64 @@ func MapContainerImage2Name(n RenderableNode, _ report.Networks) RenderableNodes
...
+		return RenderableNodes{}
+	}
+
+	// Add container-<id> key to NMD, which will later be counted to produce the minor label
+	result := NewRenderableNodeWith(id, "", "", id, n)
+	result.Node.Counters[containersKey] = 1
+	if s := strings.SplitN(id, "/", 2); len(s) == 2 {
+		result.LabelMajor = s[1]
+		result.Node.Metadata[kubernetes.Namespace] = s[0]
+		result.Node.Metadata[kubernetes.PodName] = s[1]
```
tomwilkie: Why are these needed?

tomwilkie:
Quick first pass; it looks pretty good.  Shame we can't make the connections work properly; something to come back to perhaps.

app/router.go
paulbellamy: The thought was to have it show the button if there had *ever* been k8s data received, so that if there was no k8s data in one report the button wouldn't hide again. But, I suppose we could handle that in the UI. Adding topologies makes sense from a plugin sense, but for something baked in it *does* seem a bit haphazard.

render/mapping.go
paulbellamy: There is a bug in k8s where addon pods on the master are not returned via the API. The containers are still labeled (so still show up in here). This is a workaround until the bug is fixed. Will add a comment.

paulbellamy:
@tomwilkie I suspect the connections is something obvious I'm missing, but yeah. Definitely to revisit.

tomwilkie:
Make sure we have a bug to track the connection improvements.

probe/main.go
peterbourgon: Yeah, once you `defer`, `log.Fatal` is off the table.

app/router.go
```diff
@@ -96,6 +97,9 @@ func makeReportPostHandler(a xfer.Adder) http.HandlerFunc {
 			return
 		}
 		a.Add(rpt)
+		if len(rpt.Pod.Nodes) > 0 {
+			addKubernetesTopologies()
+		}
```
peterbourgon: Yeah, agree with @tomwilkie here, the dynamism feels a bit awkward... will have to think a bit on an alternative...

render/render.go
```diff
@@ -291,6 +293,14 @@ func FilterSystem(r Renderer) Renderer {
...
 				return false
 			}
+			if node.Metadata[kubernetes.Namespace] == "kube-system" {
+				fmt.Printf("[DEBUG] Filtered k8s pod/svc: %s %s\n", node.Metadata[kubernetes.PodID], node.Metadata[kubernetes.ServiceID])
+				return false
+			}
+			if strings.HasPrefix(node.Metadata[docker.LabelPrefix+"io.kubernetes.pod.name"], "kube-system/") {
+				fmt.Printf("[DEBUG] Filtered k8s container: %s\n", node.Metadata[docker.ContainerID])
+				return false
+			}
```
peterbourgon: Lingering DEBUG

probe/kubernetes/client.go
```diff
@@ -0,0 +1,91 @@
...
+// Client keeps track of running kubernetes pods and services
+type Client interface {
+	Stop()
+	WalkPods(f func(Pod) error) error
+	WalkServices(f func(Service) error) error
+}
+
+type client struct {
+	quit             chan struct{}
+	client           cache.Getter
```
peterbourgon: What happens when the client loses connectivity to the k8s store?

probe/kubernetes/service.go
```diff
@@ -0,0 +1,105 @@
...
+	return result
+}
+
+func (s *service) ips() []string {
+	ips := []string{s.Spec.ClusterIP}
+	if s.Spec.Type == api.ServiceTypeClusterIP {
+		return ips
+	}
+
+	// TODO: Get node ips here
```
paulbellamy: Fix this

probe/main.go
```diff
@@ -36,6 +37,9 @@ func main() {
 		dockerEnabled      = flag.Bool("docker", false, "collect Docker-related attributes for processes")
 		dockerInterval     = flag.Duration("docker.interval", 10*time.Second, "how often to update Docker attributes")
 		dockerBridge       = flag.String("docker.bridge", "docker0", "the docker bridge name")
+		kubernetesEnabled  = flag.Bool("kubernetes", false, "collect kubernetes-related attributes for containers, should only be enabled on the master node")
+		kubernetesAPI      = flag.String("kubernetes.api", "http://localhost:8080", "Address of kubernetes master api")
+		kubernetesInterval = flag.Duration("kubernetes.interval", 10*time.Second, "how often to update Kubernetes attributes")
```
paulbellamy: Make this more explicit

render/mapping.go
```diff
@@ -18,11 +19,16 @@ const (
 	UncontainedID    = "uncontained"
 	UncontainedMajor = "Uncontained"
 
+	UnmanagedID    = "unmanaged"
+	UnmanagedMajor = "Unmanaged"
```
paulbellamy: Remove these

render/mapping.go
```diff
@@ -463,6 +559,64 @@ func MapContainerImage2Name(n RenderableNode, _ report.Networks) RenderableNodes
...
+// It does not have enough info to do that, and the resulting graph
+// must be merged with a container graph to get that info.
+func MapContainer2Pod(n RenderableNode, _ report.Networks) RenderableNodes {
+	// Propogate all pseudo nodes
+	if n.Pseudo {
+		return RenderableNodes{n.ID: n}
+	}
+
+	// Otherwise, if some some reason the container doesn't have a pod_id
+	// (maybe slightly out of sync reports), just drop it
```
paulbellamy: (or it's not in a pod)

render/mapping.go
```diff
@@ -430,6 +494,38 @@ func MapContainer2ContainerImage(n RenderableNode, _ report.Networks) Renderable
...
+// It does not have enough info to do that, and the resulting graph
+// must be merged with a pod graph to get that info.
+func MapPod2Service(n RenderableNode, _ report.Networks) RenderableNodes {
+	// Propogate all pseudo nodes
+	if n.Pseudo {
+		return RenderableNodes{n.ID: n}
+	}
+
+	// Otherwise, if some some reason the container doesn't have a service_ids
+	// (maybe slightly out of sync reports), just drop it
```
paulbellamy: (or not a service)

render/mapping.go
```diff
@@ -430,6 +494,38 @@ func MapContainer2ContainerImage(n RenderableNode, _ report.Networks) Renderable
...
+	ids, ok := n.Node.Metadata[kubernetes.ServiceIDs]
+	if !ok {
+		return RenderableNodes{}
+	}
+
+	result := RenderableNodes{}
+	for _, id := range strings.Fields(ids) {
+		n := NewDerivedNode(id, n)
+		n.Node.Counters[podsKey] = 1
+		result[id] = n
```
paulbellamy: maybe use AddNode here, too?

render/mapping.go
```diff
@@ -463,6 +559,64 @@ func MapContainerImage2Name(n RenderableNode, _ report.Networks) RenderableNodes
 	return RenderableNodes{name: node}
 }
 
+// MapService2Name maps container images RenderableNodes to
+// RenderableNodes for each service name.
+//
+// This mapper is unlike the other foo2bar mappers as the intention
+// is not to join the information with another topology.  Therefore
+// it outputs a properly-formed node with labels etc.
+func MapService2Name(n RenderableNode, _ report.Networks) RenderableNodes {
```
paulbellamy: Unused.

render/mapping.go
```diff
@@ -271,6 +311,30 @@ func MapContainer2IP(m RenderableNode, _ report.Networks) RenderableNodes {
 	return result
 }
 
+// MapKubeProxy2IP maps the kube-proxy process to it's IP addresses (outputs
+// multiple nodes).  This allows container to be joined directly with the
+// endpoint topology. All Kubernetes service traffic goes through kube-proxy,
+// so we split it up, because that is not a helpful thing to display in the
+// kubernetes view.
+func MapKubeProxy2IP(m RenderableNode, _ report.Networks) RenderableNodes {
```
paulbellamy: Unused

probe/kubernetes/client.go
paulbellamy: If it fails when listing or resyncing, it will retry every resync period (default: 10s). If it fails when trying to watch, it will retry every second.

paulbellamy:
I believe I've addressed all concerns with this, so if you could have another quick look, and let's merge this.

probe/kubernetes/client.go
peterbourgon: :+1:

tomwilkie:
I haven't tested the code myself (involved setting up k8s).  And I'm hesitant as I don't think anyone will be regularly running this, so I very much expect it to bit rot.

But given its not enabled by default, the risks of merging this are minimal.  LGTM.

tomwilkie:
Pls squash!

peterbourgon:
I'll give it a review at EOD. Please wait until then :)

app/router.go
```diff
@@ -133,50 +137,84 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
...
+	defer r.RUnlock()
+	t, ok := r.items[name]
+	return t, ok
+}
+
+func (r *registry) list() map[string]topologyView {
+	r.RLock()
+	defer r.RUnlock()
+	return r.items
+}
```
peterbourgon: If the consumer modifies a returned topologyView, this can be racy :\

probe/main.go
```diff
@@ -135,6 +139,16 @@ func main() {
...
 
+	if *kubernetesEnabled {
+		client, err := kubernetes.NewClient(*kubernetesAPI, *kubernetesInterval)
+		if err != nil {
+			log.Printf("Kubernetes: failed to start client: %v", err)
+		} else {
+			defer client.Stop()
+			reporters = append(reporters, kubernetes.NewReporter(client))
+		}
+	}
```
peterbourgon: Very sorry for the nit, but

```go
		if client, err := kubernetes.NewClient(*kubernetesAPI, *kubernetesInterval); err != nil {
			defer client.Stop()
			reporters = append(reporters, kubernetes.NewReporter(client))
		} else {
			log.Printf("Kubernetes: failed to start client: %v", err)
		}
```

render/mapping.go
```diff
@@ -463,6 +532,45 @@ func MapContainerImage2Name(n RenderableNode, _ report.Networks) RenderableNodes
...
+		return RenderableNodes{}
+	}
+
+	// Add container-<id> key to NMD, which will later be counted to produce the
+	// minor label
+	result := NewRenderableNodeWith(id, "", "", id, n)
+	result.Node.Counters[containersKey] = 1
+	// Due to a bug in kubernetes, addon pods on the master node are not returned
+	// from the API. This is a workaround until
+	// https://github.com/kubernetes/kubernetes/issues/14738 is fixed.
```
peterbourgon: :+1:

app/api_topologies_test.go
```diff
@@ -1,9 +1,16 @@
...
+	"bytes"
+	"encoding/gob"
 	"encoding/json"
 	"net/http/httptest"
 	"testing"
+
+	"github.com/weaveworks/scope/probe/kubernetes"
+	"github.com/weaveworks/scope/report"
+	"github.com/weaveworks/scope/test"
+	"k8s.io/kubernetes/pkg/api"
```
peterbourgon: Move to another block between stdlib and weaveworks

peterbourgon:
re: the racy thing, in the best case, you could return a deep copy, but I recognize that's super annoying... maybe just a warning comment on get and list methods to not modify the return values?

Other than those little nits, LGTM.

app/router.go
paulbellamy: Yeah.. But.. Don't do that?

tomwilkie: In other cases we've added a Walk method to solve this.

Issue #442, opened 2015-09-04
Applications view goes blank after a few seconds


2opremio:
This is recored right after loading the view in the ECS demo:

![bye bye](http://g.recordit.co/VIWx9AGv9V.gif)

The gif loop gives the impression of it coming back, but it doesn't. I had to reload the page in order to get the view back.

PR #443, opened 2015-09-04
Fix JS error on missing edge target

davkal:
* replace mergeDeep with merge on node updates to set adjacency to
  updated value
* added json-loader to make json import of websocket frames easier
* console message for this was `Cannot read property 'id' of undefined`


davkal:
Fixes #442

tomwilkie:
Can confirm it didn't go blank, even whilst graph was updating.

Issue #157, opened 2015-06-02
Nodes in graph are animated, but edges are not


tomwilkie:
This looks weird when resizing, and when your graph is changing.

davkal:
True. I left that out. Animating paths has a pretty high performance penalty because it cannot use the GPU like I do for nodes. But maybe we can make it a bit nicer what we have.

davkal:
Fixed by #445

PR #445, opened 2015-09-04
Animate edges

davkal:
Please test this with big topos first.


davkal:
Fixes #157

peterbourgon:
As I understand it, #473 includes this PR, so once that's merged, this can be closed.

PR #449, opened 2015-09-06
Prevent short lived connections appearing as pseudo nodes in the applications view.

### LINKED ISSUE #447: Short lived connections are showing up as pseudo nodes in the the applications view.
screen shot 2015-09-07 at 11 07 03

tomwilkie:
Fixes #447

### DIFF
--- a/render/mapping.go
+++ b/render/mapping.go
@@ -45,9 +45,16 @@ func MapEndpointIdentity(m RenderableNode, local report.Networks) RenderableNode
 		return RenderableNodes{}
 	}
 
+	// We need to filter out short lived connections from this view,
+	// if they don't also have a pid; see #447
+	pid, pidOK := m.Metadata[process.PID]
+	_, conntracked := m.Metadata[endpoint.Conntracked]
+	if !pidOK && conntracked {
+		return RenderableNodes{}
+	}
+
 	// Nodes without a hostid are treated as psuedo nodes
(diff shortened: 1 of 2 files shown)


peterbourgon:
For context,

```
<Tom> For short lived connections, we don't know the pid, so they can't be merged
<Tom> so they just shouldn't be there
<Tom> which is what the PR does; filters out shortlived connections which have no pids
<Peter> Ah, and in the containers view, it's OK, because while we don't know the PID we do know the [IP address, which is enough to join to a container]
```

render/mapping.go
```diff
@@ -45,9 +45,16 @@ func MapEndpointIdentity(m RenderableNode, local report.Networks) RenderableNode
 		return RenderableNodes{}
 	}
 
+	// We need to filter out short lived connections from this view,
+	// if they don't also have a pid; see #447
+	pid, pidOK := m.Metadata[process.PID]
+	_, conntracked := m.Metadata[endpoint.Conntracked]
+	if !pidOK && conntracked {
```
peterbourgon: OK, and you only squelch the node if you see it's conntracked, otherwise pseudo is created.

peterbourgon:
LGTM.

render/mapping.go
tomwilkie: Yes the previous behaviour (from procspy) is preserved - pseudo nodes are created for non-contracked connections.

PR #450, opened 2015-09-07
Optimize /proc traversal

inercia:
The objective of this PR is to reduce the number of syscalls when collecting info from `/proc`. This PR does that by: 

* reading the required files (`/proc/*/[cmdline, comm, stat, fd/*]`, `/proc/net/tcp*`) once per cycle (`spy.interval`), parsing and caching the processes/connections information
* keeping a cache of open files for `/proc/*/[cmdline, comm, stat]` so we do not `open()`/`close()` these files so often.
* removing unnecessary `stat()`s when reading `/proc` directories

Changes:

* the required code for reading/parsing `/proc` has been proved from `procspy` to `/probe/proc`
* the `Walker` has been replaced by a `proc.Reader` that provides both `Processes()` and `Connections()`.  Users can walk the list of processes/connections in the same way by providing a function to these functions. A `CachingReader` can be used for caching and getting this info, updated when `Update()`. The code for reading and parsing connection and processes has been simplified.
* a `CachingReader` is initialized in `/probe/main.go` and used in both `endpoint.Reporter` and `process.Reporter` for reporting about processed and connections, and updated periodically by calling `Update()` every `spy.interval` seconds.
* unit tests have been improved by mocking things like file system operations
* things like `Proc` and `Process` have been unified, and many things from `/probe/process` have been moved to `/probe/proc`

This is part of the solution to #284


2opremio:
@inercia I don't know if it provides all you need but maybe you could use (and even extend)  https://github.com/mitchellh/go-ps

2opremio:
Actually, I don't think `go-ps` provides connection/file information, but https://github.com/c9s/goprocinfo certainly does and it seems more mature.

inercia:
@2opremio thanks with the hint... `goprocinfo` seems interesting. The only problem I see is that they use `Read()`s instead of our current mechanism where we `buf.ReadFrom(someFile)` and the `buf` is obtained from a `Pool`, so we do not generate too much garbage... it's worth some investigation, though...

probe/proc/reader.go
```diff
@@ -0,0 +1,116 @@
...
+	Cmdline   string
+	Threads   int
+	Inodes    []uint64
+}
+
+// ProcReader is something that reads the /proc directory and
+// returns some info like processes and connections
+type ProcReader interface {
+	// Processes walks through the processes
+	Processes(func(Process)) error
```
inercia: Maybe `WalkProcesses` and `WalkConnections`?

inercia:
@tomwilkie or @peterbourgon could you please review this?

peterbourgon:
@inercia Sure. Would you mind giving a short overview of what this PR does, including motivation & pointers to key types/functions/changes?

probe/proc/reader_linux.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	return nil
+}
+
+var bufPool = sync.Pool{
+	New: func() interface{} {
+		return bytes.NewBuffer(make([]byte, 0, 5000))
+	},
+}
```
inercia: Not sure we should keep this `Pool` anymore: we do not read the `/proc` concurrently from multiple goroutines (we only do it when `Update()`ing the `CachingReader`), so we could use only one `bytes.Buffer` and get rid of all the `Pool.Get`/`Pool.Put`/`Buffer.Reset` cycle...

inercia:
@peterbourgon I have updated the PR description, so please let me know what you think about this...

inercia:
This is the latest profile graph from this branch (using go 1.5). Comapred with the original graph in #284, we can see that the number of `syscalls` has been vastly reduced, but we sped a lot of time in `runtime.*` now (too much garbage, maybe?)...

profile

peterbourgon:
@inercia Thanks, that helps a lot. I'll give it a look.

probe/proc/net.go
```diff
@@ -0,0 +1,42 @@
...
+func (c Connection) Copy() Connection {
+	dupIP := func(ip net.IP) net.IP {
+		dup := make(net.IP, len(ip))
+		copy(dup, ip)
+		return dup
+	}
+
+	c.LocalAddress = dupIP(c.LocalAddress)
+	c.RemoteAddress = dupIP(c.RemoteAddress)
+	return c
```
peterbourgon: What about the Process.Inodes?

probe/proc/reader_linux.go
peterbourgon: Yes, putting the buffer into the reader struct? If so, +1.

probe/proc/reader_linux.go
```diff
@@ -0,0 +1,203 @@
...
+
+// Connections walks through all the connections in the "/proc"
+func (w *reader) Connections(withProcs bool, f func(Connection)) error {
```
peterbourgon: Presumably `r *reader` is meant.

probe/proc/reader_linux.go
```diff
@@ -0,0 +1,203 @@
...
+		return err
+	}
+
+	buf := bytes.NewBuffer(make([]byte, 0, 5000))
+
+	for _, subdir := range dirEntries {
+		readIntoBuffer := func(filename string) error {
+			buf.Reset()
+			res := w.handles.ReadInto(path.Join(subdir, filename), buf)
+			return res
```
peterbourgon: `return w.handles.ReadInto(...)`

probe/proc/reader_linux.go
```diff
@@ -0,0 +1,203 @@
...
+// instances of Process, which it then passes one-by-one to the
+// supplied function. Processes() is only made public so that is
+// can be tested.
+func (w *reader) Processes(f func(Process)) error {
+	dirEntries, err := w.proc.ReadDirNames(w.proc.Root())
+	if err != nil {
+		return err
+	}
+
+	buf := bytes.NewBuffer(make([]byte, 0, 5000))
```
peterbourgon: Why the explicit initialization?

probe/proc/reader_linux.go
```diff
@@ -0,0 +1,203 @@
...
+
+		fdBase := path.Join(w.proc.Root(), strconv.Itoa(pid), "fd")
+		fdNames, err := w.proc.ReadDirNames(fdBase)
+		if err != nil {
+			return err
+		}
+
+		inodes := []uint64{}
+		for _, fdName := range fdNames {
+			var fdStat syscall.Stat_t
```
peterbourgon: Can this be safely moved out of the loop and re-used, perhaps?

probe/proc/reader_linux.go
```diff
@@ -0,0 +1,203 @@
...
+	filesCacheExpiration = 60 * time.Second
+)
+
+var tcpFiles = []string{
+	"net/tcp",
+	"net/tcp6",
+}
+
+// A cache for files handles
+// Note: not intended to be used from multiple goroutines
```
peterbourgon: Wouldn't it be better to make it goroutine safe, than put this warning?

probe/proc/reader_linux.go
```diff
@@ -0,0 +1,203 @@
...
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/bluele/gcache"
+)
+
+const (
+	filesCacheLen        = 512
+	filesCacheExpiration = 60 * time.Second
```
peterbourgon: Just to be clear, a single file is for example /proc/<pid>/something, is that correct? If so, is 60 seconds perhaps too long?

probe/proc/reader_darwin.go
```diff
@@ -44,6 +44,59 @@ func (walker) Walk(f func(Process)) error {
 	return nil
 }
 
+func (w *reader) Connections(withProcs bool, f func(Connection)) error {
```
peterbourgon: Presumably `r *reader` is meant here (further comments like this elided)

probe/main.go
```diff
@@ -186,7 +186,7 @@ func main() {
 				r = report.MakeReport()
 
 			case <-spyTick:
-				if err := processCache.Update(); err != nil {
+				if err := procReader.Update(*spyProcs); err != nil {
```
peterbourgon: Great. This will help.

peterbourgon:
The shape of it is good. Back to @inercia for those few comments (and rebasing, I guess).

probe/proc/reader_linux.go
inercia: Done.

probe/proc/net.go
inercia: Fixed. Added a `Process.Copy()`

probe/proc/reader_linux.go
inercia: Right. Fixed.

probe/proc/reader_linux.go
inercia: Fixed.

probe/proc/reader_linux.go
inercia: Because I copied&pasted... :smile: Fixed.

probe/proc/reader_linux.go
inercia: Done.

probe/proc/reader_linux.go
inercia: Ok, but we are adding a mutex for something that will probably never be used concurrently... but not a big deal.

probe/proc/reader_linux.go
inercia: Yes, it is a cache for `/proc/something`, but for file handles, not for content. The most common case is when we read from `/proc/<some_proc_id>/[comm, cmdline, stat]`: we try to get from the cache and hopefully we will get an already-open file handle. A `60 * time.Second` expiration time would mean that we could have entries in the cache for up to 60 seconds that will never be used...

Further improvements would be to cache the content for `comm` and `cmdline`, and we could even cache the `stat` info for `/proc/<some_proc_id>/fd/*`.

probe/proc/reader_darwin.go
inercia: Fixed.

probe/proc/reader_linux.go
peterbourgon: Gotcha!

peterbourgon:
:+1: pending Coveralls. (Maybe Coveralls is broken?)

inercia:
@peterbourgon I think the coveralls hook has failed (the jobs is finisished, but the "Coverage pending" is still here). I'm merging....

PR #451, opened 2015-09-07
Small improvements to reverse resolver

tomwilkie:
Don't rate limit requests that have already been resolved, and cache negative responses.

We do this as we generate so many reverse-ip lookups that we quickly overwhelm the the channel and start dropping them.  At a rate of 10 lookups per second, we never get through the queue.

By not rate limiting requests that are in the cache we don't actually do any more requests, it just allows us to burn through the queue quicker.  And by caching negative responses, we prevent the queue for growing massive to begin with.

I tested this by doing a ```telnet google.com 80``` and then checking we got the right hostname in the applications view.  Before this fix it would never show any hostname, after the fix it shows something.

### DIFF
--- a/probe/endpoint/resolver.go
+++ b/probe/endpoint/resolver.go
@@ -1,6 +1,7 @@
 package endpoint
 
 import (
+	"fmt"
 	"net"
 	"strings"
 	"time"
@@ -13,6 +14,8 @@ const (
 	rAddrBacklog         = 1000
 	rAddrCacheExpiration = 30 * time.Minute
 )
+
+var errNotFound = fmt.Errorf("Not found")
 
 type revResFunc func(addr string) (names []string, err error)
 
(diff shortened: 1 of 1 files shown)


probe/endpoint/resolver.go
```diff
@@ -52,20 +58,22 @@ func (r *ReverseResolver) Get(address string) (string, error) {
...
 
 func (r *ReverseResolver) loop() {
 	for request := range r.addresses {
-		<-r.Throttle // rate limit our DNS resolutions
-		// and check if the answer is already in the cache
+		// check if the answer is already in the cache
 		if _, err := r.cache.Get(request); err == nil {
 			continue
 		}
+		<-r.Throttle // rate limit our DNS resolutions
```
inercia: Good point. Throttling should happen _after_ the `Get`.

inercia:
LGTM.

Issue #432, opened 2015-09-03
Allow users to apply a transform to container names


tomwilkie:
See #413

2opremio:
As discussed with @peterbourgon the idea (for now) is to render names based on the container's context. For instance, in the case of ECS, [container names are mangled](https://github.com/weaveworks/guides/tree/5a70632985c5f1c64dcb807e35cb33cab6a7745e/aws-ecs#whats-happening-in-the-hosts) but ECS provides the original container name in a Docker container label: `com.amazonaws.ecs.container-name`

The probes will obtain the container's cloud context and include a Container report metadata entry, say `"provider"="ECS"` which the renderer can use to adapt the container name to display (from the `com.amazonaws.ecs.container-name`) and tweak other presentation details.

2opremio:
@paulbellamy I would be interested to know if this would work well for kubernetis too.

paulbellamy:
A transform would work ok. There's not really a container-specific label that would work well in the container's view. There's just the pod name label, which will be shared between multiple containers.

tomwilkie:
Sounds fine.

> The probes will obtain the container's cloud context and include a Container report metadata entry, say "provider"="ECS"

How will it know this?  Is it enough to 'just' use com.amazonaws.ecs.container-name as the name if its present?

2opremio:
I actually agree with @tomwilkie we should just use the label if present for now. We can use a `provider` metadata entry when/if the cloud-provider-specific rendering gets more sophisticated.

tomwilkie:
@fons do you think we can close this?  I don't think we need anything more general, do we?

2opremio:
Not for now, I'll close it

PR #456, opened 2015-09-07
Use label "com.amazonaws.ecs.container-name" to render container names

2opremio:
Addresses https://github.com/weaveworks/scope/issues/432 in the case of Amazon's ECS.

### DIFF
--- a/render/detailed_node.go
+++ b/render/detailed_node.go
@@ -334,7 +334,7 @@ func containerOriginTable(nmd report.Node, addHostTag bool) (Table, bool) {
 
 	var (
 		title           = "Container"
-		name, nameFound = nmd.Metadata[docker.ContainerName]
+		name, nameFound = GetRenderableContainerName(nmd)
 	)
 	if nameFound {
 		title += ` "` + name + `"`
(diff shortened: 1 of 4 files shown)


render/container.go
```diff
@@ -0,0 +1,36 @@
...
+	// Task Definition "task-example" with version 5 will lead to a container name
+	// "task-example-5-httpserver-aceb93e2f2b797caba01"
+	//
+	// Note how the dash in http-server is removed not allowing to recover the original
+	// Container Definition name.
+	//
+	// Luckily, the ecs-agent provides label
+	// "com.amazonaws.ecs.container-name" containing the original Container
+	// Definition name, which we will use for rendering.
+	//
```
tomwilkie: Thats a pretty large comment for something that pretty obvious from the code.

render/container.go
```diff
@@ -0,0 +1,36 @@
+package render
```
tomwilkie: There is about 5 lines of 'real' code in this file - I'd drop the comment and stick this function in mapping.go

2opremio: I will simplify the comment, but I am not in favor of completely dropping it.

tomwilkie:
LGTM

PR #459, opened 2015-09-07
Make report propagation more reliable

### LINKED ISSUE #458: Scope view intermittently goes blank
Don't think this is a UI issue - the app is sending a message to the UI deleting all the nodes.

### LINKED ISSUE #467: Need to bring back connection backoff (when probe pushes to app)

tomwilkie:
Fixes #458, fixes #467 

- Make scope push the reports asynchronously
- Backoff when report push fails

### DIFF
--- a/xfer/publisher.go
+++ b/xfer/publisher.go
@@ -17,6 +17,7 @@ import (
 // Publisher is something which can send a report to a remote collector.
 type Publisher interface {
 	Publish(report.Report) error
+	Stop()
 }
 
 // HTTPPublisher publishes reports by POST to a fixed endpoint.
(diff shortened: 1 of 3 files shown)


report/id.go
```diff
@@ -1,8 +1,13 @@
 package report
 
 import (
+	"hash"
+	"hash/fnv"
 	"net"
 	"strings"
+	"sync"
+
+	cache "github.com/hashicorp/golang-lru"
```
peterbourgon: Alvaro is already using a different LRU cache in one of the probe packages. Can we settle on one dependency? And can we not rename imports unless there's a conflict?

report/id.go
```diff
@@ -21,9 +26,37 @@ const (
...
+var (
+	idCache, _ = cache.New(1024)
+	hashers    = sync.Pool{
+		New: func() interface{} {
+			return fnv.New64a()
+		},
+	}
+)
+
+func lookupID(_1, _2, _3 string, f func() string) string {
```
peterbourgon: Please use different var names :\

report/topology.go
```diff
@@ -20,16 +20,15 @@ func MakeTopology() Topology {
 	}
 }
 
-// WithNode produces a topology from t, with nmd added under key nodeID; if a
-// node already exists for this key, nmd is merged with that node. Note that a
-// fresh topology is returned.
-func (t Topology) WithNode(nodeID string, nmd Node) Topology {
+// AddNode adds node to the topology under key nodeID; if a
+// node already exists for this key, nmd is merged with that node.
+// The same topology is returned to enable chaining.
```
peterbourgon: Can you add some language here that indicates this method is different from all the other similar methods (it mutates the Topology) to solve issues of GC pressure? Otherwise it's liable to be removed in a refactor.

report/id.go
tomwilkie: The package name differs from the last part of the URL (see https://github.com/hashicorp/golang-lru/blob/master/lru.go#L4) and I thought this made it more obvious.

tomwilkie:
The problem was we were pushing reports on the main probe thread, one-by-one.  If one of the ips was unreachable, it would block the pushing of all others, and the probe would stall and the UI go blank.  And tcp connection time out is quite long.

xfer/publisher.go
```diff
@@ -83,12 +94,73 @@ func (p HTTPPublisher) Publish(rpt report.Report) error {
 	return nil
 }
 
+// Stop implements Publisher
+func (p HTTPPublisher) Stop() {}
```
paulbellamy: Should we track the current request and do `http.Transport.CancelRequest(r)` here?

xfer/publisher.go
```diff
@@ -83,12 +94,73 @@ func (p HTTPPublisher) Publish(rpt report.Report) error {
...
+	publisher Publisher
+	reports   chan report.Report
+	quit      chan struct{}
+}
+
+// NewBackgroundPublisher creates a new BackgroundPublisher with the given publisher
+func NewBackgroundPublisher(p Publisher) *BackgroundPublisher {
+	result := &BackgroundPublisher{
+		publisher: p,
+		reports:   make(chan report.Report),
```
paulbellamy: On the second time around, won't this just fill up, and block the other publishers anyway?

paulbellamy:
When full, this seems to attempt to throw away new reports, and keep retrying the original one which failed. Don't we really only care about the latest report?

xfer/publisher.go
tomwilkie: File a bug for it; this change is not a regression against the existing codebase, but we should do it in the future.

tomwilkie:
We don't retry the sending of any reports - why do you think it does?

paulbellamy:
You are right, misread the backoff waiting bit.

paulbellamy:
Aside from making the http handler cancel the existing request (if that is important), LGTM.

Issue #446, opened 2015-09-04
Missing edges between the ECS-demo containers


2opremio:
Steps to reproduce.

1. Launch the scope variant of the ECS demo (which uses `scope:latest`)

    ```
git clone -b aws-ecs-scope git@github.com:weaveworks/guides.git
cd guides/aws-ecs
./setup
```

   Which should output three urls let's call them `u1`, `u2`, `u3`.

2. Make sure that the Http Servers are communicating with the Data Producers:

  ```bash
  while true; do 
    for I in u1 u2 u3; do 
      curl $I
    done
  done
  ```

3. Open Scope in any of the three instances (e.g. `u1:4040`)
    
    The HttpServer and DataProducers ( the `ecs-weave-ecs...` containers) should be connected since they are communicating with each other. However they are not.
<img width="1134" alt="screen shot 2015-09-04 at 4 11 28 pm" src="https://cloud.githubusercontent.com/assets/2362916/9687147/b02d4b78-531f-11e5-8a12-5d260cebcf11.png">

2opremio:
In fact, I also expect an edge between all the weavescope containers since every probe is reporting to all the apps.

tomwilkie:
@2opremio attaching the contents of /api/report and /api/topology/containers to bugs like this will massively help narrow it down.

2opremio:
There you go https://gist.github.com/2opremio/9494a1f48b340a620526

tomwilkie:
I've figured this out.

My bad.

2opremio:
Great that you figured it out. Out of curiosity, can you elaborate on what's causing it?

tomwilkie:
I thought I had it but I didn't. Sorry.  Still working on it.

peterbourgon:
Any update here? This is the only remaining release-blocker.

tomwilkie:
The attached or fixes some issues but doesn't fix it 100%.

I've got two test cases: cross host communication over the weave network,
and cross host communication via docker port mapping.  (1) works now, (2)
doesn't.

I don't know why it doesn't work yet.

Feel free to review the pr, I see no reason why it can't go in.

tomwilkie:
This is mostly fixes - pseudo nodes are working again, and short lived connections from containers to the internet are working.

But the short lived connection code doesn't deal with connections coming from the internet, as they go to the hosts IP, not the containers.  We know the port mapping for containers, so we should be able to fix this.  See #493

PR #465, opened 2015-09-08
Fix regression in missing edges

tomwilkie:
Fix for #446

- When applying NAT mappings, merge with existing nodes
- Catch NAT mappings which already exist when scope is started

### DIFF
--- a/probe/endpoint/conntrack.go
+++ b/probe/endpoint/conntrack.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"encoding/xml"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"strings"
@@ -59,21 +60,28 @@ type Flow struct {
 	Original, Reply, Independent *Meta `xml:"-"`
 }
 
+type conntrack struct {
+	XMLName xml.Name `xml:"conntrack"`
+	Flows   []Flow   `xml:"flow"`
+}
+
 // Conntracker uses the conntrack command to track network connections
 type Conntracker struct {
 	sync.Mutex
 	cmd           exec.Cmd
 	activeFlows   map[int64]Flow // active flows in state != TIME_WAIT
 	bufferedFlows []Flow         // flows coming out of activeFlows spend 1 walk cycle here
+	existingConns bool
 }
 
 // NewConntracker creates and starts a new Conntracter
-func NewConntracker(args ...string) (*Conntracker, error) {
+func NewConntracker(existingConns bool, args ...string) (*Conntracker, error) {
 	if !ConntrackModulePresent() {
 		return nil, fmt.Errorf("No conntrack module")
 	}
 	result := &Conntracker{
-		activeFlows: map[int64]Flow{},
+		activeFlows:   map[int64]Flow{},
+		existingConns: existingConns,
 	}
 	go result.run(args...)
 	return result, nil
@@ -105,6 +113,19 @@ var ConntrackModulePresent = func() bool {
 
 // NB this is not re-entrant!
 func (c *Conntracker) run(args ...string) {
+	if c.existingConns {
+		// Fork another conntrack, just to capture existing connections
+		// for which we don't get events
+		existingFlows, err := c.existingConnections(args...)
+		if err != nil {
+			log.Printf("conntrack existingConnections error: %v", err)
+			return
+		}
+		for _, flow := range existingFlows {
+			c.handleFlow(flow, true)
+		}
+	}
+
 	args = append([]string{"-E", "-o", "xml", "-p", "tcp"}, args...)
 	cmd := exec.Command("conntrack", args...)
 	stdout, err := cmd.StdoutPipe()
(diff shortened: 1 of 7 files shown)


2opremio:
@tomwilkie I know this is not done yet, but, continuing our conversation from yesterday, I am missing an edge from `The Internet` node to the `httpserver` node when connecting to it. 

screen shot 2015-09-09 at 11 02 34


It could also be that I don't understand how the `The Internet` node is supposed to work.

tomwilkie:
Thanks alvaro; please send the output of /api/report for this kind of bug.

2opremio:
There you go. https://gist.github.com/2opremio/322736b35abceadf75a6

Probably unrelated, but I had rebased this PR on master when I saw this.

tomwilkie:
@2opremio please try again, I believe this to be almost* fixed.

*almost in that I'm tracking down a small bug where the links 'flutter' - come and go.

tomwilkie:
Flutter bug is fixed by the Nodes.Merge cset.  This has introduce a bug in pseudo node rendering.  The whackamole continues.

tomwilkie:
Pseudo node rendering & Hosts view rendering both fixed now too.

app/router.go
```diff
@@ -76,6 +77,11 @@ func Router(c collector) *mux.Router {
 	return router
 }
 
+func badRequest(r *http.Request, w http.ResponseWriter, err error) {
+	http.Error(w, err.Error(), http.StatusBadRequest)
+	log.Printf("Error procressing request for %s: %v", r.URL.Path, err)
+}
```
peterbourgon: [Rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) dictates this shouldn't be factored out yet :)

probe/endpoint/conntrack.go
```diff
@@ -143,15 +164,45 @@ func (c *Conntracker) run(args ...string) {
 		return
 	}
 
+	defer func() {
+		log.Printf("contrack exiting")
+	}()
```
peterbourgon:
```go
defer log.Printf("...")
```

probe/endpoint/conntrack.go
```diff
@@ -143,15 +164,45 @@ func (c *Conntracker) run(args ...string) {
...
 		if err := decoder.Decode(&f); err != nil {
 			log.Printf("conntrack error: %v", err)
+			return
+		}
+		c.handleFlow(f, false)
+	}
+}
+
+func (c *Conntracker) existingConnections(args ...string) ([]Flow, error) {
+	var conntrack conntrack
```
peterbourgon: Potentially dangerous shadowing

probe/endpoint/conntrack.go
```diff
@@ -143,15 +164,45 @@ func (c *Conntracker) run(args ...string) {
...
+	args = append([]string{"-L", "-o", "xml", "-p", "tcp"}, args...)
+	cmd := exec.Command("conntrack", args...)
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return conntrack.Flows, err
```
peterbourgon: If you return non-nil err, best to return `[]Flow{}`, to prevent ambiguity in the caller.

probe/endpoint/nat.go
```diff
@@ -53,14 +53,21 @@ func toMapping(f Flow) *endpointMapping {
...
+		)
 		if !ok {
 			return
 		}
 
-		rpt.Endpoint.Nodes[copyEndpointID] = node.Copy()
+		node = node.Copy()
+		node.Metadata[Addr] = mapping.rewrittenIP
+		node.Metadata[Port] = copyEndpointPort
+		node.Metadata["copy_of"] = realEndpointID
```
peterbourgon: Just for debugging?

xfer/buffer.go
```diff
@@ -2,40 +2,73 @@ package xfer
...
 	"bytes"
+	"io"
 	"sync"
 	"sync/atomic"
 )
 
 // A Buffer is a reference counted bytes.Buffer, which belongs
 // to a sync.Pool
-type Buffer struct {
+type Buffer interface {
```
peterbourgon: Please cut the buffer stuff from this PR somehow. (Or, extract the non-buffer stuff to a different PR.)

peterbourgon:
Back to you @tomwilkie

probe/endpoint/nat.go
tomwilkie: It is.  Super useful when viewing /api/report and trying to figure out how the report came to be, so I though I'd leave it in.

app/router.go
tomwilkie: This is part of #484 and won't be committed as part of this change; I'll rebase it away now.

tomwilkie:
@peterbourgon PTAL

2opremio:
I confirm that I get the expected edges both for application and system containers in the ECS demo, however, I am missing edges between the Internet node and the `httpserver`s . I created a separate issue #449 at the request of @tomwilkie since this PR already improves the situation a lot.

peterbourgon:
LGTM

PR #469, opened 2015-09-09
Only post to one app (prototype).

### LINKED ISSUE #463: Are we doing anything to prevent the same report being send to the same app twice?

tomwilkie:
Fixes #463

This is a prototype, this approach probably isn't the way to go.


peterbourgon:
Noted :)

PR #470, opened 2015-09-09
Reduce probe cpu usage

### LINKED ISSUE #397: Do fewer RPCs to Weave
In the weave integration, we do two calls to the /report endpoint every probe interval (few seconds).

We should cache the result and just do one per cycle, or less.

tomwilkie:
2 issues - report generation is taking too long, and cpu usage is too high.

Reducing CPU usage (~40% to <20%), see #284
- WithNode -> AddNode, don't copy all the nodes every time we add a single one.
- Cache generated endpoint IDs to relieve some more GC pressure.
- Only gob serialise and gz compress reports once per publish cycle, not once per app. 
- In weave tagger, stop iterating through all containers to find a container by prefix.  Build an index instead.

Report generation:
- Log a warning when it takes too long
- Generate them in parallel
- Only fetch weave status json once per cycle (fixes #397)

### DIFF
--- a/report/id.go
+++ b/report/id.go
@@ -1,8 +1,13 @@
 package report
 
 import (
+	"hash"
+	"hash/fnv"
 	"net"
 	"strings"
+	"sync"
+
+	"github.com/bluele/gcache"
 )
 
 // TheInternet is used as a node ID to indicate a remote IP.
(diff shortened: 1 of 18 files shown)


peterbourgon:
Sorry about commenting on the commits rather than the diff view. That makes it a little harder to act on.

Back over to you.

tomwilkie:
> synchronisation

Is it needed?  Tick and {Report, Tag} are never call concurrently.  If it is, then we'll also need a pointer receiver I think?

> what else does conntrack do?

It lists (or streams) connections; udp, tcp etc, and will list of stream nat mappings.  I don't think it does anything else.

tomwilkie:
@peterbourgon I think I addressed all the issues.

peterbourgon:
> Is it needed? Tick and {Report, Tag} are never call concurrently.

That's incidental and can easily change :)

> If it is, then we'll also need a pointer receiver I think?

Yes.

xfer/publisher.go
```diff
@@ -22,7 +17,7 @@ const (
 
 // Publisher is something which can send a report to a remote collector.
 type Publisher interface {
```
peterbourgon: Comment is out-of-sync with what's happening.

peterbourgon:
@tomwilkie I made the changes I proposed. Give me a :+1: and I'll merge.

tomwilkie:
LGTM

report/id.go
```diff
@@ -21,9 +26,38 @@ const (
...
+)
+
+func lookupID(part1, part2, part3 string, f func() string) string {
+	h := hashers.Get().(hash.Hash64)
+	h.Write([]byte(part1))
+	h.Write([]byte(part2))
+	h.Write([]byte(part3))
+	sum := h.Sum64()
+	var result string
+	if id, err := idCache.Get(sum); id != nil && err != nil {
```
rade: That condition surely is wrong and results in this branch becoming unreachable; I suspect you want `id != nil && err == nil`.

Given that this bug renders the cache ineffective, any performance gains from this PR cannot be attributed to it. So some new measurements should be taken to ascertain whether the cache actually produces a noticeable gain.

Issue #324, opened 2015-07-16
Consider adding some sort of highlight around the selected node


tomwilkie:
If I have a bunch of sshds, and the graph is moving around beneath me, it sometimes becomes unclear to which node the details panel is referring.

davkal:
Fixed by #473

PR #473, opened 2015-09-09
Highlight selected node

davkal:
Highlights selected nodes like in Cello, with a circular layout of the adjacent nodes and a fade out of the non-adjacent nodes.

screen shot 2015-09-09 at 18 27 49

### DIFF
--- a/client/app/scripts/charts/edge.js
+++ b/client/app/scripts/charts/edge.js
@@ -1,5 +1,7 @@
+const _ = require('lodash');
 const d3 = require('d3');
 const React = require('react');
+const Spring = require('react-motion').Spring;
 
 const AppActions = require('../actions/app-actions');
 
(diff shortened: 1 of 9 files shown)


davkal:
This is based on PR #445 and will fix #324

peterbourgon:
Brilliant. Basically this only works when the detail pane is up, correct?

davkal:
Correct. It will revert to the old layout when you close it.

peterbourgon:
Looks great. One thing: when you click on a node, the canvas stays fixed, so some of the adjacent nodes can get moved to annoying places, like just offscreen, under the detail pane, or into (underneath) the menu bar at the top. If that could be avoided, would be nice.

davkal:
Please have another look.

peterbourgon:
Ah, works great! Nice job. :+1:

peterbourgon:
Can we get a merge?

PR #477, opened 2015-09-10
Backend build container

peterbourgon:
With :love_letter: to @davkal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,4 +52,5 @@ experimental/_integration/_integration
 *sublime-project
 *sublime-workspace
 *npm-debug.log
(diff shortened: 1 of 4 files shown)


davkal:
The building works well. However, it builds a recent master, and not the current checkout. This would be important if I want to include my latest UI changes that have been compiled into static.go.
Any chance that could be made to work?

peterbourgon:
Try that.

davkal:
Seems to work! Let me try some more.

davkal:
Yeah, works well, just tried it again. Really helpful.

PR #479, opened 2015-09-14
Makefile improvements.

tomwilkie:
- remove ephemeral containers after use
- don't rebuild the backend build container every time; make it work out its dependancies.


peterbourgon:
Works for me. :+1:

PR #481, opened 2015-09-14
Don't close the request body.

tomwilkie:
Fixes #480

PR #484, opened 2015-09-14
Can't share buffers like that!

### LINKED ISSUE #471: Manual scope setup (ie without weave) not working
report from @bboreham

tomwilkie:
As part of #470 I introduces reference counted, shared buffers such that we only serialise and compress a report once.  I got it a bit wrong: byte.Buffers contain a cursor, and therefore can only be used 'once'.

This change adapts that code such that a 'new' byte buffer is created for each report publisher, but they share the underlying []bytes, and use the reference counting to track the usage of that.

This manifested itself as a panic in the app as it couldn't decompress reports (as they were empty) - the panic was fixed in #481 without identifying the underly cause.

This change also adds error reporting on the app for why requests fail.

Fixes #471

### DIFF
--- a/app/router.go
+++ b/app/router.go
@@ -3,6 +3,7 @@ package main
 import (
 	"compress/gzip"
 	"encoding/gob"
+	"log"
 	"net/http"
 	"net/url"
 	"strings"
@@ -76,6 +77,11 @@ func Router(c collector) *mux.Router {
 	return router
 }
 
+func badRequest(r *http.Request, w http.ResponseWriter, err error) {
+	http.Error(w, err.Error(), http.StatusBadRequest)
+	log.Printf("Error procressing request for %s: %v", r.URL.Path, err)
+}
+
 func makeReportPostHandler(a xfer.Adder) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		var (
@@ -86,13 +92,13 @@ func makeReportPostHandler(a xfer.Adder) http.HandlerFunc {
 		if strings.Contains(r.Header.Get("Content-Encoding"), "gzip") {
 			reader, err = gzip.NewReader(r.Body)
 			if err != nil {
-				http.Error(w, err.Error(), http.StatusBadRequest)
+				badRequest(r, w, err)
 				return
 			}
 		}
 
 		if err := gob.NewDecoder(reader).Decode(&rpt); err != nil {
-			http.Error(w, err.Error(), http.StatusBadRequest)
+			badRequest(r, w, err)
 			return
 		}
 		a.Add(rpt)
--- a/xfer/buffer.go
+++ b/xfer/buffer.go
@@ -2,38 +2,71 @@ package xfer
 
 import (
 	"bytes"
+	"io"
 	"sync"
 	"sync/atomic"
 )
 
 // A Buffer is a reference counted bytes.Buffer, which belongs
 // to a sync.Pool
-type Buffer struct {
+type Buffer interface {
+	io.Reader
+	io.Writer
+
+	// Get returns a new buffer sharing the contents of this
+	// buffer, but with its own cursor.  Get also increases
+	// the reference count.  It is safe for concurrent calls.
+	Get() Buffer
+
+	// Put decreases the reference count, and when it hits zero, puts the
+	// buffer back in the pool.
+	Put()
+}
+
+type baseBuffer struct {
 	bytes.Buffer
 	pool *sync.Pool
 	refs int32
 }
 
+type dependentBuffer struct {
+	*bytes.Buffer
+	buf *baseBuffer
+}
+
 // NewBuffer creates a new buffer
-func NewBuffer(pool *sync.Pool) *Buffer {
-	return &Buffer{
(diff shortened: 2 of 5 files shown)


xfer/buffer.go
```diff
@@ -2,40 +2,73 @@ package xfer
...
 	"sync"
 	"sync/atomic"
 )
 
 // A Buffer is a reference counted bytes.Buffer, which belongs
 // to a sync.Pool
-type Buffer struct {
+type Buffer interface {
+	io.Reader
+	io.Writer
```
tomwilkie: Also implements ByteReader and BytesWriter; could add it here to help with compression (see http://golang.org/src/compress/gzip/gunzip.go?s=2453:2497#L74) but I don't think its necessary.

xfer/buffer.go
```diff
@@ -2,40 +2,73 @@ package xfer
...
+	Put()
+}
+
+type baseBuffer struct {
 	bytes.Buffer
 	pool *sync.Pool
 	refs int32
 }
 
+type dependantBuffer struct {
```
paulbellamy: *dependent

xfer/buffer.go
```diff
@@ -2,40 +2,73 @@ package xfer
...
+func NewBuffer(pool *sync.Pool) Buffer {
+	return &baseBuffer{
 		pool: pool,
 		refs: 0,
 	}
 }
 
-// Get increases the reference count.  It is safe for concurrent calls.
-func (b *Buffer) Get() {
+// Get implements Buffer
```
paulbellamy: This doesn't seem true?

tomwilkie: Whats not true about it?

paulbellamy: Buffer is an io.Reader and io.Writer..

tomwilkie: The embedded bytes.Buffer handles that for us.  We just need to implement Get and Put.

paulbellamy: > We just need to implement Get and Put.

You might need to, but not in order to implement Buffer.

tomwilkie: I must be missing something.  For *baseBuffer to implement buffer, we only need to implement Get and Put.  The code wouldn't compile if this wasn't the case.

paulbellamy: Nope, I was missing Get and Put being part of Buffer, just saw the reader/writer. D'oh!

app/router.go
```diff
@@ -76,6 +77,11 @@ func Router(c collector) *mux.Router {
 	return router
 }
 
+func badRequest(r *http.Request, w http.ResponseWriter, err error) {
+	http.Error(w, err.Error(), http.StatusBadRequest)
+	log.Printf("Error procressing request for %s: %v", r.URL.Path, err)
+}
```
peterbourgon: [Rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)):

- If you need it once, write it.
- If you need it twice, copy it.
- If you need it three times, factor it out.

You're only at step 2, ergo this function shouldn't exist yet :)

xfer/publisher.go
```diff
@@ -59,9 +59,7 @@ func (p HTTPPublisher) String() string {
 }
 
 // Publish publishes the report to the URL.
-func (p HTTPPublisher) Publish(buf *Buffer) error {
-	defer buf.Put()
-
+func (p HTTPPublisher) Publish(buf Buffer) error {
```
peterbourgon: If you're just using the io.Reader parts of the buffer, here, maybe you can have it take an io.Reader?

xfer/publisher.go
```diff
@@ -106,7 +104,7 @@ type BackgroundPublisher struct {
 func NewBackgroundPublisher(p Publisher) *BackgroundPublisher {
 	result := &BackgroundPublisher{
 		publisher: p,
-		reports:   make(chan *Buffer),
+		reports:   make(chan Buffer),
```
peterbourgon: This isn't reports, anymore, it's `readers: make(chan io.Reader)`

peterbourgon:
@tomwilkie in #470 I thought I understood what you were doing with the buffers but after reading this PR a couple times it's clear that I don't. Specifically, I don't understand why multiple report publishers are sharing the same buffer instance, nor the need for the manual reference counting. Can you walk me through the "why" part of this change? Happy to Skype if that would be easier.

peterbourgon:
Or @paulbellamy if you've got a handle on it happy to take a lesson from you...

tomwilkie:
The whole point of the original change was to only serialise and compress the report once per publish cycle, instead of once-per-upstream-app-per-publish cycle.  This accounted for about 1/4 of the probes CPU usage.

We could have created a fresh buffer to serialise into, and then shared that buffer (via NewByteBuffers) with the publisher, but the next biggest hit for the probe was GC, and these buffers are large.  So I introduce this mechanism for publishers to all read form the same buffer, and when they are done with it, return it to the pool, for reuse in the next cycle.  The reference counting was needed as the publishers all publish in parallel, and can all take different lengths of time to publish (especially if the app is not reachable - it take a ~30s timeout to return the buffer).

tomwilkie:
Closing in favour of #492

PR #485, opened 2015-09-14
Dedupe probe POSTs

peterbourgon:
Addresses #463


peterbourgon:
@tomwilkie PTAL

probe/resolver.go
```diff
@@ -92,13 +92,15 @@ func (r staticResolver) resolveHosts() {
...
+		endpoints := make([]string, 0, len(addrs))
 		for _, addr := range addrs {
 			// For now, ignore IPv6
 			if addr.To4() == nil {
 				continue
 			}
-			r.add(net.JoinHostPort(addr.String(), peer.port))
+			endpoints = append(endpoints, net.JoinHostPort(addr.String(), peer.port))
 		}
+		r.set(peer.hostname, endpoints)
```
tomwilkie: Hmm a temporary glitch in DNS will stop us pushing to a app for ~1 min.  I guess you're doing this to stop us trying to push to apps which have gone away?  I think we probably want to decouple some sort of app-timeout from the resolving (perhaps this should instead be coupled to the backoff?  n strikes at backoff then it gets removed?)

tomwilkie:
If I read this right, you're doing the get for /api on a single goroutine (the resolver goroutine), and you're doing them every 1 min.  These gets will block for ~30s for endpoints that don't exist; I'm worried that when scope is 'joining' a large cluster with multiple unreachable endpoints, it will take a long time (many minutes) for the data to show up in all the apps.

I suggest we have a cache of ip->id (which expires its entire after some longer timeout, say 5mins) and we do these lookups in parallel, on background goroutines.

probe/resolver.go
peterbourgon: Yes, I'm doing this so that app processes that go away are removed. Yes, if DNS stops returning an IP for a hostname, we will stop pushing reports there until it comes back. But of course we should treat DNS as a source of authority.

tomwilkie: Actually looks like set never actually removes existing ones, do above comment is incorrect.

Its not clear to me the behaviour of the code when a app restarts (and its id changes).  Could you elaborate please?

peterbourgon:
Will add a pool of goroutines for lookups, but I'm not sure adding a cache makes sense, I'm not sure it will make a significant difference to the behavior and it's an extra layer of complexity. I'll sketch it out and see how it feels.

probe/resolver.go
peterbourgon: > Its not clear to me the behaviour of the code when a app restarts (and its id changes). 

Ah, good catch. Will fix.

peterbourgon:
@tomwilkie OK, that's a bit of a refactor, but it's nice and straightforward (and tested) :)

probe/resolver.go
```diff
@@ -15,93 +15,115 @@ var (
...
+	for t, endpoints := range resolveMany(r.sema, r.targets) {
+		r.set(t.String(), endpoints)
+	}
+}
+
+func resolveMany(s semaphore, targets []target) map[target][]string {
+	result := map[target][]string{}
+	for _, t := range targets {
+		c := make(chan []string)
+		go func(t target) { s.p(); defer s.v(); c <- resolveOne(t) }(t)
```
tomwilkie: why are you doing the dns lookups on a separate goroutine?  I think its pretty safe to do these on this routine; go will timeout requests after 5 seconds if they don't get a reply.

xfer/http_publisher.go
```diff
@@ -0,0 +1,80 @@
...
+type HTTPPublisher struct {
+	url     string
+	token   string
+	probeID string
+}
+
+// NewHTTPPublisher returns an HTTPPublisher ready for use.
+func NewHTTPPublisher(target, token, probeID string) (string, *HTTPPublisher, error) {
+	targetAPI := sanitize.URL("http://", 0, "/api")(target)
+	resp, err := http.Get(targetAPI)
```
tomwilkie: This, OTOH, will block for upto 30 secs, for each host, on the resolver goroutine.

tomwilkie:
@peterbourgon do you plan on doing the /api lookups on separate goroutines?  You seem to have moved the dns lookups on separate goroutines (but they max out at 5s), but the /api lookups could take 30s and are all on one goroutine?

peterbourgon:
(sigh) When you said "lookup" I interpreted it as "DNS lookup" and not "GET /api". I'll change it.

tomwilkie:
The http calls could block for 30s if, say, the host is unreachable.  Which is why we did all the weave integration, so you could deploy scope on a cluster which spans two datacenters - where ips from one aren't reachable from the other.

peterbourgon:
@tomwilkie 8602132 should do it. I hope.

xfer/publisher.go
```diff
@@ -1,215 +1,10 @@
...
-	maxBackoff     = 60 * time.Second
-)
-
-// Publisher is something which can send a buffered set of data somewhere,
-// probably to a collector.
+// Publisher is something which can send a stream of data somewhere, probably
+// to a remote collector.
 type Publisher interface {
-	Publish(*bytes.Buffer) error
+	Publish(io.Reader) error
```
tomwilkie: Why?  This necessitates an extra read, which is unnecessary.

tomwilkie:
Have tested and seems to work well.

xfer/publisher.go
peterbourgon: The primary reason is that a concrete bytes.Buffer is an overspecification for the interface. Fundamentally, all the Publisher is doing is reading bytes from the thing. Thus, io.Reader.

The "cost" is that we have to ioutil.ReadAll in the MultiPublisher, which invokes an extra copy() builtin compared to just bytes.Buffer.Bytes. But I judge it's a good tradeoff for a more-tightly-defined API.

peterbourgon:
Agh, should have squashed. Sorry. Not enough coffee yet.

PR #487, opened 2015-09-14
New sidebar in the bottom left

### LINKED ISSUE #207: Add a status bar to show topology statistics
When we start showing containers without connections, we'll want an option to not do that.

Also means we could start probing for processes without connections, and optionally show/hide them.

davkal:
* status bar (fixes #207)
* moved error reporting down to statusbar, error is shown in status field
* moved topology options to sidebar
* render topology option like snackbar
* option items stack on top
* upgrade material-ui to 0.11

### DIFF
--- a/app/router.go
+++ b/app/router.go
@@ -128,6 +128,8 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 	respondWith(w, http.StatusOK, APIDetails{Version: version})
 }
 
+// Topology option labels should tell the current state. The first item must
+// be the verb to get to that state
 var topologyRegistry = map[string]topologyView{
 	"applications": {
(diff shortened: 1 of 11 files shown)


davkal:
screen shot 2015-09-14 at 20 47 48

davkal:
Gotta change the wording now to reflect current state, not the action. Eg. "Showing system containers".

davkal:
screen shot 2015-09-14 at 20 55 02

tomwilkie:
Nice.  I'd like to include in the status bar how many nodes were filtered out (to make it clear when you see a blank page that scope is working).

davkal:
@tomwilkie yeah, could you add this to the stats object? Then I can render it.

PR #488, opened 2015-09-15
Compile flags compatible with 1.4

tomwilkie:
In #440 we changed the build flags that inject the version number to work with go1.5; but the new format doesn't work with go1.4.

The old format still works in 1.5, just spits a warning.  Lets stick with the old format until everyone has moved to go1.5.


tomwilkie:
Symptom is version number doesn't appear in UI.

peterbourgon:
There's no reason not to tell everyone to upgrade to 1.5. There's no value in backwards compatibility.

tomwilkie:
Yeah I was thinking after 0.7 is out we should all upgrade to go1.5.  Only reason not to do it now is it will require some CI work.

PR #490, opened 2015-09-15
Hopefully simplified buffering Publisher

peterbourgon:
**Note** this is a PR into the **buffers** branch!

Two concepts:

- Publisher: takes Report, serializes to ReadCloser
- Sender: takes ReadCloser, emits it to a remote

Three implementations:

- SendingPublisher uses an encoder (gob + gzip) to serialize into a bytes.Buffer (managed by a sync.Pool). It synchronously hands that buffer to the underlying Sender, and on completion, returns it to the pool.

- MultiSender takes a ReadCloser, demuxes (copies) it into individual bytes.Buffers (managed by a sync.Pool) and passes ownership of those buffers to each underlying (HTTP) Publisher. It then closes the original ReadCloser (which is a no-op) and returns.

- HTTPSender takes a ReadCloser and puts it in a queue of size 1. (Dropped ReadClosers are closed and thereby returned to the pool of the MultiPublisher.) It continuously tries to POST the most recent ReadCloser to a remote URL.

This removes reference-counting and a lot of scaffolding. It also adds some tests.


peterbourgon:
@tomwilkie PTAL. This gets rid of reference counting and saves 80% of the optimizations of the other one. 80% being a very scientific guess ;)

xfer/sender.go
```diff
@@ -0,0 +1,248 @@
...
+	s := &HTTPSender{
+		target:    sanitize.URL("http://", AppPort, "/api/report")(target),
+		authToken: authToken,
+		probeID:   probeID,
+		recv:      make(chan io.ReadCloser),
+		send:      make(chan io.ReadCloser),
+		quit:      make(chan struct{}),
+	}
+	go s.recvLoop()
+	go s.sendLoop()
```
tomwilkie: Its not clear to me why we need these two go routines here; I think this can be achieved with a non blocking put to s.recv, which closes the rc if it fails.

tomwilkie:
I can't say I keen on the copying aspect of this, and I think the sender code could be further simplified.  Also, I get a ton of "dropping reports" with this version of the code, which doesn't happen with the others - worth investigating.  And we will need backoff code - the minute you run this with unreachable apps (say one on the cloud, one on your laptop, like I run) it massively spams the logs.

I thought I'd take a step back and analyse the options we have in terms of lines of code, cpu usage and GC impact.  I used pprof, but there is probably a wide margin of confidence on these numbers, as I'm running on a VM on a hot laptop.  Setup is two VMs, running over weave, with the example app running.

1. buffers-peter: 
  - ~406 lines (w/o tests) + 26 for sanitize refactor
  - 22% CPU usage
  - of which 11% GC 
2. buffers:
  - 429 lines (w/o tests)
  - 24% CPU
  - 11% GC
3. buffers-simple: (see #492)
  - 336 lines (w/o tests)
  - 24% CPU 
  - 11% GC

Given all these figures are easily the same within the margins, I'd say we go with the simplest approach - no sync.Pool, no reference counting, no copying - I put the code in #492.  I'd also like to pull out the sanitize refactoring from this PR.  What do you think?

PR #491, opened 2015-09-16
Show pseudo nodes & connections found through procspy

### LINKED ISSUE #489: Missing edges from the Internet node
Related to #446

After the fixes in #465, I get edges between the `httpserver`s and `dataproducer`s in the [ECS demo](http://weave.works/guides/service-discovery-with-weave-aws-ecs.html) but I am missing edges between the Internet node and the `httpserver`s

screen shot 2015-09-15 at 11 47 08 am

Here's the report: https://gist.github.com/2opremio/0a80f1a86138c6828468

tomwilkie:
Previously I'd filtered out nodes (and connections) found through conntrack, as we didn't want all the cruft to appear in the process view.  But this had also filtered out a subset of those nodes which were also found by procspy, and merged with a conntracked node.

This change stops us filtering those nodes, restoring the pseudo node behaviour to that of 0.6.

Fixes #489

### DIFF
--- a/probe/endpoint/reporter.go
+++ b/probe/endpoint/reporter.go
@@ -17,6 +17,7 @@ const (
 	Addr        = "addr" // typically IPv4
 	Port        = "port"
 	Conntracked = "conntracked"
+	Procspied   = "procspied"
 )
 
 // Reporter generates Reports containing the Endpoint topology.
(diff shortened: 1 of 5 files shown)


render/mapping.go
```diff
@@ -48,10 +48,14 @@ func MapEndpointIdentity(m RenderableNode, local report.Networks) RenderableNode
...
-	// if they don't also have a pid; see #447
+	// if they don't also have a pid; see #447.  Note if they
+	// we're introduced by proc spy then they are guaranteed to
+	// have a pid on the other end of the adjacency, so we include them
+	// no matter what.
 	pid, pidOK := m.Metadata[process.PID]
 	_, conntracked := m.Metadata[endpoint.Conntracked]
-	if !pidOK && conntracked {
+	_, procspied := m.Metadata[endpoint.Procspied]
+	if !procspied && !pidOK && conntracked {
```
tomwilkie: Note this could now probably just be !procspied, as no conntracked nodes have a pid, they were only getting pids through merges with procspied nodes.  Should I simplify?

render/mapping.go
```diff
@@ -64,8 +68,8 @@ func MapEndpointIdentity(m RenderableNode, local report.Networks) RenderableNode
 		}
 
 		// We are a 'client' pseudo node if the port is in the ephemeral port range.
-		// Linux uses 32768 to 61000.
-		if p, err := strconv.Atoi(port); err == nil && len(m.Adjacency) > 0 && p >= 32768 && p < 61000 {
+		// Linux uses 32768 to 61000, IANA suggests 49152 to 65535.
+		if p, err := strconv.Atoi(port); err == nil && len(m.Adjacency) > 0 && p >= 32768 && p < 65535 {
```
tomwilkie: Macs use a wider ephemeral port ranges.

render/mapping.go
peterbourgon: That conntracked nodes don't have PIDs is an implementation detail, not an invariant. The logic here should reflect the invariants, not the implementation.

report/topology.go
```diff
@@ -102,7 +102,7 @@ func MakeNodeWith(m map[string]string) Node {
 // WithMetadata returns a fresh copy of n, with Metadata set to m
```
peterbourgon: Please update the comment.

probe/endpoint/reporter.go
```diff
@@ -95,26 +96,31 @@ func (r *Reporter) Report() (report.Report, error) {
...
-				report.HostNodeID: hostNodeID,
-			})
+	{
+		conns, err := procspy.Connections(r.includeProcesses)
+		if err != nil {
+			return rpt, err
+		}
+		commonNodeInfo := report.MakeNode().WithMetadata(report.Metadata{
+			Procspied: "true",
+		})
```
peterbourgon: The commonNode is passed as the extraRemoteNode to addConnection, is it true that the remote node is considered to be procspied?

tomwilkie: This is the crux of the change; we only want pseudo nodes to appear for remote procspied nodes.  So yes.

peterbourgon: OK, the meaning of "procspied" could be "found via spying on /proc" i.e. local only, or "found via the proc spy mechanism" which could conceivably include remote nodes. So, cool.

peterbourgon:
LGTM.

probe/endpoint/reporter.go
tomwilkie: Remote addresses appear in /proc/net/tcp, so those two things are the same in my view.

PR #492, opened 2015-09-16
Even simpler buffer management for report publishing.

### LINKED ISSUE #471: Manual scope setup (ie without weave) not working

tomwilkie:
Turns out reusing the buffers has no impact on CPU usage; I guess this is because the buffers aren't as big as I thought (1-2k) and are only used at a rate of one every 3 seconds (publish interval)

See #484 and #490 for more info.

Fixes #471


peterbourgon:
Fine by me.

PR #493, opened 2015-09-16
In containers view, show short lived connections to/from the internet.

tomwilkie:
Problem: short-lived, incoming connections from the internet to a container don't show up in the UI.  This is because the local end of the connection is to the hosts IP (not a container), and we drop nodes without a container on them.

Outgoing short lived connections to the internet do work.

We should take into account the container port mappings when translating this node to a container ID.  This can be done unambiguously.

### DIFF
--- a/integration/300_internet_edge_test.sh
+++ b/integration/300_internet_edge_test.sh
@@ -1,0 +1,26 @@
+#! /bin/bash
+
+. ./config.sh
+
+start_suite "Test short lived connections from the Internet"
+
+weave_on $HOST1 launch
+scope_on $HOST1 launch
+docker_on $HOST1 run -d -p 80:80 --name nginx nginx
+
+do_connections() {
+	while true; do
+		curl -s http://$HOST1:80/ >/dev/null
+		sleep 1
+	done
+}
+do_connections&
+
+sleep 5 # give the probe a few seconds to build a report and send it to the app
+
+has_container $HOST1 nginx 1
+has_connection $HOST1 "The Internet" nginx
+
+kill %do_connections
+
+scope_end_suite
--- a/integration/config.sh
+++ b/integration/config.sh
@@ -31,12 +31,12 @@ weave_on() {
 	DOCKER_HOST=tcp://$host:$DOCKER_PORT $WEAVE "$@"
 }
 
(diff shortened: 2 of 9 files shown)


tomwilkie:
This doesn't work yet

tomwilkie:
@fons would you mind giving this a test?  I can't test easily from singapore.

@peterbourgon its a bit untidy, so I'd welcome suggestions on how to clean this up.

tomwilkie:
This should finally fix #446

2opremio:
@tomwilkie Ops, I missed this (I wish I could have the fons user though  :) ) 

Unfortunately it still doesn't work. The Internet node is now gone from the non-system containers view:
screen shot 2015-09-23 at 10 38 24

Report: https://gist.github.com/2opremio/2d03e27de9a3bf39bfc2

Here's the view with system containers:
screen shot 2015-09-23 at 10 40 11

tomwilkie:
Okay I'll have to take a look next week.  Copying binaries to ec2 takes too long from here.

tomwilkie:
@fons I've got this working for my test case, could you confirm it works for yours?

2opremio:
@tomwilkie Same result :S

tomwilkie:
@2opremio I'm confident I've got it this time.  Just writing a integration test.  Fancy giving it a try?

2opremio:
@tomwilkie Yes, you did!

screen shot 2015-10-01 at 19 39 37

screen shot 2015-10-01 at 19 39 57

2opremio:
Test failures aside, LGTM

tomwilkie:
@2opremio does your lgtm extend to the code, or just saying it works?

I've added a unit test as well, to ensure this case is covered.  Please take a look.

integration/300_internet_edge_test.sh
```diff
@@ -0,0 +1,26 @@
...
+	done
+}
+do_connections&
+
+sleep 5 # give the probe a few seconds to build a report and send it to the app
+
+has_container $HOST1 nginx 1
+has_connection $HOST1 "The Internet" nginx
+
+kill %do_connections
```
2opremio: I didn't know about the percentage syntax, nice.

PR #497, opened 2015-09-16
Optimize /proc traversal (again)

### LINKED ISSUE #486: Panic in proc reader
I am getting this when running the [ECS demo](http://weave.works/guides/service-discovery-with-weave-aws-ecs.html)

```
panic: interface conversion: interface is string, not proc.filesCacheEntry
goroutine 489 [running]:
github.com/weaveworks/scope/probe/proc.func·005(0x93bf80, 0xc208e3fb20, 0x93bf80, 0xc2094a0790)
        /home/vagrant/src/github.com/weaveworks/scope/probe/proc/reader_linux.go:44 +0x3e
created by github.com/bluele/gcache.(*ARC).Get
        /home/vagrant/src/github.com/bluele/gcache/arc.go:156 +0x5c3
```

inercia:
This is the same code found in #450, but replacing the caching library. This should fix #486.

In addition to the changes in #450, this PR also caches _some_ file contents. In particular, `/proc/<pid>/[comm, cmdline]`, as they do not change during the lifetime of a process (this is not always true, but I don't think we should care about these corner cases)

--------------------------

_From the original #450 description_:

The objective of this PR is to reduce the number of syscalls when collecting info from `/proc`. This PR does that by: 

* reading the required files (`/proc/*/[cmdline, comm, stat, fd/*]`, `/proc/net/tcp*`) once per cycle (`spy.interval`), parsing and caching the processes/connections information
* keeping a cache of open files for `/proc/*/[cmdline, comm, stat]` so we do not `open()`/`close()` these files so often.
* removing unnecessary `stat()`s when reading `/proc` directories

Changes:

* the required code for reading/parsing `/proc` has been proved from `procspy` to `/probe/proc`
* the `Walker` has been replaced by a `proc.Reader` that provides both `Processes()` and `Connections()`.  Users can walk the list of processes/connections in the same way by providing a function to these functions. A `CachingReader` can be used for caching and getting this info, updated when `Update()`. The code for reading and parsing connection and processes has been simplified.
* a `CachingReader` is initialized in `/probe/main.go` and used in both `endpoint.Reporter` and `process.Reporter` for reporting about processed and connections, and updated periodically by calling `Update()` every `spy.interval` seconds.
* unit tests have been improved by mocking things like file system operations
* things like `Proc` and `Process` have been unified, and many things from `/probe/process` have been moved to `/probe/proc`

This is part of the solution to #284


inercia:
@tomwilkie @peterbourgon please take a look at this when you have some time. The only change since #450 is the [cache.go](https://github.com/weaveworks/scope/pull/497/files#diff-80c130d611fddac41d844424f83144e5R1) file.

probe/proc/cache.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+type filesCacheEntry struct {
+	sync.RWMutex
+	file     File
+	contents *bytes.Buffer
+	creation time.Time
+}
+
+func (fce *filesCacheEntry) Pinned() bool { return fce.contents != nil }
```
peterbourgon: Need a read lock here

probe/proc/cache.go
```diff
@@ -0,0 +1,123 @@
...
+
+func newFilesCache(proc Dir) *filesCache {
+	evictionFunc := func(key, value interface{}) {
+		entry := value.(filesCacheEntry)
+		if entry.file != nil {
+			entry.file.Close()
+		}
+	}
+
+	cache, _ := lru.NewWithEvict(filesCacheLen, evictionFunc)
```
peterbourgon: Don't ignore the error...

probe/proc/cache.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	entry := e.(filesCacheEntry)
+	return entry.ReadInto(buf)
+}
+
+// Close closes all the handles in the cache
+func (fc *filesCache) Close() error {
+	fc.Cache.Purge()
+	fc.quit <- struct{}{}
```
peterbourgon:
```go
close(fc.quit)
```

probe/proc/cache.go
```diff
@@ -0,0 +1,123 @@
...
+// Go through all the entries in the cache and force the removal of
+// the ones that were created too long ago
+func (fc *filesCache) loop() {
+	purge := time.Tick(filesCacheExpiration)
+
+	for {
+		select {
+		case <-purge:
+			now := time.Now()
+			timeLimit := now.Add(-filesCacheExpiration)
```
peterbourgon:
```go
timeLimit := time.Now().Add(-filesCacheExpiration)
```

probe/proc/cache.go
inercia: To propagate this error was a mess, and `NewWithEvict` only raises this error when `len<0` (and we use a constant>0), so I don't know it's worth all the trouble...

probe/proc/cache.go
```diff
@@ -0,0 +1,123 @@
...
+			for _, key := range fc.Cache.Keys() {
+				e, _ := fc.Cache.Peek(key)
+				entry := e.(filesCacheEntry)
```
peterbourgon: Don't ignore the error...

probe/proc/cache.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil
+	}
+	return fce.file.ReadInto(buf)
+}
+
+// A cache for files handles
+type filesCache struct {
+	*lru.Cache
+	proc Dir
+	quit chan (struct{})
```
peterbourgon: Don't need parens here

probe/proc/cache.go
inercia: Hmmm, I'm not sure... cache entries are created either with a valid `contents` or with no contents at all, and that does not change during the lifetime of the entry. My only concern is whether this access could be interleaved with the cache entry removal...

probe/proc/cache.go
```diff
@@ -0,0 +1,123 @@
...
+		buf.Write(fce.contents.Bytes())
+		return nil
+	}
+	return fce.file.ReadInto(buf)
+}
+
+// A cache for files handles
+type filesCache struct {
+	*lru.Cache
+	proc Dir
```
peterbourgon: filesCache requires that all implementations of Dir are guaranteed to be safe for concurrent access by multiple goroutines. It would be good to put that requirement in the documentation for type Dir interface.

probe/proc/cache.go
peterbourgon: OK, don't need to propagate, but an explicit panic would be better.

probe/proc/cache.go
inercia: Right.

probe/proc/cache.go
peterbourgon: OK, I see.

peterbourgon:
Thanks for the change. Can we have a unit test which fails using bluele/gcache and passes using hashicorp/golang-lru? Otherwise, it's hard to have confidence this is a fix — and hard to prevent regressions.

Also, let's avoid introducing the tech debt of removing bluele/gcache from other packages, by proactively eliminating the dependency throughout the repo, and making the required changes in this PR.

probe/proc/cache.go
inercia: Fixed.

probe/proc/cache.go
inercia: Fixed.

probe/proc/cache.go
inercia: Done.

probe/proc/cache.go
inercia: Ok, I'll use a `panic()`

probe/proc/cache.go
inercia: I think we don't really need that. `Dir` is only used here for `Open()`ing files, and calling `Open()` from several goroutines for the same file is probably safe. So the thing that really must be goroutines-safe is the `File` returned by `Dir.Open()`, and the cache protects accesses to that `FIle` with the `RWLock` entries hold...

tomwilkie:
@inercia please give me a chance to test & review this before submitting.  I should be able to look at it first think next week.

tomwilkie:
@inercia I'm happy to remove the use of gcache in ids.go as part of this PR, if it helps.

inercia:
@peterbourgon The problem with adding a unit test is that I did never have a problem with `bluele/gcache`, and I couldn't reproduce the crash described in #486 either. Looking at the code in ` bluele/gcache` I discovered that it triggered `go evictionFunction()` and it could be problematic when evicting many cache items (_hashicorp_ library does not do that). This would be difficult to test with unit tests anyway... So I think the best way to test this would be to reproduce the scenario where we originally found the crash...

@tomwilkie I will wait until a LGTM from you.

inercia:
@tomwilkie @peterbourgon would you like to replace gcache in this PR or in a different one?

peterbourgon:
I would much rather replace gcache in this PR than another one, to avoid creating technical debt.

inercia:
@peterbourgon The last two commits remove all the references to `gcache` (there is a remaining reference in `experimental`, but I guess that s not important, right?)

probe/endpoint/resolver.go
```diff
@@ -44,41 +63,64 @@ func NewReverseResolver() *ReverseResolver {
...
 		}
+		return "", errNotFound
+	}
+	if _, ok := val.(struct{}); ok {
+		return "", errNotFound
 	}
-	return "", errNotFound
+	entry, ok := val.(*resolverCacheEntry)
+	if !ok {
+		panic(fmt.Sprintf("unknown entry type in cache: %+v", val))
```
tomwilkie: This is a programmer error if this ever happens; just let the val.(*resolverCacheEntry) panic above (ie remove the ok)

probe/endpoint/resolver.go
```diff
@@ -44,41 +63,64 @@ func NewReverseResolver() *ReverseResolver {
...
+			<-r.Throttle // rate limit our DNS resolutions
+			names, err := r.Resolver(request)
+			if err == nil && len(names) > 0 {
+				name := strings.TrimRight(names[0], ".")
+				r.cache.Add(request, newResolverCacheEntry(name))
+			} else {
+				r.cache.Add(request, struct{}{})
+			}
+
+		case <-purge:
```
tomwilkie: Shame this isn't included in the library.  We'll probably want this feature elsewhere, so I'd consider how we might break this functionality out.

probe/proc/cache.go
```diff
@@ -0,0 +1,132 @@
...
+			return err
+		}
+		if pin {
+			entry.contents = &b // save the file contents
+			f.Close()           // and close the file
+		} else {
+			entry.file = f
+		}
+		buf.Write(b.Bytes())
+		fc.Cache.Add(filename, entry)
```
tomwilkie: What happens to the other filesCacheEntry if this function is called concurrently?  You'll end up calling add twice, which will probably loose all the references to the fce, leaking open files.

tomwilkie: Oh I see there is an eviction function.  My bad.

tomwilkie: Actually looking at the code (https://github.com/hashicorp/golang-lru/blob/master/lru.go#L61) this doesn't evict the old value if you overwrite it, so there is a leak.

probe/proc/cache.go
```diff
@@ -0,0 +1,132 @@
...
+	*lru.Cache
+	proc Dir
+	quit chan struct{}
+}
+
+func newFilesCache(proc Dir) *filesCache {
+	evictionFunc := func(key, value interface{}) {
+		entry, ok := value.(filesCacheEntry)
+		if ok && entry.file != nil {
+			entry.file.Close()
```
tomwilkie: This close can be concurrently called with ReadInto above, and will race.

probe/proc/reader_linux.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return nil
+}
+
+// Connections walks through all the connections in the "/proc"
+func (r *reader) Connections(withProcs bool, f func(Connection)) error {
+	// create a map of inode->Process
+	procs := make(map[uint64]Process)
+	if withProcs {
+		r.Processes(func(p Process) {
```
tomwilkie: The whole point of this change was to only walk the proc filesystem once; this does an extra walk here.

tomwilkie:
There are too many races / leaks in this PR.

Going forward I'd drop all the caching, and focus on making a single walk of the proc filesystem, gathering both processes and connections at the same time, and porting the existing procspy code onto that.

We can potentially add caching on top of that.

probe/proc/cache.go
inercia: I think there is no leak, as we never overwrite values in this cache. You can see in the code that the only `fc.Cache.Add()` is invoked when `fc.Cache.Get()` returns `!ok`. Maybe it should be documented...

probe/endpoint/resolver.go
inercia: You mean to let the conversion panic, with no `ok` check? I think @peterbourgon had the opposite suggestion...

probe/proc/cache.go
inercia: I forgot to use the `fileCacheEntries` mutex here... fixed.

probe/proc/reader_linux.go
inercia: I forgot to move this to the CachingReader. I'll fix it.

inercia:
@tomwilkie the usage pattern for this code would be to call the `Update()` once a second, so there is no concurrency problems in practice. I've been reluctant to add locks in the cache entries because YAGNI and, in fact, I think we could perfectly remove all the mutexes in this code and maybe replace them with some big locks in `Connections()`/`Processes()`. Anyway, and as you have noticed, there is a missing lock when evicting entries in the cache, and I think that should be enough for making the cache goroutines-safe...

The performance gain of caching file handles and contents reduces the number of syscalls dramatically, and that is a main part of objective of this PR. I will fix the second `/proc` traversal and then let me know what you think...

probe/proc/cache.go
tomwilkie: But not under any lock, hence 'if this function is called concurrently'

probe/endpoint/resolver.go
tomwilkie: https://golang.org/ref/spec#Type_assertions

This is a programmer error if it fails, so I would just let the runtime panic (instead of checking ok... and panicing anyway).

inercia:
I've just realized that moving the `Mutex` from the cache entries to the `Cache`, with the expected usage pattern, reduces the numer of `Lock`s/`Unlock`s... I'll change it.

probe/endpoint/resolver.go
inercia: I think this is out of the scope of this PR: we are not trying to create a new caching library... :confused:

probe/endpoint/resolver.go
inercia: Ok, fixed.

probe/proc/cache.go
inercia: It is protected at the `Cache` level now.

probe/proc/cache.go
inercia: Moved all the locks at the `Cache` level.

inercia:
Please take another look @tomwilkie

probe/proc/cache.go
```diff
@@ -0,0 +1,139 @@
...
+	filesCacheLen        = 1024
+	filesCacheExpiration = 60 * time.Second
+)
+
+var (
+	errInvalidCacheEntry = errors.New("invalid cache entry")
+)
+
+type filesCacheEntry struct {
+	file     File
```
tomwilkie: Whats the justification for keeping the fds open?  Do you have a trace that shows its saves a lot of CPU?  Could we just cache file contents for certain files without the fds?

I'm not convinced its safe to keep files open in proc for long periods.  There little mention of it on the web (http://stackoverflow.com/questions/1623731/open-close-strategy-for-proc-pseudo-file), but unless we can find something that says its safe (ie the kernel source) I'd prefer not to.

probe/endpoint/resolver.go
tomwilkie: I'm not asking for a new caching library.  I'm asking for a wrapper around the caching library we use so that we can share the timeout logic.

probe/proc/cache.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+		buf.Write(b.Bytes())
+		fc.Cache.Add(filename, entry)
+		return nil
+	}
+
+	entry, ok := e.(filesCacheEntry)
+	if !ok {
+		return errInvalidCacheEntry
+	}
```
tomwilkie: This is a programmer error (if the cast fails).  Is should panic.

probe/proc/reader_linux.go
```diff
@@ -0,0 +1,136 @@
...
+
+	return nil
+}
+
+// Connections walks through all the connections in the "/proc"
+// Note: we never perform the process<->connection matching in Linux:
+//       use the CachingReader instead
+func (r *reader) Connections(f func(Connection)) error {
+	buf := bytes.Buffer{}
+	for _, tcpFile := range tcpFiles {
```
tomwilkie: Is this no longer reading the /proc/<pid>/net/tcp files?  Or have I misunderstood?

tomwilkie:
Whats the motivation for a separate proc and process directory in probe?

probe/proc/cache.go
inercia: The justification is to save `open()/close()` cycles and the subsequent syscalls. What I do is to open the file and then rewind (`seek(0)`) every time we want to re-read from those files (as suggested in the [second answer](http://stackoverflow.com/questions/1623731/open-close-strategy-for-proc-pseudo-file/1623931#1623931) in your link, or [here](http://stackoverflow.com/questions/19949283/seek0-on-linux-proc-sys-pseudo-files) and [here](http://stackoverflow.com/questions/4788462/is-it-safe-to-use-lseek-when-reading-from-proc-fs-files-second-time)) (and it is verified by the mere fact that we are really getting updated values and no errors with this mechanism). Files are closed after 60 seconds no matter what (so we can get rid of entries that are not being used).

The only drawback could be the exhaustion of the number of open files, but I think that `filesCacheLen` has a very conservative value (considering that most of the entries will have cached file content and not files handles, so in practice we will keep 1024/3 open files).

probe/proc/cache.go
inercia: You should unify criteria with @peterbourgon ... :smiley:

inercia: I'll change it.

inercia:
The `proc` is the library for reading from the `/proc`, while the `process` is the processes reporter. The `proc` library is used by several reporters. Maybe the `proc` could be moved to some other location... I'm open to suggestions.

probe/proc/reader_linux.go
inercia: Maybe I'm wrong, but I think we have never used `/proc/[pid]/net/tcp` but `/proc/net/tcp`... (see [these](https://github.com/weaveworks/procspy/blob/master/spy_linux.go#L48) lines)

probe/proc/cache.go
tomwilkie: Do you have a trace that shows its saves a lot of CPU?

tomwilkie:
Yes please merge these two directories, keeping the original directory.

probe/proc/reader_linux.go
tomwilkie: Your approach doesn't work for containerised apps (/proc/net/tcp is just the host net namespace).  There is a ton of log for dealing with other network namespaces, which you've completely omitted:

https://github.com/weaveworks/procspy/blob/master/proc.go#L71

inercia: Oh, my fault. Easy to fix though...

inercia: @tomwilkie do we still need to read the `/proc/net/tcp*` files?

probe/proc/cache.go
inercia: Done

inercia:
I've merged the `proc` and `process` directories. I've also merged the `Reader` and `CachingReader` structs (as it was being a pain to keep the same interface for both). I've also added the missing `tcp`/`tcp6` reads for namespaces other than the global one.

I've also added an experimental feature where the Linux `/proc` reader will skip uninteresting processes, based on the process name found in `comm` and a list of well-known system processes/tasks. This reduces the number of syscalls dramatically. This would be the new perf graph:

profile

@peterbourgon or @tomwilkie , would you mind taking a look at this?

probe/endpoint/resolver.go
inercia: I would leave that for a different PR: adding a _wrapper-around-the-cache_ would make this PR too long (it is already getting too long!)...

tomwilkie:
Alvaro - I've had a think about this PR, and a chat with Peter, and I think its best if we take a different approach.  This is a big change to the very core of Scope, and the risks associated with getting it wrong are large.  I want you to avoid rewriting large swathes of procspy as the current code is mature, and its behaviour is sometimes subtle.

So instead of proceding this PR, I suggest we take a much more incremental approach:
- 'vendor' in procspy as is, into probe/proc (do this in 1 PR, without any other changes)
- remove any usused / dead code, in another PR
- make minimal adaptations to procspy to avoid the double walk of /proc, in another PR
- evaluate adding a cache on top of this, again, in a fresh PR

From there we can evaluate other optimisations one PR at a time.  Please avoid needless renaming to make this process go faster.

Thank you for working on this, we do appreciate your contribution here.

inercia:
No problem, I'll change the code in several PRs. I'll close this PR then...

PR #503, opened 2015-09-21
Shutdown behavior

### LINKED ISSUE #424: Ensure we `Stop()` everything when the main `probe` process exits

peterbourgon:
Resolves #424.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ docker/scope-app
 docker/scope-probe
 docker/docker*
 docker/weave
+docker/runsvinit
 experimental/bridge/bridge
 experimental/demoprobe/demoprobe
 experimental/fixprobe/fixprobe
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ SCOPE_VERSION=$(shell git rev-parse --short HEAD)
 DOCKER_VERSION=1.3.1
 DOCKER_DISTRIB=docker/docker-$(DOCKER_VERSION).tgz
 DOCKER_DISTRIB_URL=https://get.docker.com/builds/Linux/x86_64/docker-$(DOCKER_VERSION).tgz
+RUNSVINIT=docker/runsvinit
 RM=--rm
 
 all: $(SCOPE_EXPORT)
@@ -28,12 +29,15 @@ docker/weave:
 	curl -L git.io/weave -o docker/weave
 	chmod u+x docker/weave
 
-$(SCOPE_EXPORT): $(APP_EXE) $(PROBE_EXE) $(DOCKER_DISTRIB) docker/weave docker/entrypoint.sh docker/Dockerfile docker/run-app docker/run-probe
+$(SCOPE_EXPORT): $(APP_EXE) $(PROBE_EXE) $(DOCKER_DISTRIB) docker/weave $(RUNSVINIT) docker/Dockerfile docker/run-app docker/run-probe
 	@if [ -z '$(DOCKER_SQUASH)' ] ; then echo "Please install docker-squash by running 'make deps' (and make sure GOPATH/bin is in your PATH)." && exit 1 ; fi
 	cp $(APP_EXE) $(PROBE_EXE) docker/
 	cp $(DOCKER_DISTRIB) docker/docker.tgz
 	$(SUDO) docker build -t $(SCOPE_IMAGE) docker/
 	$(SUDO) docker save $(SCOPE_IMAGE):latest | sudo $(DOCKER_SQUASH) -t $(SCOPE_IMAGE) | tee $@ | $(SUDO) docker load
+
+$(RUNSVINIT): vendor/runsvinit/*.go
+	go build -o $@ github.com/weaveworks/scope/vendor/runsvinit
 
 $(APP_EXE): app/*.go render/*.go report/*.go xfer/*.go
 
--- a/app/main.go
+++ b/app/main.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
@@ -28,6 +29,7 @@ func main() {
 	var (
 		window       = flag.Duration("window", 15*time.Second, "window")
 		listen       = flag.String("http.address", ":"+strconv.Itoa(xfer.AppPort), "webserver listen address")
+		logPrefix    = flag.String("log.prefix", "<app>", "prefix for each log line")
 		printVersion = flag.Bool("version", false, "print version number and exit")
 	)
 	flag.Parse()
@@ -37,20 +39,24 @@ func main() {
 		return
 	}
 
+	if !strings.HasSuffix(*logPrefix, " ") {
+		*logPrefix += " "
+	}
+	log.SetPrefix(*logPrefix)
+
+	defer log.Print("app exiting")
+
 	rand.Seed(time.Now().UnixNano())
 	uniqueID = strconv.FormatInt(rand.Int63(), 16)
 	log.Printf("app starting, version %s, ID %s", version, uniqueID)
 
 	c := xfer.NewCollector(*window)
 	http.Handle("/", Router(c))
-	irq := interrupt()
 	go func() {
 		log.Printf("listening on %s", *listen)
 		log.Print(http.ListenAndServe(*listen, nil))
-		irq <- syscall.SIGINT
 	}()
-	<-irq
-	log.Printf("shutting down")
+	log.Printf("%s", <-interrupt())
 }
 
 func interrupt() chan os.Signal {
--- a/integration/110_shutdown_test.sh
+++ b/integration/110_shutdown_test.sh
@@ -1,0 +1,16 @@
+#! /bin/bash
+
+. ./config.sh
+
+start_suite "Check scope exits cleanly within 10 seconds"
+
+scope_on $HOST1 launch
+scope_on $HOST1 stop
+
+sleep 10
+
+assert_raises "docker_on $HOST1 logs weavescope 2>&1 | grep 'app exiting'"
+assert_raises "docker_on $HOST1 logs weavescope 2>&1 | grep 'probe exiting'"
+assert_raises "docker_on $HOST1 inspect --format='{{.State.Running}}' weavescope" "false"
+
+end_suite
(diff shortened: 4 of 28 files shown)


probe/main.go
```diff
@@ -55,129 +54,152 @@ func main() {
...
 		}
+		return docker.NewTagger(registry, processCache), docker.NewReporter(registry, hostID), registry
+	}()
+	if dockerTagger != nil {
+		taggers = append(taggers, dockerTagger)
+	}
+	if dockerReporter != nil {
+		reporters = append(reporters, dockerReporter)
+	}
+	if dockerRegistry != nil {
```
tomwilkie: Can you hoist this stuff into the function?  the function has access to the taggers, reporters etc.  Also, why is the function even necessary?  Why not fatalf?

tomwilkie:
Tests I've done:

1. run ./docker/scope-probe and ctrl-c; I see the probe exit cleanly (bye!)
2. run it under docker with the -d removed from docker run in the scope script (so it launches in the foregraound), then ctrl-c; I do not see anything exit cleanly
3. run it under docker normally, send sighup (see http://smarden.org/runit/runsvdir.8.html) to the runsv process; I see it exist cleanly.
4. run it normally, use docker stop; I do not see it exit cleanly

As it looks like the only way to gracefully shutdown runsvdir is with sighup, we need to make ./scope stop send sighup for docker stop:

```diff
diff --git a/scope b/scope
index a973f23..9a724c6 100755
--- a/scope
+++ b/scope
@@ -189,7 +189,7 @@ case "$COMMAND" in
 
     stop)
         [ $# -eq 0 ] || usage
-        if ! docker stop $SCOPE_CONTAINER_NAME >/dev/null 2>&1 ; then
+        if ! docker kill -s HUP $SCOPE_CONTAINER_NAME >/dev/null 2>&1 ; then
             echo "Weave Scope is not running." >&2
         fi
         docker rm -f $SCOPE_CONTAINER_NAME >/dev/null 2>&1 || true
```

app/main.go
```diff
@@ -32,20 +32,19 @@ func main() {
 		return
 	}
 
+	defer log.Print("bye!")
```
tomwilkie: Would be nice to make it clear this is the app exiting: logs are mixed with probe when running under docker.

probe/main.go
```diff
@@ -55,129 +54,152 @@ func main() {
 		return
 	}
 
+	defer log.Print("bye!")
```
tomwilkie: Would be nice to make it clear this is the probe exiting: logs are mixed with app when running under docker.

tomwilkie:
I think its probably time we broke the probe main() function up into smaller pieces; perhaps have a probe struct to contain the reporters, taggers & tickers, make add{Reporter, Probe}, loop() and quit() functions on that?

probe/main.go
peterbourgon: No, Fatalf doesn't unwind the defer stack.

app/main.go
peterbourgon: Check.

probe/main.go
tomwilkie: Seriously?  Huh.

peterbourgon:
> I think its probably time we broke the probe main() function up into smaller pieces; 

I prototyped that first, and this was cleaner.

peterbourgon:
> As it looks like the only way to gracefully shutdown runsvdir is with sighup, we need to make ./scope stop send sighup for docker stop:

I will add this patch to the PR, thanks.

tomwilkie:
Also need to consider how we add a test to ensure we never break this
again; seem like the kinda of thing it would be easy to regress on and not
notice.

tomwilkie:
Test added; Note when I run it locally I don't see the app or probe shutdown cleanly, so I'm expecting the test to fail.

peterbourgon:
```
>>> Running ./110_shutdown_test.sh on [host1-1382-0.us-central1-a.scope-integration-tests]
>>> Test ./110_shutdown_test.sh finished with success after 3.2 secs
```

tomwilkie:
I fixed the test - it fails now (probe doesn't exit cleanly).

peterbourgon:
Update. Good news: `docker stop` and `docker kill` now reliably terminate the container. Bad news: they don't reliably wait for the probe and app processes to return. Sending SIGHUP to runsvdir causes it to send SIGTERM to all supervised processes (good), but then [immediately exit with 111](http://smarden.org/runit/runsvdir.8.html) (bad). Rarely, I will see "probe exiting" or "app exiting" before Docker destroys the container.

[Other Docker+runit solutions](https://github.com/phusion/baseimage-docker) appear to solve this with their own init script, which [explicitly waits for services to exit before returning](https://github.com/phusion/baseimage-docker/blob/14ec533a164cdb495e1c6ab10b82ebe96695a971/image/bin/my_init#L251). I guess that can work, though it feels stupid to wrap a supervisor with a supervisor. Still researching...

peterbourgon:
This definitely works, now:

```
$ ./scope launch
$ ./scope stop
$ docker logs weavescope 2>&1 | grep 'exiting'
<app> 2015/09/24 12:07:01 app exiting
<probe> 2015/09/24 12:07:01 probe exiting
```

@tomwilkie any ideas here?

tomwilkie:
I'll take a look at the test tomorrow to see why its failing.

tomwilkie:
Hmm when I run it, it works:
```
vagrant@ubuntu-14:~/src/github.com/weaveworks/scope/integration$ ./110_shutdown_test.sh 
Check scope exits cleanly within 5 seconds
  * http://10.240.201.151:4040/
host1.us-central1-a.positive-cocoa-90213> a5ee473e10dfae5690c10c4ae7b683227ab02016e66d786540cbc0a84303c8e6
all 3 tests passed in 28.188s.
vagrant@ubuntu-14:~/src/github.com/weaveworks/scope/integration$ ./110_shutdown_test.sh 
Check scope exits cleanly within 5 seconds
  * http://10.240.201.151:4040/
host1.us-central1-a.positive-cocoa-90213> eacbbb01b4e3411a1a429ad215818dab8f7f87e3b9b6cb1a3bbe9f184e391fb5
all 3 tests passed in 28.526s.
vagrant@ubuntu-14:~/src/github.com/weaveworks/scope/integration$ ./110_shutdown_test.sh 
Check scope exits cleanly within 5 seconds
  * http://10.240.201.151:4040/
host1.us-central1-a.positive-cocoa-90213> 5476200975d3539bb08fb527545c528b4d8caac858db613e7bba2db18c0aba4f
all 3 tests passed in 28.332s.
```

tomwilkie:
Test fixed.  I'm blown away by the signal handling in the entrypoint.sh script, and I'm concerned we're leaving ourselves open to bugs like this https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/ (which is why we we're exec'ing runit to begin with).

Lets get this change in (so LGTM from me) and then give it some more thought.  Perhaps a different supervisor?  Perhaps we should merge the app and probe into a single process?  That would remove a bunch of code and problems - and we could do it by effectively leaving them as completely separate, but just run in the same process?

peterbourgon:
>  I'm blown away by the signal handling in the entrypoint.sh script, and I'm concerned we're leaving ourselves open to bugs like this

You're right. We should do it properly. I wouldn't like to switch to another supervisor. I would propose we adopt the phusion my_init script. What do you think?

peterbourgon:
@tomwilkie Is this acceptable to you? See [peterbourgon/runsvinit](https://github.com/peterbourgon/runsvinit).

tomwilkie:
- peterbourgon/runsvinit needs test to show it actually reaps orphaned children.
- I'd prefer it it wasn't included as a binary, but built each time (I'm concern about it bit rotting).
- Also it needs the usual CI integrations, and should probably live under the weaveworks github account, and needs a thorough code review.

Also, needs a rebase so we can see how this has affected coverage.  I've got a feeling its going to have gone down, as probe/main.go is basically untested.

peterbourgon:
> peterbourgon/runsvinit needs test to show it actually reaps orphaned children.

OK, will add.

> I'd prefer it it wasn't included as a binary, but built each time (I'm concern about it bit rotting).

OK, will change.

> Also it needs the usual CI integrations, and should probably live under the weaveworks github account, and needs a thorough code review.

I'll add CI. I'm ambivalent on moving it to Weaveworks; it has nothing to do with our core mission. Once I add the reaping tests, I'll ask you to review it.

peterbourgon:
- I've added the test, [see here](https://github.com/peterbourgon/runsvinit/tree/master/zombietest).
- I've added CI integration, [see here](https://circleci.com/gh/peterbourgon/runsvinit).
- I've rebased this repo.
- I've manually vendored peterbourgon/runsvinit, [see here](https://github.com/weaveworks/scope/tree/shutdown/vendor/runsvinit).

I hope that's enough to earn your LGTM.

Makefile
```diff
@@ -17,6 +17,8 @@ SCOPE_VERSION=$(shell git rev-parse --short HEAD)
 DOCKER_VERSION=1.3.1
 DOCKER_DISTRIB=docker/docker-$(DOCKER_VERSION).tgz
 DOCKER_DISTRIB_URL=https://get.docker.com/builds/Linux/x86_64/docker-$(DOCKER_VERSION).tgz
+RUNSVINIT=docker/runsvinit
+RUNSVINIT_TGZ=docker/runsvinit-linux-amd64.tgz
```
tomwilkie: Is this line redundant now?

.gitignore
```diff
@@ -42,6 +42,8 @@ docker/scope-app
 docker/scope-probe
 docker/docker*
 docker/weave
+docker/runsvinit-linux-amd64.tgz
```
tomwilkie: I guess this is redundant too.

integration/110_shutdown_test.sh
```diff
@@ -0,0 +1,16 @@
+#! /bin/bash
+
+. ./config.sh
+
+start_suite "Check scope exits cleanly within 5 seconds"
```
tomwilkie: 10 seconds.

tomwilkie:
Some minor comments.  Should also squash this down to 1 changeset.  And I think you should get someone like @dpw to review the supervisor.

Other than that, LGTM.

PR #505, opened 2015-09-21
Show message if topology is empty

davkal:
screen shot 2015-09-21 at 19 03 43

### DIFF
--- a/client/app/scripts/charts/nodes-chart.js
+++ b/client/app/scripts/charts/nodes-chart.js
@@ -11,6 +11,7 @@ const Edge = require('./edge');
 const Naming = require('../constants/naming');
 const NodesLayout = require('./nodes-layout');
 const Node = require('./node');
+const NodesError = require('./nodes-error');
 
 const MARGINS = {
   top: 130,
@@ -151,6 +152,27 @@ const NodesChart = React.createClass({
     }, this);
   },
 
+  renderMaxNodesError: function(show) {
+    return (
+      <NodesError faIconClass="fa-ban" hidden={!show}>
+        <div className="centered">Too many nodes to show in the browser.<br />We're working on it, but for now, try a different view?</div>
+      </NodesError>
+    );
+  },
+
+  renderEmptyTopologyError: function(show) {
+    return (
+      <NodesError faIconClass="fa-circle-thin" hidden={!show}>
+        <div className="heading">Nothing to show. This can have any of these reasons:</div>
+        <ul>
+          <li>We haven't received any reports from probes recently. Are the probes properly configured?</li>
+          <li>There are nodes, but they're currently hidden. Check the view options in the bottom-left if they allow for showing hidden nodes.</li>
+          <li>Containers view only: you're not running Docker, or you don't have any containers.</li>
+        </ul>
+      </NodesError>
+    );
+  },
+
   render: function() {
     const nodeElements = this.renderGraphNodes(this.state.nodes, this.state.nodeScale);
     const edgeElements = this.renderGraphEdges(this.state.edges, this.state.nodeScale);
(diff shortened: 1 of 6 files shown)


2opremio:
Nice design. I have tried to test it but miserably failed:

screen shot 2015-09-29 at 1 54 24 pm


To force an empty graph, I made the probe not report to the app, providing an unassigned IP to scope:

```
./scope launch 10.0.0.1
```

Here's the `/api/report`:

```
{"Endpoint":{"Nodes":{}},"Address":{"Nodes":{}},"Process":{"Nodes":{}},"Container":{"Nodes":{}},"ContainerImage":{"Nodes":{}},"Host":{"Nodes":{}},"Overlay":{"Nodes":{}},"Sampling":{"Count":0,"Total":0},"Window":0}
```

tomwilkie:
@2opremio try ```make static```

2opremio:
I did that

client/app/scripts/charts/nodes-chart.js
```diff
@@ -151,6 +152,27 @@ const NodesChart = React.createClass({
...
+    );
+  },
+
+  renderEmptyTopologyError: function() {
+    return (
+      <NodesError faIconClass="fa-circle-thin" hidden={_.size(this.state.nodes) > 0}>
+        <div className="heading">Nothing to show. This can have any of these reasons:</div>
+        <ul>
+          <li>We haven't received any reports from probes recently. Are the probes properly configured?</li>
+          <li>There are nodes, but they're currently hidden. Check the hidden settings in the bottom-left.</li>
```
tomwilkie: s/hidden settings/filters/?

tomwilkie:
Works nicely for me:

screen shot 2015-10-29 at 18 06 15

Mind if I make the 1 suggested change and merge it?

tomwilkie:
NB I rebased this.

tomwilkie:
David wanted to solve some 'flickering' issues with this before it goes in.

client/app/scripts/stores/app-store.js
```diff
@@ -44,6 +44,7 @@ function makeNode(node) {
 
 // Initial values
 
+let changingTopologies = false;
```
tomwilkie: Thinking out loud: instead of not-showing-the-warning based on this flag, which I think could be a pain to maintain, could you use the stats from the regular fetch of /api/topology to indicate if the warning should be shown?

davkal: What a great idea! Sometimes one needs to take a step back to see the simple solution.

davkal:
@tomwilkie the implementation based on `/api/topology` is tricky, because the stats of a container topology return `node_count: 0`, but then send two nodes over the websocket: `theinternet` and `uncontained`. So those are rendered.

So either we adjust the `node_count` to reflect what would come over the websocket, or I dont render the graph when `node_count` is 0 as a hard, sufficient condition.

davkal:
Just found another stats field: `nonpseudo_node_count`. I could use that as a sufficient condition.

davkal:
Alright, a mix of stats and nodes map size does the trick:
```
currentTopology && currentTopology.stats && currentTopology.stats.node_count === 0 && nodes.size === 0;
```

client/app/scripts/stores/__tests__/app-store-test.js
```diff
@@ -101,6 +106,15 @@ describe('AppStore', function() {
     }
   };
 
+  const ReceiveNodesDeltaEmptyAction = {
```
tomwilkie: I think this is unused now?

tomwilkie:
One minor nit; otherwise LGTM.

I'm fixing master right now; once I have, could you rebase & squash please?   Once you get a green off circle, go ahead and merge.

PR #512, opened 2015-09-23
Remove pseudo nodes that are made unconnected by filtering.

### LINKED ISSUE #483: When filtering out system containers, we should also filter out pseudo nodes if they were only connected to system containers.
Below, the internet pseudo node should be filtered out:

screen shot 2015-09-14 at 13 14 10
screen shot 2015-09-14 at 13 14 24

tomwilkie:
Fixes #483

### DIFF
--- a/render/render.go
+++ b/render/render.go
@@ -178,10 +178,12 @@ type Filter struct {
 // Render implements Renderer
 func (f Filter) Render(rpt report.Report) RenderableNodes {
 	output := RenderableNodes{}
+	inDegrees := map[string]int{}
 	for id, node := range f.Renderer.Render(rpt) {
 		if f.FilterFunc(node) {
 			output[id] = node
 		}
+		inDegrees[id] = 0
 	}
 
 	// Deleted nodes also need to be cut as destinations in adjacency lists.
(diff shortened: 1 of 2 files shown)


render/render.go
```diff
@@ -178,10 +178,12 @@ type Filter struct {
 // Render implements Renderer
 func (f Filter) Render(rpt report.Report) RenderableNodes {
 	output := RenderableNodes{}
+	inDegrees := map[string]int{}
 	for id, node := range f.Renderer.Render(rpt) {
 		if f.FilterFunc(node) {
 			output[id] = node
 		}
+		inDegrees[id] = 0
```
peterbourgon: Is this like the initial set? If so, it's not necessary: the zero value of an int is zero.

peterbourgon:
LGTM.

render/render.go
tomwilkie: Indeed, except if we didn't do the initial set the iteration through all the in degrees below would only yield those with in degree > 0, and we wouldn't filter out those with no incoming edges...

peterbourgon: Gotcha.

PR #515, opened 2015-09-23
Remove some duplicate functionality, add some basic tests.

tomwilkie:


probe/endpoint/nat.go
```diff
@@ -16,16 +16,18 @@ type endpointMapping struct {
...
-type natmapper struct {
-	*Conntracker
+// NATMapper rewrites a report to deal with NAT's connections
+type NATMapper struct {
+	Conntracker
 }
 
-func newNATMapper() (*natmapper, error) {
+// NewNATMapper is exposed for testing
+func NewNATMapper() (*NATMapper, error) {
```
peterbourgon: I think now you can have all the NATMapper methods on a value receiver, since Conntracker is an interface...

probe/endpoint/reporter.go
```diff
@@ -62,7 +62,7 @@ func NewReporter(hostID, hostName string, includeProcesses bool, useConntrack bo
 		}
 	}
 	if conntrackModulePresent {
-		natmapper, err = newNATMapper()
+		natmapper, err = NewNATMapper()
```
peterbourgon: It looks like the only time you invoke the NewNATMapper constructor is within `package endpoint`, and the only reason it's exported is so you can test it in the `endpoint_test` package. If that's true, I'm not sure it's a good idea... the API boundary of a package should be dictated by its use, and only changed to improve testability if it's necessary. But in this case, I think it's fine (better) to leave it unexported, and put its tests in `endpoint_internal_test.go` with `package endpoint`.

Not a huge deal—just something that confused me as I was reading.

probe/endpoint/conntrack.go
```diff
@@ -65,8 +65,14 @@ type conntrack struct {
 	Flows   []Flow   `xml:"flow"`
 }
 
+// Conntracker is somethin that tracks connections.
```
peterbourgon: something

probe/endpoint/conntrack.go
```diff
@@ -75,11 +81,11 @@ type Conntracker struct {
 }
 
 // NewConntracker creates and starts a new Conntracter
-func NewConntracker(existingConns bool, args ...string) (*Conntracker, error) {
+var NewConntracker = func(existingConns bool, args ...string) (Conntracker, error) {
```
peterbourgon: As a general statement, I think we've passed the threshold where we're overusing package-level construction mocks/stubs. I'd like to issue a moratorium on the creation of new stubs, and instead pass factory functions to components that need them as parameters. What do you think?

peterbourgon:
Just some minor feedback. LGTM otherwise.

probe/endpoint/reporter.go
tomwilkie: I agree - but the test need to be in a separate package in this instance, as testing imports endpoint, so endpoint can't input testing.  In general though I much prefer tests in the same package as the code for this very reason.

probe/endpoint/conntrack.go
tomwilkie: Whats the downsides of this approach?  I'm not a huge fan of the factory style, if only because it pollutes potentially clean APIs (it would work better if golang had optional parameters...)

I guess there are probably cases (this being one) where we would just pass a conntracker to natmapper; that seems clean & obvious, so I'll make that change.

PR #519, opened 2015-09-24
Report number of filtered nodes in topology stats.

### LINKED ISSUE #509: Show number of filtered nodes in the bottom left.
To make it more obvious when you get an empty view that nodes have been filtered.

tomwilkie:
@davkal this adds a filtered_nodes field to the /api/topology endpoint.  You'll need to ensure you pass the same query parameters (ie system=show) to this endpoint to get accurate stats.  Could you do the UI pieces?

Fixes #509

### DIFF
--- a/app/router.go
+++ b/app/router.go
@@ -100,6 +100,17 @@ func makeReportPostHandler(a xfer.Adder) http.HandlerFunc {
 	}
 }
 
+func decorateTopologyForRequest(r *http.Request, topology *topologyView) {
+	for param, opts := range topology.options {
+		value := r.FormValue(param)
+		for _, opt := range opts {
+			if (value == "" && opt.def) || (opt.value != "" && opt.value == value) {
+				topology.renderer = opt.decorator(topology.renderer)
+			}
+		}
+	}
+}
+
 func captureTopology(rep xfer.Reporter, f func(xfer.Reporter, topologyView, http.ResponseWriter, *http.Request)) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		topology, ok := topologyRegistry[mux.Vars(r)["topology"]]
@@ -107,14 +118,7 @@ func captureTopology(rep xfer.Reporter, f func(xfer.Reporter, topologyView, http
 			http.NotFound(w, r)
 			return
 		}
-		for param, opts := range topology.options {
-			value := r.FormValue(param)
-			for _, opt := range opts {
-				if (value == "" && opt.def) || (opt.value != "" && opt.value == value) {
-					topology.renderer = opt.decorator(topology.renderer)
-				}
-			}
-		}
+		decorateTopologyForRequest(r, &topology)
 		f(rep, topology, w, r)
 	}
 }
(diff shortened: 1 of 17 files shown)


davkal:
Fixes #207

app/router.go
```diff
@@ -100,6 +100,17 @@ func makeReportPostHandler(a xfer.Adder) http.HandlerFunc {
 	}
 }
 
+func decorateTopoloyForRequest(r *http.Request, topology *topologyView) {
```
peterbourgon: topology

peterbourgon:
Looks great! I'd propose replacing "0 filtered" with just nothing, i.e. "11 nodes, 0 filtered" becomes just "11 nodes".

davkal:
screen shot 2015-09-25 at 15 19 25

Alright, the presence of `0 filtered` was confusing.

tomwilkie:
When clicking hide system containers, it doesn't hide them any more.  No error reported in logs.

Also, we're not counting filtered psuedo nodes.  I'll fix.

davkal:
Show/hide works for me. Got a report?

tomwilkie:
Report: https://gist.github.com/tomwilkie/bfda494869cfc95a14c1

Screenshot:
screen shot 2015-09-28 at 10 36 18

I'm also seeing an occasional "Uncaught TypeError: Cannot read property 'get' of undefined" in the console.

tomwilkie:
Have tested and can confirm UI works as expected: LGTM for those bits.

Can someone give the go bits a quick look please?

tomwilkie:
@davkal got another issue - "System container hidden" but "2 nodes" showing in status - should be "0 nodes (2 filtered)"

screen shot 2015-09-28 at 11 58 24

To reproduce:
- load UI
- click show (next to system containers hidden)
- switch to applications view
- switch back to containers view.

render/render.go
```diff
@@ -54,6 +75,14 @@ func (m Map) Render(rpt report.Report) RenderableNodes {
 	return output
 }
 
+// Stats implements Renderer
+func (m Map) Stats(rpt report.Report) Stats {
```
paulbellamy: You mean "Map implements Renderer"?

render/render.go
```diff
@@ -11,6 +11,18 @@ import (
 type Renderer interface {
 	Render(report.Report) RenderableNodes
 	EdgeMetadata(rpt report.Report, localID, remoteID string) report.EdgeMetadata
+	Stats(report.Report) Stats
+}
```
paulbellamy: What about:
```Go
type Renderer interface {
  Render(report.Report) RenderedReport
}

type RenderedReport struct {
  Nodes RenderableNodes
  Edges EdgeMetadata
  Stats Stats
}
```

tomwilkie: Why? That would be a huge refactor - for what purpose?

paulbellamy: per f2f discussion: I was under the impression that the Nodes, Edges, and Stats were all returned via the same api call, so would continue travelling together.

davkal:
@paulbellamy @tomwilkie  if I click hide system containers in the containers topology, do I expect system nodes to be hidden in other topologies, too? Or should that be independent of other topologies?

tomwilkie:
@davkal independent right now.  Each topologies defines their own filters.

tomwilkie:
That works well @davkal.

I'll rebase, and we're just waiting on an LGTM from @paulbellamy

peterbourgon:
LGTM.

PR #521, opened 2015-09-25
Test for BackgroundPublisher

tomwilkie:


peterbourgon:
LGTM. For the record, runtime.Gosched isn't guaranteed to do anything. I don't know a better pattern, however.

tomwilkie:
Yeah me neither.  I just need the goroutine in background publisher to run, and as goroutines are cooperatively multitasking, they need some sleep or sched call.  This shouldn't be as fragile as other uses.

Issue #417, opened 2015-09-01
Add NodeMetadata.Sets


tomwilkie:
Some values in nmd.Metadata are actually sets encoded as string separated lists, with set-merge semantics.

Like NodeMetadata.Counters, we should put these in NodeMetada.Sets (of type map[string]IDList) and define appropriate merge semantics.

Examples would be container IP addresses + hostnames, host networks.  Others?

peterbourgon:
Not map[string]IDList but map[string]StringSet, otherwise :+1:

tomwilkie:
IDList is a stringset; can we just typedef it?

peterbourgon:
Sure, just want to make it distinct.

PR #522, opened 2015-09-28
NodeMetadata Sets

peterbourgon:
Addresses #417

### DIFF
--- a/report/topology.go
+++ b/report/topology.go
@@ -2,6 +2,7 @@ package report
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 )
 
@@ -78,8 +79,9 @@ func (n Nodes) Merge(other Nodes) Nodes {
 // given node in a given topology, along with the edges emanating from the
 // node and metadata about those edges.
 type Node struct {
-	Metadata  `json:"metadata,omitempty"`
-	Counters  `json:"counters,omitempty"`
+	Metadata  Metadata      `json:"metadata,omitempty"`
+	Counters  Counters      `json:"counters,omitempty"`
+	Sets      Sets          `json:"sets,omitempty"`
 	Adjacency IDList        `json:"adjacency"`
 	Edges     EdgeMetadatas `json:"edges,omitempty"`
 }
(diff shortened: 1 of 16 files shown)


report/topology.go
```diff
@@ -78,8 +79,9 @@ func (n Nodes) Merge(other Nodes) Nodes {
 // given node in a given topology, along with the edges emanating from the
 // node and metadata about those edges.
 type Node struct {
-	Metadata  `json:"metadata,omitempty"`
-	Counters  `json:"counters,omitempty"`
+	Metadata  Metadata      `json:"metadata,omitempty"`
+	Counters  Counters      `json:"counters,omitempty"`
+	Sets      Sets          `json:"sets,omitempty"`
```
2opremio: I would put Sets inside Metadata (it's just a different way to structure metadata). Also, I wonder If we can avoid the heterogeneous representation (maps and sets) by using interfaces with getters instead of plain strings.

peterbourgon: The goal is to eventually eliminate metadata.

I'm not sure what you're proposing. `map[string]interface{}`?

tomwilkie: Sets is akin to Counters and belongs outside Metadata.  The different fields (Metadata, Sets, Counters) represent different kinds of data to associate with a Node, and different ways to merge said data.

report/topology.go
```diff
@@ -195,6 +197,100 @@ func (c Counters) Copy() Counters {
...
+	result := Sets{}
+	for k, v := range s {
+		result[k] = v.Copy()
+	}
+	return result
+}
+
+// StringSet is a sorted set of unique strings. Clients must use the Add
+// method to add strings.
+type StringSet []string
```
2opremio: I am not convinced that the ordering provided by `[]string` is a real advantage over using `map[string]struct{}`. 

`map[string]struct{}` would make the implementation much simpler plus ordering can be guaranteed for tests by sorting explicitly.

peterbourgon: The implementation is contained to this file, and well-tested. The public API is Make, Add, Merge, and Contains. Those methods would remain with the other implementation.

Also, another important benefit of []string is that the JSON serialization is intuitive: `["a", "b", "c"]`. If we switched to map[string]struct{}, we would need to implement Marshal/Unmarshal JSON methods to serialize to, and deserialize from, this form.

tomwilkie: Another advantage is the zero type if valid if we use []string.

report/topology.go
```diff
@@ -195,6 +197,100 @@ func (c Counters) Copy() Counters {
...
+func (s Sets) Copy() Sets {
+	result := Sets{}
+	for k, v := range s {
+		result[k] = v.Copy()
+	}
+	return result
+}
+
+// StringSet is a sorted set of unique strings. Clients must use the Add
+// method to add strings.
```
tomwilkie: Given this implementation is identical to IDList (AFAICT), could we just type def it instead of copying the code?

peterbourgon: The idea was to typedef IDList (specific) to StringSet (general), just didn't finish the implementation.

tomwilkie: Makes sense, I'll do it that way round.

tomwilkie: Okay I had a stab at that but its not very clean - it would seem the type aliasing doesn't propagate methods.  I tried embedding it in a struct too - that didn't work either.

https://github.com/weaveworks/scope/commit/b867b0465a252a06329df4ed199990d8fd757bb5

Thoughts?

report/topology.go
paulbellamy: Though you could implement the methods on a map so that the zero type is still valid:
```Go
func (s set) add(v string) set {
	if s == nil {
		s = make(map[string]struct{})
	}
	s[v] = struct{}{}
	return s
}
```

report/topology.go
```diff
@@ -195,6 +215,100 @@ func (c Counters) Copy() Counters {
...
+type StringSet []string
+
+// MakeStringSet makes a new StringSet with the given strings.
+func MakeStringSet(strs ...string) StringSet {
```
paulbellamy: This function modifies the input, which could cause gotchas later on.

e.g.
```Go
a := []string{"b", "a", "c"}
MakeStringSet(a...)
fmt.Println(a) // [a b c]
```

paulbellamy:
A niggle about modifying arguments, but other than that, LGTM.

report/topology.go
peterbourgon: Seems dangerous to me, too...

tomwilkie: Nice catch! I'll fix this first thing.

Issue #329, opened 2015-07-20
make getting the docker run command easier


progrium:
The launcher is nice, but it won't work for how we deploy stuff. And I imagine many others. I had to pick through the script to see how the container is actually run. It would be nice to see the docker run command as an alternative way to run it in the readme, and/or a subcommand that will return the docker run command (that could be passed into a subshell or just copy and pasted from the command line). 

I can submit a PR for the former if that sounds ok.

tomwilkie:
Hi Jeff; a contribution like that would be welcome.

The scope script is also used to register scope with weave DNS, which is
done so scopes can find each other and automatically cluster.

Thanks

Tom

progrium:
It's a privileged container so you could register from inside the
container, right?

On Tuesday, July 21, 2015, Tom Wilkie <notifications@github.com> wrote:

> Hi Jeff; a contribution like that would be welcome.
>
> The scope script is also used to register scope with weave DNS, which is
> done so scopes can find each other and automatically cluster.
>
> Thanks
>
> Tom
>
> On Monday, 20 July 2015, Jeff Lindsay <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
>
> > The launcher is nice, but it won't work for how we deploy stuff. I had to
> > pick through the script to see how the container is actually run. It
> would
> > be nice to see the docker run command as an alternative way to run it in
> > the readme, and/or a subcommand that will return the docker run command
> > (that could be passed into a subshell or just copy and pasted from the
> > command line).
> >
> > I can submit a PR for the former if that sounds ok.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/weaveworks/scope/issues/329>.
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/weaveworks/scope/issues/329#issuecomment-123218616>.
>


-- 
Jeff Lindsay
http://progrium.com

tomwilkie:
Yes thats a good point; we should do that.

Issue #510, opened 2015-09-23
Make scope deal with weave restarts (or weave starts after scope)


tomwilkie:
```
2015/09/23 06:07:17 error doing ticker: Get http://172.17.0.1:6784/report: dial tcp 172.17.0.1:6784: no route to host
2015/09/23 06:07:18 report generation took too long (4.194702594s)
2015/09/23 06:07:20 error doing ticker: Get http://172.17.0.1:6784/report: dial tcp 172.17.0.1:6784: no route to host
2015/09/23 06:07:21 report generation took too long (3.137524219s)
2015/09/23 06:07:24 error doing ticker: Get http://172.17.0.1:6784/report: dial tcp 172.17.0.1:6784: no route to host
2015/09/23 06:07:25 report generation took too long (4.42816017s)
2015/09/23 06:07:27 error doing ticker: Get http://172.17.0.1:6784/report: dial tcp 172.17.0.1:6784: no route to host
2015/09/23 06:07:28 report generation took too long (2.781171793s)
2015/09/23 06:07:31 error doing ticker: Get http://172.17.0.1:6784/report: dial tcp 172.17.0.1:6784: no route to host
2015/09/23 06:07:32 report generation took too long (4.325375019s)
2015/09/23 06:07:34 error doing ticker: Get http://172.17.0.1:6784/report: dial tcp 172.17.0.1:6784: no route to host
2015/09/23 06:07:36 report generation took too long (3.229815901s)
```

tomwilkie:
I'm also missing a connection from scope -> weave after the restart.

PR #524, opened 2015-09-29
Move bulk of scope script logic into entrypoint.sh, extend testing

### LINKED ISSUE #523: Weave integration has false negatives
The `scope` script checks that the weave script is in the user's path (via `command_exists`). But this won't be the case in general; and in particular, if weave is being run remotely.

The result is that scope doesn't associate weave IPs with containers, and thereby doesn't detect short-lived connections over the weave network.

### LINKED ISSUE #531: MultiPublisher leaks go routines
When Set is called, we replace any existing publishers for a given target.  The old publishers are not Stopped.

tomwilkie:
Fixes #523 by moving the detection and communication with Weave into the Scope container (and out of the scope script).

Will also help towards #329.

Also
- Add test for clustering without weave
- Update test image to docker 1.8
- Stop publishers when they are replaces (fixes #531)
- Poll DNS every 10s to speed up scope clustering

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,12 +1,56 @@
 #!/bin/sh
 
 usage() {
-	echo "$0 --dns <IP> --hostname <NAME> --searchpath <SEARCHPATH> --app.foo bar --probe.foo bar"
+	echo "$0 --app.foo bar --probe.foo bar"
 	exit 1
 }
 
 # This script exists to modify the network settings in the scope containers
 # as docker doesn't allow it when started with --net=host
+
+WEAVE_CONTAINER_NAME=weave
+DOCKER_BRIDGE=docker0
+HOSTNAME=scope
+DOMAIN=weave.local
+IP_REGEXP="[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}"
+
+container_ip() {
+    if ! status=$(docker inspect --format='{{.State.Running}} {{.NetworkSettings.IPAddress}}' $1 2>/dev/null); then
+        echo "$2" >&2
+        return 1
+    fi
+    case "$status" in
+        "true ")
+            echo "$1 container has no IP address; is Docker networking enabled?" >&2
+            return 1
+            ;;
+        true*)
+            CONTAINER_IP="${status#true }"
+            ;;
+        *)
+            echo "$3" >&2
+            return 1
+            ;;
+    esac
+}
+
+is_running() {
+    status=$(docker inspect --format='{{.State.Running}}' $1 2>/dev/null) && [ "$status" = "true" ]
+    return $?
+}
+
+docker_bridge_ip() {
+    local DOCKER_BRIDGE_IP=$(ip -f inet address show dev $DOCKER_BRIDGE | grep -m1 -o 'inet \([.0-9]\)*')
+    echo ${DOCKER_BRIDGE_IP#inet }
+}
+
+# Run `weave expose` if it's not already exposed.
+weave_expose() {
+    status=$(weave --local ps weave:expose | awk '{print $3}' 2>/dev/null)
+    if [ "$status" = "" ]; then
+        weave --local expose
+    fi
+}
 
 mkdir -p /etc/weave
 APP_ARGS=""
@@ -14,16 +58,6 @@ PROBE_ARGS=""
 
 while true; do
     case "$1" in
-        --dns)
-            [ $# -gt 1 ] || usage
-            DNS_SERVER="$2"
-            shift
-            ;;
-        --searchpath)
-            [ $# -gt 1 ] || usage
-            SEARCHPATH="$2"
-            shift
-            ;;
         --app.*)
             if echo "$1" | grep "=" 1>/dev/null; then
                 ARG_NAME=$(echo "$1" | sed 's/\-\-app\.\([^=]*\)=\(.*\)/\1/')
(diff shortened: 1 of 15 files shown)


integration/setup.sh
```diff
@@ -15,4 +15,7 @@ echo Installing weave
 for HOST in $HOSTS; do
     run_on $HOST "sudo curl -sL git.io/weave -o /usr/local/bin/weave"
     run_on $HOST "sudo chmod a+x /usr/local/bin/weave"
+	weave_on $HOST setup
```
paulbellamy: whitespace alignment is off

paulbellamy:
Why not move most of the logic straight into the go code? instead of the bash? Just expediency?

scope
```diff
@@ -20,33 +20,16 @@ fi
 IMAGE_VERSION=${VERSION:-$IMAGE_VERSION}
 SCOPE_IMAGE=weaveworks/scope:$IMAGE_VERSION
 SCOPE_CONTAINER_NAME=weavescope
-WEAVE_CONTAINER_NAME=weave
-WEAVEDNS_CONTAINER_NAME=weavedns
-HOSTNAME=scope
-DOMAINNAME=weave.local
-FQDN=$HOSTNAME.$DOMAINNAME
 DOCKER_BRIDGE=${DOCKER_BRIDGE:-docker0}
```
paulbellamy: This is unused, now.

integration/200_clustering_2_test.sh
```diff
@@ -2,24 +2,30 @@
...
-done
-assert "echo $SUCCESS" "1"
-
-end_suite
+start_suite "Launch 2 scopes and check they cluster automatically"
+
+weave_on $HOST1 launch $HOST1 $HOST2
+weave_on $HOST2 launch $HOST1 $HOST2
+
+sleep 10 # let weave settle
```
paulbellamy: Is there any way we could poll and check that it has "settled"?

tomwilkie:
> Why not move most of the logic straight into the go code? instead of the bash? Just expediency?

Yeah we should do that at some point; that would help with #510.  But thats a bigger change, and this gets us some of the way there for now.

docker/entrypoint.sh
```diff
@@ -73,15 +107,34 @@ while true; do
     shift
 done
 
-echo "$APP_ARGS" >/etc/weave/scope-app.args
-echo "$PROBE_ARGS" >/etc/weave/scope-probe.args
+if is_running $WEAVE_CONTAINER_NAME; then
```
paulbellamy: We're in the container on the weave network now. Couldn't we just check for ethwe?

paulbellamy: Ah, we need the IP anyway, to hit `/report`...

docker/entrypoint.sh
```diff
@@ -1,29 +1,63 @@
...
 	exit 1
 }
 
 # This script exists to modify the network settings in the scope containers
 # as docker doesn't allow it when started with --net=host
 
+WEAVE_CONTAINER_NAME=weave
+DOCKER_BRIDGE=docker0
+HOSTNAME=scope
+DOMAIN=weave.local
```
paulbellamy: What if they've done `weave launch --dns-domain=foo.bar.` ?

docker/entrypoint.sh
```diff
@@ -73,15 +107,34 @@ while true; do
...
-if [ -n "$DNS_SERVER" -a -n "$SEARCHPATH" ]; then
-    echo "domain $SEARCHPATH" >/etc/resolv.conf
-    echo "search $SEARCHPATH" >>/etc/resolv.conf
-    echo "nameserver $DNS_SERVER" >>/etc/resolv.conf
+    DOCKER_BRIDGE_IP=$(docker_bridge_ip)
+    echo "Weave container detected at $CONTAINER_IP, Docker bridge at $DOCKER_BRIDGE_IP"
+
+    echo "domain $DOMAIN" >/etc/resolv.conf
+    echo "search $DOMAIN" >>/etc/resolv.conf
+    echo "nameserver $DOCKER_BRIDGE_IP" >>/etc/resolv.conf
```
paulbellamy: Won't this have been set up when the scope probe was launched into the weave network? Or are we not assuming that it is in the weave network?

integration/105_launch_sans_weave_test.sh
```diff
@@ -4,11 +4,12 @@
...
 start_suite "Launch scope (without weave installed) and check it boots"
 
-assert_raises "run_on $HOST1 \
-  PATH=/usr/local/scope/bin:/usr/sbin:/usr/bin:/sbin:/bin \
-  DOCKER_HOST=tcp://$HOST1:$DOCKER_PORT \
-  scope launch"
+scope_on $HOST1 launch
 
-assert_raises "curl $HOST1:4040"
+sleep 5
```
paulbellamy: Comment on sleeps to say what we're waiting for, please.

integration/config.sh
```diff
@@ -10,21 +10,38 @@ export HOSTS
...
+  local host=$1
   shift 1
   [ -z "$DEBUG" ] || greyly echo "Weave on $host: $@" >&2
-  run_on $host DOCKER_HOST=tcp://$host:$DOCKER_PORT weave "$@"
+  DOCKER_HOST=tcp://$host:$DOCKER_PORT $WEAVE "$@"
+}
+
+# this checks we have a weavescope container
+has_container() {
+	local host=$1
```
paulbellamy: Match the indentation throughout the file, please (2 spaces).

scope
```diff
@@ -75,113 +58,19 @@ check_not_running() {
...
-                exit 1
-            fi
-            weave_dns_add $CONTAINER $FQDN $IP_ADDRS
-        fi
-
+            $WEAVESCOPE_DOCKER_ARGS $SCOPE_IMAGE --probe.docker true "$@")
         echo $CONTAINER
 
         if ! echo "$@" | grep -E "\-\-no\-app|\-\-service\-token" 1>/dev/null; then
+			IP_ADDRS=$(docker run --rm --net=host --entrypoint /bin/sh $SCOPE_IMAGE -c "$IP_ADDR_CMD")
```
paulbellamy: Whitespace

docker/entrypoint.sh
tomwilkie: Also, not on the weave network just yet - we're in the host network namespace, and we haven't done the expose.

docker/entrypoint.sh
tomwilkie: Yeah that won't work.  That was broken before and this change has made it no worse - I'll file a bug for it.

tomwilkie: #532

docker/entrypoint.sh
tomwilkie: Its launched with --net=host (and won't work if launched any other way).

integration/config.sh
tomwilkie: I went with tabs if thats okay.

integration/200_clustering_2_test.sh
tomwilkie: Now thats weave pushed the DNS polling to 10s it might not be necessary.  Let me try.

PR #525, opened 2015-09-29
vendor/runsvinit: update vendored copy

peterbourgon:
Will de-spam the logs.

### DIFF
--- a/vendor/runsvinit/main.go
+++ b/vendor/runsvinit/main.go
@@ -13,52 +13,66 @@ import (
 
 const etcService = "/etc/service"
 
+var (
+	debugf = log.Printf
+	info   = log.Print
+	infof  = log.Printf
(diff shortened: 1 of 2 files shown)


vendor/runsvinit/main.go
```diff
@@ -13,52 +13,66 @@ import (
 
 const etcService = "/etc/service"
 
+var (
+	debugf = log.Printf
+	info   = log.Print
+	infof  = log.Printf
+	fatal  = log.Fatal
+	fatalf = log.Fatalf
```
tomwilkie: given that fatal and info are never overridden, it doesn't seem necessary to alias them here.

peterbourgon: It's just for consistency — I don't want log.Foo(...) and bar(...) mixed.

PR #526, opened 2015-09-29
Refactor probe/endpoint for export and dependency cleanliness

### LINKED ISSUE #589: Nil pointer dereference in TestSpyNoProcesses when run in Darwin
https://gist.githubusercontent.com/peterbourgon/a9825b1f218ea078e451/raw/c65f687deb9aafbe50086ab033adabf2164dff1d/gistfile1.txt

peterbourgon:
Fixes #589

Motivation: https://gist.github.com/peterbourgon/4185284b3b2c2e014b0f


probe/endpoint/nat.go
```diff
@@ -50,7 +50,10 @@ func toMapping(f Flow) *endpointMapping {
 // ApplyNAT duplicates Nodes in the endpoint topology of a
 // report, based on the NAT table as returns by natTable.
 func (n NATMapper) ApplyNAT(rpt report.Report, scope string) {
-	n.WalkFlows(func(f Flow) {
+	if n.Conntracker == nil {
+		return
+	}
```
tomwilkie: I'd put this test in reporter.go; it creating the conntracker fails, don't both to create the nat mapper.

There is already a test for the natmapper being nil: https://github.com/weaveworks/scope/blob/master/probe/endpoint/reporter.go#L142

peterbourgon: NATMapper cannot be nil anymore as it's a value-type. Further, I'd rather not use nil as an ersatz Option type. It feels like the right way to do this isn't to make nil checks deep in component logic, but rather to assume all dependencies are valid, and provide no-op implementations when they may be disabled, or fail to be constructed.

peterbourgon: I would like to merge this change, so that my tests no longer segfault. If you're not :+1:, I'll switch my attention to refactoring this package so that these nil conditions don't occur. Please let me know.

tomwilkie: >  It feels like the right way to do this isn't to make nil checks deep in component logic,

It anything this is pushing the nil check further in to the logic.

And in either case the nil check is virtually the same place (it was before the call to ApplyNat, now its the first thing that happens ~after the call).

If you want to provide a noop implementation, then I'm fine with that.  But that would entail making NATMapper an interface, which can be nil, so kinda defeats the whole purpose.

tomwilkie:
I LGTM for peter's changes, just need an LGTM on my last one.

probe/endpoint/conntrack.go
```diff
@@ -14,87 +13,92 @@ import (
...
+	bufferedFlows []flow         // flows coming out of activeFlows spend 1 walk cycle here
 	existingConns bool
 	args          []string
 	quit          chan struct{}
 }
 
-// NewConntracker creates and starts a new Conntracter
-func NewConntracker(existingConns bool, args ...string) (Conntracker, error) {
+// newConntracker creates and starts a new conntracker.
+func newConntrackFlowWalker(useConntrack, existingConns bool, args ...string) flowWalker {
```
paulbellamy: `existingConns` only seems to be used by the tests. any way we can get rid of it?

tomwilkie: We could extend the tests to deal with that behaviour.  But thats work....

Oh go on then I'll do it.

probe/endpoint/reporter.go
```diff
@@ -49,44 +48,21 @@ var SpyDuration = prometheus.NewSummaryVec(
...
-	}
 	return &Reporter{
 		hostID:           hostID,
 		hostName:         hostName,
 		includeProcesses: includeProcesses,
-		conntracker:      conntracker,
-		natmapper:        &natmapper,
-		revResolver:      NewReverseResolver(),
+		flowWalker:       newConntrackFlowWalker(useConntrack, true),
+		natMapper:        makeNATMapper(newConntrackFlowWalker(useConntrack, true, "--any-nat")),
```
paulbellamy: `useConntrack` here should be `true`, otherwise this is a change in behaviour.

probe/endpoint/conntrack.go
```diff
@@ -14,87 +13,92 @@ import (
...
+	walkFlows(f func(flow))
+	stop()
 }
 
-// Conntracker uses the conntrack command to track network connections
-type conntracker struct {
+type nilFlowWalker struct{}
+
+func (n *nilFlowWalker) stop()                  {}
+func (n *nilFlowWalker) walkFlows(f func(flow)) {}
```
paulbellamy: These methods probably don't need to be on the pointer, then you probably don't need the `&` then initializing, right?

tomwilkie: Huh, I was labouring under the misunderstanding that only pointer-to-structs could implement interfaces.  I was obviously wrong.

probe/endpoint/reporter.go
tomwilkie: change in behaviour here is fine I think, no?

tomwilkie:
@paulbellamy PTAL

tomwilkie:
Test did pass: https://circleci.com/gh/weaveworks/scope/1747

paulbellamy:
LGTM

tomwilkie:
Updated to latest tool to overcome a bug in rebuild-image, well spotted @davkal!

PR #540, opened 2015-10-03
Change type

webwurst:
Fixes compile error for me

### DIFF
--- a/probe/host/system_linux.go
+++ b/probe/host/system_linux.go
@@ -12,7 +12,7 @@ import (
 // Uname is swappable for mocking in tests.
 var Uname = syscall.Uname
 
-func charsToString(ca [65]int8) string {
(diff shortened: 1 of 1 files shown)


webwurst:
Ok, somehow on my system `utsname.Release` returns `uint8`? I am using `golang:1.4` on `armv7l`.

tomwilkie:
Hi @webwurst.  The build failed on this, as on x86 and utsname.Release is definitely a ```int8```.  The complete fix would be to have a separate file that exposes GetKernelVersion and has build tags for arm.  Docs on how to do this are here: https://golang.org/pkg/go/build/

Do you think you could have a go at that?

tomwilkie:
Hi @webwurst, I've updated this branch with my comments here: #547

PR #545, opened 2015-10-05
Add containers by hostname view.

### LINKED ISSUE #544: Add containers-by-hostname view to scope
This is better than container-by-image, as it approximates services more closely (containers with the same host name being considered replicas for the same service)

tomwilkie:
Fixes #544

### DIFF
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -71,6 +71,7 @@ type Container interface {
 	ID() string
 	Image() string
 	PID() int
+	Hostname() string
 	GetNode([]net.IP) report.Node
 
 	StartGatheringStats() error
@@ -99,6 +100,15 @@ func (c *container) Image() string {
 
 func (c *container) PID() int {
 	return c.container.State.Pid
+}
+
+func (c *container) Hostname() string {
+	if c.container.Config.Domainname == "" {
+		return c.container.Config.Hostname
+	}
+
+	return fmt.Sprintf("%s.%s", c.container.Config.Hostname,
+		c.container.Config.Domainname)
 }
 
 func (c *container) StartGatheringStats() error {
@@ -222,7 +232,7 @@ func (c *container) GetNode(localAddrs []net.IP) report.Node {
 		ImageID:          c.container.Image,
 		ContainerIPs: strings.Join(append(c.container.NetworkSettings.SecondaryIPAddresses,
 			c.container.NetworkSettings.IPAddress), " "),
-		ContainerHostname: c.container.Config.Hostname,
+		ContainerHostname: c.Hostname(),
 	})
 	AddLabels(result, c.container.Config.Labels)
 
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -22,11 +22,12 @@ import (
 
 // These constants are keys used in node metadata
 const (
-	ContainerName    = "docker_container_name"
-	ContainerCommand = "docker_container_command"
-	ContainerPorts   = "docker_container_ports"
-	ContainerCreated = "docker_container_created"
-	ContainerIPs     = "docker_container_ips"
+	ContainerName     = "docker_container_name"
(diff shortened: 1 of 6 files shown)


tomwilkie:
@davkal what do you think re:UX?  Maybe make the sub topologies a drop down if there are more than 1?

screen shot 2015-10-05 at 16 04 42

davkal:
-1 on dropdowns, especially if it's only for 2 items.

Long term we could have groupings in one line like the time filter in google finance charts:
screen shot 2015-10-09 at 15 21 32

tomwilkie:
Fair enough; are you happy with UX as is?

davkal:
Yup. Sorry, should have said that.

tomwilkie:
How about, as you hover over Containers, the different groupings appear?  Then if you select one it stays visible?

davkal:
Hover only works with mouse. Leaves out ipad et al. for cool demos.

davkal:
Or not on hover, but click on main topology will show sub topologies. That would work on tablets too.

tomwilkie:
Ah yes.  Good point.  How about, on tablets, we use a 3d UI?

probe/docker/container.go
```diff
@@ -221,6 +222,7 @@ func (c *container) GetNode(localAddrs []net.IP) report.Node {
 		ImageID:          c.container.Image,
 		ContainerIPs: strings.Join(append(c.container.NetworkSettings.SecondaryIPAddresses,
 			c.container.NetworkSettings.IPAddress), " "),
+		ContainerHostname: c.container.Config.Hostname,
```
paulbellamy: This should include the domainname.

render/mapping.go
```diff
@@ -594,6 +594,28 @@ func MapContainer2Pod(n RenderableNode, _ report.Networks) RenderableNodes {
...
 
+// MapContainer2Hostname maps container RenderableNodes to 'hostname' renderabled nodes..
+func MapContainer2Hostname(n RenderableNode, _ report.Networks) RenderableNodes {
+	// Propogate all pseudo nodes
+	if n.Pseudo {
+		return RenderableNodes{n.ID: n}
+	}
+
+	// Otherwise, if some some reason the container doesn't have a hostname
+	// (maybe slightly out of sync reports), just drop it
```
paulbellamy: Is that really a reason the container wouldn't have a hostname?

render/mapping.go
```diff
@@ -594,6 +594,28 @@ func MapContainer2Pod(n RenderableNode, _ report.Networks) RenderableNodes {
...
+	id, ok := n.Node.Metadata[docker.ContainerHostname]
+	if !ok {
+		return RenderableNodes{}
+	}
+
+	// Add container-<id> key to NMD, which will later be counted to produce the minor label
```
paulbellamy: This comment is not what we are doing.

render/topologies.go
```diff
@@ -171,6 +171,25 @@ var ContainerImageRenderer = Map{
...
+				MapFunc:  MapContainer2ContainerImage,
+				Renderer: ContainerRenderer,
+			},
+			Map{
+				MapFunc:  MapContainerImageIdentity,
+				Renderer: SelectContainerImage,
+			},
+		),
+	},
+}
```
paulbellamy: Are we just grouping containers by hostname, or by container image, then hostname? It looks like the latter, but checking whether that is what we want.

render/mapping.go
tomwilkie: Yes; if the containers collect gets out of sync for some reason with the endpoint collection, you can end up with a container node with little metadata.  We can safely ignore them in this case.

render/topologies.go
tomwilkie: Yeah this is a bad merge I think.

tomwilkie:
@paulbellamy PTAL

paulbellamy:
LGTM

PR #547, opened 2015-10-09
Make Scope build on arm

tomwilkie:
This is a follow on from #540


tomwilkie:
Testing with: 
```
docker run -ti -e GOARCH=arm -v $(pwd):/go/github.com/weaveworks/scope golang:1.5 /bin/sh -c "cd github.com/weaveworks/scope && make app/scope-app probe/scope-probe"
```

Don't want to add it to circle, with all the problem we had compiling on darwin.  I'll try and think of a better way to though - #550

Issue #546, opened 2015-10-09
Kubernetes PR broke 300_internet_edge_test.sh 


tomwilkie:
```
vagrant@ubuntu-14:~/src/github.com/weaveworks/scope/integration$ ./300_internet_edge_test.sh 
Test short lived connections from the Internet
339ed943494b97835d0585333d94ec2ef551d091d412b734535d46de7fd6db85
Weave Scope is reachable at the following URL(s):
  * http://10.32.0.1:4040/
  * http://10.240.0.23:4040/
11845b1722455e5ccf885051a202ac797cc199ec76e75e9676d60446b1f9653e
/home/vagrant/src/github.com/weaveworks/weave/test/assert.sh: line 76: 15729 Terminated              do_connections
test #1 "curl -s http://host1.us-central1-a.positive-cocoa-90213:4040/api/topology/containers?system=show | jq -r '[.nodes[] | select(.label_major == "nginx")] | length'" failed:
	expected "1"
	got "0"
test #2 "curl -s http://host1.us-central1-a.positive-cocoa-90213:4040/api/topology/containers?system=show | jq -r '.nodes[""].adjacency | contains([""])'" failed:
	expected "true"
	got nothing
2 of 2 tests failed in 15.675s.
```

tomwilkie:
I suspect something has changed such that it is taking more that 5s to send the first report.  If you extend the timeout to 20s it passes.  Would be good to get to the bottom of though, probe shouldn't take that long to send first report.

paulbellamy:
This test only works on GCE (which is awful), but even on GCE, I can't replicate this. I would be *very* surprised if it is related to the Kubernetes changes.

PR #542, opened 2015-10-05
Add tests for short-lived connection between hosts and containers.

tomwilkie:
Fixes #448

PR #549, opened 2015-10-09
Hacky fix for integration/300

paulbellamy:
To help diagnose #546, and as a general improvement to the tests.

### DIFF
--- a/integration/100_launch_test.sh
+++ b/integration/100_launch_test.sh
@@ -9,8 +9,8 @@ scope_on $HOST1 launch
 
 sleep 5 # give the probe a few seconds to build a report and send it to the app
 
-has_container $HOST1 weave 1
-has_container $HOST1 weaveproxy 1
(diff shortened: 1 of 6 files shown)


tomwilkie:
Let see this pass, but I think you might have to add a "and wait for this set of container to exist or 5s" as well, as some of the cluster tests will give you 2 containers, and not the 4 you need.

paulbellamy:
Not using `wait_for_reports` in the clustering tests, as checking a set of containers is in the report thing involves a bit of jq madness.

paulbellamy:
Ditched wait_for_reports, as even on 300, it was returning after there were reports, but before there was nginx, and trying to wait for specific containers in the output descended into madness.

tomwilkie:
@paulbellamy I figured out why 300 was failing; the curl was being run to quickly and erroring, and killing the subshell.

I'll fix in #542 and incorporate your changes too.

PR #553, opened 2015-10-10
Make getting the docker run command easier.

### LINKED ISSUE #329: make getting the docker run command easier

tomwilkie:
Fixes #329

### DIFF
--- a/scope
+++ b/scope
@@ -6,6 +6,7 @@ usage() {
     echo "Usage:"
     echo "scope launch [<peer> ...]"
     echo "scope stop"
+    echo "scope command"
     echo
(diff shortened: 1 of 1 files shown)


tomwilkie:
Test failures should be fixed by #542

PR #558, opened 2015-10-13
Vendor "procspy"

inercia:
"procspy" has been copied to the `vendor` directory.
Unit tests have been moved to the `procspy_test` (in order to pass `lint` checks)


inercia:
Errors on CircleCI seem unrelated... :confused:

tomwilkie:
Please rebase, the errors should go away.

inercia:
@tomwilkie I think there is nothing to rebase...

tomwilkie:
@inercia yes you're right, sorry.

I've fixed the test breakage in #570, lets get that reviewed and merged and then rebase on top of that please.

paulbellamy:
Rebased to see if it passes.

paulbellamy:
Should we use the go 1.5 vendor thing (i.e. `./vendor/github.com/weaveworks/procspy`)? @peterbourgon ?

peterbourgon:
@paulbellamy Ideally, yes, but I'm fine with this as an intermediate solution, given I set the precedent with runsvinit. I've filed #573 to migrate to GO15VENDOREXPERIMENT.

tomwilkie:
The aim here is more than just vendor procspy: subsequent changes are going to integrate it into the probe a bit more (using the probe to walk the proc fs, to save a double walk, for instance).

paulbellamy:
So it's not really vendoring it, but forking it, then, yeah?

tomwilkie:
Yeah

paulbellamy:
Then, I'd rather just put it in `./procspy`. It's not really vendoring...

tomwilkie:
Yes, or even under probe/endpoint.

inercia:
I will move the code to `./procspy`, as @paulbellamy has suggested.

inercia:
_"Coverage decreased (-1.1%) to 73.912%"_  :confused:

tomwilkie:
Could you rebase this please?  And remove the vendored copy of procspy from vendor/, and vendor/manifest.

Also we need to decide what to do re:coverage before this is merged.

tomwilkie:
Here is the coverage report: https://circle-artifacts.com/gh/weaveworks/scope/1656/artifacts/0/tmp/circle-artifacts.SEylqlo/coverage.html

I think adding a test for proc.go would solve our problems.

inercia:
> I think adding a test for proc.go would solve our problems.

@tomwilkie the objective of this PR was just to _copy & paste_ the `procspy` code, with no modifications, so it would be trivial to review and it would be the the first step for subsequent modifications. Adding a unit test for `proc.go` would involve mocking some OS files operations, and I think that would mess the whole PR...

inercia:
Any thoughts on this @tomwilkie ?

tomwilkie:
I've fought hard to get the code coverage up to ~75%, and I'm reluctant to LGTM any change which drops it this significantly.  Alternatively you could put a new test into procspy to increase coverage, then vendor it here.

Also you haven't addressed this comment:
> And remove the vendored copy of procspy from vendor/, and vendor/manifest.

And I don't think this deservers a top level directory, it should be under probe somewhere.

inercia:
> I've fought hard to get the code coverage up to ~75%, and I'm reluctant to LGTM any change which drops it this significantly. Alternatively you could put a new test into procspy to increase coverage, then vendor it here.

The only reason because it has decreased is because this brings the `procspy` code here, but the code coverage has not changed at all. The only thing that  has changed is _the number_, but the code is covered in the same way as it was before. And adding unit tests in `/procspy` is a waste of time as they will change once the _major surgery_ takes place here...

> Also you haven't addressed this comment:
> > And remove the vendored copy of procspy from vendor/, and vendor/manifest.

Hmmm, I do not see any files left in `/vendor`...

> And I don't think this deservers a top level directory, it should be under probe somewhere.

This could have been suggested when I [stated my intention of moving it to `/procspy`](https://github.com/weaveworks/scope/pull/558#issuecomment-149331113). Where would you prefer exactly to move this to?

tomwilkie:
If you don't want to add the unit test, I will add it.  I'm keen on keeping this number up.

> Hmmm, I do not see any files left in /vendor

You need to rebase to see that (see previous comment "Could you rebase this please? And remove the vendored copy of procspy from vendor/, and vendor/manifest.")

> This could have been suggested when I stated my intention of moving it to /procspy. Where would you prefer exactly to move this to?

See my previous comment "Yes, or even under probe/endpoint."

If you use [gomove](https://github.com/KSubedi/gomove), it should make it much easier.  I haven't tried it myself though.

Under `probe` or `probe/endpoint` are both fine by me.

PR #562, opened 2015-10-14
Sort DNS answers before returning a name with Get()

### LINKED ISSUE #504: Unstable reverse DNS lookups in UI
When viewing the connections of a container talking to servers with multiple A-records, the view switches the displayed server names frequently making it confusing.

I tried to record the example but I wasn't fast enough since the names displayed ended up stabilizing. Will add it later if I see it again.

inercia:
Reated to #504 (it mitigates it) by sorting the DNS answers and getting the first one on `Get()`

### DIFF
--- a/probe/endpoint/resolver.go
+++ b/probe/endpoint/resolver.go
@@ -3,6 +3,7 @@ package endpoint
 import (
 	"fmt"
 	"net"
+	"sort"
 	"strings"
 	"time"
 
@@ -70,6 +71,7 @@ func (r *ReverseResolver) loop() {
 		<-r.Throttle // rate limit our DNS resolutions
 		names, err := r.Resolver(request)
 		if err == nil && len(names) > 0 {
+			sort.Strings(names)
 			name := strings.TrimRight(names[0], ".")
 			r.cache.Set(request, name)
 		} else {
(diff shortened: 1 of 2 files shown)


tomwilkie:
I'm not sure this really fixes the problem; different hosts can have different ideas about what an IP reverse resolves to, and without making the end type a set, the merge will always be nondeterministic.  Does that make sense?

I think we need to make the reverse resolver store all the hostnames for a given IP, return all of them, and define proper merge semantics for this field, ideally using #417

inercia:
@tomwilkie I know this does not fix the problem, that is the reason why this _"Partially fixes #504..."_ :wink: This is a quick fix until we have a proper solution. I assume that, in the most common case, most of the `probe `s will be under the same upstream (caching) DNS server, so they will get the same reverse resolution, and a `sort` on those reverse resolutions should be enough...

But, as you said, a complete solution should follow in a different PR...

tomwilkie:
I'm not sure this bug is urgent enough to warrant such a partial fix; I'd prefer us putting in a full fix.

That being said, I don't have a problem with this change going in, so LGTM.

Be careful with saying 'partially fixes #xxx' and github will just see that as 'fixes #xxx'.

PR #565, opened 2015-10-16
Support wss for secure websockets when page was loaded over ssl

### LINKED ISSUE #563: Client should support SSL for websockets
Should use the same scheme the page was loaded on (tls or not).

paulbellamy:
Fixes #563


client/app/scripts/local.js
```diff
@@ -2,6 +2,7 @@
 // see main.js for the real entrypoint
 
 // Inject websocket url to dev backend
-window.WS_URL = 'ws://' + location.hostname + ':4040';
+window.WS_PROTO = (location.protocol === 'https:' ? 'wss' : 'ws');
+window.WS_URL = window.WS_PROTO + '://' + location.hostname + ':4040';
```
tomwilkie: Also the 4040 won't for the service...

tomwilkie: Actually scratch that, this is just for the local test server.

davkal:
LGTM

PR #566, opened 2015-10-16
Split the internet node for incoming vs outgoing connections.

### LINKED ISSUE #360: Consider differentiating internet node between clients and servers
With the directionality change, the internet node commonly appears at the top (for incoming connections) and bottom (for outgoing connections).  It might help users to duplicate the node, with a minor label for incoming/outgoing.

tomwilkie:
Fixes #360, includes #570

Before this change:

screen shot 2015-10-16 at 17 19 56

After this change:

screen shot 2015-10-16 at 17 23 30

### DIFF
--- a/app/api_topology_test.go
+++ b/app/api_topology_test.go
@@ -167,7 +167,7 @@ func TestAPITopologyWebsocket(t *testing.T) {
 	if err := decoder.Decode(&d); err != nil {
 		t.Fatalf("JSON parse error: %s", err)
 	}
-	equals(t, 7, len(d.Add))
+	equals(t, 8, len(d.Add))
 	equals(t, 0, len(d.Update))
 	equals(t, 0, len(d.Remove))
 }
--- a/integration/300_internet_edge_test.sh
+++ b/integration/300_internet_edge_test.sh
@@ -22,11 +22,11 @@ do_connections() {
 }
 do_connections&
 
-wait_for_containers $HOST1 60 nginx "The Internet"
+wait_for_containers $HOST1 60 nginx "Inbound"
 
 has_container $HOST1 nginx
-has_container $HOST1 "The Internet"
-has_connection containers $HOST1 "The Internet" nginx
+has_container $HOST1 "Inbound"
+has_connection containers $HOST1 "Inbound" nginx
 
 kill %do_connections
 
--- a/integration/config.sh
+++ b/integration/config.sh
@@ -67,7 +67,6 @@ has_connection() {
 	local timeout="${5:-60}"
 	local from_id=$(node_id "${view}" "${host}" "${from}")
 	local to_id=$(node_id "${view}" "${host}" "${to}")
-
 	for i in $(seq $timeout); do
 		local nodes="$(curl -s http://$host:4040/api/topology/${view}?system=show)"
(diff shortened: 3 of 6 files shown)


peterbourgon:
:-1:

- Two internet nodes in the same graph is immediately confusing
- The heuristic of incoming vs. outgoing connections is nonobvious, nondeterministic, and confuses the codebase
- The "fix" to the rendered topology is superficial; I believe we can get better results by continuing to iterate on the UI

pidster:
On the basis that the outbound node will be split into labelled services (somehow) and that the two internet nodes are now visibly separate; the two nodes primary label should be "Inbound" and "Outbound" (connections) - not "The Internet" (inbound/outbound). 

So reversing the title/subtitle labels in effect.

tomwilkie:
Screen shot showing outgoing internet lower than other nodes (David's work):
screen shot 2015-11-10 at 18 03 07

tomwilkie:
Renamed to "Inbound" / "Outbound" Requests:

screen shot 2015-11-10 at 18 03 07

As this has been an contentious PR, I'd like to get an LGTM from @pidster for the concept, and a second LGTM from someone (@Paul?) for the code.

paulbellamy:
The code looks fine. Conceptually I'm far from convinced this is a good idea.

- Inbound/Outbound is a complete guess. Right now we have no way of *actually* telling this.
- `the outbound node will be split into labelled services (somehow)` I'd like to see a proof-of-concept that this is can be done usefully before we make this the goal.

pidster:
LGTM

Sorry for the delay. 
I think we need to make progress and then iterate based on feedback.

tomwilkie:
I've rebased this up to date, but the pinning of the internet nodes doesn't seem to be working properly @davkal:
screen shot 2016-02-29 at 15 14 46

davkal:
- added JS tests, now for the real world...
- outbound node positioning seems to work
- @tomwilkie got a report with incoming nodes to test?
- still thinking about positioning relative to unconnected node field (which can be big, pushing the out node far below)

davkal:
This is the unconnected field I'm talking about:
screen shot 2016-03-01 at 13 10 27

davkal:
Seems to work:
screen shot 2016-03-01 at 13 12 55

Click on Force Re-layout balances layout nicely, too.

davkal:
Inbound node positioning seems to shift nodes, but leave edges untouched. Fixing that...

tomwilkie:
@davkal inbound from the internet to the scope container (will need to show them)

[report.json.zip](https://github.com/weaveworks/scope/files/152786/report.json.zip)

davkal:
screen shot 2016-03-01 at 14 28 56

tomwilkie:
thanks @davkal!  I don't know about you, but I don't think the pinning looks that good:
screen shot 2016-03-01 at 13 42 56

pidster:
I am +1 for splitting The Internet into Outgoing & Incoming per the above. 
See also: https://github.com/weaveworks/scope/issues/1082

I understand that improving the rendering algo that pins the two nodes to the top & bottom respectively, to make a less odd looking diagram will require more work - so let's park that for now.

LGTM without pinning, for now.

PR #571, opened 2015-10-17
Sort the structs returned by /api/topology

### LINKED ISSUE #561: Containers > By Image / By Hostname swaps around
We probably need to return these in the backend in an array rather than a map

tomwilkie:
Fixes #561 

Also dedupe some of of the structs in the app, topologyView -> APITopolgyDesc, optionValue -> APITopologyOption.

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -38,21 +38,21 @@ func makeTopologyList(rep xfer.Reporter) func(w http.ResponseWriter, r *http.Req
 			rpt        = rep.Report()
 			topologies = []APITopologyDesc{}
 		)
-		topologyRegistry.walk(func(name string, def topologyView, subDefs map[string]topologyView) {
+		topologyRegistry.walk(func(def topologyViewAndID, subDefs topologyViewAndIDs) {
 			describedSubDefs := []APITopologyDesc{}
-			for subName, subDef := range subDefs {
+			for _, subDef := range subDefs {
 				describedSubDefs = append(describedSubDefs, APITopologyDesc{
 					Name:    subDef.human,
-					URL:     "/api/topology/" + subName,
-					Options: makeTopologyOptions(subDef),
+					URL:     "/api/topology/" + subDef.id,
+					Options: makeTopologyOptions(subDef.topologyView),
 					Stats:   stats(subDef.renderer, rpt),
 				})
 			}
 			topologies = append(topologies, APITopologyDesc{
 				Name:          def.human,
-				URL:           "/api/topology/" + name,
+				URL:           "/api/topology/" + def.id,
 				SubTopologies: describedSubDefs,
-				Options:       makeTopologyOptions(def),
+				Options:       makeTopologyOptions(def.topologyView),
 				Stats:         stats(def.renderer, rpt),
 			})
 		})
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -123,6 +123,12 @@ type APITopologyDesc struct {
 	Stats         *topologyStats    `json:"stats,omitempty"`
 }
 
+type byName []APITopologyDesc
+
+func (a byName) Len() int           { return len(a) }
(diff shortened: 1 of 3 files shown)


app/api_topologies.go
```diff
@@ -31,6 +139,52 @@ type topologyStats struct {
...
+
+func (r *registry) add(ts ...APITopologyDesc) {
+	r.Lock()
+	defer r.Unlock()
+	for _, t := range ts {
+		t.URL = apiTopologyURL + t.id
+
+		if t.parent != "" {
+			parent := r.items[t.parent]
+			parent.SubTopologies = append(r.items[t.parent].SubTopologies, t)
```
paulbellamy: Probably `append(parent.SubTopologies, t)`, no?

app/api_topologies.go
```diff
@@ -31,6 +139,52 @@ type topologyStats struct {
...
+			sort.Sort(byName(parent.SubTopologies))
+			r.items[t.parent] = parent
+		}
+
+		r.items[t.id] = t
```
paulbellamy: Shouldn't we only do this if it's not a sub-toplogy?

app/api_topologies.go
```diff
@@ -31,6 +139,52 @@ type topologyStats struct {
...
+	for _, desc := range r.items {
+		if desc.parent != "" {
+			continue
+		}
+		descs = append(descs, desc)
+	}
+	sort.Sort(byName(descs))
+	for _, desc := range descs {
+		f(desc)
+	}
```
paulbellamy: Why not `sort.Sort(byName(t.items))` in `add`?

app/api_topologies.go
```diff
@@ -31,6 +139,52 @@ type topologyStats struct {
 	FilteredNodes      int `json:"filtered_nodes"`
 }
 
+type byName []APITopologyDesc
```
paulbellamy: Let's move this up next to APITopologyDesc, to group it with the thing it sorts.

app/api_topologies.go
```diff
@@ -31,6 +139,52 @@ type topologyStats struct {
...
+		descs = append(descs, desc)
+	}
+	sort.Sort(byName(descs))
+	for _, desc := range descs {
+		f(desc)
+	}
+}
+
 // makeTopologyList returns a handler that yields an APITopologyList.
 func makeTopologyList(rep xfer.Reporter) func(w http.ResponseWriter, r *http.Request) {
```
paulbellamy: This would be cleaner as a method on `registry`.

app/api_topologies.go
tomwilkie: Ah yes, good catch.

app/api_topologies.go
tomwilkie: Unfortunately not; sub topologies are still looked up by id (they are not nested in their URL).

app/api_topologies.go
tomwilkie: Add puts them in a map, not a list.

I considered having add put them in a list as well, and keeping that sorted, but it wasn't clear to me if the k8s topologies were added multiple times, and if add was supposed to dedupe.

app/api_topologies.go
tomwilkie: Can do.  I put it there to be near the functions, and typically we don't like to mix functions and type definitions.

app/api_topologies.go
tomwilkie: Will do.

app/api_topologies.go
paulbellamy: Ah, yeah.

app/api_topologies.go
paulbellamy: I meant move the whole `byName` and it's fns. Either way, though.

app/api_topologies.go
tomwilkie: Discussed offline, not worth it.

tomwilkie:
@paulbellamy PTAL

paulbellamy:
LGTM

PR #574, opened 2015-10-19
Make scope work with weave master.

### LINKED ISSUE #560: Scope can't talk to master weave
Probably due to FDP merge putting the weave router in the host netns

```
vagrant@ubuntu-14:~/src/github.com/weaveworks/scope$ docker logs weavescope
weave container has no IP address; is Docker networking enabled?
weave container has no IP address; is Docker networking enabled?
Weave container detected at , Docker bridge at 172.17.42.1
weave container has no IP address; is Docker networking enabled?
weave container has no IP address; is Docker networking enabled?
```

tomwilkie:
Fixes #560

- Don't use the embedded weave script to drive dns or expose; run the appropriate weaveexec image instead
- Deal with weave being in the host net namespace
- Embedded weave script still needed to get container ips.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -15,20 +15,21 @@ DOMAIN=weave.local
 IP_REGEXP="[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}"
 
 container_ip() {
-    if ! status=$(docker inspect --format='{{.State.Running}} {{.NetworkSettings.IPAddress}}' $1 2>/dev/null); then
-        echo "$2" >&2
+    if ! status=$(docker inspect --format='{{.State.Running}} {{.HostConfig.NetworkMode}}' $1 2>/dev/null); then
+        echo "Container $1 not found" >&2
         return 1
     fi
     case "$status" in
-        "true ")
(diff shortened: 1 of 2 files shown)


tomwilkie:
Current problem is we need to run weave ps periodically to tag containers with weave ips for conntrack.

This PR has removed the weave script, and we don't want to run the weaveexec container every few seconds.

docker/entrypoint.sh
```diff
@@ -15,20 +15,21 @@ DOMAIN=weave.local
...
-        echo "$2" >&2
+    if ! status=$(docker inspect --format='{{.State.Running}} {{.HostConfig.NetworkMode}}' $1 2>/dev/null); then
+        echo "Container $1 not found" >&2
         return 1
     fi
     case "$status" in
-        "true ")
-            echo "$1 container has no IP address; is Docker networking enabled?" >&2
-            return 1
+        "true host")
```
paulbellamy: What about `"true container:"*`?

paulbellamy:
Question about the container_ip case, but other than that, LGTM.

docker/entrypoint.sh
tomwilkie: weave will never be run like that, so I didn't think it was necessary.

paulbellamy: Fine.

PR #575, opened 2015-10-19
Add ssl support for the probe

### LINKED ISSUE #559: Probe posting should follow redirects

paulbellamy:
Hosts ending in `:443` are treated as SSL.

There is also a `--probe.insecure` flag, for using it in development
environments where SSL cannot be verified.

Probably needs a unit test for the HTTPPublisher written. Will do that if the code/concept looks ok.

Fixes #559 in a way.


xfer/http_publisher.go
```diff
@@ -15,24 +17,39 @@ type HTTPPublisher struct {
...
-func NewHTTPPublisher(target, token, probeID string) (string, *HTTPPublisher, error) {
+func NewHTTPPublisher(target, token, probeID string, insecure bool) (string, *HTTPPublisher, error) {
+	_, port, err := net.SplitHostPort(target)
+	if err != nil {
+		return "", nil, err
+	}
+	scheme := "http"
+	if port == "443" {
+		scheme = "https"
+	}
```
tomwilkie: Should we push this logic into sanitise.URL?

tomwilkie:
Do we have the certificate bundle installed in the scope container for this to work?

paulbellamy:
Oh, for an authority to validate certs. Probably not.

xfer/http_publisher.go
paulbellamy: yes, probably, actually.

tomwilkie:
LGTM

Issue #520, opened 2015-09-25
Edges showing where they shouldn't


tomwilkie:
See screen shot:

<img width="1792" alt="screen shot 2015-09-25 at 12 59 44" src="https://cloud.githubusercontent.com/assets/444037/10093474/671898cc-6385-11e5-9b26-6f9be8a5745c.png">

Edge between scope & app should not be there.  Its hard to reproduce and only happens occasionally

squaremo:
I see this in my example app. This may help reproduce it (assumes you are using the weave API proxy, and Scope):

```bash
git clone https://github.com/squaremo/micro-wiki
cd micro-wiki
make -C pages
make -C conf
make -C preso
./microwikictl up
# will probably download a lot of things
./microwikictl load
# now go watch Scope
```

NB it's all quite rickety but feel free to ask for help :)

tomwilkie:
@squaremo I get an error:
```
vagrant@ubuntu-14:~/src/github.com/squaremo/micro-wiki$ make -C pages
make: Entering directory `/home/vagrant/src/github.com/squaremo/micro-wiki/pages'
npm install
npm WARN package.json micro-wiki-pages-service@0.0.3 No README data
npm ERR! git clone git@github.com:file:../us Cloning into bare repository '/home/vagrant/.npm/_git-remotes/git-github-com-file-us-d1dcada0'...
npm ERR! git clone git@github.com:file:../us fatal: remote error: 
npm ERR! git clone git@github.com:file:../us    is not a valid repository name
npm ERR! git clone git@github.com:file:../us   Email support@github.com for help
npm ERR! notarget No compatible version found: us@'file:../us'
npm ERR! notarget Valid install targets:
npm ERR! notarget ["1.0.0","1.0.1","1.0.2","1.0.3","1.1.0","1.1.1"]
npm ERR! notarget 
npm ERR! notarget This is most likely not a problem with npm itself.
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.

npm ERR! System Linux 3.13.0-24-generic
npm ERR! command "/usr/bin/node" "/usr/bin/npm" "install"
npm ERR! cwd /home/vagrant/src/github.com/squaremo/micro-wiki/pages
npm ERR! node -v v0.10.40
npm ERR! npm -v 1.4.28
npm ERR! code ETARGET
npm ERR! 
npm ERR! Additional logging details can be found in:
[...]

PR #577, opened 2015-10-20
Include host scope for container joins based on IP.

tomwilkie:
Part of #520 - can explain false edges between containers on different hosts, but the bug has been seen between containers on the same host, and this shouldn't affect that.

### DIFF
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -22,12 +22,13 @@ import (
 
 // These constants are keys used in node metadata
 const (
-	ContainerName     = "docker_container_name"
-	ContainerCommand  = "docker_container_command"
-	ContainerPorts    = "docker_container_ports"
-	ContainerCreated  = "docker_container_created"
-	ContainerIPs      = "docker_container_ips"
-	ContainerHostname = "docker_container_hostname"
+	ContainerName          = "docker_container_name"
+	ContainerCommand       = "docker_container_command"
+	ContainerPorts         = "docker_container_ports"
+	ContainerCreated       = "docker_container_created"
+	ContainerIPs           = "docker_container_ips"
+	ContainerHostname      = "docker_container_hostname"
+	ContainerIPsWithScopes = "docker_container_ips_with_scopes"
 
 	NetworkRxDropped = "network_rx_dropped"
 	NetworkRxBytes   = "network_rx_bytes"
@@ -72,7 +73,7 @@ type Container interface {
 	Image() string
 	PID() int
 	Hostname() string
-	GetNode([]net.IP) report.Node
+	GetNode(string, []net.IP) report.Node
 
 	StartGatheringStats() error
 	StopGatheringStats()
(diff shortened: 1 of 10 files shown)


probe/docker/container.go
```diff
@@ -219,20 +220,28 @@ func (c *container) ports(localAddrs []net.IP) string {
 	return strings.Join(ports, ", ")
 }
 
-func (c *container) GetNode(localAddrs []net.IP) report.Node {
+func (c *container) GetNode(hostID string, localAddrs []net.IP) report.Node {
 	c.RLock()
 	defer c.RUnlock()
 
+	ips := append(c.container.NetworkSettings.SecondaryIPAddresses,
+		c.container.NetworkSettings.IPAddress)
```
peterbourgon: Either

```go
ips := append(c.container.NetworkSettings.SecondaryIPAddresses, c.container.NetworkSettings.IPAddress)
```

or

```go
ips := append(
	c.container.NetworkSettings.SecondaryIPAddresses,
	c.container.NetworkSettings.IPAddress,
)
```

probe/docker/container.go
```diff
@@ -219,20 +220,28 @@ func (c *container) ports(localAddrs []net.IP) string {
...
+	// Treat all Docker IPs as local scoped.
+	ipsWithScopes := []string{}
+	for _, ip := range ips {
+		ipsWithScopes = append(ipsWithScopes, fmt.Sprintf("%s:%s", hostID, ip))
```
peterbourgon: We have ID constructors for operations like this; can you use one?

probe/docker/container.go
```diff
@@ -268,3 +277,9 @@ func (c *container) GetNode(localAddrs []net.IP) report.Node {
 func ExtractContainerIPs(nmd report.Node) []string {
 	return strings.Fields(nmd.Metadata[ContainerIPs])
 }
+
+// ExtractContainerIPsWithScopes returns the list of container IPs, prepended
+// with scopes, given a Node from the Container topology.
+func ExtractContainerIPsWithScopes(nmd report.Node) []string {
+	return strings.Fields(nmd.Metadata[ContainerIPsWithScopes])
+}
```
peterbourgon: Likewise above, we have ID parsers instead of falling back to raw strings.Fields...

render/mapping.go
```diff
@@ -283,23 +283,22 @@ func MapEndpoint2IP(m RenderableNode, local report.Networks) RenderableNodes {
...
-	// Emit addr:port nodes as well, so connections from the internet to containers
-	// via port mapping also works.
-	port, ok := m.Metadata[endpoint.Port]
-	if ok {
-		id := fmt.Sprintf("%s:%s", addr, port)
-		result[id] = NewRenderableNodeWith(id, "", "", "", m)
+	// Emit 2 nodes: scope:addr and scope:addr:port, so connections from the
+	// internet to containers via port mapping also works.
+	id := fmt.Sprintf("%s:%s", scope, addr)
+	idWithPort := fmt.Sprintf("%s:%s:%s", scope, addr, port)
```
peterbourgon: Likewise above, can you use constructors?

render/mapping.go
```diff
@@ -283,23 +283,22 @@ func MapEndpoint2IP(m RenderableNode, local report.Networks) RenderableNodes {
...
+	return RenderableNodes{
+		id:         NewRenderableNodeWith(id, "", "", "", m),
+		idWithPort: NewRenderableNodeWith(idWithPort, "", "", "", m),
```
peterbourgon: Why do you need both? Isn't a scoped-X strictly more useful than, and usable in the place of, X alone?

render/mapping.go
```diff
@@ -309,20 +308,20 @@ var portMappingMatch = regexp.MustCompile(`([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.
...
 	}
 
-	// also output all the host:port port mappings
+	// Also output all the host:port port mappings.
+	// In this case, we assume this doesn't need a scope,
+	// as they are for host IPs.
 	for _, mapping := range portMappingMatch.FindAllStringSubmatch(m.Metadata[docker.ContainerPorts], -1) {
 		ip, port := mapping[1], mapping[2]
-		id := fmt.Sprintf("%s:%s", ip, port)
+		id := fmt.Sprintf(":%s:%s", ip, port)
```
peterbourgon: Likewise above, please use a constructor...

render/mapping.go
tomwilkie: We don't always know what port a container is listening on, and container-to-container communications can be unambiguously identified without ports.  OTOH, connections to the host IPs which have been port mapped to a container can only be unambiguously identified with the port.  So we need both, for two different cases.

peterbourgon: I understand. Thanks for the explanation!

peterbourgon:
LGTM.

render/mapping.go
tomwilkie: I should probably put it in a comment...

tomwilkie:
Crap; there is actually a good reason to use the id generators - they embed the logic for whether a scope is appended or not, and really are only for use within the probe.

Lets see if can factor this out.

tomwilkie:
```MakeEndpointNodeIDAlwaysWithScope``` seems a little verbose; any better suggestions?

peterbourgon:
MakeScopedEndpointNodeID

or the general form

Make[Scoped]_Topology_NodeID

tomwilkie:
Okay done.  Hopefully this should all pass now.

PR #583, opened 2015-10-22
Build inside the build container, using go1.5, by default

tomwilkie:
Fixes #550, fixes #530

And build for GOOS=darwin and GOARCH=arm

With this change, `make` should build the scope container on a correctly configured Mac, using Docker Machine.

PR #585, opened 2015-10-23
[WIP] Make tools a subtree so we can control what version we use.

tomwilkie:
I'll put instructions on how to update it in the tools repo.


tomwilkie:
This got included as part of #583

Issue #419, opened 2015-09-01
"Dancing" nodes


2opremio:
I recall someone addressed this one before   ( @tomwilkie ? ) but we are still suffering from it. The whole layout changes when a new container is spawned.

The following gif was recorded from the Scope-variant of the ECS guide I am working on ( You can reproduce it by following https://github.com/weaveworks/guides/tree/aws-ecs-scope/aws-ecs ).

![Dancing containers](http://g.recordit.co/hXSMIHPSVe.gif)

Context: A new container is being spawned every 2 seconds due to  `weave status` being called repeatedly in from an Upstart script (to see whether Weave died and re-launch it). For the time being I will workaround the problem by checking Weave's status differently but we should still fix this.

peterbourgon:
Another end-run around this problem is using the "hide system containers" option we're working on in #420.

davkal:
We're currently using the dagre library which gives ok results for layered graphs. It's perfect for single runs, but sub-optimal for changing graphs, e.g. adding a single node may change your edge cross count [which might lead to a different layer matrix being chosen](https://github.com/cpettitt/dagre/blob/master/lib/order/index.js#L43). There are no options to configure this.

Apart from writing our own library or finding a different one that suits us better, an approach with dagre (which is no longer maintained) could be to only do single runs per topology, and then implement our own transitions for adding/removing nodes.

davkal:
A previous attempt was made in #379 by reusing chart engine instances, but that wont help if a change in the topology causes a heuristic to choose a different layout.

2opremio:
@davkal Sorry if this question is naive, but have you looked at the other graph libraries? This github project seems to provide a pretty good summary of what's out there: https://github.com/anvaka/graph-drawing-libraries

2opremio:
It actually happens even when no new containers are seemingly added. This is also a gif from the ECS demo, once I modified it to avoid creating new containers:

![More dancing containers](http://g.recordit.co/S35ssg13LQ.gif)

2opremio:
And it also happens periodically in the hosts view, as @rade pointed out:

![hosts cha cha cha](http://g.recordit.co/LZI6qpu3T2.gif)

paulbellamy:
It's particularly bad on a redis cluster. They ping each other every few seconds, which sometimes short-lived connections catch, and sometimes not.

paulbellamy:
[...]

PR #593, opened 2015-10-26
Mitigate dancing nodes

davkal:
This is to mitigate the dancing node problem in #419 

Heuristics to implement towards iterative layout engine (minimize re-layouts):

* dont layout on edge removal
* dont layout on edge adding
* dont layout on node removal
* relayout on node adding
* track degradation of layouts via scoring (edge crossings, wrong edge direction) and only re-layout when exceeding a threshold

Be smart about node adding:
* track recent layouts for reuse of previous node locations of same node (instead of layout on node adding) 
* add node to 1st neighbour and stretch rest on same level

### DIFF
--- a/client/app/scripts/charts/__tests__/node-layout-test.js
+++ b/client/app/scripts/charts/__tests__/node-layout-test.js
@@ -1,0 +1,50 @@
+jest.dontMock('../nodes-layout');
+jest.dontMock('../../constants/naming'); // edge naming: 'source-target'
+
+describe('NodesLayout', () => {
+  const NodesLayout = require('../nodes-layout');
+
+  function scale(val) {
+    return val * 3;
+  }
+  const topologyId = 'tid';
+  const width = 80;
+  const height = 80;
+  const margins = {
+    left: 0,
+    top: 0
+  };
+
+  const nodeSets = {
+    initial4: {
+      nodes: {
+        n1: {id: 'n1'},
+        n2: {id: 'n2'},
+        n3: {id: 'n3'},
+        n4: {id: 'n4'}
+      },
+      edges: {
+        'n1-n3': {id: 'n1-n3', source: {id: 'n1'}, target: {id: 'n3'}},
+        'n1-n4': {id: 'n1-n4', source: {id: 'n1'}, target: {id: 'n4'}},
+        'n2-n4': {id: 'n2-n4', source: {id: 'n2'}, target: {id: 'n4'}}
+      }
+    }
+  };
+
+  it('lays out initial nodeset', () => {
+    const nodes = nodeSets.initial4.nodes;
+    const edges = nodeSets.initial4.edges;
+    NodesLayout.doLayout(nodes, edges, width, height, scale, margins, topologyId);
+    expect(nodes.n1.x).toBeLessThan(nodes.n2.x);
+    expect(nodes.n1.y).toEqual(nodes.n2.y);
+
+    expect(nodes.n1.x).toEqual(nodes.n3.x);
+    expect(nodes.n1.y).toBeLessThan(nodes.n3.y);
+
+    expect(nodes.n3.x).toBeLessThan(nodes.n4.x);
+    expect(nodes.n3.y).toEqual(nodes.n4.y);
+
+    console.log(nodes, nodeSets.initial4.nodes);
+  });
+
+});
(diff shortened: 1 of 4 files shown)


paulbellamy:
All seem reasonable. My only concern is that, while this will lead to fewer re-layouts, the re-layouts it does do will be bigger.  i.e. It will have added and removed a bunch of edges, removed some nodes, and not done a re-layout. Then when a new node comes along it will trigger a massive (confusing) re-layout, however the the degradataion scoring could probably help mitigate this.

IMO, degradation should be tracked strictly as a diff against the previous layout (or data-set). Tracking it as an objective score of edge-crossings, etc, could behave badly on layouts which are difficult to render (i.e. it will keep trying to lay it out, all with bad results). Would it be possible to compare the current layout's score, with the potential re-layout's score, and only apply the re-layout if the new score is "much higher"?

Also, IMO, (as mentioned on #419) it's not strictly an issue of nodes moving, as much as unpredictability.

davkal:
Simple heuristics are implemented. Once caveat with reusing old node locations is that they may overlap with locations of a different layout run. Reusing whole sets of old node locations wont get very far though because they only apply if the node sets are the same. Some postprocessing may be needed to spread out overlapping nodes.
Hard numbers: with a probe running on my mac for 10 minutes, 55 of 123 were layed out with the heuristic and prevented jumping. That's 44%.

davkal:
> All seem reasonable. My only concern is that, while this will lead to fewer re-layouts, the re-layouts it does do will be bigger. i.e. It will have added and removed a bunch of edges, removed some nodes, and not done a re-layout. Then when a new node comes along it will trigger a massive (confusing) re-layout, however the the degradataion scoring could probably help mitigate this.

There is no perceived correlation between the extent of node/edge changes and the subsequent layout change. A single edge can throw everything around.

> IMO, degradation should be tracked strictly as a diff against the previous layout (or data-set). Tracking it as an objective score of edge-crossings, etc, could behave badly on layouts which are difficult to render (i.e. it will keep trying to lay it out, all with bad results). Would it be possible to compare the current layout's score, with the potential re-layout's score, and only apply the re-layout if the new score is "much higher"?

Good idea with diffing the score against a re-layout.

tomwilkie:
I can't really follow the code I'm afraid; let me know if you want me to try harder.

I have tested on the dev env and it dramatically reduces dancing.  So massive +1 from me there.

Local testing shows there is some strange behaviour when selecting and unselecting nodes; the selected node doesn't always get centered properly, and the canvas doesn't always go back to where I'd expect.  Perhaps this is unrelated?

client/app/scripts/charts/__tests__/node-layout-test.js
```diff
@@ -0,0 +1,238 @@
...
+    nodes = result.nodes.toJS();
+
+    expect(nodes.n1.x).toBeLessThan(nodes.n2.x);
+    expect(nodes.n1.y).toEqual(nodes.n2.y);
+    expect(nodes.n1.x).toEqual(nodes.n3.x);
+    expect(nodes.n1.y).toBeLessThan(nodes.n3.y);
+    expect(nodes.n3.x).toBeLessThan(nodes.n4.x);
+    expect(nodes.n3.y).toEqual(nodes.n4.y);
+  });
+
```
davkal: Test exact positions against the first result.

client/app/scripts/charts/nodes-chart.js
```diff
@@ -27,8 +29,9 @@ const NodesChart = React.createClass({
 
   getInitialState: function() {
     return {
-      nodes: {},
-      edges: {},
+      nodes: makeMap(),
+      edges: makeMap(),
+      history: makeList(),
```
davkal: Not used.

client/app/scripts/charts/nodes-chart.js
```diff
@@ -166,7 +190,7 @@ const NodesChart = React.createClass({
       wasShifted = true;
     }
     const errorClassNames = this.state.maxNodesExceeded ? 'nodes-chart-error' : 'nodes-chart-error hide';
-    const svgClassNames = this.state.maxNodesExceeded || _.size(nodeElements) === 0 ? 'hide' : '';
+    const svgClassNames = this.state.maxNodesExceeded || nodeElements.size === 0 ? 'hide' : '';
 
     return (
       <div className="nodes-chart">
```
davkal: Replace `'` with a unicode character that doesnt mess up syntax highlighting.

tomwilkie:
Did a review on hangouts with david, and other than the comments he just added this LGTM.

PR #594, opened 2015-10-26
[docs] Refer to different Scope setups as standalone and service modes

2opremio:
@pidster @kpoplin @abuehrle @mjlodge PTAL

This stems from the internal terminology discussion. It will make the Scope documentation consistent with https://github.com/weaveworks/integrations/pull/27 (which is awaiting for terminology consistency to be merged).

Both @abuehrle and I agree that *standalone mode* is more intuitive than *multi-host setup*. We also agree on this being ultimately a product/marketing decision.

CC: @peterbourgon @tomwilkie @paulbellamy

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ serving the UI, and pushing these topologies to the UI.
 +-----------------------+
 ```
 
-## Multi host setup
+## Standalone Mode
 
 When running Scope in a cluster, each probe sends reports to each app.
 The App merges the reports from each probe into a more complete report.
@@ -99,7 +99,7 @@ acceptable, both with and without ports:
 Hostnames will be regularly resolved as A records, and each answer used as a
 target.
 
-## Using Scope Service
+## Service Mode
 
 Scope can also be used to feed reports to the Scope Service. The Scope Service
 allows you centrally manage and share access to your Scope UI. In this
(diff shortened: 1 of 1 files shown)


kpoplin:
I agree with the standalone/hosted mode use cases for the single product called Weave Scope.

I think the section titles could be even more clear, i.e. “Using Weave Scope in Hosted Mode” and “Using Weave Scope in Standalone Mode”

(Dumb question but the "setup" in the older "Multi-host setup" doesn't mean the section includes setup instructions correct? If it did, I'd suggest adding that to the above titles, like "Using Weave Scope in Hosted Mode: Setup Instructions"...)

@2opremio @abuehrle @pidster @mjlodge what do you guys think?

mjlodge:
Agree with @kpoplin that Standalone and Service make sense as terms.

2opremio:
> I think the section titles could be even more clear, i.e. “Using Weave Scope in Hosted Mode” and “Using Weave Scope in Standalone Mode”

I have updated the PR to reflect this.

> (Dumb question but the "setup" in the older "Multi-host setup" doesn't mean the section includes setup instructions correct? If it did, I'd suggest adding that to the above titles, like "Using Weave Scope in Hosted Mode: Setup Instructions"...)

It doesn't really include setup instructions, no.

PR #596, opened 2015-10-27
pass the hostname through to xfer.HttpPublisher, so we can verify ssl certs

paulbellamy:


probe/main.go
```diff
@@ -90,8 +90,8 @@ func main() {
 	}
 	log.Printf("publishing to: %s", strings.Join(targets, ", "))
 
-	factory := func(endpoint string) (string, xfer.Publisher, error) {
-		id, publisher, err := xfer.NewHTTPPublisher(endpoint, *token, probeID, *insecure)
+	factory := func(endpoint, hostname string) (string, xfer.Publisher, error) {
```
tomwilkie: A bit of a nit, but other method go (hostname, endpoint) - mind changing?

probe/main.go
```diff
@@ -90,8 +90,8 @@ func main() {
...
+		id, publisher, err := xfer.NewHTTPPublisher(endpoint, hostname, *token, probeID, *insecure)
```
tomwilkie: Ditto

tomwilkie:
Just one nit, otherwise LGTM (assuming circle is happy)

PR #598, opened 2015-10-27
Add controls form container start, stop, pause, etc

### LINKED ISSUE #315: Add controls

tomwilkie:
Add control interface for Scope.  Allows UI to do RPCs all the way back to the probe, with the app doing the routing in the middle.  Add metadata in the API so these controls can be appropriately rendered, and add some basic support in the UI.

This is a very hacky, very early implementation.  Fixes #315 (see bug for some design notes).

Also:
- Docker integration adds controls for container lifecycle
- include non-running docker containers in the report, dynamically update docker container states
- add a filter to filter out non-running containers

Todo
- [x] Make it lint & pass tests
- [x] Try and use bits and bobs from golang rpc
- [x] Separate control http mux/router in the app
- [x] Add tests for capabilities
- [x] Reorganise change sets.
- [x] Make mutli client delete old apps
- [x] Document / comment code
- [x] Integration test
- [x] Make UI reflect result of control RPC (@davkal?)

Punted out of this PR:
- #628 LatestMap on reports (cause of multiple states and delays)
- #629 State updates (for containers) could trigger earlier report pushes (suggestion from demo)
- #630 Consider merging AppClient and HTTP Publisher.


client/app/scripts/components/node-details-table.js
```diff
@@ -11,6 +12,15 @@ const NodeDetailsTable = React.createClass({
           {this.props.title}
         </h4>
 
+
+        <div className="node-details-table-row">
+        {this.props.capabilities.map(function(capability) {
```
davkal:
```diff
--- a/client/app/scripts/components/node-details-table.js
+++ b/client/app/scripts/components/node-details-table.js
@@ -14,7 +14,7 @@ const NodeDetailsTable = React.createClass({
 
 
         <div className="node-details-table-row">
-        {this.props.capabilities.map(function(capability) {
+        {this.props.capabilities && this.props.capabilities.map(function(capability) {
           return (
             <NodeCapabilityButton capability={capability} />
           );
```

probe/capabilities/capabilities.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/weaveworks/scope/xfer"
+)
+
+var (
+	mtx      = sync.Mutex{}
+	handlers = map[string]xfer.CapabilityHandler{}
+)
+
+// HandleCapabilityRequest performs a capability request.
```
paulbellamy: this comment is stupid.

probe/docker/capabilities.go
```diff
@@ -0,0 +1,20 @@
...
+
+	"github.com/weaveworks/scope/probe/capabilities"
+	"github.com/weaveworks/scope/xfer"
+)
+
+// RestartContainer is the capability ID used by the docker intergation.
+const RestartContainer = "docker_restart_container"
+
+func init() {
+	capabilities.Register(RestartContainer, func(req xfer.Request) xfer.Response {
```
paulbellamy: In general it's probably better form to pull the func out, so we can unit-test it. Is there such a thing as a capability that will need initialization at (or after) probe boot time?

xfer/app_client.go
```diff
@@ -0,0 +1,154 @@
...
+				log.Printf("Error parsing message from %s: %v", err)
+				return
+			}
+
+			incoming <- request
+		}
+	}(conn)
+
+	go func(c *websocket.Conn) {
+		defer close(quit)
```
paulbellamy: one of these closes will panic.

xfer/app_client.go
```diff
@@ -0,0 +1,154 @@
...
+	defer resp.Body.Close()
+	return result, json.NewDecoder(resp.Body).Decode(&result)
+}
+
+func (c *appClient) Capabilities(handler CapabilityHandler) error {
+	dialer := websocket.Dialer{}
+	headers := http.Header{}
+	c.authoriseHeaders(headers)
+	// TODO(twilkie) need to update sanitize to work with wss
+	url := sanitize.URL("ws://", 0, "/api/capabilities")(c.target)
```
paulbellamy: will need to handle `wss://` eventually.

xfer/app_client.go
```diff
@@ -0,0 +1,154 @@
...
+				log.Printf("Error sending response to %s: %v", err)
+				return
+			}
+		}
+	}(conn)
+
+	for {
+		select {
+		case req := <-incoming:
+			// TODO process more than one request at once.
```
paulbellamy: Not sure this is worth it. Will lead to some race conditions, and having the commands could be serialized kind of nice.

xfer/multi_client.go
```diff
@@ -0,0 +1,102 @@
...
+			_, ok := c.clients[details.ID]
+			if !ok {
+				c.clients[details.ID] = client
+				go c.capabilityLoop(endpoint, client)
+			}
+		}(endpoint)
+	}
+}
+
+func (c *multiClient) capabilityLoop(endpoint string, client AppClient) {
```
paulbellamy: I feel like most of this should be pushed into the AppClient.

paulbellamy:
Do we need to implement our own jsonrpc instead of using: https://golang.org/pkg/net/rpc/jsonrpc/#NewClientCodec ?

paulbellamy:
Just curious how you see the request/response model working with longer-running (interactive) capabilities, like shell? Bridging through a websocket to the user's browser? or separate request/response for each chunk of data?

paulbellamy:
Also, if it will be called controls, let's call it that everywhere.

app/capabilities.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/gorilla/websocket"
+
+	"github.com/weaveworks/scope/report"
+	"github.com/weaveworks/scope/xfer"
+)
+
+const capabilityKey = "capability"
+
+// CapabilityRouter handles the impedence mismatch between posted Capability calls,
+// and websocket connections.
```
paulbellamy: What is the impedance mismatch?

app/capabilities.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+type capabilityRouter struct {
+	sync.Mutex
+	quit   chan struct{}
+	probes map[string]xfer.CapabilityHandler
+}
+
+// NewCapabilitiyRouter makes a new CapabilitiyRouter
+func NewCapabilitiyRouter() CapabilityRouter {
```
paulbellamy: typo

client/app/scripts/components/node-details-table.js
tomwilkie: Done, thanks!

probe/docker/capabilities.go
tomwilkie: > In general it's probably better form to pull the func out, so we can unit-test it.

Ack, done.

>  Is there such a thing as a capability that will need initialization at (or after) probe boot time?

Not that I can think of.

xfer/app_client.go
tomwilkie: Yeah needs more thoughts.  The request / responses can be interleaved on the wire, so thats not a problem.  Otherwise yeah, handling multiple requests at once might be a challenge.

xfer/multi_client.go
tomwilkie: Could do yeah; layering is not 100% clear right now.

tomwilkie:
> Also, if it will be called controls, let's call it that everywhere.

Agreed; have renamed everything.

tomwilkie:
> Just curious how you see the request/response model working with longer-running (interactive) capabilities, like shell?

I don't think we can build shell / logs with this just yet.  I think we need to layer some more abstraction/plumbing on top first.

I'm currently thinking that a request could return a 'pipe'  instead a response.  The pipe would have a different lifecycle, and use different messages on the same websocket.  We might be able to use app<->client websocket too.

Anyway I suggest we get the short-lived RPCs bit (start, stop, restart etc) right first, then think about logs & terminals.

app/controls.go
```diff
@@ -0,0 +1,106 @@
...
+
+// HandleControl ...
+func (cr *controlRouter) HandleControl(w http.ResponseWriter, r *http.Request) {
+	var (
+		vars    = mux.Vars(r)
+		probeID = vars["probeID"]
+		nodeID  = vars["nodeID"]
+		control = vars["control"]
+	)
+	handler, ok := cr.probes[probeID]
```
paulbellamy: Probably missing a `Lock()` here.

tomwilkie: Thanks!

app/router.go
```diff
@@ -75,8 +75,13 @@ func Router(c collector) *mux.Router {
 	get.MatcherFunc(URLMatcher("/api/origin/host/{id}")).HandlerFunc(
 		gzipHandler(makeOriginHostHandler(c)))
 	get.HandleFunc("/api/report", gzipHandler(makeRawReportHandler(c)))
+	get.HandleFunc("/api/controls", controlRouter.HandleProbeWS)
 	get.PathPrefix("/").Handler(http.FileServer(FS(false))) // everything else is static
 
+	post := router.Methods("POST").Subrouter()
+	post.HandleFunc("/api/report", makeReportPostHandler(c)).Methods("POST")
+	post.MatcherFunc(URLMatcher("/api/control/{probeID}/{nodeID}/{control}")).HandlerFunc(
+		controlRouter.HandleControl)
```
paulbellamy: Might separate concerns a bit to invert the registration? Not sure...

```Go
controlRouter.RegisterRoutes(router.PathPrefix("/api/control").Subrouter())
```
or even
```Go
NewControlRouter(router.PathPrefix("/api/control").Subrouter())
```

app/controls.go
```diff
@@ -0,0 +1,106 @@
...
+
+	cr.Lock()
+	cr.probes[probeid] = func(req xfer.Request) xfer.Response {
+		var res xfer.Response
+		client.Call("control.Handle", req, &res)
+		return res
+	}
+	cr.Unlock()
+
+	defer func() {
```
paulbellamy: why does this need to be a defer? why not just do it after the `CloseNotify`?

app/controls.go
```diff
@@ -0,0 +1,106 @@
...
+		cr.Lock()
+		defer cr.Unlock()
+		delete(cr.probes, probeid)
+		client.Close()
+	}()
+
+	notify := w.(http.CloseNotifier).CloseNotify()
```
paulbellamy: What if this isn't a `CloseNotifier`?

paulbellamy:
Would it make sense to have this? i.e. treating controls as a "plugin" with app and probe components.
```
.
└── controls
    ├── app
        └── *.go
    └── probe
        └── *.go
```

Thinking about it, they are fairly tightly integrated with the docker probe plugin and everything else, so maybe this doesn't make sense...

app/controls.go
tomwilkie: Then its a programming error, and it should panic.

app/controls.go
tomwilkie: Can do.

tomwilkie:
re: treating as a plugin; as you say, the code it pretty tightly integrated, so I think the division you suggest would be pretty artificial.

I'm trying to do the following:
- have the common bits (ie on the wire format, codecs etc) in xfer/
- have the app specific bits (accepting a control from the UI, routing it to the right probe) in the app/
- have the probe specific bits (handling the control, routing to appropriate handler) in probe/

The bit that doesn't fit it the probe's connection to the app logic, which I've put in xfer/, as thats where the http publisher are.  On could argue this belongs in probe/, but I'm not too fussed either way.

app/controls.go
tomwilkie: Actually there is a race here; if a probe reconnects before the close notify drops the last connection, we'll end up deleting the new connection from the map, not the old one.

probe/docker/controls.go
```diff
@@ -0,0 +1,70 @@
...
+	if !ok {
+		return xfer.ResponseErrorf("Invalid ID: %s", req.NodeID)
+	}
+
+	container, ok := r.getContainer(containerID)
+	if !ok {
+		return xfer.ResponseErrorf("Invalid ID: %s", req.NodeID)
+	}
+
+	return xfer.ResponseError(container.Stop())
```
paulbellamy: I would merge these into:
```Go
return xfer.ResponseError(r.stopContainer(containerID))
```

Which keeps knowledge of the client in the registry.

probe/docker/container.go
```diff
@@ -77,18 +85,32 @@ type Container interface {
...
+	client      Client
 	container   *docker.Container
 	statsConn   ClientConn
 	latestStats *docker.Stats
 }
 
 // NewContainer creates a new Container
-func NewContainer(c *docker.Container) Container {
-	return &container{container: c}
+func NewContainer(client Client, c *docker.Container) Container {
```
paulbellamy: Don't think any of these need to be public, right?

app/controls.go
paulbellamy: Or it's a tls connection, possibly.

app/controls.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	conn, err := upgrader.Upgrade(w, r, nil)
+	if err != nil {
+		log.Printf("Error upgrading to websocket: %v", err)
+		return
+	}
+	defer conn.Close()
+
+	codec := xfer.NewJSONWebsocketCodec(conn)
```
paulbellamy: Superficially, the codec and rpc wiring stuff seem more closely related to `xfer`, than to `app`. Maybe some of this could be better-encapsulated there.

xfer/multi_client.go
```diff
@@ -0,0 +1,104 @@
...
+		quit:    make(chan struct{}),
+	}
+}
+
+// Set ...
+func (c *multiClient) Set(hostname string, endpoints []string) {
+	for _, endpoint := range endpoints {
+		go func(endpoint string) {
+			c.sema.p()
+			defer c.sema.v()
```
paulbellamy: these method names are terrible.

probe/docker/container.go
tomwilkie: For the tests, _sigh_

probe/docker/controls.go
tomwilkie: roger that.

app/controls.go
tomwilkie: Fixed now.

xfer/multi_client.go
tomwilkie: Yup.  They were not introduced in this change, so I suggest we file a techdebt ticket and fix them up lated.

app/controls.go
tomwilkie: I\m pretty sure, even for tls connections, the thing we get passed is a still a CloseNotifier.  TLS connections are handled at the transport layer in go; at the http layer we've got a connection, which implements CloseNotifier.

I'd love to be proved wrong through...

app/controls.go
tomwilkie: The codec does live in xfer/.  Are you suggesting something like `xfer.NewControlClient(websocker)`?

xfer/multi_client.go
```diff
@@ -0,0 +1,104 @@
...
+		case <-time.After(backoff):
+		case <-c.quit:
+			return
+		}
+		backoff *= 2
+		if backoff > maxBackoff {
+			backoff = maxBackoff
+		}
+	}
+}
```
paulbellamy: A lot of this logic feels like it should be done in the client.

tomwilkie: Roger roger.

xfer/app_client.go
```diff
@@ -0,0 +1,120 @@
...
+		target:      target,
+		client: http.Client{
+			Transport: httpTransport,
+		},
+	}, nil
+}
+
+func (c *appClient) Stop() {
+	close(c.quit)
+}
```
paulbellamy: At the moment this is unused. (unless we move the controlLoop here).

xfer/controls.go
```diff
@@ -0,0 +1,111 @@
...
+
+// Response ...
+type Response struct {
+	ID    int64
+	Value interface{}
+	Error string
+}
+
+// ControlHandler ...
+type ControlHandler func(Request) Response
```
paulbellamy: ControlHandlerFunc (to match net/http)

xfer/app_client.go
tomwilkie: Yup, using it now the control loop is here.

xfer/controls.go
tomwilkie: roger.

xfer/multi_client.go
paulbellamy: created https://github.com/weaveworks/scope/issues/610

app/controls.go
paulbellamy: Yes, that is what I was proposing.

xfer/app_client.go
```diff
@@ -0,0 +1,166 @@
...
+
+type appClient struct {
+	ProbeConfig
+
+	quit     chan struct{}
+	target   string
+	insecure bool
+	client   http.Client
+
+	controlServerCodecMtx sync.Mutex
```
paulbellamy: Why do we need this mutex here now, when we didn't before? or did we also need this before?

xfer/app_client.go
```diff
@@ -0,0 +1,166 @@
...
+
+	server.ServeCodec(codec)
+	return nil
+}
+
+func (c *appClient) ControlConnection(handler ControlHandlerFunc) {
+	backoff := initialBackoff
+
+	for {
+		err := c.controlConnection(handler)
```
paulbellamy: Maybe a better name for this method, that describes what it does or what it returns? `setupConnection`? And, if it were to return the servercodec, would we still need the mutex? or could we contain the codec to this method, and clean it up when quit is closed?

app/router.go
paulbellamy: and swap `/api/controls` to `/api/control` so they match.

xfer/app_client.go
tomwilkie: This was added so when we stop an appclient, we can stop the associated control server.  As this requires us checking for a current app server, we need a mutex, as its modified concurrently.  Also synchronises access to controlServerCodec and quit.

paulbellamy:
There's a period when stopping or starting the container that you get all three controls available.
screen shot 2015-11-02 at 13 13 47

paulbellamy:
Should probably show the container state in the details panel.

xfer/app_client.go
tomwilkie: The call to ServeCodec blocks, so its not particularly useful to return anything here.  I also think codec should be contained within the client.

We can't close the mutex when quit is closed in this function, as ServeCodec blocks.  The call to codec.Close() is the thing that unblocks that.

paulbellamy:
Needs a `make static`, also.

tomwilkie:
> Should probably show the container state in the details panel.

done.

tomwilkie:
> There's a period when stopping or starting the container that you get all three controls available.

After discussion, this has been punted to the next PR.

app/router.go
tomwilkie: Updated the prefixes (its /api/control/ws and /api/control/foo/bar/baz now).  Broken them out into a separate registration function.

The problem with PathPrefix & subrouters was two fold: (1) the path prefix from / static files was owning everything, and (2) path prefixes and our custom matcher funcs don't mix.  So we can't use PathPrefix here.

probe/controls/controls_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/weaveworks/scope/probe/controls"
+	"github.com/weaveworks/scope/test"
+	"github.com/weaveworks/scope/xfer"
+)
+
+func TestControls(t *testing.T) {
+	controls.Register("foo", func(req xfer.Request) xfer.Response {
```
paulbellamy: Not critical, but this leaks into global state.

tomwilkie: Fixed.

tomwilkie:
+0.2% on coverage with 1.2k new loc!

paulbellamy:
Backend LGTM

tomwilkie:
Frontend LGTM.

PR #603, opened 2015-10-29
Don't keep stacking filters onto renderers, close idle http connections when we close a http publisher.

### LINKED ISSUE #602: App eventually uses 100% CPU
Reported by david & peter.

tomwilkie:
Fixes #602


peterbourgon:
For context: previously, we were decorating the same instance over and over again, leading to tons of layers over time. Now, we leave the original instance alone. LGTM.

peterbourgon:
Merging with permission via Slack.

PR #618, opened 2015-11-03
Speed up build & tests on Circle

### LINKED ISSUE #587: Parallelise/shard Circle, now that its taking >20mins
this arose from the 1.5 switch.

tomwilkie:
Fixes #587 

Lots of changes:
- do the darwin build on one shard, the arm build on another.
- Ensure we use up to date build container when running tests
- Parallelise host setup for integration tests. (~30s saved)
- Pass some more circle env vars through to ./tools/tests so the scheduler works. (~2mins saved)
- Gather coverage from multiple shards.
- Create files as the right owner inside the container (so we can compile the coverage stats)
- Parallelise the circle.yaml
- Build experimental in parallel. (~1min saved)


tomwilkie:
This disappointingly only seems to shave a few minutes off.  There is scope for a couple more, but after that we'll have to get more aggressive.

Now is a good time to stick this in.

PR #622, opened 2015-11-05
Sparklines

paulbellamy:
Initial work on sparklines. For this, they are just an in-memory set of [timestamp, float] tuples. Merge semantics are easy, as it is just set-union with timestamp for uniqueness. Timestamps expire as reports expire.

The graphs are a bit sub-optimal, but probably good enough to be useful, so will keep relatively basic for first release.

Todo:
- [x] Clean up code
- [x] Add tests
- [x] Add other sparklines (will need some generalizing of the code)
- [x] UI work (in particular, render "latest value" inline after sparkline)
- [x] Squash commits

Potential other sparklines to add:
- container memory usage
- cpu usage (do we report this)
- # Threads
- # TCP Connections
- stuff from docker?

At the moment, because we get new data every 15 seconds (and the sparklines are 15 seconds long), every 15 seconds the sparklines entirely redraw. It would be nice to smooth that out.


app/origin_host.go
```diff
@@ -16,7 +16,9 @@ type OriginHost struct {
 	Hostname string   `json:"hostname"`
 	OS       string   `json:"os"`
 	Networks []string `json:"networks"`
-	Load     string   `json:"load"`
+	Load1    string   `json:"load1"`
+	Load5    string   `json:"load5"`
+	Load15   string   `json:"load15"`
```
tomwilkie: Is this file even used for anything?

report/topology.go
```diff
@@ -131,6 +134,21 @@ func (n Node) WithSets(sets Sets) Node {
...
+	result.Metrics[key] = existing.Merge(metric)
+	return result
+}
+
+// WithMetrics returns a fresh copy of n, with metrics merged in.
+func (n Node) WithMetrics(metrics Metrics) Node {
+	result := n.Copy()
+	result.Metrics = result.Metrics.Merge(metrics)
+	return result
+}
```
tomwilkie: Nice!

tomwilkie:
Another thought: we might want to come up with a way to deal with units - memory usage will be in bytes, but we might want to display it in bytes.  Similarly, CPU in %, with min and max pegged at 0 and 100, or latency in ms, or QPS in... QPS.

probe/docker/container.go
```diff
@@ -220,6 +222,48 @@ func (c *container) ports(localAddrs []net.IP) report.StringSet {
 	return report.MakeStringSet(ports...)
 }
 
+func (c *container) cpuPercentMetric() report.Metric {
+	result := report.MakeMetric()
+	if len(c.pendingStats) < 2 {
```
paulbellamy: We should probably still return the sample (if we have one), so that we can write out the "latest value", but not draw a sparkline, I guess.

paulbellamy: Ah, I guess because it's all in deltas, we don't actually know it without 2 samples. blah.

probe/docker/container.go
```diff
@@ -220,6 +222,48 @@ func (c *container) ports(localAddrs []net.IP) report.StringSet {
...
+	result := report.MakeMetric()
+	for _, s := range c.pendingStats {
+		result = result.Add(s.Read, f(s))
+	}
+	return result
+}
+
+func (c *container) metrics() report.Metrics {
+	result := report.Metrics{
+		MemoryUsage:   c.metric(func(s *docker.Stats) float64 { return float64(s.MemoryStats.Usage) }),
```
paulbellamy: `c.memoryUsageMetric()` and eliminate `metric(...)`

tomwilkie: The idea is that would be reusable for other (absolute) docker metrics, if we find any.

probe/docker/container.go
tomwilkie: Indeed.

render/detailed_node.go
```diff
@@ -321,17 +324,26 @@ func containerOriginTable(nmd report.Node, addHostTag bool) (Table, bool) {
...
 	if addHostTag {
 		rows = append([]Row{{Key: "Host", ValueMajor: report.ExtractHostID(nmd)}}, rows...)
 	}
 
+	for _, tuple := range []struct {
+		key, human, format string
+		scale              float64
+	}{
+		{docker.MemoryUsage, "Memory Usage", "%0.2f MB", 1024 * 1024.},
+		{docker.CPUTotalUsage, "CPU Usage", "%0.2f%%", 1.},
```
paulbellamy: Should set the min and max on this graph.

paulbellamy:
First two commits are pulled out into: https://github.com/weaveworks/scope/pull/631

@tomwilkie @davkal , PTAL, thanks.

client/app/scripts/components/node-details-table.js
```diff
@@ -15,14 +14,12 @@ const NodeDetailsTable = React.createClass({
...
-                {row.value_minor}
-              </div>}
+              { row.value_type === 'numeric' && <div className="node-details-table-row-value-scalar">{row.value_major}</div> }
+              { row.value_type === 'numeric' && <div className="node-details-table-row-value-unit">{row.value_minor}</div> }
+              { row.value_type === 'sparkline' && <div className="node-details-table-row-value-sparkline"><Sparkline data={row.metric.samples} min={0} max={row.metric.max} first={row.metric.first} last={row.metric.last} interpolate="none" />{row.value_major}</div> }
+              { row.value_type === 'sparkline' && <div className="node-details-table-row-value-unit">{row.value_minor}</div> }
+              { row.value_type !== 'numeric' && row.value_type !== 'sparkline' && <div className="node-details-table-row-value-major truncate" title={row.value_major}>{row.value_major}</div> }
+              { row.value_type !== 'numeric' && row.value_type !== 'sparkline' && row.value_minor && <div className="node-details-table-row-value-minor truncate" title={row.value_minor}>{row.value_minor}</div> }
```
tomwilkie: Pls file a techdebt ticket to tidy this up (subcomponents etc)

render/detailed_node.go
```diff
@@ -300,6 +303,14 @@ func processOriginTable(nmd report.Node, addHostTag bool, addContainerTag bool)
 	}, len(rows) > 0 || commFound || pidFound
 }
 
+func sparklineRow(human, format string, scale float64, metric report.Metric) Row {
```
tomwilkie: Wonder if, instead of passing scale & format, we should pass a function (float64 -> string) that does the formatting.  Then have a bunch of standard function, one for percents, one for amounts (kb/mb/gb etc) that picks the right scale.  Thoughts?

render/detailed_node.go
```diff
@@ -321,17 +332,17 @@ func containerOriginTable(nmd report.Node, addHostTag bool) (Table, bool) {
...
 		rows = append([]Row{{Key: "Host", ValueMajor: report.ExtractHostID(nmd)}}, rows...)
 	}
 
+	if val, ok := nmd.Metrics[docker.MemoryUsage]; ok {
+		rows = append(rows, sparklineRow("Memory Usage", "%0.2f MB", 1024*1024, val))
+	}
+	if val, ok := nmd.Metrics[docker.CPUTotalUsage]; ok {
+		rows = append(rows, sparklineRow("CPU Usage", "%0.2f%%", 1, val))
+	}
+
```
tomwilkie: Stick this is a for loop with the anonymous type?  I have the feeling we'll be adding more here shortly.

client/app/scripts/components/node-details-table.js
paulbellamy: Filed: https://github.com/weaveworks/scope/issues/637

render/detailed_node.go
paulbellamy: :+1:

render/detailed_node.go
paulbellamy: Let's wait until we have more to add, then refactor.

report/topology.go
```diff
@@ -311,6 +331,119 @@ func (s StringSet) Copy() StringSet {
...
+
+// Metric is a list of timeseries data. Clients must use the Add
+// method to add values.
+type Metric struct {
+	Samples []Sample  `json:"samples"`
+	Min     float64   `json:"min"`
+	Max     float64   `json:"max"`
+	First   time.Time `json:"first"`
+	Last    time.Time `json:"last"`
+}
```
tomwilkie: Could we make this type immutable?  I think we just need to make the Metric returned by Add a copy.  And it we swap the sample list of a ps.List, arranged in reverse order, then we can share the list of sample between copies too.

report/topology.go
```diff
@@ -311,6 +331,119 @@ func (s StringSet) Copy() StringSet {
...
+		m.First = t
+	}
+	if m.Last.IsZero() || t.After(m.Last) {
+		m.Last = t
+	}
+	return m
+}
+
+// Merge combines the two Metrics and returns a new result.
+func (m Metric) Merge(other Metric) Metric {
```
tomwilkie: please make the metric you return a copy.

client/app/scripts/components/node-details-table.js
tomwilkie: Thanks!

client/app/scripts/components/sparkline.js
```diff
@@ -0,0 +1,125 @@
...
+      y.domain([
+        this.props.min || d3.min(data),
+        this.props.max || d3.max(data)
+      ]);
+
+      lastX = x(data.length - 1);
+      lastY = y(data[data.length - 1]);
+    }
+
+    let svg = d3.select(this.getDOMNode()).
```
davkal: The rendering should really by done by react, and not d3. d3 should only provide the data-to-path translation.
This could be done its own ticket later, e.g. when animating sparklines (#638)

paulbellamy: :+1: Let's do it as a second step.

report/topology.go
paulbellamy: Sure, will do once #628 is merged (adds ps).

davkal:
Metric data are sent even for non-metric fields:
```json
{
            "key": "Operating system",
            "value_major": "darwin",
            "metric": {
              "samples": null,
              "min": 0,
              "max": 0,
              "first": "0001-01-01T00:00:00Z",
              "last": "0001-01-01T00:00:00Z"
            }
          },
```

This roughly doubles the response size of a node details request compared to `master`.

davkal:
I'd change the width to `48`:
screen shot 2015-11-09 at 12 42 34


Otherwise they are too dominant.

davkal:
I'd also add `title` attribute so that a tooltip shows the length of the timeseries.

paulbellamy:
@davkal by "length" do you mean number of samples, timespan, or both?

davkal:
@paulbellamy whatever we can say for sure. At the minimum "Last 15 secs", even better "Last 15 sec, 1 sample/sec, min: 0.6, max: 1.8, avg: 1.2".

report/topology.go
paulbellamy: No need for a copy, as everything in it is immutable, and `m` is passed-by-value.

report/topology.go
tomwilkie: Just merged that.

Makefile
```diff
@@ -43,7 +43,7 @@ $(PROBE_EXE): probe/*.go probe/docker/*.go probe/kubernetes/*.go probe/endpoint/
 
 ifeq ($(BUILD_IN_CONTAINER),true)
 $(APP_EXE) $(PROBE_EXE) $(RUNSVINIT): $(SCOPE_BACKEND_BUILD_UPTODATE)
-	$(SUDO) docker run $(RM) -v $(shell pwd):/go/src/github.com/weaveworks/scope -e GOARCH -e GOOS \
+	$(SUDO) docker run $(RM) -ti -v $(shell pwd):/go/src/github.com/weaveworks/scope -e GOARCH -e GOOS \
```
tomwilkie: I removed these as they didn't play well then you try and run make under ssh.

paulbellamy: ah, damn.

render/detailed_node.go
```diff
@@ -43,10 +44,12 @@ type Table struct {
...
-	Key        string `json:"key"`                   // e.g. Ingress
-	ValueMajor string `json:"value_major"`           // e.g. 25
-	ValueMinor string `json:"value_minor,omitempty"` // e.g. KB/s
-	Expandable bool   `json:"expandable,omitempty"`  // Whether it can be expanded (hidden by default)
+	Key        string         `json:"key"`                   // e.g. Ingress
+	ValueMajor string         `json:"value_major"`           // e.g. 25
+	ValueMinor string         `json:"value_minor,omitempty"` // e.g. KB/s
+	Expandable bool           `json:"expandable,omitempty"`  // Whether it can be expanded (hidden by default)
+	ValueType  string         `json:"value_type,omitempty"`  // e.g. sparkline
+	Metric     *report.Metric `json:"metric,omitempty"`      // e.g. KB/s
```
tomwilkie: Comment is probably a little misleading now.  Maybe remove?

render/detailed_node.go
```diff
@@ -343,6 +346,21 @@ func processOriginTable(nmd report.Node, addHostTag bool, addContainerTag bool)
...
+	}
+	lastStr := ""
+	if last := metric.LastSample(); last != nil {
+		lastStr = format(last)
+	}
+	return Row{Key: human, ValueMajor: lastStr, Metric: &metric, ValueType: "sparkline"}
+}
+
+func formatDefault(s *report.Sample) string { return fmt.Sprintf("%0.2f", s.Value) }
+func formatMB(s *report.Sample) string      { return fmt.Sprintf("%0.2f MB", s.Value) }
```
tomwilkie: Can we make this FormatMemory, like this:

```go
type ByteSize float64

const (
    _           = iota // ignore first value by assigning to blank identifier
    KB ByteSize = 1 << (10 * iota)
    MB
    GB
    TB
    PB
)

func formatMB(s *report.Sample) string { 
  switch {
  case s.Value < KB:
    return fmt.Sprintf("%0.2f bytes", s.Value)
  case s.Value < MB:
    return fmt.Sprintf("%0.2f KB", s.Value / KB)
  case s.Value < GB:
    return fmt.Sprintf("%0.2f MB", s.Value / MB)
  case s.Value < TB:
    return fmt.Sprintf("%0.2f GB", s.Value / GB)
  case s.Value < PB:
    return fmt.Sprintf("%0.2f TB", s.Value / TB)
  default:
    return fmt.Sprintf("%0.2f PB", s.Value / PB)
  }
}
```

render/detailed_node.go
```diff
@@ -441,9 +459,39 @@ func getDockerLabelRows(nmd report.Node) []Row {
...
 	rows := []Row{}
 	for _, tuple := range []struct{ key, human string }{
-		{host.Load, "Load"},
+		{host.Load1, "Load (1m)"},
+		{host.Load5, "Load (5m)"},
+		{host.Load15, "Load (15m)"},
+	} {
+		if val, ok := nmd.Metrics[tuple.key]; ok {
+			val.Max = maxLoad
+			val.First = lastLoad.Add(-15 * time.Second) // TODO(paulbellamy): This should be based on the duration flag, maybe? or just auto-scale to data.
```
tomwilkie: This defined by the -window argument to app main.  Shouldn't really hard code it.  Plumbing it though it going to be fun though....

report/latest_map.go
```diff
@@ -79,7 +79,7 @@ func (m LatestMap) toIntermediate() map[string]LatestEntry {
 	return intermediate
 }
 
-func fromIntermediate(in map[string]LatestEntry) LatestMap {
+func (m LatestMap) fromIntermediate(in map[string]LatestEntry) LatestMap {
```
tomwilkie: Why?

report/metrics.go
```diff
@@ -0,0 +1,268 @@
...
+	"bytes"
+	"encoding/gob"
+	"encoding/json"
+	"time"
+
+	"github.com/mndrix/ps"
+)
+
+// Metrics is a string->metric map.
+type Metrics map[string]Metric
```
tomwilkie: If you were feeling super keen, you could make this a ps.Map....

tomwilkie: In fact, we should break out the changes to ps.Map and ps.List (marshalling etc) into our own structs, and then reuse them.

davkal:
Tooltip looks great! I'd still vote for smaller width, e.g. 48px. Otherwise LGTM! Well done, jsx :heart: Paul

report/metrics.go
```diff
@@ -0,0 +1,268 @@
...
+	if v < m.Min {
+		m.Min = v
+	}
+	if m.First.IsZero() || t.Before(m.First) {
+		m.First = t
+	}
+	if m.Last.IsZero() || t.After(m.Last) {
+		m.Last = t
+	}
+	return m
```
tomwilkie: This is not immutable; you should return a new Metric, not the same one.

report/metrics.go
```diff
@@ -0,0 +1,268 @@
...
+	s := m.Samples.Head()
+	return &s
+}
+
+// Samples is an immutable list of timeseries data. We have this to implement
+// proper marshalling for the ps.List, as well as fixing some if ps.List's
+// behaviour.
+type Samples struct {
+	ps.List
+}
```
tomwilkie: I'm debating where we need the extra nesting (Metric -> Samples -> ps.List) and the ability to deal with Nil.  Mind if I have a play with this code?

report/metrics.go
tomwilkie: Okay I tried that and couldn't make it work nicely.  Lets leave for now, and I'll make this immutable in the future.

report/metrics.go
tomwilkie: I've fixed all the immutability in #645

tomwilkie:
When you're ready, I think this need to be squashed and split into 3 PRs - the plumbing (changes to report and render/ etc), the exposing of metrics in Docker and Host probe packages, and the UI work.

Makefile
paulbellamy: Even with `ssh -t`?

report/metrics.go
paulbellamy: It is a new metric. The methods are on the passed-by-value, not the pointer.

Makefile
tomwilkie: Even with `ssh -t`

tomwilkie: You could stick them in a Makefile var like RM.

render/detailed_node.go
paulbellamy: There is a `report.Report.Window` attribute, but that seems to be partially unused, and have weird merge semantics? Let's just base the x-axis on the data for now, like we do with the other graphs. Needs a better general solution.

tomwilkie: Roger that.  Report.Window was added for the packet capture stuff, and I don't think should be used here.

tomwilkie:
Assuming this is green, LGTM.

PR #623, opened 2015-11-05
Moving xfer.Collector into App (the only place its used)

paulbellamy:


tomwilkie:
Assuming tests pass, LGTM

Issue #502, opened 2015-09-17
Investigate extending immutability


tomwilkie:
Discussion TBD

tomwilkie:
We had to remove the Copy() from Topology.AddNode, as it was a very expensive thing to do when building a topology - O(n^2) in number of nodes in the topology.

Instead, we could use something like https://github.com/mndrix/ps, and add explicit Nodes.Get, Nodes.Remove and Nodes.Walk.

Similarly we could do the same for the Node struct itself.

paulbellamy:
Additionally, we should only be doing a `Copy()` when modifying the node, which should (almost) always happen in methods on the node.

PR #628, opened 2015-11-05
Add 'latest' CRDT, make UI update quicker.

tomwilkie:
The idea is to have an alternative to the metadata map on each node, where the merge results are deterministic (pick the latest value).  This should allow us to propagate state changes faster in the UI, as we don't have to wait for the 15s window to pass.

Also using this as an opportunity to try some of the persistent datatypes, to help address #502.

### DIFF
--- a/render/renderable_node.go
+++ b/render/renderable_node.go
@@ -134,10 +134,12 @@ func (rn RenderableNode) Copy() RenderableNode {
 // Specifically, that means cutting out parts of the Node.
 func (rn RenderableNode) Prune() RenderableNode {
 	cp := rn.Copy()
-	cp.Node.Metadata = report.Metadata{}   // snip
-	cp.Node.Counters = report.Counters{}   // snip
-	cp.Node.Edges = report.EdgeMetadatas{} // snip
-	cp.Node.Sets = report.Sets{}           // snip
+	cp.Node.Metadata = report.Metadata{}     // snip
+	cp.Node.Counters = report.Counters{}     // snip
+	cp.Node.Edges = report.EdgeMetadatas{}   // snip
+	cp.Node.Sets = report.Sets{}             // snip
+	cp.Node.Controls = report.NodeControls{} // snip
+	cp.Node.Latest = report.LatestMap{}      // snip
 	return cp
 }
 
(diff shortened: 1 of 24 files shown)


render/renderable_node.go
```diff
@@ -134,10 +134,12 @@ func (rn RenderableNode) Copy() RenderableNode {
...
-	cp.Node.Metadata = report.Metadata{}   // snip
-	cp.Node.Counters = report.Counters{}   // snip
-	cp.Node.Edges = report.EdgeMetadatas{} // snip
-	cp.Node.Sets = report.Sets{}           // snip
+	cp.Node.Metadata = report.Metadata{}     // snip
+	cp.Node.Counters = report.Counters{}     // snip
+	cp.Node.Edges = report.EdgeMetadatas{}   // snip
+	cp.Node.Sets = report.Sets{}             // snip
+	cp.Node.Controls = report.NodeControls{} // snip
+	cp.Node.Latest = report.LatestMap{}      // snip
```
paulbellamy: At this point, it would be easier just to:
```Go
cp.Node = Node{Adjacency: rn.Node.Adjacency.Copy()}
```

report/controls.go
```diff
@@ -32,3 +36,40 @@ func (cs Controls) Copy() Controls {
 func (cs Controls) AddControl(c Control) {
 	cs[c.ID] = c
 }
+
+// NodeControls represent the individual controls that are valid
+// for a given node at a given point in time.  Its is immutable.
+type NodeControls struct {
+	Timestamp int64  `json:"timestamp"`
+	Controls  IDList `json:"controls"`
```
paulbellamy: `StringSet`

report/controls.go
```diff
@@ -32,3 +36,40 @@ func (cs Controls) Copy() Controls {
 func (cs Controls) AddControl(c Control) {
 	cs[c.ID] = c
 }
+
+// NodeControls represent the individual controls that are valid
+// for a given node at a given point in time.  Its is immutable.
+type NodeControls struct {
+	Timestamp int64  `json:"timestamp"`
```
paulbellamy: Why not a `time.Time`?

render/renderable_node.go
tomwilkie: +1

report/controls.go
tomwilkie: No reason; I'm used to thinking of time like this.  It seems the go way is to use a time.Time, as it has all the right json marshalling baked in.  I can't think of any gotchas, so will change.

paulbellamy:
LGTM

PR #631, opened 2015-11-06
General Cleanup

paulbellamy:
@davkal Safe to add eslintignore to the dockerfile? It should be in there, right?


davkal:
@paulbellamy yes you can add it. it currently ignores linting the test code. Maybe create another issue to enable linting of test code.

tomwilkie:
LGTM

PR #633, opened 2015-11-06
Refactoring of probe main

### LINKED ISSUE #537: Break up probe.main into smaller, testable functions.

paulbellamy:
Fixes #537

- Move probe/main.go to prog/probe/
- Rename probe/* the 'probe' package
- Moves resolver.go to xfer/
- Removed probe/instrumentation.go, not used anywhere
- An empty StingSet (and IDList) is now nil
- Move filter tests to filters_test.go


xfer/resolver.go
```diff
@@ -1,4 +1,4 @@
-package main
+package xfer
```
paulbellamy: Why the move of this?

paulbellamy:
Let's break out the functional changes (StringSet stuff) from the re-org, just a separate commit would be nice.

xfer/resolver.go
tomwilkie: It just didn't seem to 'fit' in the probe package; it was much more xfer-y, don't you think?

In other words: no good reason.

tomwilkie:
+1 will do.

tomwilkie:
@paulbellamy I think I've addressed your comment(s)

xfer/resolver.go
paulbellamy: It's only ever used by the probe, though. So I feel it is much more probe-y...

tomwilkie: The same could be said for {Multi,Background,HTTP}Publishers and (Multi)AppClient...  The resolver should very much be colocated this there 5 things; I'm open to arguments that these should be in a probe/ sub module...

paulbellamy: after f2f discussion, lets move all this into the probe.

tomwilkie: We'll follow up with a ticket...

paulbellamy:
LGTM

PR #635, opened 2015-11-09
Enable and fixup linting on client test js

### LINKED ISSUE #632: We should lint test js

paulbellamy:
Fixes #632 

@davkal There is one left I'm not sure how to resolve:
```
app/scripts/stores/__tests__/app-store-test.js
  75:8  error  ClickGroupingAction is defined but never used  no-unused-vars
```

Is this a missing test, or should we just remove it?


davkal:
@paulbellamy this was old functionality, can be removed.

paulbellamy:
Bah, jest globally injects a bunch of stuff, and eslint doesn't have a way to ignore it in test files. We can either ignore jest stuff everywhere, or have two npm scripts (and run both).

paulbellamy:
Currently throwing: `Error: ENOENT, no such file or directory 'app/**/*-test.js'`

davkal:
I've added test linting and upgrade linter as part of #649. All green.

PR #639, opened 2015-11-09
Resolve to a group of names

### LINKED ISSUE #504: Unstable reverse DNS lookups in UI

inercia:
Fixes #504

### DIFF
--- a/probe/endpoint/resolver.go
+++ b/probe/endpoint/resolver.go
@@ -3,7 +3,6 @@ package endpoint
 import (
 	"fmt"
 	"net"
-	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


probe/endpoint/resolver.go
```diff
@@ -72,8 +72,10 @@ func (r *reverseResolver) loop() {
 		names, err := r.Resolver(request)
 		if err == nil && len(names) > 0 {
 			sort.Strings(names)
```
tomwilkie: Sort is not needed anymore, I think.

tomwilkie:
Also pretty sure this is going to need an update in https://github.com/weaveworks/scope/blob/master/render/detailed_node.go#L256

probe/endpoint/resolver.go
inercia: Done.

tomwilkie:
LGTM!

PR #640, opened 2015-11-09
Push mini-reports when container states changes

### LINKED ISSUE #629: Make state updates on containers trigger a mini report push.

tomwilkie:
Fixes #629


tomwilkie:
Right now, this change bypasses the 3s probe publish interval for docker container state changes, pushing a mini-report containing just the container node.  As discussed, we're going to add a flag to reports (fastforward?) to indicate to the app that it should push a delta to the UI quickly too, eliminating another source of delay.

tomwilkie:
Status: when a container state changes, a probe -> app mini report is pushed immediately; this causes an app -> ui topology push to also happen, and you get sub-second updated of the rendered graph (yay!)

Unfortunately, the details panel is on a 1 sec poll, driven by the UI, so is beyond scope for these changes.  @davkal is there anything we can do?  I suspect we're going to have to rework the app <-> UI websocket when pipes come along, so we could delay the work until then.

davkal:
@tomwilkie I was waiting for something like this to come up. I've been meaning to look into GraphQL, or Netflix's Falcor for this. But then again maybe this is overblown. Until then we can send arbitrary messages along the websocket, in addition to node set changes.

tomwilkie:
I think for pipes we're going to need the websocket to be 'stationary' - ie not connected / disconnected for every topology the user clicks.  If so, the websocket protocol might need to evolve to something along the lines of the UI 'subscribing' to interesting 'channels' - such as the rendered topology updates, node detail updates, pipes etc.  Just a though, lets move this discussion off this PR.

davkal:
Yes, the current implementation with one websocket per topology is not optimal. It's a garbled REST approach. Maybe we can transition to a single socket.io connection that will also have built-in fallbacks if websockets are not supported somewhere along the wire.

tomwilkie:
The failure was unrelated (and is now fixed).  This should be ready for review.

prog/probe/main.go
```diff
@@ -134,7 +134,10 @@ func main() {
 		if registry, err := docker.NewRegistry(*dockerInterval); err == nil {
 			defer registry.Stop()
 			p.AddTagger(docker.NewTagger(registry, processCache))
-			p.AddReporter(docker.NewReporter(registry, hostID))
+
+			reporter := docker.NewReporter(registry, hostID, p)
+			registry.WatchContainerUpdates(reporter.ContainerUpdated)
```
paulbellamy: Let's do this in `NewReporter`, where we already pass in the registry.

probe/probe.go
```diff
@@ -140,6 +152,17 @@ func (p *Probe) tag(r report.Report) report.Report {
 	return r
 }
 
+func condense(rpt report.Report, rs chan report.Report) report.Report {
```
paulbellamy: merge? drain? maybe just a comment would suffice...

probe/probe.go
```diff
@@ -150,8 +173,14 @@ func (p *Probe) publishLoop() {
...
-			localReport := p.rpt.swap(report.MakeReport())
-			if err := rptPub.Publish(localReport); err != nil {
+			rpt := condense(report.MakeReport(), p.spiedReports)
+			if err := rptPub.Publish(rpt); err != nil {
+				log.Printf("publish: %v", err)
+			}
+
+		case rpt := <-p.shortcutReports:
+			rpt = condense(rpt, p.shortcutReports)
+			if err := rptPub.Publish(rpt); err != nil {
```
paulbellamy: It seems odd that we need to store and handle these so differently...

tomwilkie: Yeah I was hoping for a way to unify these a bit more, but I couldn't come up with a nice way.  I guess we could move the publish step into the condense function, and call it drainAndPublish?

app/api_topology.go
```diff
@@ -114,7 +114,11 @@ func handleWebsocket(
 	var (
 		previousTopo render.RenderableNodes
 		tick         = time.Tick(loop)
+		wait         = make(chan struct{}, 1)
 	)
+	rep.WaitOn(wait)
+	defer rep.UnWait(wait)
```
paulbellamy: I'm debating if this would be any cleaner with a watch and lock. instead of a channel. Not sure...

probe/probe.go
paulbellamy: It's not immediately clear why we need two different chans for the reports?

app/api_topology.go
tomwilkie: Yes I went round and round on this.  I'm not sure either.

I had a version that made a new channel each time and stuck it in a slice on the reporter; the reporter would then close all of them when it received a shortcut report.  The problem is that list would grow unbounded in the absent of the shortcut reports, and we would be continually creating / closing channels.

probe/probe.go
```diff
@@ -140,20 +152,33 @@ func (p *Probe) tag(r report.Report) report.Report {
...
 		select {
 		case <-pubTick:
-			localReport := p.rpt.swap(report.MakeReport())
-			if err := rptPub.Publish(localReport); err != nil {
-				log.Printf("publish: %v", err)
-			}
+			p.drainAndPublish(report.MakeReport(), p.spiedReports)
+
+		case rpt := <-p.shortcutReports:
+			p.drainAndPublish(rpt, p.shortcutReports)
```
paulbellamy: Why won't this do the following?
1) receive a shortcut report
2) publish the shortcur report
3) fire pubTick
4) publish an out-of-date spiedReport

probe/probe.go
tomwilkie: I considered using one channel - the side effect would be that shortcut reports could get merged with any pending 'full' reports, potentially making the thing we push much larger, more often.  Given we had to back off the publish tick to 3s due to cpu usage previously, I decided not to do that.

paulbellamy: Ah, because shortcut reports are much smaller. Gotcha.

probe/probe.go
tomwilkie: It will do exactly that.  Luckily the reports are CRDTs, so merging them together will always result in the same output, independent of order.

paulbellamy: ah, right.

paulbellamy:
LGTM, then.

PR #644, opened 2015-11-10
Report coverage correctly

tomwilkie:
Coverage reports need to contain output from both shards, but due to a change in circle the other shard was getting stopped before we attempted to collect the reports.

### DIFF
--- a/integration/320_container_edge_cross_host_2_test.sh
+++ b/integration/320_container_edge_cross_host_2_test.sh
@@ -4,8 +4,8 @@
 
 start_suite "Test short lived connections between containers on different hosts"
 
-WEAVE_NO_FASTDP=true weave_on $HOST1 launch $HOST1 $HOST2
-WEAVE_NO_FASTDP=true weave_on $HOST2 launch $HOST1 $HOST2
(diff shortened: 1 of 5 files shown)


integration/320_container_edge_cross_host_2_test.sh
```diff
@@ -4,8 +4,8 @@
 
 start_suite "Test short lived connections between containers on different hosts"
 
-WEAVE_NO_FASTDP=true weave_on $HOST1 launch $HOST1 $HOST2
-WEAVE_NO_FASTDP=true weave_on $HOST2 launch $HOST1 $HOST2
+WEAVE_NO_FASTDP=true WEAVE_DOCKER_ARGS=$ADD_HOST_ARGS weave_on $HOST1 launch $HOST1 $HOST2
+WEAVE_NO_FASTDP=true WEAVE_DOCKER_ARGS=$ADD_HOST_ARGS weave_on $HOST2 launch $HOST1 $HOST2
```
paulbellamy: What's up with these bits?

tomwilkie: This was the cause of the breakage on master:

https://github.com/weaveworks/weave/commit/499d6312b7e6159ab94c116127b9c020fe109118

We still pull the master weave repo; I'm working on moving the bits we need (gce.sh and run_all.sh) into tools, so we can pin their version properly.

paulbellamy:
LGTM

PR #645, opened 2015-11-10
Make metrics immutability more obvious

tomwilkie:


report/metrics_test.go
```diff
@@ -74,21 +74,19 @@ func TestMetricAdd(t *testing.T) {
...
+	want := report.MakeMetric().
+		Add(s[0].Timestamp, s[0].Value).
+		Add(s[1].Timestamp, s[1].Value).
+		Add(s[2].Timestamp, 0.5)
 
-	want := report.Samples{}.Cons(s[0]).Cons(s[1]).Cons(s[2])
-	if !reflect.DeepEqual(want, have.Samples) {
-		t.Errorf("diff: %s", test.Diff(want, have.Samples))
+	if !reflect.DeepEqual(want, have) {
+		t.Errorf("diff: %s", test.Diff(want, have))
```
paulbellamy: The `intermediate` stage was to check that `Add` does not modify the existing metric.

report/topology.go
```diff
@@ -145,8 +145,10 @@ func (n Node) WithSets(sets Sets) Node {
 // WithMetric returns a fresh copy of n, with metric merged in at key.
 func (n Node) WithMetric(key string, metric Metric) Node {
 	result := n.Copy()
-	existing := n.Metrics[key]
-	result.Metrics[key] = existing.Merge(metric)
+	if existing, ok := n.Metrics[key]; ok {
+		metric = existing.Merge(metric)
+	}
+	n.Metrics[key] = metric
```
paulbellamy: If you're doing this verbosely, then this should be in an else.

report/metrics.go
```diff
@@ -63,81 +78,98 @@ func (m Metric) Len() int {
...
-	m.Max = m.Max / n
-	m.Min = m.Min / n
-	return m
+	var div func(ps.List) ps.List
+	div = func(ss ps.List) ps.List {
+		if ss.IsNil() {
+			return ss
+		}
+		s := ss.Head().(Sample)
+		return div(ss.Tail()).Cons(Sample{s.Timestamp, s.Value / n})
```
paulbellamy: Shouldn't be done recursively, as go has no tail recursion.

report/metrics.go
```diff
@@ -63,81 +78,98 @@ func (m Metric) Len() int {
...
-	}
-	if v < m.Min {
-		m.Min = v
-	}
-	if m.First.IsZero() || t.Before(m.First) {
-		m.First = t
+		if currSample.Timestamp.Before(t) {
+			return ss.Cons(Sample{t, v})
+		}
+		return insert(ss.Tail()).Cons(currSample)
```
paulbellamy: Again, recursion is a bad plan.

report/metrics.go
```diff
@@ -63,81 +78,98 @@ func (m Metric) Len() int {
...
-	for !newSamples.IsNil() {
-		head := newSamples.Head()
-		if head.Timestamp.Equal(t) {
-			// The list already has the element.
-			return m
+	// Find the first element which is before you element, and insert
+	// your new element in the list.  NB we want to dedupe entries with
+	// equal timestamps.
+	var insert func(ps.List) ps.List
+	insert = func(ss ps.List) ps.List {
```
paulbellamy: `:=`

report/metrics.go
```diff
@@ -63,81 +78,98 @@ func (m Metric) Len() int {
...
-	oldSamples := m.Samples
-	m.Samples = Samples{}
-	oldSamples.ForEach(func(s Sample) {
-		m = m.Add(s.Timestamp, s.Value/n)
-	})
-	m.Max = m.Max / n
-	m.Min = m.Min / n
-	return m
+	var div func(ps.List) ps.List
+	div = func(ss ps.List) ps.List {
```
paulbellamy: Let's move this method to `Samples`.

report/metrics.go
```diff
@@ -17,7 +18,10 @@ type Metrics map[string]Metric
 func (m Metrics) Merge(other Metrics) Metrics {
 	result := m.Copy()
 	for k, v := range other {
-		result[k] = result[k].Merge(v)
+		if o, ok := result[k]; ok {
+			v = v.Merge(o)
+		}
+		result[k] = v
```
paulbellamy: Again, this needs to be an else.

report/topology.go
paulbellamy: Ah, I missed that `metric` is being shadowed. That's quite non-obvious.

paulbellamy:
Added a failing test for the sample ordering in output.

paulbellamy:
If it helps you understand the code better, it's fine. I'm on the fence about `Samples` -> `WireThingy`. It is less code, but we lose sensible zero-values for Metric.

report/metrics.go
```diff
@@ -49,13 +51,26 @@ type Sample struct {
...
 func (m Metric) WithFirst(t time.Time) Metric {
-	m.First = t
-	return m
+	return Metric{
+		Samples: m.Samples,
+		Max:     m.Max,
+		Min:     m.Min,
+		First:   t,
+		Last:    m.Last,
+	}
```
paulbellamy: If we're going to explicitly make another copy we should probably put all the methods on `*Metric`, to avoid copying everything twice.

report/topology.go
tomwilkie: (not strictly shadowed, but reused) but yeah, non obvious.

I'll add nil-value handling in for metrics, then we can remove this little if.

report/metrics.go
tomwilkie: Thinking about it, I think that might have been ps's motivation for putting everything behind an interface?

tomwilkie:
Will rebase to get new test.

tomwilkie:
I had a go at making the metrics methods take pointers, but it tends to invade the entire codebase (can't call methods with pointer receivers when they are used as non-pointer values in a map).  So I decided against.  A smart compiler will notice that we don't modify the receiver, and pass it by references.  Not that go has a smart compiler...

PR #646, opened 2015-11-10
Always center selected node

### LINKED ISSUE #615: Selecting node and moving canvas results in canvas not being in the place you'd expect
I find if I select & deselect nodes, moving the canvas around all the time, it quickly starts getting out of hand.  This interaction need some more thought.

Suggestion: when you have a node selected, don't allow the canvas to move?  Then when you deselect, put the canvas back to where it was before.  This might be more predictable.

davkal:
* dont allow panning while node is selected
* restore old panning position after deselect

Fixes #615

### DIFF
--- a/client/app/scripts/charts/node.js
+++ b/client/app/scripts/charts/node.js
@@ -12,7 +12,7 @@ const Node = React.createClass({
 
   render: function() {
     const props = this.props;
-    const nodeScale = this.props.nodeScale;
+    const nodeScale = props.focused ? props.selectedNodeScale : props.nodeScale;
     const zoomScale = this.props.zoomScale;
     let scaleFactor = 1;
     if (props.focused) {
--- a/client/app/scripts/charts/node.js
+++ b/client/app/scripts/charts/node.js
@@ -12,15 +12,16 @@ const Node = React.createClass({
 
   render: function() {
     const props = this.props;
-    const scale = this.props.scale;
+    const scale = this.props.nodeScale;
+    const zoomScale = this.props.zoomScale;
     let scaleFactor = 1;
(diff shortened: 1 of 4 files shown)


tomwilkie:
A few comments from tests - Its a bit jerky going into and out of selected mode, and it doesn't deal with window resizing in selected mode very well - the nodes 'pop':

0ncltekrfe

tomwilkie:
My understanding is:
- on click
- move canvas so focused node is in the middle
- move other nodes so they are arranged around it
- render them bigger, others as translucent
- on ‘un’ click, revert those changes

Would it be easier/look better is we didn’t move the canvas, just the nodes?  Eg on click, ‘lock’ the canvas, move the nodes to where we want them, render them bigger etc.  On ‘un’ click, put them back and unlock the canvas?

davkal:
@tomwilkie I implemented the suggestions. PTAL

tomwilkie:
With latest changes, I can zoom and drag in selected mode again.  Also, selected nodes seem to have gotten smaller, and I managed to get stuck in selected mode somehow, although I can't reproduce that.

However the not-moving-the-canvas-moving-the-node approach seems to work really well, and the resizing of the window not longer jumps around like before.  In fact its all much smoother.

davkal:
@tomwilkie PTAL

tomwilkie:
Yup, looking better.  1 issue - if you zoom (with two fingers on the track pad) whilst in selected mode, nothing happens (as expected).  But then, when you leave selected mode, and trigger a redraw (resize the window), the zoom gets applied.

tomwilkie:
LGTM!

PR #650, opened 2015-11-11
Pipes

### LINKED ISSUE #365: Pipes: Bidirectional streams from UI<-->App<-->Probe
(Was: Combine GoTTY and docker exec)

https://github.com/yudai/gotty

One click open a terminal with a shell in the container. Similarly, one click could attach the terminal of a container.

### LINKED ISSUE #648: Race condition in test
Could indicate a real race condition.

```
--- FAIL: TestControl (0.12s)
panic: sync: WaitGroup is reused before previous Wait has returned [recovered]
	panic: sync: WaitGroup is reused before previous Wait has returned

goroutine 84 [running]:
testing.tRunner.func1(0xc8202b2d80)
	/usr/local/go/src/testing/testing.go:450 +0x171
sync.(*WaitGroup).Wait(0xc820492d30)
	/usr/local/go/src/sync/waitgroup.go:128 +0x114
net/http/httptest.(*Server).Close(0xc820492d00)
	/usr/local/go/src/net/http/httptest/server.go:168 +0x5c
github.com/weaveworks/scope/app.TestControl(0xc8202b2d80)
	/home/vagrant/src/github.com/weaveworks/scope/app/controls_test.go:69 +0xadb
testing.tRunner(0xc8202b2d80, 0x1c1b0b8)
	/usr/local/go/src/testing/testing.go:456 +0x98
created by testing.RunTests
	/usr/local/go/src/testing/testing.go:561 +0x86d

goroutine 1 [chan receive]:
testing.RunTests(0x14183b8, 0x1c1afe0, 0xc, 0xc, 0xc820292b00)
	/usr/local/go/src/testing/testing.go:562 +0x8ad
testing.(*M).Run(0xc8202cdef8, 0x7fcc44085038)
	/usr/local/go/src/testing/testing.go:494 +0x70
main.main()
	github.com/weaveworks/scope/app/_test/_testmain.go:76 +0x116

goroutine 17 [syscall, locked to thread]:
runtime.goexit()
	/usr/local/go/src/runtime/asm_amd64.s:1696 +0x1

goroutine 20 [chan receive]:
github.com/weaveworks/scope/vendor/github.com/golang/glog.(*loggingT).flushDaemon(0x1c39ae0)
	/home/vagrant/src/github.com/weaveworks/scope/vendor/github.com/golang/glog/glog.go:882 +0x67
created by github.com/weaveworks/scope/vendor/github.com/golang/glog.init.1
	/home/vagrant/src/github.com/weaveworks/scope/vendor/github.com/golang/glog/glog.go:410 +0x297

goroutine 22 [syscall]:
os/signal.loop()
	/usr/local/go/src/os/signal/signal_unix.go:22 +0x18
created by os/signal.init.1
	/usr/local/go/src/os/signal/signal_unix.go:28 +0x37

goroutine 63 [IO wait]:
net.runtime_pollWait(0x7fcc444e8048, 0x72, 0xc820068080)
	/usr/local/go/src/runtime/netpoll.go:157 +0x60
net.(*pollDesc).Wait(0xc8201220d0, 0x72, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:73 +0x3a
net.(*pollDesc).WaitRead(0xc8201220d0, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:78 +0x36
net.(*netFD).Read(0xc820122070, 0xc820414000, 0x1000, 0x1000, 0x0, 0x7fcc444e2028, 0xc820068080)
	/usr/local/go/src/net/fd_unix.go:232 +0x23a
net.(*conn).Read(0xc820498088, 0xc820414000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/net.go:172 +0xe4
net/http.noteEOFReader.Read(0x7fcc44085130, 0xc820498088, 0xc820412058, 0xc820414000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/http/transport.go:1370 +0x67
net/http.(*noteEOFReader).Read(0xc82058e760, 0xc820414000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
	<autogenerated>:126 +0xd0
bufio.(*Reader).fill(0xc820408300)
	/usr/local/go/src/bufio/bufio.go:97 +0x1e9
bufio.(*Reader).Peek(0xc820408300, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/bufio/bufio.go:132 +0xcc
net/http.(*persistConn).readLoop(0xc820412000)
	/usr/local/go/src/net/http/transport.go:876 +0xf7
created by net/http.(*Transport).dialConn
	/usr/local/go/src/net/http/transport.go:685 +0xc78

goroutine 64 [select]:
net/http.(*persistConn).writeLoop(0xc820412000)
	/usr/local/go/src/net/http/transport.go:1009 +0x40c
created by net/http.(*Transport).dialConn
	/usr/local/go/src/net/http/transport.go:686 +0xc9d

goroutine 72 [IO wait]:
net.runtime_pollWait(0x7fcc444e8108, 0x72, 0xc820068080)
	/usr/local/go/src/runtime/netpoll.go:157 +0x60
net.(*pollDesc).Wait(0xc82025e680, 0x72, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:73 +0x3a
net.(*pollDesc).WaitRead(0xc82025e680, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:78 +0x36
net.(*netFD).accept(0xc82025e620, 0x0, 0x7fcc440850e0, 0xc8204a28e0)
	/usr/local/go/src/net/fd_unix.go:408 +0x27c
net.(*TCPListener).AcceptTCP(0xc82006e068, 0xc820492b00, 0x0, 0x0)
	/usr/local/go/src/net/tcpsock_posix.go:254 +0x4d
net.(*TCPListener).Accept(0xc82006e068, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/tcpsock_posix.go:264 +0x3d
net/http/httptest.(*historyListener).Accept(0xc82027c750, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/http/httptest/server.go:48 +0x63
net/http.(*Server).Serve(0xc8202f4780, 0x7fcc444e8348, 0xc82027c750, 0x0, 0x0)
	/usr/local/go/src/net/http/server.go:1887 +0xb3
created by net/http/httptest.(*Server).Start
	/usr/local/go/src/net/http/httptest/server.go:109 +0x380

goroutine 65 [IO wait]:
net.runtime_pollWait(0x7fcc444e7f88, 0x72, 0xc820068080)
	/usr/local/go/src/runtime/netpoll.go:157 +0x60
net.(*pollDesc).Wait(0xc820122220, 0x72, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:73 +0x3a
net.(*pollDesc).WaitRead(0xc820122220, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:78 +0x36
net.(*netFD).Read(0xc8201221c0, 0xc820212000, 0x1000, 0x1000, 0x0, 0x7fcc444e2028, 0xc820068080)
	/usr/local/go/src/net/fd_unix.go:232 +0x23a
net.(*conn).Read(0xc8204980b8, 0xc820212000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/net.go:172 +0xe4
net/http.(*liveSwitchReader).Read(0xc8204120f8, 0xc820212000, 0x1000, 0x1000, 0xc820125000, 0x0, 0x0)
	/usr/local/go/src/net/http/server.go:219 +0xa4
io.(*LimitedReader).Read(0xc82058e820, 0xc820212000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
	/usr/local/go/src/io/io.go:427 +0xbd
bufio.(*Reader).fill(0xc8202abaa0)
	/usr/local/go/src/bufio/bufio.go:97 +0x1e9
bufio.(*Reader).Peek(0xc8202abaa0, 0x4, 0x0, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/bufio/bufio.go:132 +0xcc
net/http.(*conn).readRequest(0xc8204120b0, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/http/server.go:629 +0x2bb
net/http.(*conn).serve(0xc8204120b0)
	/usr/local/go/src/net/http/server.go:1319 +0x727
created by net/http.(*Server).Serve
	/usr/local/go/src/net/http/server.go:1910 +0x3f6

goroutine 44 [chan receive]:
github.com/weaveworks/scope/app.(*controlRouter).handleProbeWS(0xc820285bb0, 0x7fcc444e8560, 0xc820106420, 0xc820432000)
	/home/vagrant/src/github.com/weaveworks/scope/app/controls.go:117 +0x3c1
github.com/weaveworks/scope/app.(*controlRouter).(github.com/weaveworks/scope/app.handleProbeWS)-fm(0x7fcc444e8560, 0xc820106420, 0xc820432000)
	/home/vagrant/src/github.com/weaveworks/scope/app/controls.go:19 +0x3e
net/http.HandlerFunc.ServeHTTP(0xc820285fe0, 0x7fcc444e8560, 0xc820106420, 0xc820432000)
	/usr/local/go/src/net/http/server.go:1422 +0x3a
github.com/weaveworks/scope/vendor/github.com/gorilla/mux.(*Router).ServeHTTP(0xc820476a50, 0x7fcc444e8560, 0xc820106420, 0xc820432000)
	/home/vagrant/src/github.com/weaveworks/scope/vendor/github.com/gorilla/mux/mux.go:100 +0x29e
net/http/httptest.(*waitGroupHandler).ServeHTTP(0xc82012f200, 0x7fcc444e8560, 0xc820106420, 0xc820432000)
	/usr/local/go/src/net/http/httptest/server.go:200 +0xc2
net/http.serverHandler.ServeHTTP(0xc8202f5e00, 0x7fcc444e8560, 0xc820106420, 0xc820432000)
	/usr/local/go/src/net/http/server.go:1862 +0x19e
net/http.(*conn).serve(0xc820106370)
	/usr/local/go/src/net/http/server.go:1361 +0xbee
created by net/http.(*Server).Serve
	/usr/local/go/src/net/http/server.go:1910 +0x3f6

goroutine 87 [IO wait]:
net.runtime_pollWait(0x7fcc444e7ec8, 0x72, 0xc820068080)
	/usr/local/go/src/runtime/netpoll.go:157 +0x60
net.(*pollDesc).Wait(0xc82048c610, 0x72, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:73 +0x3a
net.(*pollDesc).WaitRead(0xc82048c610, 0x0, 0x0)
	/usr/local/go/src/net/fd_poll_runtime.go:78 +0x36
net.(*netFD).Read(0xc82048c5b0, 0xc820438000, 0x1000, 0x1000, 0x0, 0x7fcc444e2028, 0xc820068080)
	/usr/local/go/src/net/fd_unix.go:232 +0x23a
net.(*conn).Read(0xc8200cc0e0, 0xc820438000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/net.go:172 +0xe4
bufio.(*Reader).fill(0xc820064ba0)
	/usr/local/go/src/bufio/bufio.go:97 +0x1e9
bufio.(*Reader).Read(0xc820064ba0, 0xc8203e8be0, 0x2, 0x8, 0xc8203de960, 0x0, 0x0)
	/usr/local/go/src/bufio/bufio.go:207 +0x260
github.com/weaveworks/scope/vendor/github.com/gorilla/websocket.(*Conn).readFull(0xc82043a000, 0xc8203e8be0, 0x2, 0x8, 0x0, 0x0)
	/home/vagrant/src/github.com/weaveworks/scope/vendor/github.com/gorilla/websocket/conn.go:546 +0xa9
github.com/weaveworks/scope/vendor/github.com/gorilla/websocket.(*Conn).advanceFrame(0xc82043a000, 0xc81ffe1095, 0x0, 0x0)
	/home/vagrant/src/github.com/weaveworks/scope/vendor/github.com/gorilla/websocket/conn.go:570 +0x159
github.com/weaveworks/scope/vendor/github.com/gorilla/websocket.(*Conn).NextReader(0xc82043a000, 0xc8203ded50, 0x0, 0x0, 0x0, 0x0)
	/home/vagrant/src/github.com/weaveworks/scope/vendor/github.com/gorilla/websocket/conn.go:705 +0x76
github.com/weaveworks/scope/vendor/github.com/gorilla/websocket.(*Conn).ReadJSON(0xc82043a000, 0xe00fc0, 0xc8203ded50, 0x0, 0x0)
	/home/vagrant/src/github.com/weaveworks/scope/vendor/github.com/gorilla/websocket/json.go:45 +0x37
github.com/weaveworks/scope/xfer.(*JSONWebsocketCodec).ReadResponseHeader(0xc8203e5240, 0xc8203ded50, 0x0, 0x0)
	/home/vagrant/src/github.com/weaveworks/scope/xfer/controls.go:93 +0x55
net/rpc.(*Client).input(0xc8203c0960)
	/usr/local/go/src/net/rpc/client.go:109 +0xbf
created by net/rpc.NewClientWithCodec
	/usr/local/go/src/net/rpc/client.go:201 +0xd2

goroutine 86 [runnable]:
syscall.Syscall(0x0, 0x6, 0xc8203f1000, 0x1000, 0x81, 0x1000, 0x0)
	/usr/local/go/src/syscall/asm_linux_amd64.s:18 +0x5
syscall.read(0x6, 0xc8203f1000, 0x1000, 0x1000, 0x72, 0x0, 0x0)
	/usr/local/go/src/syscall/zsyscall_linux_amd64.go:783 +0x5f
syscall.Read(0x6, 0xc8203f1000, 0x1000, 0x1000, 0xffffffffffffff01, 0x0, 0x0)
	/usr/local/go/src/syscall/syscall_unix.go:160 +0x4d
net.(*netFD).Read(0xc820157730, 0xc8203f1000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/fd_unix.go:228 +0x18b
net.(*conn).Read(0xc82006e0f0, 0xc8203f1000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/net.go:172 +0xe4
bufio.(*Reader).fill(0xc8203c08a0)
	/usr/local/go/src/bufio/bufio.go:97 +0x1e9
bufio.(*Reader).ReadSlice(0xc8203c08a0, 0xc81ffd4a0a, 0x0, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/bufio/bufio.go:328 +0x21a
bufio.(*Reader).ReadLine(0xc8203c08a0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/bufio/bufio.go:357 +0x53
net/textproto.(*Reader).readLineSlice(0xc8200c1820, 0x0, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/textproto/reader.go:55 +0x81
net/textproto.(*Reader).ReadLine(0xc8200c1820, 0x0, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/textproto/reader.go:36 +0x40
net/http.ReadResponse(0xc8203c08a0, 0xc8203b3180, 0xc82006e0f0, 0x0, 0x0)
	/usr/local/go/src/net/http/response.go:128 +0xf6
github.com/weaveworks/scope/vendor/github.com/gorilla/websocket.(*Dialer).Dial(0xc8203e0f50, 0xc8203dec30, 0x23, 0xc8203debd0, 0x0, 0x0, 0x0, 0x0)
	/home/vagrant/src/github.com/weaveworks/scope/vendor/github.com/gorilla/websocket/client.go:304 +0x156b
github.com/weaveworks/scope/xfer.(*appClient).controlConnection(0xc82056af30, 0x7fcc45b34370, 0xc8203e8840, 0x0, 0x0)
	/home/vagrant/src/github.com/weaveworks/scope/xfer/app_client.go:90 +0x158
github.com/weaveworks/scope/xfer.(*appClient).controlConnectionLoop(0xc82056af30, 0x7fcc45b34370, 0xc8203e8840)
	/home/vagrant/src/github.com/weaveworks/scope/xfer/app_client.go:130 +0x168
created by github.com/weaveworks/scope/xfer.(*appClient).ControlConnection
	/home/vagrant/src/github.com/weaveworks/scope/xfer/app_client.go:150 +0x49
FAIL	github.com/weaveworks/scope/app	0.484s
```

tomwilkie:
Fixes #365, Fixes #648

UI Todo:
- [x] Popout term to new browser window.
- [x] Handle `ESC` gracefully when not-popped out.
- [x] Handle multiple presses of the attach button. (For popped out window case too).
- [x] ~~Handle presses of other buttons? (Close on pause or restart?).~~
- [x] Close w/ node-details panel (close w/ popped out window too?)
- [x] Highlighting stuff in the terminal window (to copy-paste) immediately deselects my selection, which is a bit annoying.
- [x] Persist pipes/terminals across UI refresh (in the hash fragment)
- [x] Handle error codes.
- [x] Store rawTty in state hash.
- [x] Fix error handling in the popped out window.

Backend Todo:
- [x] Pipes that don't have a shell on the other end are garbled
- [x] Add docker exec control.
- [x] Test for all the timeout logic.
- [x] Make close propagate from the probe -> UI.
- [x] Timeout pipes for probes that disappear.


xfer/controls.go
```diff
@@ -20,6 +22,35 @@ type Response struct {
 	ID    int64
 	Value interface{}
 	Error string
+	Pipe  int64
```
tomwilkie: Json tags here.

tomwilkie:
I'll put all the feedback in one comment for now:

- x to close the terminal window is on the right hand side for the details panel; would be nice to be consistent.
- cannot drag window around by the text area on the title.
- want a way to 'pop out' the terminal into another window/tab
- resize is broken (doesn't seem to do the right thing if you load up a vi, for instance).  Might just want to disable it for now.
- should we offer a way to minimize the windows?  closing and reopening them doesn't do the right thing, as the contents is lost, and the new terminal doesn't seem to work.
- I can't drag the terminal out beyond the outer window edge, which seems counter-intuitive (with normal windows on the desktop, for instance)

In general I'm not super-keen on the windows-within-windows approach, maybe you should put this in front of @pidster, @rade and @monadic to see what they think?

Also we need to nail down the lifecycle of these things.  I'm thinking:
- When user closes the window, UI should DELETE the /api/pipe/foo - this will cause the pipe to get closed all the way down the stack
- UI<->App reconnects should work fine (as the app won't delete the pipe when you disconnect, it will just buffer up stuff)
- There are buffers at two stages in the process - one in the app for data coming from the probe, and one in the probe from data coming from the app.  If either buffer hits a limit (a few KB I think), I will make it tear down the pipe.  In this case, I will close the UI<->App ws.  You will need to retry the ws when ever it the connection dropped, and if you get a 404, you'll need to show an error to the user.
- If probe<->app communication is disrupted, the associated pipes will be destroyed.  This should deal with apps or probes going away.  You will see your ws get closed, and then a 404 on retry.

Note with this I'm not proposing we garbage collect pipes at any stage - so if you open a pipe and then never use it, it will stay around for ever.  They don't consume much resource, so this shouldn't be a problem.

What do you think? (@paul would be good to get your input too)

tomwilkie:
Also, if you stop scope underneath the UI, it resized an open terminal very small.  If you put try to reconnect (maybe with a 'reconectting...' overlay on the terminal) with a backoff, and if you get an error code (404, as opposed to bad hostname or can't connect) it should show an error ("this pipe has gone away.  Sooooory")

tomwilkie:
Another thing: click-and-drag doesn't move the window to the foreground - need to move it to foreground on mouse down I think.

tomwilkie:
And a nice-to-have would be some kind of tiling window management...  or something like the aero-snap feature.

davkal:
Comprehensive feedback!
Before implementing a full-blown window manager, I'd like to focus on correctness of display and input being sent, as well as reconnection issues.
When that's working reliably, I'd like to show it to the rest of the team for further UI feedback.
Re window manager, I think we should keep it rudimentary but allow a popout to a new tab.

tomwilkie:
I've talked to @pidster, @bboreham, @awh, @dpw, @monadic and @rade about the window-manger-in-a-browser-window 'feature' for feedback, which summarized was:
- having the terminal 'pop out' into a separate tab is a must have.
- only having the pop out (ie ditching the current window-in-a-window approach) was not desirable, as it would be 'surprising' to the users (everyone agreed here)
- @pidster suggested instead of a movable, overlappable set of windows we only allow a single window to be open at once, and it wouldn't be moveable, and would resize with the browser window:

screen shot 2015-11-30 at 15 29 15

This way, if the user wants to have multiple terminals (or topology view) side-by-side, they have to use their own windows manager and separate scope tabs/windows.  Everyone I spoke to (except @monadic) preferred this idea.  Note it doesn't involved us implementing a window manager ourselves, and fits nicely with the 'window management' of the details panel.
- a final suggestions (from @monadic) was not not allow them to move or resize, but fix them to the left hand side of the window, maybe allowing some accordion-style minimisation.

paulbellamy:
re: lifecycle

I'm not sure about having a buffer overflow teardown the pipe. Some sort of pipe-specific backpressure would be better, but not sure on the feasibility of implementing that in probe/UI.

I'd rather have some sort of timeout/teardown for unused pipes, with not connections open. Even if they're not active. It's just bad form to leave them around.

Other than that, sounds roughly sane.

common/bufferedpipe/bufferedpipe.go
```diff
@@ -0,0 +1,49 @@
...
+func New() io.ReadWriteCloser {
+	r, w := io.Pipe()
+	p := &pipe{
+		bufs:   make(chan []byte, 100),
+		Reader: r,
+		closer: r,
+	}
+	go p.loop(w)
+	return p
+}
```
paulbellamy: I guess this has slightly different behaviour than `bufio.NewWriter(w)`, correct? In that it's never explicitly flushed...

common/bufferedpipe/bufferedpipe.go
```diff
@@ -0,0 +1,49 @@
...
+func (p *pipe) Write(b []byte) (int, error) {
+	// copy the buffer
+	buf := make([]byte, len(b))
+	copy(buf, b)
+
+	select {
+	case p.bufs <- buf:
+		return len(buf), nil
+	default:
+		return 0, fmt.Errorf("Pipe full")
```
paulbellamy: Is there a standard error we can use here? Or variable-ize this for comparisons...

tomwilkie:
After discussion with @Paul, I've been convinced the previous pipe-lifecycle design was floored and we need to introduce some for of back pressure to ensure `cat bigfile` doesn't cause pipes to close.  Initial ideas where:
- make pipe read/writes blocking functions calls across the WS using the existing control/rpc mechanisms.  We believe this would make the keyboard echo unbearable on high latency connections.
- implement some kind of fixes window of pipe ios and sequence numbers, blocking sending of new ones when the window fills up, and adding messages to the WS to ack up to a certain sequences number.  Basically reimplement TCP.
As neither of these two ideas are particularly attractive, @Paul & I think going for a separate WS from the app <-> probe per pipe is the best way forward.  We we then make all the operations (read/write) blocking, and rely on the OS's TCP stack for back pressure etc.

This brings us on to the pipe lifecycle.  Again, @Paul convinced me we need to timeout pipes that are unused.  This need to deal with App and Probe restarts (not leaking resources in either), but also long lived idle pipes - we want pipes to stay open when you go home for the night, for instance.
- Timeouts will be done by the App, based on no UI connected to the pipe for some period (10mins?).
- If the app gets a connection from the probe for a pipe it doesn't know about, it will 'create' a pipe for it.  If it gets a connections for a pipe it doesn't know about from the UI (or one that has been deleted), it will 404.
- Deleting / timing out a pipe will mark it as 'tombstoned' in the App, and close any connections to the pipe (from both the UI and the App).
- The probe will gracefully deal with dropped connections to the app, reconnecting if necessary.
- When the probe reconnects to a pipe, if the pipe is known and tombstoned, it will get a 404.  The probe will treat 404 as 'this pipe has been deleted' and stop the pipe locally.

This should deal nicely with app restarts (pipes will eventually get reconnected) and probe restarts (pipes will timeout in the app).

peterbourgon:
> we want pipes to stay open when you go home for the night, for instance

This seems incredibly odd to me. Pipes are a quick-and-easy way to get at specific parts of your infrastructure, ephemeral and not long-lived.

tomwilkie:
> I'd rather expect the moment a user closes the UI element that spawned the pipe,

Yes we'll add this as well, using the same mechanism for timeouts.  Sorry that wasn't clear.

tomwilkie:
Just did the demo with @squaremo and he had the suggestion that ctrl-click on the attach button should open in new window.  Otherwise I think he was 50/50 between option (3) & (4).

squaremo:
> Otherwise I think he was 50/50 between option (3) & (4).

I think

 1. things you can move and things you can't move should look different;
 2. MDI is awful

so that leads me to suggest that the info pane is attached to the side of the window, and the console is attached to that.

foot:
- [ ] Popout term to new browser window.
- [ ] Handle <ESC> gracefully.
- [ ] Handle multiple presses of the attach button
- [ ] Handle presses of other buttons? (Close on pause or restart?).

foot:
What happens if we fail to send the DELETE /api/pipe/:pid in some cases? (E.g. user quitting the browser). Time's out after a few minutes and no hard done?

paulbellamy:
Or, user hitting the back button...

tomwilkie:
Yes, we'll timeout the pipes after 10mins or so.  I think that was in the design.

foot:
Came across a nifty API in [chrome/firefox (and can be reasonably polyfilled elsewhere)](http://caniuse.com/#feat=beacon) for sending fire-and-forget requests as the browser is unloading the page: [`sendBeacon`](http://w3c.github.io/beacon/)! Played with it this morning and it works as advertised.

It has a very simple api: `sendBeacon(url, data)` and only `POST`s. Could we add a POSTable endpoint for cleaning up pipes? Something terrible like this perhaps:
 - `POST /api/pipe/:pid/_delete`.

paulbellamy:
I'd suggest `POST /api/pipe/:pid?_method=delete`

foot:
+1 Good call much nicer.

app/pipes.go
```diff
@@ -0,0 +1,174 @@
...
+func (pr *pipeRouter) timeoutPipes() {
+	pr.Lock()
+	defer pr.Unlock()
+	now := time.Now()
+	for id, lastPut := range pr.lastPut {
+		if now.Sub(lastPut) > pipeTimeout {
+			log.Printf("Timeing out pipe %s", id)
+			delete(pr.lastPut, id)
+			pipe := pr.pipes[id]
+			pipe.Close(now)
```
paulbellamy: Can this hang?

app/pipes.go
```diff
@@ -0,0 +1,174 @@
...
+	now := time.Now()
+	for pipeID, pipe := range pr.pipes {
+		if pipe.Closed() && now.Sub(pipe.ClosedAt()) > gcTimeout {
+			log.Printf("GCing pipe %s", pipeID)
+			delete(pr.pipes, pipeID)
+		}
+	}
+}
+
+func (pr *pipeRouter) handleUIPipeWS(w http.ResponseWriter, r *http.Request) {
```
paulbellamy: It would be nice to combine this with `handleProbePipeWS`

app/pipes.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func (pr *pipeRouter) deletePipe(w http.ResponseWriter, r *http.Request) {
+	var (
+		vars   = mux.Vars(r)
+		pipeID = vars["pipeID"]
+	)
+	pipe, ok := pr.getOrCreatePipe(pipeID, true)
+	if !ok {
+		http.NotFound(w, r)
```
paulbellamy: Delete should probably be idempotent.

app/pipes.go
```diff
@@ -0,0 +1,174 @@
...
+	pipe, ok := pr.getOrCreatePipe(pipeID, true)
+	if !ok {
+		log.Printf("Pipe %s is not connected right now...", pipeID)
+		http.NotFound(w, r)
+		return
+	}
+	defer func() {
+		pr.Lock()
+		defer pr.Unlock()
+		pr.lastPut[pipeID] = time.Now()
```
paulbellamy: My understanding of this, is it will timeout any pipes 10 minutes after a UI disconnects (so long as there has not been another connection established since then). Correct?

Is there a case like this?
- ui1 connects
- ui2 connects
- ui1 disconnects
- pipe is gced and ui2 is booted

client/app/scripts/components/terminal.js
```diff
@@ -0,0 +1,180 @@
...
+      term.write(input);
+    };
+
+    this.term = term;
+    this.socket = socket;
+
+    const {pixelPerCol, pixelPerRow} = terminalCellSize(
+      innerNode, this.state.rows, this.state.cols);
+
+    window.addEventListener('resize', this.handleResize, this);
```
paulbellamy: Isn't the third argument to `addEventListener` a boolean?

app/pipes.go
tomwilkie: There is that case, however the implicit assumption throughout the whole PR is that there is only one UI connected to each pipe.  If you managed to connect two, all hell will break loose.

client/app/scripts/components/terminal.js
foot: Yes! Good catch.

client/app/scripts/stores/app-store.js
```diff
@@ -303,8 +321,11 @@ export class AppStore extends Store {
       break;
 
     case ActionTypes.HIT_ESC_KEY:
-      deSelectNode();
-      this.__emitChange();
+      // disable ESC when terminals are open
+      if (!controlPipes.size) {
```
paulbellamy: `if (controlPipes.size > 0) {`, please... (Or whatever this means)

paulbellamy:
Repeatedly pressing the "pipe" button leads to a stack of pipes with garbled output:
screen shot 2015-12-03 at 13 35 04

You can't quite see the semi-transparent output of the previous pipe behind this one.

client/app/scripts/components/terminal.js
paulbellamy: Same for removeEventListener.

paulbellamy:
Actually, in general input/output is getting garbled between pipes:

screen shot 2015-12-03 at 13 38 48

tomwilkie:
Yeah thats got nothing to do with repeated pipes, its 'just' pipes that don't have a shell on the other end are garbled.  If you do `docker run -dti alpine /bin/sh` it works.  We do need to get to the bottom of this though.

app/pipes.go
paulbellamy: Ok, I'm just thinking of a refresh/reconnect, or such. Not sure if it is an issue or not...

probe/controls/pipes.go
```diff
@@ -0,0 +1,34 @@
...
+	// Client is the thing the probe uses to make pipe connections.
+	Client PipeClient
+)
+
+// PipeClient is the thing the probe uses to make pipe connections.
+type PipeClient interface {
+	PipeConnection(string, xfer.Pipe, io.ReadWriter) error
+}
+
+// NewPipe destroys pipes.
```
paulbellamy: wat?

app/pipes.go
tomwilkie: Yeah I should protect against this case.  Maybe keep reference counts for 'open' pipes instead?

probe/controls/pipes.go
```diff
@@ -0,0 +1,34 @@
...
+const (
+	maxOutstandingPIOs = 32
+)
+
+var (
+	// Client is the thing the probe uses to make pipe connections.
+	Client PipeClient
+)
+
+// PipeClient is the thing the probe uses to make pipe connections.
```
paulbellamy: Is `PipeClient` "the thing the probe uses to make pipe connections", or is that `Client`?

probe/controls/pipes.go
tomwilkie: Oh yeah... me getting annoyed at stoopid comments we have to write for golint.

probe/docker/controls.go
```diff
@@ -15,69 +17,76 @@ const (
...
+			Stream:       true,
+			Stdin:        true,
+			Stdout:       true,
+			Stderr:       true,
+			InputStream:  pipe.Starboard(),
+			OutputStream: pipe.Starboard(),
+			ErrorStream:  pipe.Starboard(),
+		}); err != nil {
+			log.Printf("Error from docker attach: %v", err)
+		}
```
paulbellamy: Should probably return this error to the requester somehow, no? Could we do this synchonously and return the error up front?

xfer/pipes.go
```diff
@@ -0,0 +1,143 @@
...
+	Close(time.Time)
+	Closed() bool
+	ClosedAt() time.Time
+	OnClose(func())
+}
+
+type pipe struct {
+	mtx             sync.Mutex
+	id              string
+	port, starboard io.ReadWriter
```
paulbellamy: Not a fan of these names (especially `port`, as it's overloaded enough)

probe/docker/registry.go
```diff
@@ -56,11 +56,13 @@ type Client interface {
 	ListImages(docker_client.ListImagesOptions) ([]docker_client.APIImages, error)
 	AddEventListener(chan<- *docker_client.APIEvents) error
 	RemoveEventListener(chan *docker_client.APIEvents) error
+
 	StopContainer(string, uint) error
 	StartContainer(string, *docker_client.HostConfig) error
 	RestartContainer(string, uint) error
 	PauseContainer(string) error
 	UnpauseContainer(string) error
+	AttachToContainer(docker_client.AttachToContainerOptions) error
```
paulbellamy: Would it make sense to model "deleting a pipe" as "DetachContainer"?

xfer/pipes.go
```diff
@@ -0,0 +1,143 @@
...
+		},
+		quit: make(chan struct{}),
+	}
+}
+
+func (p *pipe) ID() string {
+	return p.id
+}
+
+func (p *pipe) Port() io.ReadWriter {
```
paulbellamy: What about `func (p *pipe) Ends() (io.ReadWriter, io.ReadWriter)`, and leave naming the ends up to the caller?

xfer/pipes.go
```diff
@@ -0,0 +1,143 @@
...
+	return p.port
+}
+
+func (p *pipe) Starboard() io.ReadWriter {
+	return p.starboard
+}
+
+func (p *pipe) Close(t time.Time) {
```
paulbellamy: How do you feel about the ends being a `ReadWriteCloser`, and set the tombstoneTime in `NewPipe(id, tombstoneTime)` ?

xfer/pipes.go
```diff
@@ -0,0 +1,143 @@
...
+func (p *pipe) ClosedAt() time.Time {
+	return p.tombstone
+}
+
+func (p *pipe) OnClose(f func()) {
+	p.onClose = f
+}
+
+// CopyToWebsocket copies pipe data to/from a websocket.  It blocks.
+func (p *pipe) CopyToWebsocket(pipe io.ReadWriter, conn *websocket.Conn) error {
```
paulbellamy: This function signature seems muddled. Why is pipe taking another pipe?

Does this need to be a method on pipe? Isn't it just "Copy this readwriter to a websocket"?

xfer/pipes.go
```diff
@@ -0,0 +1,143 @@
...
+			io.Reader
+			io.Writer
+		}{
+			r2, w1,
+		},
+		quit: make(chan struct{}),
+	}
+}
+
+func (p *pipe) ID() string {
```
paulbellamy: Not really sure why pipe needs to track it's own ID? Doesn't seem like it is the pipe's concern, only the caller's.

paulbellamy:
Have done a first pass. Needs another go over (particularly the JS, as I haven't fully understood it).

paulbellamy:
It seems like the "being attached to a pipe" doesn't modify the url fragment (meaning attachments go away on refresh). Should it?

paulbellamy:
re: garbled output, the messages when connecting seem quite badly formatted (added the `[]` at the end for the cursor):
```
W<probe> 2015/12/03 14:11:40 Pipe pipe-4116411269973356742 connecting to 127.0.0
.1:4040                                                                         
       D<app> 2015/12/03 14:11:40 Creating pipe id pipe-4116411269973356742     
                                                                           D<app
> 2015/12/03 14:12:02 Timeing out pipe pipe-5753019419086106492                 
                                                                <probe> 2015/12/
03 14:12:02 Error reading websocket for pipe pipe-5753019419086106492: read tcp 
127.0.0.1:41606->127.0.0.1:4040: read: connection reset by peer                 
                                                               µ<probe> 2015/12/
03 14:12:02 Closing pipe pipe-5753019419086106492 connection to 127.0.0.1:4040  
                                                                              <p
robe> 2015/12/03 14:12:02 Error doing pipe pipe-5753019419086106492 for 127.0.0.
1:4040, backing off 1s: read tcp 127.0.0.1:41606->127.0.0.1:4040: read: connecti
on reset by peer                                                                
                <probe> 2015/12/03 14:12:02 Error writing websocket for pipe pip
e-5753019419086106492: write tcp 127.0.0.1:41606->127.0.0.1:4040: use of closed 
network connection                                                              
                  W<probe> 2015/12/03 14:12:03 Pipe pipe-5753019419086106492 con
necting to 127.0.0.1:4040                                                       
                         L<app> 2015/12/03 14:12:03 Pipe pipe-575301941908610649
2 has been deleted...                                                           
                     _<probe> 2015/12/03 14:12:03 Pipe pipe-5753019419086106492 
connection to 127.0.0.1:4040 exiting                                            
                                                                           []
```

paulbellamy:
Highlighting stuff in the terminal window (to copy-paste) immediately deselects my selection, which is a bit annoying.

paulbellamy:
As a user "Timeing out" (spelling), arguably is something I care about, but GCing it is definitely not.

paulbellamy:
When I try to connect to a container it seems to hang on connecting.

paulbellamy:
Ah, so this is doing a "container attach" not a "shell exec", right? Won't that cause the container to exit when I detach?

xfer/pipes.go
tomwilkie: Yes, it doesn't really.  I found it useful for logging statements, although I can get rid of those too now.

xfer/pipes.go
tomwilkie: Indeed the only use was for logging, which I just removed.  Will fix.

tomwilkie: Ah no its need the pipe reference so it can unblock when the pipe is closed.

xfer/pipes.go
tomwilkie: Thats a nice idea.

probe/docker/registry.go
tomwilkie: We could do that on top of everything we have, but we'll always need to deal with the case of the UI going away with an open pipe.  I think I'd prefer to leave it as the generic mechanism for now.

probe/docker/controls.go
tomwilkie: Not sure how? AttachToContainer blocks; we could put the error down a pipe and return it to the client, but we'd need to put a timeout on it.

Its an unfortunate API, as some of the errors will get returned immediately, and some will only get returned after the attach has exited.

app/pipes.go
tomwilkie: Yes would be nice, but I can't think of a way that isn't more horrific than having two similar functions.  The key differences are (1) they talk to different ends of the pipe and (2) one does reference counting to trigger pipe timeouts.

app/pipes.go
tomwilkie: I don't think so - it just closes a channel.

xfer/pipes.go
paulbellamy: Ok, so I would prefer:
```Go
func CopyToWebsocket(pipe io.ReadWriter, conn *websocket.Conn, cancel <-chan struct{}) error
```

not sure I have a better name for this than `CopyXXXXXXX` which sounds similar to `io.Copy`, except it does a two-way wiring, not just one-way. Which makes the name seem slightly misleading.

xfer/pipes.go
paulbellamy: If the ends are `ReadWriteCloser`s does that let us simplify any of the usage? It seems like it would *almost* let us use `io.Copy` for wiring it up to the websockets...

tomwilkie: re: tombstoneTime, I'm not sure what you mean?  When that field is set, the pipe is considered closed.

Also not clear where calling close on one end of the pipe would help.

app/pipes.go
tomwilkie: Done.

app/pipes.go
tomwilkie: Added reference counting.

xfer/pipes.go
paulbellamy: I was expecting that calling close on either end would close the entire thing.

I was using `tombstoneTime` to refer to the duration that the tombstone would exist for, but looking now, that's never referenced by the pipe, so wouldn't need to be passed in.

Again, not sure it's a good idea, but my hunch is that it seems like it *should* simplify usage.

Total strawman:
```Go
func duplex(a, b io.ReadWriteCloser, cancel <-chan struct{}) error {
  // This implementation is totally made up
  errs := make(chan error)

  // close everything when one exists
  defer a.Close()
  defer b.Close()

  // copy everything
  go func() { errs <- io.Copy(a, b) }()
  go func() { errs <- io.Copy(b, a) }()

  // wait for one to exit
  select {
  case err := <-errs:
    return err
  case <-cancel:
    return nil
  }
}

// in the probe, maybe:
cancel := make(chan struct{})
p := NewPipe()
local, remote := p.Ends()
go duplex(local, docker, cancel)
go duplex(remote, websocket, cancel)
p.Wait() // Wait for pipe to be closed. Pipe is closed when either end is closed.
```

Maybe that won't work, though...

tomwilkie: That would work, except a websocket connection is not a ReadWriteCloser.

paulbellamy: Could it be adapted to one? Would that be a colossal pain?

foot:
Sorry I just realised I haven't understood this aspect of the pipes very well!: Is there any state in the pipe?

E.g. If a browser loses connectivity and this causes the websocket to be disconnected, is there a benefit to trying to re-establish a connection to the same pipe vs. DELETEing it and creating a new one? stdout for the disconnected pipe is buffered by scope-app perhaps?

Follow up question, does this connection lifecycle stuff on the client sounds okay:

1. As long as a terminal window stays open we will try and reconnect to the same pipe until we receive a 404.

2. When a terminal window is closed in any way, send a DELETE and abandon the pipe. Closing a window can be triggered by:
 - the terminal "x" button being pressed.
 - the browser back button being pressed.
 - the browser forward button being pressed.
 - the browser refresh button being pressed.

tomwilkie:
> Is there any state in the pipe?

No not really; there might be pending output buffered up, but not state per-se.

> is there a benefit to trying to re-establish a connection to the same pipe vs. DELETEing it and creating a new one?

Yes I think so - the commands to create a pipe aren't necessarily idempotent (exec won't be), so doing the same command twice might give you a different shell, with no scroll back, for instance.

Whats more, the browser terminal keeps state (previously typed commands) and you want to preserve that if possible, ie you don't want the screen to occasionally turn black,

> does this connection lifecycle stuff on the client sounds okay

That looks fine, although I'm not sure about refreshing the browser window - should this connect to the same pipe or not?  I don't think you should ever reissue the command to recreate a pipe.

foot:
> the commands to create a pipe aren't necessarily idempotent (exec won't be), so doing the same command twice might give you a different shell, with no scroll back, for instance.

`attach` might actually give you different shells? Or if we implement some future functionality that uses `exec` we might get different shells?

> Whats more, the browser terminal keeps state (previously typed commands) and you want to preserve that if possible, ie you don't want the screen to occasionally turn black,

Agreed!

> I'm not sure about refreshing the browser window - should this connect to the same pipe or not?

We can't really differentiate between the browser being closed and the page being refreshed. The JS env is being destroyed in both cases, in the 2nd its immediately recreated. Cleanest way to handle it would be to DELETE the pipe in both cases, and create a new one as we load up after a refresh if the url-state says we should. (So we'd store a `{terminal: nodeId}` rather than `{terminal: pipeId}` to create a new pipe for that node).

tomwilkie:
> attach might actually give you different shells? Or if we implement some future functionality that uses exec we might get different shells?

attach won't, but exec will.  And exec will be implemented before this PR gets merged, so this isn't hypothetical future functionality.

> We can't really differentiate between the browser being closed and the page being refreshed....

True, but that breaks the never-issue-the-command-to-create-a-pipe-again edict.  So I guess the options are:
1. don't close pipes when the user leaves the page, include the pipe id in the hash fragment and reattach said pipes, or
2. close pipes when the user leaves the page, don't include anything in the has fragment and don't reopen terminal windows when the user refreshes the page (ie they disappear)

Given we have code for timing out pipes if they aren't closed, (1) doesn't sound too bad.  (2) sounds surprising from a users pov.

foot:
> exec will be implemented before this PR gets merged

ahh

> the never-issue-the-command-to-create-a-pipe-again edict

Ah hah, this is for this `exec` use case?

> 1) don't close pipes when the user leaves the page, include the pipe id in the hash fragment and reattach said pipes.

Yeah lets give this a go. If we get too many old pipes lying around we can try.. something else. But Go is good at this stuff I think.

tomwilkie:
Roger roger.

peterbourgon:
> 1. don't close pipes when the user leaves the page, include the pipe id in the hash fragment and reattach said pipes, or
> 2. close pipes when the user leaves the page, don't include anything in the has fragment and don't reopen terminal windows when the user refreshes the page (ie they disappear)

I'm sorry to pipe in here unannounced (drums) but I find option 1 very surprising, and strongly prefer option 2. Pipes are fragile by their nature and should be understood as ephemeral. And what happens if a user shares a URL with a pipe fragment—we can't mux users on a single pipe!

foot:
Good points! I'm leaning towards option 2 now.

Option 1 has a lot of "wow" factor I think.
Option 2 is very simple and yeah perhaps a little less surprising in the end.

foot:
David's gonna PR w/ the already implemented Option 1 and then we'll chat about simplifying it down to Option 2.

foot:
@tomwilkie Working really nicely!

Only issue I've come across is that 404s on deleted pipes don't hang around very long and we don't get them for pipes that have timed out.

In the UI we might try and reconnect to a deleted/timed-out pipe and will get a "successful" websocket connection (instead of a 404) that doesn't respond to anything as the pipe doesn't actually exist anymore.

Do you think it would be an easy fix to get /api/pipe/foo to respond with a 404 for example?

tomwilkie:
> Only issue I've come across is that 404s on deleted pipes don't hang around very long and we don't get them for pipes that have timed out.

Yes, its only set for 1 minute.  I can increase that.

> In the UI we might try and reconnect to a deleted/timed-out pipe and will get a "successful" websocket connection (instead of a 404) that doesn't respond to anything as the pipe doesn't actually exist anymore.

Yes thats right - after a minute that pipe will also timeout and start returning 404

This is the design I'm afraid - I can increase the tombstone retention period, but beyond that we haven't come up with a better scheme.

paulbellamy:
This PR is far too large.

app/pipes.go
```diff
@@ -0,0 +1,194 @@
...
+	router.Methods("GET").Path("/api/pipe/{pipeID}").HandlerFunc(pipeRouter.handleWs(func(p *pipe) (*end, io.ReadWriter) {
+		uiEnd, _ := p.Ends()
+		return &p.ui, uiEnd
+	}))
+	router.Methods("GET").Path("/api/pipe/{pipeID}/probe").HandlerFunc(pipeRouter.handleWs(func(p *pipe) (*end, io.ReadWriter) {
+		_, probeEnd := p.Ends()
+		return &p.probe, probeEnd
+	}))
+	router.Methods("DELETE").Path("/api/pipe/{pipeID}").HandlerFunc(pipeRouter.deletePipe)
+	router.Methods("POST").Path("/api/pipe/{pipeID}").HandlerFunc(pipeRouter.deletePipe)
```
paulbellamy: Let's multiline these, please (just so they don't wrap massively):
```Go
router.
  Methods("POST").
  Path("/api/pipe/{pipeID}").
  HandlerFunc(pipeRouter.deletePipe)
```

app/pipes.go
```diff
@@ -0,0 +1,194 @@
...
+
+type pipe struct {
+	ui, probe     end
+	tombstoneTime time.Time
+
+	xfer.Pipe
+}
+
+// RegisterPipeRoutes registers the pipe routes
+func RegisterPipeRoutes(router *mux.Router) *PipeRouter {
```
paulbellamy: It seems like this function conflates `NewPipeRouter()`, and `pipeRouter.RegisterRoutes(mux)`. Maybe that's ok in this case... Thoughts?

paulbellamy:
in `xfer/pipes.go`, `pipe.Close` will call the `onClose` callback every time, even if the pipe was already previously closed. Presumably the user only wants `onClose` to happen once (when the pipe is closed).

tomwilkie:
> in xfer/pipes.go, pipe.Close will call the onClose callback every time, even if the pipe was already previously closed. Presumably the user only wants onClose to happen once (when the pipe is closed).

Oh no it wont - the function's local onClose is only assigned it the pipe is open.

app/pipes.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		pr.pipes[id] = p
+	}
+	if p.Closed() {
+		return nil, false
+	}
+	return p, true
+}
+
+func (pr *PipeRouter) getPipeRef(id string, pipe *pipe, end *end) bool {
```
paulbellamy: There might be a better name for `getPipeRef` and `putPipeRef`...  Maybe `retainPipe` and `releasePipe`?

paulbellamy: Also, do we ever call `getOrCreatePipe` and *not* follow it by `getPipeRef`?

app/pipes.go
```diff
@@ -0,0 +1,194 @@
...
+		if (pipe.ui.refCount == 0 && now.Sub(pipe.ui.lastUsedTime) >= pipeTimeout) ||
+			(pipe.probe.refCount == 0 && now.Sub(pipe.probe.lastUsedTime) >= pipeTimeout) {
+			log.Printf("Timing out pipe %s", id)
+			pipe.Close()
+			pipe.tombstoneTime = now
+		}
+	}
+}
+
+func (pr *PipeRouter) gcPipes() {
```
paulbellamy: What would a `PipeRouter` GC, if not pipes? (Same for timeout)

Edit: And a lot of the other methods in here.

app/controls_test.go
```diff
@@ -47,7 +41,13 @@ func TestControl(t *testing.T) {
 		return xfer.Response{
 			Value: "foo",
 		}
-	}))
+	})
+	client, err := xfer.NewAppClient(probeConfig, ip+":"+port, ip+":"+port, controlHandler)
+	if err != nil {
+		t.Fatal(err)
+	}
+	client.ControlConnection()
```
paulbellamy: Why not roll this into `xfer.NewAppClient`?

probe/controls/controls.go
```diff
@@ -18,15 +17,10 @@ func HandleControlRequest(req xfer.Request) xfer.Response {
 	handler, ok := handlers[req.Control]
 	mtx.Unlock()
 	if !ok {
-		return xfer.Response{
-			ID:    req.ID,
-			Error: fmt.Sprintf("Control '%s' not recognised", req.Control),
-		}
+		return xfer.ResponseErrorf("Control '%s' not recognised", req.Control)
```
paulbellamy: `%q` is probably what you want...

probe/controls/pipes.go
```diff
@@ -0,0 +1,48 @@
...
+var Client interface {
+	PipeConnection(string, string, xfer.Pipe) error
+	PipeClose(string, string) error
+}
+
+// Pipe the probe-local type for a pipe, extending
+// xfer.Pipe with the appID and a custom closer method.
+type Pipe interface {
+	xfer.Pipe
+}
```
paulbellamy: This comment doesn't seem to match the code?

probe/docker/container.go
```diff
@@ -327,7 +332,9 @@ func (c *container) GetNode(hostID string, localAddrs []net.IP) report.Node {
 	if c.container.State.Paused {
 		result = result.WithControls(UnpauseContainer)
 	} else if c.container.State.Running {
-		result = result.WithControls(RestartContainer, StopContainer, PauseContainer)
+		result = result.WithControls(
+			RestartContainer, StopContainer, PauseContainer, AttachContainer, ExecContainer,
```
paulbellamy: I'm assuming this is unrelated to the display order in the UI?

prog/probe/main.go
```diff
@@ -101,12 +101,19 @@ func main() {
...
-	}, xfer.ControlHandlerFunc(controls.HandleControlRequest), xfer.NewAppClient)
+	}
+	clients := xfer.NewMultiAppClient(func(hostname, endpoint string) (xfer.AppClient, error) {
+		return xfer.NewAppClient(
+			probeConfig, hostname, endpoint,
+			xfer.ControlHandlerFunc(controls.HandleControlRequest),
+		)
+	})
 	defer clients.Stop()
+	controls.Client = clients
```
paulbellamy: Setting a global var cross-packages is a bit naff.

paulbellamy:
Could put `xfer/pipes.go` into `xfer/pipe/pipe.go`, giving you `pipe.New()`. Also, why do you need the `Pipe` interface, if there's only ever one implementation?

paulbellamy:
Resize and text wrapping still seem broken.

paulbellamy:
re: onClose

Ah, I'd missed that, yeah.  It should still probably only exec once close is fully finished? or do we care?

davkal:
First round of UI feedback. Will look at code after second round.

1. increase margin to details panel. to me it clashes with the proportions of the outer margins (i'd recommend half the size) and it's in the uncanny closeness distance where you cant tell if it should have been attached or detached

screen shot 2015-12-09 at 15 19 31

2. Reconnect on reload works great, but it's odd to see no prompt, maybe send a new line char to force a prompt

screen shot 2015-12-09 at 15 23 09

3.Status elements should have rounded corners, see topology status in lower left, maybe reduce size too.

screen shot 2015-12-09 at 15 24 51

4. Space out tools like in controls

screen shot 2015-12-09 at 15 27 14

5. after resize of browser window, the terminal is too high

screen shot 2015-12-09 at 15 45 32

6. Dont expose implementation details, user opened a console or terminal, they dont care about pipes. Also no period at the end of titles.

screen shot 2015-12-09 at 15 46 45

7. Cant Exec on an Attached pipe. I'm always getting the error from above. Maybe change the wording?

8. When I get the delete error from above and click Exec a couple of times, and then try to close the one I see, a new terminal shows up, probably from a previous click (that is still showing the error message). This kind of stacking (?) is unexpected. I'm expecting a replace of the terminal when I open a new one.

9. After that if I click on Attach, I'm getting 400s without an error being displayed. I also cant seem to Exec on that container anymore (maybe old Attach pipe lying around?) and there is no way to get into a good state.

screen shot 2015-12-09 at 15 55 49

foot:
> (2) Reconnect on reload works great, but it's odd to see no prompt, maybe send a new line char to force a prompt

Bit dangerous sending commands but I agree its annoying not to have a prompt there. It would be interesting to save stdout to localStorage and repopulate the terminal when you re-open but perhaps there might be sensitive stuff in there.

> (5) after resize of browser window, the terminal is too high

Does the terminal height recover if you wiggle the size of the window a bit more or is it completely off? Chrome?

> (7) Cant Exec on an Attached pipe. I'm always getting the error from above. Maybe change the wording?
> (9) After that if I click on Attach, I'm getting 400s without an error being displayed. I also cant seem to Exec on that container anymore (maybe old Attach pipe lying around?) and there is no way to get into a good state.

Having trouble repro'ing. Which node are you exec'ing attaching on?

davkal:
> Bit dangerous sending commands
Yeah, makes sense.

> but I agree its annoying not to have a prompt there. It would be interesting to save stdout to localStorage and repopulate the terminal when you re-open but perhaps there might be sensitive stuff in there.
I remember trying that, but the result was not good. But give it another go.

> Does the terminal height recover if you wiggle the size of the window a bit more or is it completely off? Chrome?

Yes, Chrome. Reducing the window height always shows this behaviour. Extending the window height over its original height fixes it, resizing to the side has no effect.

> Having trouble repro'ing. Which node are you exec'ing attaching on?

Seems to work for me too now (restarted scope though, will keep an eye out).

app/pipes.go
tomwilkie: A little, yeah.  The returned pipe router is only used in tests TBH, to ensure the background goroutine shut sdown.  I don't think we need to worry about that in production though.

app/pipes.go
tomwilkie: > There might be a better name for getPipeRef and putPipeRef... Maybe retainPipe and releasePipe?

Roger.

> Also, do we ever call getOrCreatePipe and not follow it by getPipeRef?

We do not, but we need to know which end of the pipe we're holding the reference too, hence I made them two functions.

app/pipes.go
tomwilkie: Roger that.

app/controls_test.go
tomwilkie: > Why not roll this into xfer.NewAppClient?

Not all users of appclient want to start a control connection.  Tests are one example, another is using app client to fetch the ID of the app, and then using that to decide if we should start a new control connection.

probe/controls/controls.go
tomwilkie: Thanks!

probe/controls/pipes.go
tomwilkie: Roger, I think that interface is redundant.

probe/docker/container.go
tomwilkie: That correct - I think the UI is sorting them alphabetically, which happens to looks good right now,

prog/probe/main.go
tomwilkie: Agreed; but plumbing through a PipeFactory into the docker registry seems more naff.  Any other suggestions?

tomwilkie:
@paulbellamy 
> Could put xfer/pipes.go into xfer/pipe/pipe.go, giving you pipe.New(). 

Could do, but I've shied away from single file packages just for cosmetic reasons.  We're going to break up the xfer package soon, so mind if I skip this?

> Also, why do you need the Pipe interface, if there's only ever one implementation?

There is a 'core' pipe implementation in xfer/, then the App and the Probe both wrap this (and override close, for instance).  Its convenient.

prog/probe/main.go
paulbellamy: Pass clients into the docker registry. At least that way it's honest about the dependencies.

davkal:
2nd round of UI feedback (popouts and resizing work now)

(10) action to close should be all caps, no underline (and maybe a different color), like the ones in the topo options.

screen shot 2015-12-10 at 16 33 00

(11) popout window has url bar, which suggests that it's sharable, which does not seem to work

(12) chrome v45 crashes when doing 2 popouts in a row (will create separate issue):
```
1. exec -> popout -> close
2. exec -> popout -> crash
```

client/app/scripts/components/terminal.js
```diff
@@ -0,0 +1,293 @@
...
+    socket.onclose = () => {
+      log('socket closed');
+      this.socket = null;
+      const wereConnected = this.state.connected;
+      this.setState({connected: false});
+      if (this.term && this.props.pipe.status !== 'PIPE_DELETED') {
+        if (wereConnected) {
+          this.createWebsocket(term);
+        } else {
+          reconnectTimer = setTimeout(
```
davkal: `reconnectTimer` needs to be cleared in cwu.

client/app/scripts/components/terminal.js
```diff
@@ -0,0 +1,293 @@
...
+    });
+
+    this.createWebsocket(this.term);
+
+    const {pixelPerCol, pixelPerRow} = terminalCellSize(
+      innerNode, this.state.rows, this.state.cols);
+
+    window.addEventListener('resize', this.handleResize);
+
+    setTimeout(() => {
```
davkal: this needs to be cleared in cwu.

client/app/scripts/components/terminal.js
```diff
@@ -0,0 +1,293 @@
...
+    } else {
+      window.close();
+    }
+  }
+
+  handlePopoutTerminal(ev) {
+    ev.preventDefault();
+    const paramString = JSON.stringify(this.props);
+    clickCloseTerminal(this.getPipeId());
+    openNewWindow(`/terminal.html#!/state/${paramString}`);
```
davkal: Remove leading `/`, otherwise this wont be able to run in the service.

client/app/scripts/terminal-main.js
```diff
@@ -0,0 +1,9 @@
+require('font-awesome-webpack');
```
davkal: I dont think it's needed in the popout.

client/app/scripts/utils/router-utils.js
```diff
@@ -3,12 +3,31 @@ import page from 'page';
 import { route } from '../actions/app-actions';
 import AppStore from '../stores/app-store';
 
+//
+// TODO: move this logic somewhere else.
```
davkal: state-utils.js ?

davkal: it's very url state bound, and highly specific, I'd leave it here.

client/app/scripts/utils/router-utils.js
```diff
@@ -3,12 +3,31 @@ import page from 'page';
...
+//
+function shouldReplaceState(prevState, nextState) {
+  return (
```
davkal: isnt that just `prevState.controlPipe` or is this being explicit for a reason?

client/app/scripts/utils/web-api-utils.js
```diff
@@ -157,3 +176,42 @@ export function doControl(probeId, nodeId, control) {
...
+  const url = `/api/pipe/${encodeURIComponent(pipeId)}`;
+  reqwest({
+    method: 'GET',
+    url: url,
+    success: function(res) {
+      log('ERROR: expected responses: [400, 404]. Got:', res);
+    },
+    error: function(err) {
+      const status = {
+        400: 'PIPE_ALIVE',
```
davkal: But 400 is BAD REQUEST?

client/app/scripts/utils/router-utils.js
foot: Yes, yes it is. Very good. https://www.wolframalpha.com/input/?i=%28P+and+Q%29+or+%28P+and+not+Q%29

I like the venn diagram.

I do like the expanded form as it does reveal the idea behind it. Which is these particular transitions, (Term -> Term or Term -> No Term). I'll give them `const` names though to make that clearer.

client/app/scripts/utils/web-api-utils.js
foot: True! Yeah this is a bit of a hack :(.

1. Websockets don't receive the http status code from the server (for security reasons). Thats why we do this request.
2. If a pipe exists the server expects a WS connection, but we're xhr'ing which it considers a bad request (400).
3. If a pipe doesn't exist it returns a 404 and doesn't care that we're xhr'ing.

So it would be nice if either:
 - the server could respond w/ a 200 if we're xhr'ing and a WS handshake if we're WS'ing, (I'm not sure if thats easy).
 - We had another endpoint that was just for xhr GETting etc. e.g. `/api/pipe/:pid/status`.

We could open a new ticket for that, as this might causes scares in the future.

foot:
> (11) popout window has url bar, which suggests that it's sharable, which does not seem to work

Can't get rid of url bar these days. =/

> (12) chrome v45 crashes when doing 2 popouts in a row (will create separate issue):

Couldn't figure this out! =/

But the rest of it should be working!

davkal:
UI and JS LGTM

paulbellamy:
Backend code, LGTM.

- Text-wrapping and resize still seem a bit off. There's a blank space on the right side of the terminal that text never enters/uses.
- Should arrow keys and tab work? I'm assuming not?
- Hitting escape seems to close the terminal?
- Not sure the TTY-sizing is making it back to the terminal when execing. Vim seems to only occupy the top-left quarter of the "screen"

paulbellamy:
I've opened https://github.com/weaveworks/scope/issues/746 to deal with ncurses/popout/resize issues.

paulbellamy:
Ok, have opened issues for remaining issues. So all LGTM.

PR #655, opened 2015-11-12
Fix a couple bugs in metrics first/last calculation

### LINKED ISSUE #653: Container CPU usage sparkline not rendering correctly
These are against an alpine container running `/bin/sh`

There are a couple issues:

- `first` and `last` json fields are included even if the timestamp is "zero".
- `first` on the cpu graphs is always zero.

paulbellamy:
Fixes #653

### DIFF
--- a/report/metrics.go
+++ b/report/metrics.go
@@ -79,14 +79,14 @@ func (m Metric) Len() int {
 }
 
 func first(t1, t2 time.Time) time.Time {
-	if !t1.IsZero() && t1.Before(t2) {
+	if t2.IsZero() || (!t1.IsZero() && t1.Before(t2)) {
 		return t1
 	}
 	return t2
 }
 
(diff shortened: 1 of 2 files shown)


tomwilkie:
LGTM (assuming its green)

Why is the change to the intermediate marshalling necessary?  Is it for the UI? A couple of lines in the PR would be nice :-)

paulbellamy:
Ah, fair enough. In the js, since it's harder to check for go's `time.Time{}` value, we check the existence of a first field. So if it is zero we want to omit it, to stop rendering sparklines of all time.

paulbellamy:
The first/last calculation bug was resulting in `first(time.Now(), time.Time{})` returning `time.Time{}`. Which led to merge being non-commutative.

tomwilkie:
Thanks!

PR #658, opened 2015-11-12
Move weave ps calls to a background goroutine, instruments probe runtime.

tomwilkie:
- the weave tagger was taking 0.7s, causing report generation to overrun.
- move the call to weave ps to a background goroutine, such that calls to tag are now quick
- leave all the instrumentation in for future use; a `sudo pkill -SIGUSR1 scope-probe` will cause the probe to dump instrumentation to STDERR


paulbellamy:
Split the instrumentation and weave tagging stuff into 2 commits.

tomwilkie:
@paulbellamy PTAL

PR #672, opened 2015-11-13
Fix react key warning

davkal:
* one in TopologyOptions
* and some in details panel, @tomwilkie need report for that


tomwilkie:
I still get a few `Warning: Sparkline.getDOMNode(...) is deprecated. Please use ReactDOM.findDOMNode(instance) instead.`, but otherwise this does indeed fix the react warnings.  LGTM.

davkal:
`Sparkline.getDOMNode(...) is deprecated` is gone. Maybe you needed to rebuild.

PR #677, opened 2015-11-13
Dont treat missing node as UI error

### LINKED ISSUE #607: "Trying to reconnect" in UI if selected node disappears
For instance, select a container and then stop it.  UI stays trying to reconnect, even through its connected.

davkal:
A missing node no longer flashes the `Trying to reconnect` status.

Fixes #607


tomwilkie:
Works! LGTM

PR #679, opened 2015-11-13
Mitigate one-line-of-nodes layouts

davkal:
Previously it looked like this:

screen shot 2015-11-17 at 15 14 10

To mitigate:
* exclude 0-degree nodes from layout run
* layout 0-degree nodes separately 

With this PR it looks like this for the same report:

screen shot 2015-11-17 at 15 13 58


peterbourgon:
Ran it quickly on my box. No obvious problems. :+1:

PR #687, opened 2015-11-19
Upgraded react to 0.14.3

davkal:
also add production globals to webpack config to reduce bundle size by 10%, not bad!


2opremio:
This seems to work just fine:

screen shot 2015-11-23 at 23 18 17

PR #690, opened 2015-11-20
Procedural color creation for nodes

davkal:
Currently, colors are assigned at random order from a set of 20, leading to random color assignment across reloads.
Procedural color creation takes a string and maps it to the 360 degree color wheel, therefor colors will always be consistent across reloads.
* color hue is based on the first 2 letters of node rank
* color lightness is based on the first 2 letters of node label

screen shot 2015-11-20 at 14 48 51

PR #691, opened 2015-11-20
Show a spinner while node details are loading

davkal:
screen shot 2015-11-20 at 16 51 12

### DIFF
--- a/client/app/scripts/components/node-details.js
+++ b/client/app/scripts/components/node-details.js
@@ -21,15 +21,23 @@ const NodeDetails = React.createClass({
   },
 
   renderLoading: function() {
+    const node = this.props.nodes.get(this.props.nodeId);
+    const nodeColor = this.getNodeColorDark(node.get('rank'), node.get('label_major'));
+    const styles = {
+      header: {
+        'backgroundColor': nodeColor
+      }
+    };
+
     return (
       <div className="node-details">
-        <div className="node-details-header node-details-header-loading">
+        <div className="node-details-header" style={styles.header}>
           <div className="node-details-header-wrapper">
-            <h2 className="node-details-header-label">
-              Loading...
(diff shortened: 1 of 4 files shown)


2opremio:
I am getting the spinner but the details pane color is not matching the node's color. Wasn't that fixed?

screen shot 2015-11-24 at 4 48 34 pm


Also, as you show in your screenshot, the details panel is grey while showing the spinner. Why aren't you showing the final color? Is there a technical limitation preventing you from doing that?

davkal:
The color matching was done in a concurrent PR so it's broken in this one, but was now merged to master in #689 and #690.

But already coloring the panel in the final color is a good idea. will do that.

2opremio:
cool

davkal:
@2opremio PTAL.

2opremio:
@davkal LGTM, however, I've noticed that if you click on nodes while they are in the background, their detail requests end up somehow queuing (?) and displaying when you are no longer interested.

queued events

davkal:
You are right, when clicked, the details panel should go back to Loading mode.

2opremio:
Will test this tomorrow. Why are tests failing BTW?

2opremio:
Edge test ...

davkal:
yeah, unrelated.

2opremio:
It still suffers from the *problem* I mentioned, this video makes it more clear (wait until the end):

davkal:
Ah now I know what u mean. Will see how/if this can be fixed.

2opremio:
Much better now, although the details pane takes longer to load (it could be that my system is slower today due to the number of containers) and the Internet node for some reason is brown while it loads (although it's final color is purple):

Wrong internet details pane color while loading

2opremio:
LGTM

PR #701, opened 2015-11-30
Don't apply filters to node endpoints & update details panel logic

### LINKED ISSUE #693: Empty details panel after stopping container
Either the details panel should be implicitly hidden after stopping a container or we should show a friendlier message such as, "container stopped"

Before stopping:

screen shot 2015-11-23 at 22 54 16

After stopping:

screen shot 2015-11-23 at 22 54 30

tomwilkie:
Such that node which disappear from the topology (eg when they are stopped) don't disappear from the details panel.

Fixes #693

### DIFF
--- a/client/app/scripts/components/node-details.js
+++ b/client/app/scripts/components/node-details.js
@@ -77,14 +77,19 @@ const NodeDetails = React.createClass({
     const details = this.props.details;
     const nodeExists = this.props.nodes && this.props.nodes.has(this.props.nodeId);
 
+    if (details) {
+      return this.renderDetails();
+    }
+
     if (!nodeExists) {
       return this.renderNotAvailable();
     }
 
-    if (!details) {
-      return this.renderLoading();
-    }
+    return this.renderLoading();
+  },
 
+  renderDetails: function() {
+    const details = this.props.details;
     const nodeColor = this.getNodeColorDark(details.rank, details.label_major);
     const styles = {
       controls: {
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -245,7 +245,9 @@ func renderedForRequest(r *http.Request, topology APITopologyDesc) render.Render
 	return renderer
 }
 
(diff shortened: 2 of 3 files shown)


davkal:
We're still sending the topology options as a request parameter:
https://github.com/weaveworks/scope/blob/693-no-filter-details/client/app/scripts/utils/web-api-utils.js#L104

Should that be removed too?

tomwilkie:
Yes it can be now.  Do you mind doing that?

client/app/scripts/components/node-details.js
```diff
@@ -77,14 +77,19 @@ const NodeDetails = React.createClass({
     const details = this.props.details;
     const nodeExists = this.props.nodes && this.props.nodes.has(this.props.nodeId);
 
+    if (details) {
+      return this.renderDetails();
+    }
+
```
paulbellamy: Can we still `renderDetails` if `!nodeExists`? I'm specifically worried about that showing `undefined` in some spots...

tomwilkie: undefined is falsey, I think.  nodeExists is true iff the node exists in the currently topology, but the whole point of the bug report was that we don't want the details panel to go empty if the node disappears from the topology (due to a filter).

Do you have other proposals?  Maybe @davkal can suggest a better way?  We don't want node details to appear if the backend reports a 404 for the node.

paulbellamy: Sounds fine, was just checking that `renderDetails` wouldn't "panic" (js, lol), if `!nodeExists`, but it looks like it won't.

davkal: A 404 on getting node details will result in an empty `details` field, so that will work. Maybe rename `nodeExists` to `nodeInTopology` because it may be filtered. But to assuage paul's concerns, all fields rendered from the `details` object so they dont rely on data from the topology endpoint.

tomwilkie:
Test failures are fixed on master.  Can I merge this now?

paulbellamy:
LGTM, then.

davkal:
Looks like I came too late: 40f5040 (removes options query from node details request)

PR #705, opened 2015-11-30
Merge image metadata into container metadata so image labels can be used to hide containers.

### LINKED ISSUE #664: Consider hiding containers that are created by the weave & scope scripts from Scope
Using the works.weave.role=system hack.

tomwilkie:
Fixes #664

### DIFF
--- a/render/topologies.go
+++ b/render/topologies.go
@@ -121,7 +121,8 @@ type containerWithImageNameRenderer struct {
 }
 
 // Render produces a process graph where the minor labels contain the
-// container name, if found.
+// container name, if found.  It also merges the image node metadata into the
+// container metadata.
 func (r containerWithImageNameRenderer) Render(rpt report.Report) RenderableNodes {
 	containers := r.Renderer.Render(rpt)
 	images := Map{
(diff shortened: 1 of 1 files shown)


scope
```diff
@@ -110,7 +110,7 @@ case "$COMMAND" in
         echo $CONTAINER
 
         if ! echo "$@" | grep -E "\-\-no\-app|\-\-service\-token" 1>/dev/null; then
-            IP_ADDRS=$(docker run --rm --net=host --entrypoint /bin/sh $SCOPE_IMAGE -c "$IP_ADDR_CMD")
+            IP_ADDRS=$(docker run --rm --net=host --label=works.weave.role=system --entrypoint /bin/sh $SCOPE_IMAGE -c "$IP_ADDR_CMD")
```
paulbellamy: This uses the scope image, so why do we need this?

paulbellamy:
Other than the above, LGTM.

scope
tomwilkie: Ah yes it doesn't now.  Will fix.

PR #706, opened 2015-11-30
Filter nodes that are only connected to themselves.

### LINKED ISSUE #656: A node with only connections to itself should count as "unconnected"
screen shot 2015-11-12 at 13 33 37

tomwilkie:
Fixes #656

### DIFF
--- a/render/filters.go
+++ b/render/filters.go
@@ -22,7 +22,8 @@ func (c CustomRenderer) Render(rpt report.Report) RenderableNodes {
 }
 
 // ColorConnected colors nodes with the IsConnected key if
-// they have edges to or from them.
+// they have edges to or from them.  Edges to/from yourself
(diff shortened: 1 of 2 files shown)


davkal:
It works. But then there is no options to show unconnected nodes. The option has no action. I think if we filter nodes we should allow to unset the filter.

screen shot 2015-12-01 at 15 58 30

tomwilkie:
This is on the Applications view?  We decide a while back not to add a filter for that as there would be too many processes in the show-unconnected state.  Either way, do you mind filing a separate issue for this?   With the recent improvements in the rendering engine this is something we could revisit.

davkal:
Added #713 for the missing option. Otherwise LGTM

tomwilkie:
Thanks!

PR #712, opened 2015-12-01
JS to ES2015

foot:
Introduces shiny new ES2015 features such as:

 - import/export for modules.
 - new class syntax for React components.
 - direct export of actions. Now you gotta import actions by name.


davkal:
LGTM

PR #722, opened 2015-12-03
Add json support to app POST /api/report

paulbellamy:
This way you can save reports from /app/report, then use curl to post it
to the app later for testing.


tomwilkie:
LGTM assuming green from circle.

PR #724, opened 2015-12-04
Share /Users/twilkie/Documents/pkg with the build container in an attempt to speed up rebuilds.

### LINKED ISSUE #710: Speed up builds by sharing $GOPATH/pkg with the build container
Will need to make this optional and disabled on circle.

tomwilkie:
Fixes #710 

It does seem to have some affect, although not as much as I had hoped:

```
# make clean
# time make
...
real	2m6.012s
user	0m0.732s
sys	0m2.870s
# touch prog/probe/main.go
...
real	1m3.676s
user	0m0.527s
sys	0m2.412s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ RUNSVINIT=vendor/runsvinit/runsvinit
 RM=--rm
 RUN_FLAGS=-ti
 BUILD_IN_CONTAINER=true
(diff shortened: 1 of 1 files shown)


paulbellamy:
Sharing `pkg` but not `src` seems like a recipe for unexpected weirdness.

tomwilkie:
Actually this seems to have no gain in time, so I'm going to abandon it.

PR #725, opened 2015-12-04
Omit controls field from json if emtpy.

### LINKED ISSUE #654: Controls should be omitted from report if empty
They're being held by the timestamp:
```json
"controls": {
  "timestamp": "0001-01-01T00:00:00Z"
}
```

tomwilkie:
Fixes #654

### DIFF
--- a/report/controls.go
+++ b/report/controls.go
@@ -1,6 +1,8 @@
 package report
 
 import (
+	"bytes"
+	"encoding/json"
 	"time"
 
 	"github.com/weaveworks/scope/common/mtime"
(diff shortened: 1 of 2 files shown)


paulbellamy:
They're still there. Due to go's awkward zero semantics, you have to explicitly render the timestamp as in: https://github.com/weaveworks/scope/blob/master/report/metrics.go#L215

tomwilkie:
Definitely not there (on topologies) any more:

screen shot 2015-12-04 at 12 27 24

tomwilkie:
So I guess you mean on Nodes; let me take a look.

paulbellamy:
LGTM

PR #728, opened 2015-12-07
Make make build ui

tomwilkie:
As per discussion on slack, this removed the need to make frontend & make static.

Moved app/main.go into prog/app as per prog/probe, so that the tests do not need static.go.


app/router.go
```diff
@@ -16,6 +16,14 @@ import (
...
 )
 
+var (
+	// Version - set at buildtime.
+	Version = "dev"
+
+	// UniqueID - set at runtime.
+	UniqueID = "0"
+)
+
```
paulbellamy: Should these live in `prog/app/main.go`?

PR #737, opened 2015-12-09
Move procspy out of vendor into probe/endpoint.

tomwilkie:
Part of #284

PR #741, opened 2015-12-10
Use caching walker in procspy

tomwilkie:
Builds on #737, <s>will rebase once thats in</s>.  Part of #284.

Gives us a 1-2% cpu usage reduction on my VM.

### DIFF
--- a/common/fs/fs.go
+++ b/common/fs/fs.go
@@ -7,8 +7,8 @@ import (
 	"syscall"
 )
 
-// T is the filesystem interface type.
-type T interface {
+// Interface is the filesystem interface type.
+type Interface interface {
 	ReadDir(string) ([]os.FileInfo, error)
 	ReadFile(string) ([]byte, error)
 	Lstat(string, *syscall.Stat_t) error
@@ -19,17 +19,7 @@ type T interface {
 type realFS struct{}
 
 // FS is the way you should access the filesystem.
-var FS T = realFS{}
-
-// Mock is used to switch out the filesystem for a mock.
-func Mock(fs T) {
-	FS = fs
-}
-
-// Restore puts back the real filesystem.
-func Restore() {
-	FS = realFS{}
-}
+var fs Interface = realFS{}
 
 func (realFS) ReadDir(path string) ([]os.FileInfo, error) {
 	return ioutil.ReadDir(path)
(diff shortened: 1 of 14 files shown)


common/fs/fs.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"syscall"
+)
+
+// T is the filesystem interface type.
+type T interface {
```
paulbellamy: In the `sort` package (and other places?) this is called `type Interface interface {...}`. Not sure if I like that, but it's a bit more descriptive than `T`.

common/fs/fs.go
```diff
@@ -0,0 +1,52 @@
...
+	ReadFile(string) ([]byte, error)
+	Lstat(string, *syscall.Stat_t) error
+	Stat(string, *syscall.Stat_t) error
+	Open(string) (io.ReadWriteCloser, error)
+}
+
+type realFS struct{}
+
+// FS is the way you should access the filesystem.
+var FS T = realFS{}
```
paulbellamy: Maybe move this (and `type realFS ...`) below the mocking stuff so it's all together.

probe/process/walker_linux.go
```diff
@@ -28,7 +23,7 @@ func NewWalker(procRoot string) Walker {
 // passes one-by-one to the supplied function. Walk is only made public
 // so that is can be tested.
 func (w *walker) Walk(f func(Process)) error {
-	dirEntries, err := ReadDir(w.procRoot)
+	dirEntries, err := fs.FS.ReadDir(w.procRoot)
```
paulbellamy: Could have:
```Go
func ReadDir(...) {
  return FS.ReadDir
}
```
And that would clean up the invocations.

paulbellamy:
Other than the sugar/naming quibbles, it LGTM.

common/fs/fs.go
tomwilkie: Good idea.

common/fs/fs.go
tomwilkie: Roger.

probe/process/walker_linux.go
tomwilkie: Yes thats also a good idea - on a roll!

Issue #680, opened 2015-11-16
Consider deduping connections by client port numbers in the details panel


tomwilkie:
In lots of cases (esp. since the short lived connection stuff went in) we see LOTS of connections lists in the details panel.  We tend to be less interested in the client port number, so perhaps all lines that share a client port, connecting to a unique dst ip, dst port, could be deduped, and show as a range?

tomwilkie:
This was also suggested (or something similar) as part of the details panel redesign.

tomwilkie:
Just did hangout with @peterbourgon @davkal and @paulbellamy.

Grouping incoming connections by local port makes little sense, and the cardinality of that is likely to be very low (1-2) in most situations.  Similarly grouping outgoing connections by remote port makes little sense.

It was observed that the current connections tables lists every tcp 4 tuple for the selected node: this is likely to be too much information.  It was suggested that the connections table should instead group by the granularity of whatever topology you are currently viewing, eg when viewing the container topoloy, we  would have the following ables:

```
Incoming Connections
+------------------+
+------------------------------------------------+
|Container         |Local Port     |Count        |
+------------------------------------------------+
|Foo               |80             |12           |
|Foo               |443            |13           |
|Bar               |80             |2            |
|Baz               |80             |4            |
|Baz               |443            |8            |
+------------------------------------------------+

Show more...

Outgoing Connections
+------------------+
+------------------------------------------------+
|Container         |Remote Port    |Count        |
+------------------------------------------------+
|Foo               |80             |12           |
|Foo               |443            |13           |
|Bar               |80             |2            |
|Baz               |80             |4            |
|Baz               |443            |8            |
+------------------------------------------------+

Show more...
```

Ordering of the columns is TBD.  Port and Count columns should be sortable.

peterbourgon:
What if a container has a connection to something that isn't represented as a container? Will it fall back to the TCP 4 tuple? Hopefully the answer is yes. In which case the header of the leftmost column shouldn't be Container, but rather Remote or Other or something.

```
Incoming connections
+------------------+------------+-------+
| Remote           | Local port | Count |
|------------------|------------|-------|
| _Foo_            |       1337 |     3 |
| 8.8.4.2          |         80 |     1 |
+------------------+------------+-------+
```

tomwilkie:
Yes +1, if its a pseudo node then we display what ever that is.  Agree re: title

paulbellamy:
[...]

PR #752, opened 2015-12-11
Details panel redesign

### LINKED ISSUE #301: Table type should be more flexible
Right now the Table has a fixed and generic "schema" which isn't totally appropriate for everything we want to show in the details pane. Let's figure out a way to have different types of Tables for different types of information. Motivated by #298.

### LINKED ISSUE #382: Expandable detail-pane tables in UI
UX part of #322 

The UI will know what rows to expand based on a new optional row key: `expandable`

Sample table:

```json
{
    "title": "Connections",
    "numeric": false,
    "rows": [
        {
            "key": "TCP connections",
            "value_major": "2"
        },
        {
            "key": "Client",
            "value_major": "Server",
            "expandable": true
        },
        {
            "key": "127.0.0.1:3445",
            "value_major": "127.0.0.1:80",
            "expandable": true
        },
        {
            "key": "127.0.0.1:5647",
            "value_major": "127.0.0.1:560",
            "expandable": true
        }
    ]
}
```

davkal:
This PR is to redesign the details panel that appears when clicking on a node.  Fixes #382, fixes #301.

- [x] Links to relatives below main title
- [x] Health section with big last value and sparklines
- [x] Show-more option for grouped Health items
- [x] Tables for child nodes with metrics
- [x] Limit tables and lists to 5 entries with show more/less button
- [x] make links work (need unified node IDs)
- [x] sort table by metric
- [x] try alternative transitions when showing new details after link click
- [x] paul: fix ids so we can link between topologies (and get metrics/metadata reliably)
- [x] paul: remove "self" from "parents" table
- [x] ~~add connections summary section~~ (pushed to https://github.com/weaveworks/scope/issues/680)
- [x] paul: clean up backend code
- [x] paul: update backend tests
- [x] don't link unconnected processes in child table
- [x] try it with k8s
- [x] fix container-host-container bug
- [ ] stop containers with no image name displaying `<none>:<none>` as a parent (also containers with no name displaying as `<none>`)
- [x] fix bug with containers (w/o an image, maybe?) showing up in hosts view

screen shot 2015-12-11 at 17 52 39


tomwilkie:
Looks awesome!

A few comments on the UX (not looked at code yet):
- If you select weavescope -> scope process -> weavescope container, the details panel for it should come to the front again (it doesn't do anything right now)
- Also, the windows stack to the bottom right, so weill eventually fall off screen no?  How about stack to the bottom left?
- I want to be able to see a node in the context of the relevant topology - ie I want to be able to click a host, and move to the hosts view with that host selected.
- Host loadavg look good, I like the expando thing.
- The little triangel is cropped when sorting by memory usage: screen shot 2016-01-04 at 17 00 33 - I'd just drop the word usage (and from cpu).

Things that are missing / broken:
- When I select a host, I expect to see a table of all processes and containers.  Similarly I should be able to see a container image in the list of 'tags' for a container
- We've lost the connections table.   I liked the connections table.
- The whole selected node fan/circle is broken.
- Pipes / terminals don't work anymore.
- I'd like some of the key-values under the sparklines to be hidden, and have a 'show more there;'

This is looking really good guys - a massive improvements.  Good work.  Can I have a pony too?

tomwilkie:
Some feedback on go code (high level, not line by line):
- Pls don't commit `app/scope-app`
- In the probe, instead of adding ID to the `Node.Metadata`, we could make a field for it (`RenderableNode` has it, so I'd move it from `RenderableNode` to `Node`).  Constructors would also need updating, and `Topology.AddNode` could just take one argument.
- I like `render/metadata.go`. Perhaps the detailed node rendering deservers its own package now?
- I'm not convinced by the use of `Rank` to decide what order to merge nodes in.  Merge should be commutative.  Why was this needed again?
- I think we should promote the various 'parent' foreign keys out of `Node.Metadata` into a `Node.Parent` field, of type `[]struct{topology, nodeid string}`, with set-like merge semantics.  Doing this will allow us to efficiently gather all the parents recursively, and potentially simplify all the foreign key joins in the rendering code, and might even solve the commutative merge problem.
- please make changes to `tools/test` in tools.git

paulbellamy:
Responses inline:

> Pls don't commit app/scope-app

d'oh!

> In the probe, instead of adding ID to the Node.Metadata, we could make a field for it (RenderableNode has it, so I'd move it from RenderableNode to Node). Constructors would also need updating, and Topology.AddNode could just take one argument.

Sounds nicer, will have a look into how much shared code there is

> I like render/metadata.go. Perhaps the detailed node rendering deservers its own package now?

Agree!

> I'm not convinced by the use of Rank to decide what order to merge nodes in. Merge should be commutative. Why was this needed again?

Without the `Rank` change, merge of `Node.Metadata` is not commutative, which specifically results in the wrong value for `Node.Metadata["topology"]` (among others, I'm sure). We could possibly pull Topology and ID into fields, giving them better merge semantics.

> I think we should promote the various 'parent' foreign keys out of Node.Metadata into a Node.Parent field, of type []struct{topology, nodeid string}, with set-like merge semantics. Doing this will allow us to efficiently gather all the parents recursively, and potentially simplify all the foreign key joins in the rendering code, and might even solve the commutative merge problem.

A better structure for parents sounds nice. Unfortunately, gathering parents recursively is no good. For a container, it would grab all the parents of the container image which will include all hosts running that image.

> please make changes to tools/test in tools.git

ack

paulbellamy:
f2f discussion with @tomwilkie brought up that we should maybe try to set parents on the Node, not on RenderableNode, then aggregate them when merging. Doing it on the Probe would be nice, but not possible in all cases (k8s).

Also, that possibly `Node.Metadata[probe.Topology]` is not really used, so may be possible to swap it into a field, and add a "up-levelling" function in the `MapX2Y` funcs where we explicitly change it and the ID.

tomwilkie:
Some more feedback (its looking awesome, but...)
- clicking on a node in the detail panel thats already open in the stack of details panels doesn't work or do anything
- we used to see the pid of the selected process in the details panel in the title of the table.
- containers should have their images as a parent / tag
- container state is missing, as is IP (I think)
- container psuedo nodes are showing jibberish in the details panel (looks like process details panel)
- containers-by-hostname nodes should list the containers grouped under that

Theresmore, but it'll have to wait

tomwilkie:
More feedback:
- for nodes in the applications-by-name view, applications in the table can't be clicked.
- (for a future PR?) it would be nice if the docker labels fields were a table for container nodes
- nodes in containers-by-hostname have a container as a parent (should be a child).  Also need to ensure they can have multiple hosts and images as parents.
- I'd like to see a container images table for nodes in the hosts view, with a column for #containers.  Maybe another PR?

Going to look at code now.

probe/topology_tagger.go
```diff
@@ -19,18 +16,19 @@ func (topologyTagger) Name() string { return "Topology" }
...
+		"pod":             &(r.Pod),
+		"service":         &(r.Service),
 		"host":            &(r.Host),
 		"overlay":         &(r.Overlay),
 	} {
-		metadata := map[string]string{Topology: val}
-		for id, node := range topology.Nodes {
-			topology.AddNode(id, node.WithMetadata(metadata))
+		for id, node := range t.Nodes {
+			t.AddNode(id, node.WithID(id).WithTopology(name))
```
tomwilkie: I think I'd much prefer it if AddNode only took one argument, and the node constructors were forced to pass an ID (instead of the ID getting filled out after-the-fact).

render/detailed/metadata.go
```diff
@@ -0,0 +1,104 @@
...
+type MetadataRow struct {
+	ID    string `json:"id"`
+	Label string `json:"label"`
+	Value string `json:"value"`
+}
+
+// NodeMetadata produces a table (to be consumed directly by the UI) based on
+// an origin ID, which is (optimistically) a node ID in one of our topologies.
+func NodeMetadata(r report.Report, n render.RenderableNode) []MetadataRow {
+	renderers := map[string]struct {
```
tomwilkie: pull this out into the vars section above?

render/detailed/metadata.go
```diff
@@ -0,0 +1,104 @@
...
+	hostNodeMetadata = renderMetadata([]MetadataRow{
+		{ID: host.HostName, Label: "Hostname"},
+		{ID: host.OS, Label: "Operating system"},
+		{ID: host.KernelVersion, Label: "Kernel version"},
+		{ID: host.Uptime, Label: "Uptime"},
+	})
+)
+
+// MetadataRow is a row for the metadata table.
+type MetadataRow struct {
```
tomwilkie: This type seems to be used as both a template and as instance - we might want two types for this.

render/detailed/metrics.go
```diff
@@ -0,0 +1,169 @@
...
+		Value  float64 `json:"value"`
+		report.WireMetrics
+	}{
+		ID:          m.ID,
+		Label:       m.Label,
+		Format:      m.Format,
+		Group:       m.Group,
+		Value:       m.Value,
+		WireMetrics: m.Metric.ToIntermediate(),
+	})
```
tomwilkie: Nice!

davkal:
> we used to see the pid of the selected process in the details panel in the title of the table.

@tomwilkie you can see the parent PID on hover, good enough? Is that the same as "PID" in that context (if so, maybe we drop the "Parent")? 

screen shot 2016-01-11 at 14 37 51

render/detailed/metrics.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	if renderer, ok := renderers[n.SummaryTopology]; ok {
+		if nmd, ok := renderer.t.Nodes[n.SummaryID]; ok {
+			return renderer.r(nmd)
+		}
+	}
+	return nil
+}
+
+func processNodeMetrics(nmd report.Node) []MetricRow {
```
tomwilkie: Would be nice to make these more inline with how we render node metadata - ie a template & and function which takes the template and produces the nodes.

tomwilkie: I say this as the templates will need to be user-provided when we do plugins / extensions.

render/detailed/node.go
```diff
@@ -0,0 +1,303 @@
...
+	"github.com/weaveworks/scope/report"
+)
+
+// Node is the data type that's yielded to the JavaScript layer when
+// we want deep information about an individual node.
+type Node struct {
+	ID       string            `json:"id"`
+	Label    string            `json:"label"`
+	Rank     string            `json:"rank,omitempty"`
+	Pseudo   bool              `json:"pseudo,omitempty"`
```
tomwilkie: What are rank & pseudo used for in the UI (in the details panel)?

render/detailed/node.go
```diff
@@ -0,0 +1,303 @@
...
+type ControlInstance struct {
+	ProbeID string `json:"probeId"`
+	NodeID  string `json:"nodeId"`
+	report.Control
+}
+
+// MakeNode transforms a renderable node to a detailed node. It uses
+// aggregate metadata, plus the set of origin node IDs, to produce tables.
+func MakeNode(r report.Report, n render.RenderableNode) Node {
+	var hosts, pods, containerImages, containers, applications []ChildSummary
```
tomwilkie: I don't like this hard coding here - can we have a map[topologyid]ChildSummary?

render/detailed/node.go
```diff
@@ -0,0 +1,303 @@
...
+
+// ChildGroup is a topology-typed group of children for a Node.
+type ChildGroup struct {
+	Label      string         `json:"label"`
+	Nodes      []ChildSummary `json:"nodes"`
+	TopologyID string         `json:"topologyId"`
+}
+
+// ChildSummary is summary information about a child for a Node.
+type ChildSummary struct {
```
tomwilkie: Could this type be the same as Node?  It seems to have a lot in common, and would make generating them use the same (or some subset) of the code from MakeNode.

render/detailed/node.go
```diff
@@ -0,0 +1,303 @@
...
+			applications = append(applications, processChildSummary(child))
+		case "container":
+			containers = append(containers, containerChildSummary(child))
+		case "container_image":
+			containerImages = append(containerImages, containerImageChildSummary(child))
+		case "pod":
+			pods = append(pods, podChildSummary(child))
+		case "host":
+			hosts = append(hosts, hostChildSummary(child))
+		}
```
tomwilkie: I'd like to see if we can remove the specialisation here - perhaps a function MakeChildSummary that uses NodeMetadata and NodeMetrics etc?

render/detailed/node.go
```diff
@@ -0,0 +1,303 @@
...
+		return controlsFor(r.ContainerImage, n.ControlNode)
+	} else if _, ok := r.Host.Nodes[n.ControlNode]; ok {
+		return controlsFor(r.Host, n.ControlNode)
+	}
+	return []ControlInstance{}
+}
+
+// parents is a total a hack to find the parents of a node (which is
+// ill-defined).
+func parents(r report.Report, n render.RenderableNode) (result []Parent) {
```
tomwilkie: Paul + I discussed this offline and Paul's going to try and generate this from the nodes themselves, instead of using the heuristics below.

The basic ideas is that a report.Node will have a Parents field, which will be a set of (topology, node id) tuples.  When you merge two report.Nodes (or render.RenderableNodes), this field will be unioned.

A render.RenderableNode (which embeds a report.Node) will add an extra field, Children, which will be automatically filled whenever a node is "Up" rendered, that is the "kind" of the node is changed (ie from endpoint to process).  When this happens, the current report.nodeID is added to the children, and node.parents is erased (to prevent eg processes having every host as a parent through the container image).

The observation is you should never merge two nodes of different kinds (ie an endpoint and a process node), instead you "up" merge some nodes (from endpoint to process), then merge the result with the process node.

render/detailed/node.go
paulbellamy: Was initially named `NodeSummary`, so may go back to that.

probe/docker/container.go
```diff
@@ -331,7 +331,11 @@ func (c *container) GetNode(hostID string, localAddrs []net.IP) report.Node {
 		ContainerIPsWithScopes: report.MakeStringSet(ipsWithScopes...),
 	}).WithLatest(
 		ContainerState, mtime.Now(), state,
-	).WithMetrics(c.metrics())
+	).WithMetrics(
+		c.metrics(),
+	).WithParents(report.Sets{
+		"container_image": report.MakeStringSet(report.MakeContainerImageNodeID(hostID, c.container.Image)),
```
tomwilkie: This should also have a host as the parent, no?

probe/docker/tagger.go
```diff
@@ -84,6 +86,9 @@ func (t *Tagger) tag(tree process.Tree, topology *report.Topology) {
 
 		topology.AddNode(nodeID, report.MakeNodeWith(map[string]string{
 			ContainerID: c.ID(),
+		}).WithParents(report.Sets{
+			"container":       report.MakeStringSet(report.MakeContainerNodeID(t.hostID, c.ID())),
+			"container_image": report.MakeStringSet(report.MakeContainerImageNodeID(t.hostID, c.Image())),
```
tomwilkie: Ditto host.

probe/docker/container.go
tomwilkie: Given that almost all the taggers/reporters need to know the host id to produce their node ids, it seems a shame not to have them also add the host id to the parents and get rid of the host tagger.  OTOH, doing it in one place guarantees all nodes will have the a host as a parent.

probe/host/tagger.go
```diff
@@ -26,16 +26,21 @@ func (Tagger) Name() string { return "Host" }
...
+			report.ProbeID:    t.probeID,
+		}
+		parents = report.Sets{
+			"host": report.MakeStringSet(t.hostNodeID),
+		}
+	)
 
 	// Explicity don't tag Endpoints and Addresses - These topologies include pseudo nodes,
 	// and as such do their own host tagging
 	for _, topology := range []report.Topology{r.Process, r.Container, r.ContainerImage, r.Host, r.Overlay} {
```
tomwilkie: So the k8s reported needs to do its own host parenting?  Makes sense, but it one less argument for doing the host parenting here.

render/detailed/metrics.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// NodeMetrics produces a table (to be consumed directly by the UI) based on
+// an origin ID, which is (optimistically) a node ID in one of our topologies.
+func NodeMetrics(n report.Node) []MetricRow {
+	renderers := map[string]func(report.Node) []MetricRow{
+		"process":   processNodeMetrics,
+		"container": containerNodeMetrics,
+		"host":      hostNodeMetrics,
+	}
```
tomwilkie: I've been giving this some thought, and I think if we move this to a template way of rendering the nodes, and put those templates on the Topology itself (propagated from the Probe), it will allow for plugins to add stuff to the details panel, and make the code in the app truely generic.  In another PR, or course...

paulbellamy: :+1: We might even be able to do something like `node.Metrics()`

probe/host/tagger.go
paulbellamy: No, `r.Pod` should be added to this list as well. The k8s reporter should just need to "tag containers" with their podID.

probe/docker/container.go
paulbellamy: I agree with the latter.

paulbellamy:
Last `MakeDetailedHostNode` failure is stuck on https://github.com/weaveworks/scope/issues/815

foot:
Wow looks awesome!

- I get container-image `<none>:<none>` appearing sometimes? (see screenshot)
- `Encountered two children with the same key` after clicking *show more* in Host > Containers, and Host > Container Images (screenshot)
- Host > Container list has lots of room, doesn't need ellipsising so early. (screenshot)
- Text-size/text-color is *just* readable. Upping the base font from 13px to 14px is easier on my eyes (though breaks the layout). 13px does look more aesthetic.. 14px "feels easier".
- CPU always has its sorting arrow, that mimics the primary sorting arrow (e.g. If you click PID a couple of times, both its sorting arrow and that of CPU Usage changes). 
- Clicking away from the stack (on the canvas) I expected the entire stack to be dismissed rather than just the top card. (Though I still like <kbd>esc</kbd> to pop just the top card...)
- ~~As the stack gets higher it gets fuzzier (??) (see screenshot).~~ I was at 110% zoom, still strange though!
- Sometimes I can click on an app (it has a link), other times I cannot, though maybe they are different apps with the same name (?). Makes the UI feel a bit broken sometimes though =/. What controls whether you can click on an app?
- Edge case: If you are being silly and make a stack of 8 cards, it starts DoSing the server, the top card sometimes has to wait for other requests to complete before loading, making it feel a smidge laggy. Perhaps only poll top card.

Stuff thats already been mentioned:
- Host > Container > Host doesn't work (Not sure what cycle behaviour should be though, summon card to top?) (Tom).
- Terminal/attach not working (Tom)

Screenshot:
screen shot 2016-01-13 at 17 51 37

paulbellamy:
Also PID sorting seems to be doing it alphabetically, which is a bit strange.

paulbellamy:
> Sometimes I can click on an app (it has a link), other times I cannot, though maybe they are different apps with the same name (?). Makes the UI feel a bit broken sometimes though =/. What controls whether you can click on an app?

Whether it is a link or not is determined by the process having connections. Processes without any connections don't show up in the process view, so there's not really anything to link to. We could link them to a process details panel, but hide the link to "go to topology" (or just show the process topology with a details panel for a non-existing node).

client/app/scripts/charts/node.js
```diff
@@ -99,14 +100,14 @@ export default class Node extends React.Component {
...
-  handleMouseEnter(ev) {
-    enterNode(ev.currentTarget.id);
+  handleMouseEnter() {
+    enterNode(this.props.id);
   }
 
-  handleMouseLeave(ev) {
-    leaveNode(ev.currentTarget.id);
+  handleMouseLeave() {
+    leaveNode(this.props.id);
```
foot: nice

client/app/scripts/components/details-card.js
```diff
@@ -0,0 +1,53 @@
...
+const marginRight = 36;
+const panelWidth = 420;
+const offset = 8;
+
+export default class DetailsCard extends React.Component {
+
+  constructor(props, context) {
+    super(props, context);
+    this.state = {
+      mounted: null
```
foot: `null` vs `false`? Is one more reacty?

client/app/scripts/components/details-card.js
```diff
@@ -0,0 +1,53 @@
...
+    };
+  }
+
+  componentDidMount() {
+    setTimeout(() => {
+      this.setState({mounted: true});
+    });
```
foot: While this pattern is very clear and simple it feels a tiny bit hackish, does `ReactCSSTransitionGroup` etc give you a nice way to specify the animation states? Skimming the docs it look like it might, but might complicate things a lot too.

client/app/scripts/components/node-details.js
```diff
@@ -15,9 +36,25 @@ export default class NodeDetails extends React.Component {
     resetDocumentTitle();
   }
 
+  renderTools() {
+    const showSwitchTopology = this.props.index !== undefined && this.props.index > 0;
```
foot: `this.props.index` shouldn't ever be `undefined`

client/app/scripts/components/node-details/node-details-health-item.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import Sparkline from '../sparkline';
+import { formatMetric } from '../../utils/string-utils';
+
+export default class NodeDetailsHealthItem extends React.Component {
+  render() {
```
foot: Can make this a function!

client/app/scripts/components/node-details/node-details-relatives-link.js
```diff
@@ -0,0 +1,27 @@
...
+
+  constructor(props, context) {
+    super(props, context);
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick(ev) {
+    ev.preventDefault();
+    clickRelative(this.props.id, this.props.topologyId, this.props.label,
+      ReactDOM.findDOMNode(this).getBoundingClientRect());
```
foot: nice

client/app/scripts/components/node-details/node-details-relatives.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  render() {
+    let relatives = this.props.relatives;
+    const limited = relatives && this.state.limit > 0 && relatives.length > this.state.limit;
+    const showLimitAction = relatives && (limited || (this.state.limit === 0 && relatives.length > this.DEFAULT_LIMIT));
+    const limitActionText = limited ? 'Show more' : 'Show less';
+    if (relatives && limited) {
+      relatives = relatives.slice(0, this.state.limit);
+    }
```
foot: lots of `relatives` checking here, maybe bump those up a level and assume there are relatives if this component is being rendered.

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -1,33 +1,159 @@
...
+
+  getDefaultSortBy() {
+    // first metric
+    const nodes = this.props.nodes;
+    if (nodes && nodes.length > 0) {
+      if (nodes[0].metrics && nodes[0].metrics.length > 0) {
+        return nodes[0].metrics[0].id;
+      }
+    }
+    return null;
```
foot: `_.get(this.props.nodes, [0, 'metrics', 0, 'id'])`

returns (`value`|`undefined`) for any value of this.props.nodes!

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -1,33 +1,159 @@
...
+  }
+
+  getMetaDataSorters() {
+    // returns an array of sorters that will take a node
+    if (this.props.nodes && this.props.nodes.length > 0 && this.props.nodes[0].metadata) {
+      return this.props.nodes[0].metadata.map((field, index) => {
+        return node => node.metadata[index] ? node.metadata[index].value : null;
```
foot: more fun w/ _.get!
`return _.get(this.props.nodes, [0, 'metadata'], []).map(...);`

client/app/scripts/stores/app-store.js
```diff
@@ -360,16 +425,17 @@ export class AppStore extends Store {
       break;
 
     case ActionTypes.RECEIVE_CONTROL_PIPE:
-      controlPipe = {
+      controlPipes = controlPipes.set(payload.pipeId, {
         id: payload.pipeId,
+        nodeId: payload.nodeId,
         raw: payload.rawTty
-      };
+      });
```
foot: Gotta wrap the object in another `makeOrderedMap`.

client/app/scripts/stores/app-store.js
```diff
@@ -453,7 +536,18 @@ export class AppStore extends Store {
       setTopology(payload.state.topologyId);
       setDefaultTopologyOptions(topologies);
       selectedNodeId = payload.state.selectedNodeId;
-      controlPipe = payload.state.controlPipe;
+      if (payload.state.controlPipe) {
+        controlPipes = makeOrderedMap(
+          [payload.state.controlPipe.pipeId, payload.state.controlPipe]
```
foot: I think this has to be either:
```
makeOrderedMap(
    [[payload.state.controlPipe.pipeId, payload.state.controlPipe]]
);
```
or
```
makeOrderedMap(
    {[payload.state.controlPipe.pipeId]: payload.state.controlPipe}
);
```

foot:
Done reading the code! Looks good

client/app/scripts/components/node-details/node-details-table.js
davkal: Very nice one!

client/app/scripts/stores/app-store.js
davkal: I'll just treat those as a normal object.

client/app/scripts/stores/app-store.js
davkal: You're right, `[[payload.state.controlPipe.pipeId, payload.state.controlPipe]]` it is.

client/app/scripts/components/details-card.js
davkal: Next time around :)

tomwilkie:
@paulbellamy I think we might need to profile the app, as I've seen it maxing out the CPU whilst running this PR, which I don't think it used to do.

tomwilkie:
@paulbellamy this will need to be squashed too.  I suggest just going for two changesets, one for the go/backend code, and another for the client/javascript code.

tomwilkie:
Some more feedback on the UX:
- [ ] I'd drop the term 'Usage' from the column headers, as it doesn't fit:  screen shot 2016-01-15 at 13 10 25

tomwilkie:
Looking really good Paul!

Comments on the code
- [ ] The changes to app/api_topologies.go for k8s view could go straight into master, as a separate change.
- [ ] I think report.MakeContainerImageNodeID shouldn't take a host ID, which hopefully will mean we don't have to plumb it everywhere.
- [ ] I think we should use constants for the keys in Node.Parents
- [ ] `ltst` is a sucky name for a template
- [ ] with the way we're now showing load avaerage, I think we can remove the special casing of max, and bring it in line with all the other metrics.
- [ ] I still think metrics should be templated like the metadata, but we can leave that to another PR.
- [ ] detailed.Node should embed a NodeSummary, and use MakeNodeSummary to build it.
- [ ] the comment "// parents is a total a hack..." isn't really true anymore.
- [ ] the defer.. at the top of parents doesn't add anything - why not inline it?
- [ ] I still want to see WithID go away...
- [ ] I think we need to make NodeSet properly immutable.  Also, doesn't it belong in render?

tomwilkie:
So good luck figuring out whats using cpu:
[cpu.pdf](https://github.com/weaveworks/scope/files/93110/cpu.pdf)

paulbellamy:
> `ltst` is a sucky name for a template

I agree, but it is 4 characters, which makes all the templates vertically-aligned. Could refactor it to a struct (instead of a fn), so the "type" goes at the end.

>  I still want to see WithID go away...

Started on this, but you have to change all the `MakeNode` and `MakeNodeWith`s, got about 726 lines in, and stashed it.

paulbellamy:
> I think the numbers for host cpu usage are wrong: my cpu is close to 100%, and scope it reporting...

I hypothesize that it may be this bug: https://github.com/weaveworks/scope/issues/815

We merge all the processes and containers into the host (to get the children), but that brings the metrics with it, and since some of those keys happen to overlap, they all get merged in, and a random number is chosen.

tomwilkie:
> Could refactor it to a struct (instead of a fn), so the "type" goes at the end.

Roger, please file a ticket against 0.13 and we'll do that then.

> Started on this, but you have to change all the MakeNode and MakeNodeWiths, got about 726 lines in, and stashed it.

Ditto, we can fix in 0.13

> I hypothesise that it may be this bug: #815

Okay we need to fix this before the merge - I think dropping the metrics when we create new "Derived" nodes in the rendering will be enough for now.  I'll take a look today.

paulbellamy:
- [ ] Uncontained (in the containers view), is rendering the metadata/metrics from one of the processes, not just a list of processes.

paulbellamy:
We've LGTMed and agreed in chat to merge this, then continue working, as rebasing is becoming painful.

PR #753, opened 2015-12-14
Also proxy Websockets through the dev server while in dev mode.

foot:
WS were not using the proxy and instead were making a direct connection to the provided BACKEND_HOST.

Now they also are proxied.
- Simplifies FE code/build as we don't have a special case for websockets.
- Can more easily test WS error handling by simulating loss of connection to the BE by shutting down proxy.

### DIFF
--- a/client/.eslintrc
+++ b/client/.eslintrc
@@ -1,8 +1,5 @@
 {
   "extends": "airbnb",
-  "globals": {
-    __WS_URL__: false
-  },
   "env": {
     "browser": true,
     "jest": true,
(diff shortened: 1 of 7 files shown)


davkal:
Works well! 

Forgot one WSURL:
```diff
--- a/client/.eslintrc
+++ b/client/.eslintrc
@@ -1,8 +1,5 @@
 {
   "extends": "airbnb",
-  "globals": {
-    __WS_URL__: false
-  },
   "env": {
     "browser": true,
     "jest": true,
```

If that is fixed up, LGTM

PR #766, opened 2015-12-14
Set correct size for terminal which fixes text wrapping.

### LINKED ISSUE #765: Terminal doesn't wrap text correctly
- Sometimes the prompt disappears below the bottom of the terminal.
- Sometimes text escapes out the side of the terminal.

foot:
Sometimes it was going off the sides/bottom.

Fixes #765

### DIFF
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -13,8 +13,8 @@ const wsProto = location.protocol === 'https:' ? 'wss' : 'ws';
 const wsUrl = wsProto + '://' + location.host + basePath(location.pathname);
 const log = debug('scope:terminal');
 
-const DEFAULT_COLS = 1;
-const DEFAULT_ROWS = 1;
+const DEFAULT_COLS = 80;
+const DEFAULT_ROWS = 24;
 // Unicode points can be used in html and document.title
 // html shorthand codes (&times;) don't work in document.title.
 const TIMES = '\u00D7';
@@ -28,10 +28,11 @@ function ab2str(buf) {
 
 function terminalCellSize(wrapperNode, rows, cols) {
   const height = wrapperNode.clientHeight;
-  const firstRow = wrapperNode.querySelector('.terminal div');
 
   // Guess the width of the row.
   let width = wrapperNode.clientWidth;
+  // Now try and measure the first row we find.
+  const firstRow = wrapperNode.querySelector('.terminal div');
   if (!firstRow) {
     log("ERROR: Couldn't find first row, resizing might not work very well.");
   } else {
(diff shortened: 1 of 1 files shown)


client/app/scripts/components/terminal.js
```diff
@@ -198,15 +207,17 @@ export default class Terminal extends React.Component {
     ev.preventDefault();
     const paramString = JSON.stringify(this.props);
     clickCloseTerminal(this.getPipeId());
-    openNewWindow(`terminal.html#!/state/${paramString}`);
+
+    const minWidth = this.state.pixelPerCol * 80 + (8 * 2);
```
davkal: Why 80? Why not use the current `cols`?

client/app/scripts/components/terminal.js
```diff
@@ -29,24 +27,35 @@ function ab2str(buf) {
...
+  // Guess the width of the row.
+  let width = wrapperNode.clientWidth;
+  // Now try and measure the first row we find.
+  const firstRow = wrapperNode.querySelector('.terminal div');
+  if (!firstRow) {
+    log("ERROR: Couldn't find first row, resizing might not work very well.");
+  } else {
+    const rowDisplay = firstRow.style.display;
+    firstRow.style.display = 'inline';
+    width = firstRow.offsetWidth;
```
davkal: Why is this better than `wrapperNode.clientWidth`?

client/app/scripts/components/terminal.js
foot: I have a little laptop screen, it gets very small w/ dev tools open. So when the embedded term appears it is very small. My first thought is: lets pop that out and make it bigger! But when it pops out it is very small so I have to immediately resize it.

Popping out windows is like freeing them! 80 chars is the ideal min size i think.

davkal: Understood.

client/app/scripts/components/terminal.js
foot: wrapperNode is a block div which fills its parent and gives a bad size reading.

We wanna get the container for the the actual text characters, turn it into a span so that it "clings" to them, and measure that. I found this clever hack on term.js's issue page somewhere.

Which makes me thing actually, that maybe if the initial 80 cols don't fit in the initial container they might wrap, giving another bad size reading. Maybe we should set the initial cols to 1... it will updated immediately to fit the window size anywho.. hmm

foot: More testing and it uses &nbsp; to fill out the rest of the line so it won't wrap! \o/.

So this measurement will be accurate regardless of init ncols. I think.

davkal:
LGTM

PR #767, opened 2015-12-14
Gather per-process CPU and memory metrics.

### LINKED ISSUE #756: Gather per-process CPU and Memory usage
We pretty much already have the code in proc walker.

### LINKED ISSUE #757: Make metrics consistent between nodes
Hosts have % memory usage, containers have it in KB (or MB etc).

We probably want a way of doing both, TBH.

tomwilkie:
Fixes #756, fixes #757

### DIFF
--- a/probe/process/reporter.go
+++ b/probe/process/reporter.go
@@ -3,29 +3,37 @@ package process
 import (
 	"strconv"
 
+	"github.com/weaveworks/scope/common/mtime"
 	"github.com/weaveworks/scope/report"
 )
 
 // We use these keys in node metadata
 const (
-	PID     = "pid"
-	Comm    = "comm"
-	PPID    = "ppid"
-	Cmdline = "cmdline"
-	Threads = "threads"
+	PID         = "pid"
+	Comm        = "comm"
+	PPID        = "ppid"
+	Cmdline     = "cmdline"
+	Threads     = "threads"
+	CPUUsage    = "cpu_usage_percent"
+	MemoryUsage = "memory_usage_bytes"
 )
 
 // Reporter generates Reports containing the Process topology.
 type Reporter struct {
-	scope  string
-	walker Walker
+	scope   string
+	walker  Walker
+	jiffies Jiffies
 }
 
+// Jiffies is the type for the function used to fetch the elapsed jiffies.
+type Jiffies func() (uint64, float64, error)
+
 // NewReporter makes a new Reporter.
-func NewReporter(walker Walker, scope string) *Reporter {
+func NewReporter(walker Walker, scope string, jiffies Jiffies) *Reporter {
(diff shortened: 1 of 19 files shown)


probe/process/walker_darwin.go
```diff
@@ -92,3 +92,8 @@ func parseLSOF(output string) (map[string]Process, error) {
 	}
 	return processes, nil
 }
+
+// GetDeltaTotalJiffies returns 0 - darwin doesn't have jiffies.
+func GetDeltaTotalJiffies() (uint64, float64, error) {
+	return 0, 0.0, nil
+}
```
paulbellamy: Won't this cause a flat sparkline? Better to disable the metric.

paulbellamy: Ah, deltaTotal will be 0, so this won't display.

probe/process/walker.go
```diff
@@ -8,19 +8,22 @@ type Process struct {
...
 // CachingWalker is a walker than caches a copy of the output from another
 // Walker, and then allows other concurrent readers to Walk that copy.
 type CachingWalker struct {
-	cache     []Process
-	cacheLock sync.RWMutex
-	source    Walker
+	cache         []Process
+	previousByPID map[int]Process
+	cacheLock     sync.RWMutex
+	source        Walker
```
paulbellamy: If we don't need guaranteed ordering of `Walk(...)`, it *might* be cleaner to change the type of `cache` to match `previousByPID`, so processes are always indexed by pid.

probe/process/reporter.go
```diff
@@ -59,9 +73,18 @@ func (r *Reporter) processTopology() (report.Topology, error) {
...
 		if p.PPID > 0 {
 			node.Metadata[PPID] = strconv.Itoa(p.PPID)
 		}
+
+		if deltaTotal > 0 {
+			cpuUsage := float64(p.Jiffies-prev.Jiffies) / float64(deltaTotal) * 100.
+			node = node.WithMetric(CPUUsage, report.MakeMetric().Add(now, cpuUsage).WithMax(maxCPU))
+		}
+
+		node = node.WithMetric(MemoryUsage, report.MakeMetric().Add(now, float64(p.RSSBytes)))
```
paulbellamy: memory usage from this disagrees with top/htop/ps-aux

probe/process/walker.go
tomwilkie: We don't rely on it anywhere.  Will give it a go.

probe/process/reporter.go
paulbellamy: Ah, had forgotten to divide ps aux/etc output by 1024 (not 1000).

paulbellamy:
LGTM

PR #780, opened 2015-12-17
Fixes display of unicode characters in the terminal

### LINKED ISSUE #748: Pipe should allow me to type all characters
On UK keyboards I have to use `Right-Alt + 3` to type `#`. These don't seem to be sent.

It also seems that `Ctrl + {arrow}` keys are ignored.

Found in #650

foot:
Fixes #748


davkal:
LGTM

screen shot 2015-12-23 at 12 50 19

PR #781, opened 2015-12-17
Add integration test for procspy edges.

### LINKED ISSUE #779: Add integration test for procspied connections
To prevents the 0.11 bug from appearing again.

tomwilkie:
Fixes #779

### DIFF
--- a/integration/330_application_edge_test.sh
+++ b/integration/330_application_edge_test.sh
@@ -1,0 +1,21 @@
+#! /bin/bash
+
+. ./config.sh
+
+start_suite "Test short connections between processes"
+
+WEAVE_DOCKER_ARGS=$ADD_HOST_ARGS weave_on $HOST1 launch
+scope_on $HOST1 launch
+weave_on $HOST1 run -d --name nginx nginx
+weave_on $HOST1 run -dti --name client alpine /bin/sh -c "while true; do \
+	nc nginx.weave.local 80 || true; \
+	sleep 1; \
+done"
+
+wait_for application $HOST1 60 nginx client
+
+has applications $HOST1 nginx
+has applications $HOST1 nc
+has_connection applications $HOST1 nc nginx
+
+scope_end_suite
--- a/integration/330_application_edge_test.sh
+++ b/integration/330_application_edge_test.sh
@@ -2,10 +2,10 @@
 
 . ./config.sh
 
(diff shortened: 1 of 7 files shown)


integration/config.sh
```diff
@@ -44,58 +36,84 @@ scope_end_suite() {
 	done
 }
 
+# this checks we have a named node in the given view
+has() {
+    local view=$1
```
paulbellamy: indentation

integration/330_application_edge_test.sh
```diff
@@ -0,0 +1,21 @@
...
+weave_on $HOST1 run -dti --name client alpine /bin/sh -c "while true; do \
+	nc nginx.weave.local 80 || true; \
+	sleep 1; \
+done"
+
+wait_for application $HOST1 60 nginx client
+
+has applications $HOST1 nginx
+has applications $HOST1 nc
+has_connection applications $HOST1 nc nginx
```
paulbellamy: v nice

integration/330_application_edge_test.sh
```diff
@@ -0,0 +1,21 @@
...
+
+start_suite "Test short connections between processes"
+
+WEAVE_DOCKER_ARGS=$ADD_HOST_ARGS weave_on $HOST1 launch
+scope_on $HOST1 launch
+weave_on $HOST1 run -d --name nginx nginx
+weave_on $HOST1 run -dti --name client alpine /bin/sh -c "while true; do \
+	nc nginx.weave.local 80 || true; \
+	sleep 1; \
+done"
```
paulbellamy: This isn't actually that short of a connection, as one it is connected the `nc` will hang for a while.

tomwilkie: Not supposed to be a short connection (they don't get applied between processes) - I'm testing procspy, which only looks at long connections.  Have updated comment and turned off conntrack for this test to make it more obvious.

tomwilkie:
@paulbellamy PTAL

paulbellamy:
Ok, so the salient point of these tests is that when conntrack is off the probe is forced to use procspy, correct? if so, let's either mention procspy in the test name, or a comment on why conntrack is disabled.

tomwilkie:
@paulbellamy kinda - we're checking process to process edges, which only come from procspy.  Have added this to test name.

PR #789, opened 2016-01-04
Speed up incremental build

tomwilkie:
- Pass `-i` to `go build` to speed up builds (forces `go build` to install dependancies, thanks @bboreham)
- Share `scope/pkg` into build container as `/go/pkg` (cached intermediate packages are installed here)
- Ignore files in `./vendor` for Makefile dependancies
- Don't send two copies of the docker tarball to the docker daemon when building container
- Add `make shell` target to get a shell in the build container
- Don't install `python-pip` in the build container, just `python-requests`
- Update to go 1.5.2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ SCOPE_BACKEND_BUILD_IMAGE=$(DOCKERHUB_USER)/scope-backend-build
 SCOPE_BACKEND_BUILD_UPTODATE=.scope_backend_build.uptodate
 SCOPE_VERSION=$(shell git rev-parse --short HEAD)
 DOCKER_VERSION=1.6.2
-DOCKER_DISTRIB=docker-$(DOCKER_VERSION).tgz
+DOCKER_DISTRIB=.pkg/docker-$(DOCKER_VERSION).tgz
 DOCKER_DISTRIB_URL=https://get.docker.com/builds/Linux/x86_64/docker-$(DOCKER_VERSION).tgz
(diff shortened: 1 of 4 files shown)


tomwilkie:
With this change, incremental builds take ~30s.  Full build is 7mins.

```shell
$ make clean
$ make
...
real	6m54.132s
user	0m0.536s
sys		0m3.712s
$ # some very minor change
$ make
...
real	0m31.620s
user	0m0.124s
sys		0m2.352s
```

tomwilkie:
For reference, on master incremental builds take almost 2mins:

```shell
$ make clean
$ make
$ # some very minor change
$ make
...
real	1m49.625s
user	0m0.996s
sys		0m5.480s
```

Makefile
```diff
@@ -12,12 +12,14 @@ SCOPE_BACKEND_BUILD_IMAGE=$(DOCKERHUB_USER)/scope-backend-build
 SCOPE_BACKEND_BUILD_UPTODATE=.scope_backend_build.uptodate
 SCOPE_VERSION=$(shell git rev-parse --short HEAD)
 DOCKER_VERSION=1.6.2
-DOCKER_DISTRIB=docker/docker-$(DOCKER_VERSION).tgz
+DOCKER_DISTRIB=docker-$(DOCKER_VERSION).tgz
```
paulbellamy: would be nice if we had a "target build directory" to put all this junk in.

paulbellamy:
If you can think of a better place to put the docker tarball that would be great, otherwise LGTM.

PR #794, opened 2016-01-04
Split and move xfer package.

### LINKED ISSUE #641: Everything* in xfer/ is only used by the Probe
(except a const port number, the request & response control data types, and the Details datatype.)

Proposal is to move every except that into probe/xfer, and move the above into common/xfer

tomwilkie:
Fixes #641


paulbellamy:
Not a fan of:
```Go
import (
  "github.com/weaveworks/weave/common"
  xcommon "github.com/weaveworks/scope/common/xfer" // sometimes called just `common`
  "github.com/weaveworks/scope/probe/xfer"
)
```
Maybe:
`probeXfer "github.com/weaveworks/scope/probe/xfer"`, or
`commonXfer "github.com/weaveworks/scope/common/xfer"`

In general `common "github.com/weaveworks/scope/common/xfer"` seems a bit confusing, since `common` is already the name of another package.

tomwilkie:
@paulbellamy I agree (x)common is pretty ugly; but the cAmElCaSe seems pretty ugly too.  How about we rename `probe/xfer` to be `proble/appclient`?

paulbellamy:
`probe/xfer -> app/client` ?

tomwilkie:
@paulbellamy PTAL

PR #795, opened 2016-01-05
Animate sparklines

### LINKED ISSUE #638: Sparklines are jumpy
These could be animated, then it's much easier to tell the progress.

davkal:
* Refactored sparklines to be rendered by react
* Animate sparklines by feeding data bit by bit using a buffer
* Corrected row key for sparklines so root element does not get replaced all the time
* Tried using CSS transitions to [animate the line](http://bl.ocks.org/benjchristensen/1148374) between consecutive data feeds, but that still looked choppy

Fixes #638


tomwilkie:
Looking good 

yijibh1klw

client/app/scripts/components/sparkline.js
```diff
@@ -1,118 +1,179 @@
...
-      attr('cx', lastX).
-      attr('cy', lastY).
-      attr('fill', '#46466a').
-      attr('fill-opacity', 0.6).
-      attr('stroke', 'none').
-      attr('r', this.props.circleDiameter);
+    this.x.domain([first, last]);
+
+    this.y.domain([
+      this.props.min || d3.min(data, d => d.value),
```
foot: Lose `0` as the min domain here, maybe check for `undefined`?

foot: Or do a `Math.min` as below? Not sure if that works.

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -14,15 +14,19 @@ export default class NodeDetailsTable extends React.Component {
...
           } else if (row.value_type === 'sparkline') {
+            key = row.key;
             valueComponent = <NodeDetailsTableRowSparkline row={row} />;
           } else {
+            key = row.key + row.value_major + row.value_minor;
             valueComponent = <NodeDetailsTableRowValue row={row} />;
           }
           return (
-            <div className="node-details-table-row" key={row.key + row.value_major + row.value_minor}>
+            <div className="node-details-table-row" key={key}>
```
foot: Do we need all this stuff in the `key`? `row.key` by itself looks like its unique enough (the label).

client/app/scripts/components/sparkline.js
```diff
@@ -1,118 +1,179 @@
...
-
-      y.domain([
-        this.props.min || d3.min(data, d => d.value),
-        this.props.max || d3.max(data, d => d.value)
-      ]);
-
-      lastX = x(data[data.length - 1].date);
-      lastY = y(data[data.length - 1].value);
-      title = 'Last ' + d3.round((last - first) / 1000) + ' seconds, ' + data.length + ' samples, min: ' + d3.round(d3.min(data, d => d.value), 2) + ', max: ' + d3.round(d3.max(data, d => d.value), 2) + ', mean: ' + d3.round(d3.mean(data, d => d.value), 2);
+    // adjust scales based on data
```
foot: range not based on data.

client/app/scripts/components/sparkline.js
```diff
@@ -1,118 +1,179 @@
...
+
+  getGraphData() {
+    // data is of shape [{date, value}, ...]
+    let data = this.getDataFromBuffer();
 
-    // Do nothing if no data is passed in.
-    if (data.length === 0) {
-      return;
+    // Do nothing if no data or data w/o date are passed in.
+    if (data.length === 0 || data[0].date === undefined) {
```
foot: fail fast? +1 for exploding if we're getting some weird data.

client/app/scripts/components/node-details/node-details-table.js
davkal: Yup, in the current node details table, some rows in the connection tables have the same keys :frowning:

client/app/scripts/components/sparkline.js
```diff
@@ -1,118 +1,179 @@
...
+      let firstIndex = dates.indexOf(this.state.first);
+      if (firstIndex > -1 && firstIndex < dates.size - 1) {
+        firstIndex++;
+      } else {
+        firstIndex = 0;
+      }
+      const first = dates.get(firstIndex);
+
+      let lastIndex = dates.indexOf(this.state.last);
+      if (lastIndex > -1 && lastIndex < dates.size) {
```
foot: `lastIndex < dates.size` is always `true`?

foot:
It would be cool to have a better domain for the load averages. (They still "jump around a bit"). But I'm not sure what that would be, some dynamic rolling average perhaps.

foot:
As the sparkline updates more frequently than the label the two feel a bit out of sync sometimes. You see a spike whizz by, but the label failed to capture it. It might be a bit annoying to have the number changing too often also though.

foot:
We could separate out the logic a little more so that ticking/buffer-merging happened in an `AnimatedSparkline`, but that might get in the way again if we wanted to have a SmoothScrollingAnimatedSparkline in the future too..

Anywho, looks good!

davkal:
The latest value next to the sparkline was out of sync because it was not rendered by the bit-by-bit fed sparkline. The logic to feed the data has been extracted but this needs to be adapted when #752 is merged, because the number layout changes there. I suggest we wait for that.

BONUS: smooth out the y-domain, maybe via an exponentially dampened moving average like load itself.

davkal:
Format renderer fixes are included, fixes #902

foot:
Looks good!

Minor issues:

- If you get used to the animating sparklines its a surprise to see them sit there statically for 5s when the details panel is first opened. (cheat?: Initial "load 10", and stream in the next 5 for immediate animation?)
- Load jumping around: Maybe we could record the greatest value we've seen so far and use that as the max. Its fun to watch though, sometimes looks like a never ending staircase.
- Top memory render gently "bumps" the layout when changing values sometimes. E.g. 302MB -> 302.1MB will make the container a tiny bit bigger to fit in the "1".

foot:
Code looks good.

HOC FTW LGTM

davkal:
Final comment: the child tables are not fed by the feeder, so the big top metrics seem to change all the time, while the metrics in the tables stay the same til the next details request. This comes with 2 drawbacks: (1) they dont look like they re updating that often. I tried using the feeder there too, but it was quite CPU heavy and they quickly became out of sync. (2) as the big metric is going through the samples it may have a lower value than that of a child metric, which wont add up, e.g. host CPU < container CPU.

foot:
LGTMTM

PR #732, opened 2015-12-08
Merge probe and app binaries, run with ./scope probe or ./scope app.

tomwilkie:
The benefits here are (1) 50% reduction in build time and (2) almost 50% reduction in container size.

Fixes #569

PR #796, opened 2016-01-05
Use argv[0] for process name, differentiate scope app and probe.

tomwilkie:
As of #732 scope became a single binary, and it became hard to differentiate the app from the probe.  This PR reintroduces the naming difference, and also bring the process names in scope more inline with `ps`.


paulbellamy:
It seems like this PR does two (somewhat) unrelated things, changing the name of the scope processes, and changing what is rendered for a process name/command.  Should (at least) be 2 different commits, please.

tomwilkie:
@paulbellamy PTAL

paulbellamy:
LGTM

Issue #764, opened 2015-12-14
Control errors persist when switching details pane, and prevent closing the details pane


paulbellamy:
To Reproduce:

1) Open details pane on nodeA
2) Click a control and get an error
3) Try to close the details pane by clicking the backround (it doesn't do anything)
4) Open details pane on nodeB
5) Node that the error from nodeA is still showing

tomwilkie:
Probably related to #758

PR #797, opened 2016-01-05
Fix node controls so they behave independently across nodes

### LINKED ISSUE #758: An outstanding control on one container prevents you doing another operation on a different container.
Sometimes the docker controls (stop, start etc) take some time, and we show a spinner whilst they are outstanding.

If I select a different container, I still see the spinner, even though the outstanding control operation was on a different container.  I'd like to be able to do controls on different containers in parallel.

foot:
If one node was busy performing an action, any other nodes' controls
would also be in a 'blocked and waiting' state.

fixes #758 and #764

### DIFF
--- a/client/app/scripts/actions/app-actions.js
+++ b/client/app/scripts/actions/app-actions.js
@@ -3,7 +3,7 @@ import debug from 'debug';
 import AppDispatcher from '../dispatcher/app-dispatcher';
 import ActionTypes from '../constants/action-types';
 import { updateRoute } from '../utils/router-utils';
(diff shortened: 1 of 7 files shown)


davkal:
Restarting the second node affects the first node:
still happening

Both were alpine images.

foot:
Whoops! lets have a look..

foot:
This works for me! Cleared cache?

davkal:
Weird. works for me too now. so LGTM

foot:
Whoop!

PR #806, opened 2016-01-11
Less hacky fix for /%2F bug

### LINKED ISSUE #804: /api/topology/applications-by-name/%2Fhome%2Fweave%2Fweave gets redirected to /api/topology/applications-by-name/home/weave/weave
A bug in go / gorilla: https://github.com/gorilla/mux/issues/142

I can probably do a workaround.

tomwilkie:
Fixes #804

### DIFF
--- a/app/router.go
+++ b/app/router.go
@@ -29,48 +29,86 @@ var (
 // mistakenly unescapes the Path before parsing it.  This breaks %2F (encoded
 // forward slashes) in the paths.
 func URLMatcher(pattern string) mux.MatcherFunc {
-	matchParts := strings.Split(pattern, "/")
-
 	return func(r *http.Request, rm *mux.RouteMatch) bool {
-		path := strings.SplitN(r.RequestURI, "?", 2)[0]
-		parts := strings.Split(path, "/")
-		if len(parts) != len(matchParts) {
-			return false
+		vars, match := matchURL(r, pattern)
+		if match {
+			rm.Vars = vars
 		}
-
-		rm.Vars = map[string]string{}
-		for i, part := range parts {
-			unescaped, err := url.QueryUnescape(part)
-			if err != nil {
-				return false
-			}
-			match := matchParts[i]
-			if strings.HasPrefix(match, "{") && strings.HasSuffix(match, "}") {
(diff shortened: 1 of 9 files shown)


app/router.go
```diff
@@ -68,11 +68,28 @@ func RegisterTopologyRoutes(c Reporter, router *mux.Router) {
...
+// means the request has been handled.
+func HandleNode(c Reporter, w http.ResponseWriter, r *http.Request) bool {
+	matcher := URLMatcher("/api/topology/{topology}/{id}")
+	rm := mux.RouteMatch{}
+	if !matcher(r, &rm) {
+		return false
+	}
+	topologyID := rm.Vars["topology"]
+	nodeID := rm.Vars["id"]
+	if nodeID == "ws" {
```
paulbellamy: Won't we still have the same bug for websockets, then?

prog/app.go
```diff
@@ -18,14 +18,21 @@ import (
...
 	router.Methods("GET").PathPrefix("/").Handler(http.FileServer(FS(false)))
-	return router
+
+	// need to handle some routes in a completely custom way, because
+	// https://github.com/weaveworks/scope/issues/804
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if !app.HandleNode(c, w, r) {
+			router.ServeHTTP(w, r)
+		}
+	})
```
paulbellamy: Don't like that this is in 2 places. Maybe extract it somewhere into a function with the explanation there. Could just inline it into `app.HandleNode`, maybe.

paulbellamy:
ugly, but can't think of a better way to do it.

tomwilkie:
I think this is beyond even my ugliness threshold.  @peterbourgon have any suggestions?  Otherwise I'll take another stab as bashing it into shape.

tomwilkie:
> Won't we still have the same bug for websockets, then?

None of the topology view name start with (or contain) a slash, so we won't have this bug.

> Don't like that this is in 2 places.

Have updated and rejigged this a little so its less ugly.  Also added a test.  @paulbellamy PTAL.

PR #814, opened 2016-01-13
[WIP] Rate limit reading expensive files from /proc

### LINKED ISSUE #812: Probe using 70% CPU
@2opremio reports the probe is using 70% CPU on the service hosts.

tomwilkie:
Fixes #812

### DIFF
--- a/integration/340_process_edge_across_host_2_test.sh
+++ b/integration/340_process_edge_across_host_2_test.sh
@@ -16,7 +16,7 @@ weave_on $HOST2 run -dti --name client alpine /bin/sh -c "while true; do \
 	sleep 1; \
 done"
 
-sleep 30 # need to allow the scopes to poll dns, resolve the other app ids, and send them reports
+sleep 60 # need to allow the scopes to poll dns, resolve the other app ids, and send them reports
 
 check() {
     has processes $1 "nginx: worker process"
--- a/probe/endpoint/procspy/backgound_reader.go
+++ b/probe/endpoint/procspy/backgound_reader.go
@@ -12,9 +12,9 @@ import (
 )
 
 const (
-	timeout   = 60                     // keep contents for 60 seconds
-	size      = 10 * 1024 * 1024       // keep upto 10MB worth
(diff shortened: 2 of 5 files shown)


tomwilkie:
This seems to have some unintended side effects on the edges that get detected.  Needs more work.

tomwilkie:
Have rebased, increased limits and made it so results are deleted once used (a bit racey right now, just an experiment).  Results are improved but he connections are still flakey and sometime outright odd.

Its worth testing without conntrack (as that has known bugs) so you know any mis-connections are due to this change:

`./scope launch --probe.conntrack=false`

2opremio:
I've experimented with grouping the processes by namespace to ensure that all the reads of `/proc/PID/fds` are done immediately after reading `/proc/PID/net/tcp{,6}`.

However, I still miss a lot of connections.

The cause most likely lays in the inconsistencies resulting from caching  `/proc/PID/net/tcp{,6}`  and not `/proc/PID/fds`. I will start by caching both to see if it solves the problem.

2opremio:
This is still hacky/work in process, but works well. Now there's a background gorountine walking `/proc/PID/fd/*` and `/proc/PID/net/tcp{,6}` in a loop with dynamically adjusted ratelimiting.  

On the way I found out that we were not reading the network namespace inodes properly (the inode of the symlink is not the inode of the namespace),  resulting in reading `/proc/PID/net/tcp{,6}` for all processes (see f63b637).  

In the average case that should provide over one order of magnitude of speedup to PID walker (e.g. in my machine I hat 150 processes and only 7 net namespaces).

So, for now, maybe it's enough to simply cherrypick f63b637 and probably 2c32110 .

2opremio:
I have made some measurements to see if it's worth having a background reader after fixing the namespace reading and it doesn't seem to be the case.

With 60 network namespaces, scope is at ~15% CPU in both cases. BTW, master is over 30% CPU, which gives us a 2X :).

So, I am closing this in favor of #898 and I will cherry-pick f799dc8 into a separate PR.

tomwilkie:
+1

PR #823, opened 2016-01-15
k8s: Filter out pause as a system container

### LINKED ISSUE #733: Filter kubernetes pause container as a system container
In Kubernets every pod implicitly includes a root `pause` container, whose network namespace is referenced by the other pod containers to ensure a single, shared namespace per pod.

This containers should be considered system containers, since they don't have any app-level functionality and clutter the view.

2opremio:
This makes a huge difference when viewing containers in a kubernetes cluster since every pod contains a pause container, completely cluttering the view.

Fixes #733 

@tomwilkie @paulbellamy PTAL (although the change is trivial)

### DIFF
--- a/render/filters.go
+++ b/render/filters.go
@@ -184,10 +184,12 @@ var systemContainerNames = map[string]struct{}{
 }
 
 var systemImagePrefixes = map[string]struct{}{
-	"weaveworks/scope":        {},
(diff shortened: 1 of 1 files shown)


tomwilkie:
LGTM

paulbellamy:
Seems like a good idea to me!

PR #824, opened 2016-01-15
Correctly extract image name when using custom hub.

### LINKED ISSUE #820: containers-by-image view collapses all images provided by a docker registry on non-standard port
When using images from a docker registry on a non-standard port, it looks like the logic to collapse multiple versions of the same image is also collapsing all containers from the registry.

e.g.

    docker-registry.domain.name:5000/image1
    docker-registry.domain.name:5000/image2

are both displayed as docker-registry.domain.name in the containers-by-image view.

adding a version doesn't help:

    docker-registry.domain.name:5000/image3:latest

is also collapsed into docker-registry.domain.name.

tomwilkie:
Fixes #820

### DIFF
--- a/render/mapping.go
+++ b/render/mapping.go
@@ -541,11 +541,12 @@ func MapPod2Service(n RenderableNode, _ report.Networks) RenderableNodes {
 }
 
 func imageNameWithoutVersion(name string) string {
-	parts := strings.SplitN(name, ":", 2)
-	if len(parts) == 2 {
-		return parts[0]
+	parts := strings.SplitN(name, "/", 3)
+	if len(parts) == 3 {
+		name = fmt.Sprintf("%s/%s", parts[1], parts[2])
 	}
-	return name
+	parts = strings.SplitN(name, ":", 2)
+	return parts[0]
 }
 
 // MapContainerImage2Name maps container images RenderableNodes to
--- a/render/mapping_internal_test.go
+++ b/render/mapping_internal_test.go
@@ -1,0 +1,20 @@
+package render
+
+import (
+	"testing"
+)
+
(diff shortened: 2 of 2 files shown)


paulbellamy:
I'd like to see a few tests for ones which are already without a version.

paulbellamy:
also, that regex is awful.

tomwilkie:
> I'd like to see a few tests for ones which are already without a version.

AFAICT the docker API will never give you images without versions.  This regex expects a version.

> also, that regex is awful.

Awfully good? :-) But seriously, whats awful about it?  I might remove the names matches, but otherwise it seems pretty minimal and robust.

paulbellamy:
Also, images don't have to have a slash in them, do they?

2opremio:
> Also, images don't have to have a slash in them, do they?

Very true, `ubuntu, `busybox` etc ...

paulbellamy:
> This regex expects a version.

One of the assumptions I had when writing #752, was that this function is idempotent. It shouldn't break anything, but would be nice to maintain that if possible.

tomwilkie:
> Very true, ubuntu, busybox etc ...

Actually those ones do (they come from library/busybox)

But images you make locally won't, so that does need fixing.  Unfortunately I think that makes parsing non-deterministic (how to differentiate from hostname/foo vs org/foo).  I guess something like "first split on slash, if 2 -> first is org, if 3 -> first is repo" might work?

> One of the assumptions I had when writing #752, was that this function is idempotent. It shouldn't break anything, but would be nice to maintain that if possible.

I see - so you want versions to be optional.  Thats totally possible.

tomwilkie:
@paulbellamy PTAL

paulbellamy:
LGTM

PR #837, opened 2016-01-18
Stops scope crashing chrome when we get "long" edges.

### LINKED ISSUE #830: Web application becomes unresponsive at higher container counts
I've got scope v0.11.1 (running locally, not using web service) set up on a cluster of 5 hosts, with ~60-70 containers and the web ui seems to be completely unresponsive. The initial scope page loads, but immediately causes the browser tab to lock up and start consuming >100% cpu. I've noticed that this problem goes away when I run on only 1 or 2 hosts (so the container count is only around 20).

Has anyone else seen similar client-side performance issues with larger topologies? 

Relevant Info:
```
Docker version: Docker version 1.9.1, build a34a1d5
Weave version: 1.4.2
Scope command: docker run --privileged -d --name=weavescope --net=host --pid=host -v /var/run/docker.sock:/var/run/docker.sock weaveworks/scope:0.11.1 --probe.docker true
Browser: Chrome version 47.0.2526.106 (64-bit)
```

foot:
Long edges are those w/ a lot of points on them (control points, as in, the edge meanders around a lot). 

Which can pop up in heavily connected layouts.

Fixes #830

### DIFF
--- a/client/app/scripts/charts/edge.js
+++ b/client/app/scripts/charts/edge.js
@@ -84,11 +84,11 @@ export default class Edge extends React.Component {
   }
 
   ensureSameLength(points) {
-    // Spring needs constant list length, hoping that dagre will insert never more than 10
-    const length = 10;
+    // Spring needs constant list length, hoping that dagre will insert never more than 30
+    const length = 30;
     let missing = length - points.length;
 
-    while (missing) {
+    while (missing > 0) {
       points.unshift(points[0]);
       missing = length - points.length;
     }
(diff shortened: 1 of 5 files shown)


davkal:
@foot nice debug bar! but which hunk is fixing #830?

davkal:
screen shot 2016-01-19 at 08 51 49

foot:
1. `while (missing) {`

- dagre was supplying more than 10 points
- `missing` went negative immediately
- inifinite loop!

2. Long lines were then appearing truncated, so bumped the points up to 30.

3. Semi related fix: `if (!layout) { return layout; }` fixes the "too many nodes" message appearing, before that it was passing a null layout into the next fn which was throwing an exception.

davkal:
Neato! LGTM

PR #841, opened 2016-01-20
Rename metric keys so they don't conflict and get merged between types

### LINKED ISSUE #840: Metrics on containers and hosts are wrong
Metrics are being merged up from processes -> containers -> hosts, causing spurious data in containers and hosts metrics.

paulbellamy:
This stops us aggregating different types of metrics between nodes.

I also made the metrics in the test fixture a bit more specific, though there are no tests which use this.

Fixes #840


paulbellamy:
Actually, we could also just rename `process.CPUUsage`, so it doesn't conflict with the other keys and get merged.

paulbellamy:
I've just renamed the metric keys so they won't conflict. However the result still seem very random for me. @tomwilkie can you confirm that this fixes the issue for you?

tomwilkie:
> However the result still seem very random for me. @tomwilkie can you confirm that this fixes the issue for you?

This seems to fix for me (LGTM).  What random results where you seeing?  Its doesn't quite match top, but its close (mine varies from 35-20%, top says 25%)

paulbellamy:
Yeah, it wavers around top a bit randomly. I think that is just polling intervals/etc. For me it had quite a quite margin around top it would be in, but that could just be the stuff I was running was spiky.

tomwilkie:
Yeah we take a shorter window than top.  We can harmonise them later (when we have a TSDB).

PR #842, opened 2016-01-20
Various Refactoring

paulbellamy:
3 refactoring commits, in order of increasing contentiousness.

First, is just some missed constants.

Second, is centralizing the label rendering into one file, so all the "human-readability" mappings are easy to find.

Third, is an experiment with explicitly allowed conversions between node types, and rendering metadata/metrics on aggregated groups of nodes.

When we group each "type" of nodes (process, container, etc), by some key, we should deny all Metadata, Metrics, Latest, etc (except the key we are merging on), and explicitly aggregate the ones which make sense (so far only Sets, but Metrics could with some summation func).

To make rendering groups of nodes a bit saner, I added a rule that any metadata key beginning with `_` is ignored, (so we can automatically filter out "internal" keys), this could probably be implemented cleaner (i.e. in `render/detailed/labels.go`).

Thoughts, @tomwilkie?


tomwilkie:
(1) is fine, just stick that straight in master.

re: (2) I want the keys to be the same for the metrics across 'kinds'.  So this seems like the wrong direction, although I need to think about it some more.

re: (3) I don't think I understand what your trying to do.  I like the idea of dropping node fields when producing a group, but I'm not keen on rules like "any metadata key beginning with _ is ignored" and would prefer thing kinda thing be much more structural.  lets discuss more next week.

either way lets split this up so we can have more focused discussion.

paulbellamy:
re: (2) I agree, its just a workaround.

re: (3) There are kind of three ideas smashed into it, firstly dropping all node fields unless explicitly "converted". This is probably eventually the "real" solution to (2), but needs a lot more work to put in everywhere.

second, rendering metadata/metrics on aggregate nodes (Processes-by-name, etc), having rules for which metadata/metrics are ok to keep when grouping e.g. processes. It's not strictly needed, but lays out a pattern for doing summed metrics, and makes grouped nodes look a bit nicer.

the second implies we need a way to render summaries for aggregate nodes (which don't map back to a topology). In this implementation we just iterate over all the metadata/latest/metric keys, and render any present. Unfortunately metadata also contains stuff we don't want to render (e.g. IsConnected). The "_ is ignored" rule is a bit of a hack to work around that. A cleaner way would probably be to say that "if Labels() does not have a defined human text for a key do not render it".

tomwilkie:
> firstly dropping all node fields unless explicitly "converted".

I 100% support this.

> second, rendering metadata/metrics on aggregate node

Other than aggregated metrics, what would be useful here?

My opinion (without a huge amount of thought...) is we shouldn't be rendering any metadata on this groups of nodes - we treat them specially (with the special stack icon).  For metrics, I think we should perhaps rely on our tsdb to do the aggregation, so in effect this become a problem from how we layout the data there.

paulbellamy:
I misread which commit 2 is above.

re: (2) I don't see how this stops sharing keys across types. It's just centralizing the english-translation of names. If you have only one key which was the same across kinds you would expect it to always have the same label when rendered. Doing this was actually how I found that the process and host cpu keys were overlapping, as I got a duplicate key error.

paulbellamy:
You could also do grouped counters etc, for example "Total # Threads", or "# Containers Running"

paulbellamy:
When doing it, I actually implemented rendering groups, and the centralized labels dict kind of fell out of that. But it is nice having all that in one place, so re-ordered the commits.

paulbellamy:
Merged all except (3) into master.

PR #849, opened 2016-01-21
Fixes exec/attach from the 2nd+ card.

### LINKED ISSUE #847: Cannot exec/attach from 2nd+ card.
Terminal only works if activated from the single bottom card.

foot:
- Also fixes strange issue where hitting ESC while a terminal was
  active, would update the route, reload the top card, and break the
  terminal connection status.

Fixes #847


foot:
RFR

tomwilkie:
Can confirm, this does indeed fix the problem.  LGTM

PR #853, opened 2016-01-24
Add container uptime and restart count to details panel.

### LINKED ISSUE #844: Uptime monitoring for containers
It'd be very cool to see uptime and restart counts expose, Docker API exposes `State.StartedAt` and `RestartCount` which can be used for this quite easily.

tomwilkie:
Fixes #844

### DIFF
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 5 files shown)


paulbellamy:
Container uptime is very specific (e.g. `3h0m9.382506493s`). Let's truncate it to seconds.

Restarts seems to always be 0.. Does it only count restart policy restarts? Does it only work if there is a restart policy?

tomwilkie:
> Container uptime is very specific (e.g. 3h0m9.382506493s). Let's truncate it to seconds.

Fixed

> Restarts seems to always be 0.. Does it only count restart policy restarts? Does it only work if there is a restart policy?

This I do not know.  I couldn't get it to != 0.  Needs more work.

tomwilkie:
Actually works fine:

screen shot 2016-01-27 at 19 30 59

It only shows restarts due to restart policy, so doing this:

```
docker run -dti --restart=always alpine /bin/sh
```

And then killing the `/bin/sh` increments the counter.

PR #858, opened 2016-01-26
use immutability for the NodeSet

paulbellamy:
Part of #502, #854

Gives a small performance boost (~30%) on the HostRenderer report.json.

### DIFF
--- a/report/node_set.go
+++ b/report/node_set.go
@@ -33,7 +33,7 @@ func (n NodeSet) Add(nodes ...Node) NodeSet {
 		result = ps.NewMap()
 	}
 	for _, node := range nodes {
-		result = result.Set(fmt.Sprintf("%s %s", node.Topology, node.ID), node)
+		result = result.Set(fmt.Sprintf("%s|%s", node.Topology, node.ID), node)
 	}
 	return NodeSet{result}
 }
--- a/report/node_set.go
+++ b/report/node_set.go
@@ -1,94 +1,197 @@
 package report
 
 import (
+	"bytes"
+	"encoding/gob"
+	"encoding/json"
+	"fmt"
+	"reflect"
 	"sort"
+
+	"github.com/mndrix/ps"
 )
 
-// NodeSet is a sorted set of nodes keyed on (Topology, ID). Clients must use
+// NodeSet is a set of nodes keyed on (Topology, ID). Clients must use
 // the Add method to add nodes
-type NodeSet []Node
+type NodeSet struct {
+	psMap ps.Map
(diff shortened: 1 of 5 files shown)


tomwilkie:
I'm surprised this results in such a big performance increase, but I guess its mainly due to the cheap copy.  LGTM.

report/node_set.go
```diff
@@ -1,94 +1,197 @@
...
-		})
-		if i < len(n) && n[i].Topology == node.Topology && n[i].ID == node.ID {
-			// The list already has the element.
-			continue
-		}
-		// It a new element, insert it in order.
-		n = append(n, Node{})
-		copy(n[i+1:], n[i:])
-		n[i] = node
+		result = result.Set(fmt.Sprintf("%s %s", node.Topology, node.ID), node)
```
tomwilkie: You could consider using something other that space for the key...

report/node_set.go
```diff
@@ -1,94 +1,197 @@
...
+	"github.com/mndrix/ps"
 )
 
-// NodeSet is a sorted set of nodes keyed on (Topology, ID). Clients must use
+// NodeSet is a set of nodes keyed on (Topology, ID). Clients must use
 // the Add method to add nodes
-type NodeSet []Node
+type NodeSet struct {
+	psMap ps.Map
+}
```
tomwilkie: Why is NodeSet a set of nodes and not a set of ids?  I thought this was just used for children?

tomwilkie:
Also, shouldn't NodeSet be in render/?

report/node_set.go
paulbellamy: It is used for children. But unlike parents, we summarise children, so we have to use the aggregated nodes. This way we store the nodes as we are "rolling up" the aggregations.

report/node_set.go
paulbellamy: Sure. It doesn't really matter, aside from sorting.

paulbellamy:
> Also, shouldn't NodeSet be in render/?

It is `Set<report.Node>`, so I felt it should live in `report`.  If it were a `Set<render.RenderableNode>`, then I'd say it should live in `render`.

report/node_set.go
tomwilkie: Ah yes good point.

tomwilkie:
> It is Set<report.Node>, so I felt it should live in report. If it were a Set<render.RenderableNode>, then I'd say it should live in render.

Ah yes we had that discussion before.  Sorry I forgot.

PR #859, opened 2016-01-26
fix bug in containers being filtered by image name

### LINKED ISSUE #857: System container filtering by image is broken
Using current Scope/master:

screen shot 2016-01-26 at 13 00 46

I can see containers with image `beta.gcr.io/google_containers/pause` [whilst I shouldn't](https://github.com/weaveworks/scope/blob/6b5647576642555d91ff42f3d15b500ff818b308/render/filters.go#L208)

This must have broken recently, it was working last week. Maybe as a side-effect of https://github.com/weaveworks/scope/pull/838 ?

After some digging, it seems `node.Latest.Lookup(docker.ImageName)` gives an empty result most of the time. 

Adding `fmt.Printf("Found image prefix '%s' (image '%s') for container '%s'\n", imagePrefix, imageName, containerName)` after [this line](https://github.com/weaveworks/scope/blob/6b5647576642555d91ff42f3d15b500ff818b308/render/filters.go#L171) results in logs like:

```
Found image prefix '' (image '') for container 'k8s_controller-manager.d5a82dcf_k8s-master-127.0.0.1_default_8eff06875acb5697c821b26b569968fb_cd187338
```

paulbellamy:
Fixes #857

Was introduced in #838

### DIFF
--- a/render/topologies.go
+++ b/render/topologies.go
@@ -140,7 +140,7 @@ func (r containerWithImageNameRenderer) Render(rpt report.Report) RenderableNode
 			continue
 		}
 		c.Rank = ImageNameWithoutVersion(image.LabelMajor)
-		c.Latest = c.Latest.Merge(c.Latest)
+		c.Latest = image.Latest.Merge(c.Latest)
 		containers[id] = c
 	}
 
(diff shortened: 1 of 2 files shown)


2opremio:
LGTM (pending the tests)

PR #861, opened 2016-01-27
Refactored render/memoise and added a basic test for it

paulbellamy:
No performance impact, but encapsulates the caching mechanism a bit more.

@tomwilkie wdyt?


tomwilkie:
This is _much_ nicer.  I am embarrassed.

PR #862, opened 2016-01-27
Make render.RenderableNodes immutable

paulbellamy:
Part of #502, #854

Possibly, actually seems to slow things down, slightly... Still within range of being noise...

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -214,13 +214,13 @@ func decorateWithStats(rpt report.Report, renderer render.Renderer) topologyStat
 		realNodes int
 		edges     int
 	)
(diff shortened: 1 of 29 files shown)


tomwilkie:
I'm not convinced this is needed - RenderableNodes collections are never shared or copied, right?  There use is quite restricted to the rendering pipeline.  ps.Maps are going to be slower here than normal go maps.

Worth taking 00afc20 and making it a separate PR though.

tomwilkie:
Either way this doesn't look urgent for v0.12, so can we park it and discuss later?

paulbellamy:
Decided to extract refactoring, and remove the render.RenderableNodes immutability bits, as they do not have a performance improvement.

PR #880, opened 2016-01-29
Fixes host node/details panel color mismatch

### LINKED ISSUE #871: Mismatching details pane color for host nodes
I found this while testing the 0.12 candiate. Note how the nodes are redish and the details pane is purple:

screen shot 2016-01-28 at 12 37 52

I recall @davkal fixed this for all nodes in https://github.com/weaveworks/scope/issues/684 so it's probably a regression.

foot:
- 'label_major' has been changed to 'label' in the details api response.
- getNodeColorDark was giving up too early if a rank wasn't provided.
  The topo diagram which uses getNodeColor will select a color even w/
  no rank. Hosts don't have ranks at the mo.

Fixes #871

### DIFF
--- a/client/app/scripts/charts/node.js
+++ b/client/app/scripts/charts/node.js
@@ -25,8 +25,8 @@ export default class Node extends React.Component {
     }
     let labelOffsetY = 18;
     let subLabelOffsetY = 35;
-    const isPseudo = !!this.props.pseudo;
-    const color = isPseudo ? '' : getNodeColor(this.props.rank, this.props.label);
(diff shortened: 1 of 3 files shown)


davkal:
The `Loading...` looks a bit pale

screen shot 2016-02-01 at 13 35 39

foot:
I will do something about this!

foot:
No, I am having trouble doing something about this. Its white for my host-nodes! Clear-cache?

davkal:
It seems like this style happens while react reconciliation gets interrupted by #891 
So, this PR LGTM

PR #884, opened 2016-01-29
Node shapes

### LINKED ISSUE #659: Consider puttting icons in the fried eggs
There was a suggestion we start putting icons in the fried eggs.  We could show docker container state, or infer some application icons, put a cloud icon in the internet node etc.  Thoughts?

### LINKED ISSUE #826: Confusing 'by hostname' view when containers have their own hostnames
I don't know if this is intentional and I simply don't understand the "by hostname" view" but, the "by hostname" view of my kubernetes cluster includes the hostnames from the containers, which I find confusing:

screen shot 2016-01-15 at 5 08 28 pm

```
$ docker ps  | grep app-mapper | grep -v pause
faf115d00c15        weaveworks/app-mapper-db                             "/docker-entrypoint.s"   3 days ago          Up 3 days                               k8s_app-mapper-db.3281346e_app-mapper-db-jimdf_default_10d549d8-b93d-11e5-a532-0800279b9f1d_9d3d0a02
$ docker exec -ti faf115d00c15 bash
root@app-mapper-db-jimdf:/# hostname
app-mapper-db-jimdf
```

### LINKED ISSUE #878: Make container-by-hostname and processes-by-name icons a little stack.
This might be a dupe.

This came up in the session with Darren from Rancher.

foot:
Fixes #659.
Fixes #878.
Fixes #826.


foot:
gif of new shapes!

foot:
Shapes not finallll..

tomwilkie:
I love it.  

- Is the shape hard coded into the frontend?  How do you choose?
- I think we could add a 'shape' and 'group' (bool) field to a node details struct and have the backend send you this if you like?
- The hexagon should be use for k8s, for sure

foot:
We are dispatching on the tail of the topo's URL (`containers|containers-by-name...)`. `pseudo` nodes are drawn as circles and `theinternet` is special cased. [0]

Solutions: 
 1. Add `type: (host|container|app), grouped: (true|false)` props to the node data.
 2. Add a `shape: (circle|square|stacked-circle|cloud)` and let the BE choose.

I prefer 1. but if 2. makes more sense we could do that.

[0]
```js

const nodeShapes = {
  'hosts': NodeShapeCircle,
  'containers': NodeShapeHex,
  'containers-by-hostname': stackedShape(NodeShapeHex),
  'containers-by-image': stackedShape(NodeShapeHex),
  'applications': NodeShapeRoundedSquare,
  'applications-by-name': stackedShape(NodeShapeRoundedSquare)
};

function isTheInternet(id) {
  return id === 'theinternet';
}

function getNodeShape({id, pseudo, topologyId}) {
  if (isTheInternet(id)) {
    return NodeShapeCloud;
  } else if (pseudo) {
    return NodeShapeCircle;
  }
  return nodeShapes[topologyId];
}
```

paulbellamy:
Don't forget pods and services.

foot:
Do pods and services exist in the system at the moment?

paulbellamy:
Yes, but only if connected to a k8s cluster.

davkal:
> Don't forget pods and services.

Maybe the drop shape from the style studies for pods? 

And for services maybe a "S" that is a circle? Like
screen shot 2016-02-01 at 12 17 15
Or maybe that is trying to be too smart

More importantly, we need a fall-back shape that gets applied when developing new topologies. Maybe a non-rounded rect?

tomwilkie:
I'd prefer if the ui didn't hard code any topology names, as we want the
user to be able to define there own topologies when we add the dsl.

tomwilkie:
I'll take a stab at adding this to the backend today.

tomwilkie:
Just wondering if the stacks would look slightly better if the stacked diagonally (vs the current vertical)?

tomwilkie:
If we wanted to be super awesome, we could also provide the number of things in a stack and render appropriately?

tomwilkie:
We should also make the uncontained nodes on the containers view a stack of squares, as they are processes...

errordeveloper:
Very nice!!

tomwilkie:
Have rebased. Pseudo nodes don't seem to be honouring backend shapes.

foot:
Yes I think there is some code in there that forces pseudo = circle. Will fix. And review.

foot:
@davkal LGTM

foot:
Sooo!

Rational behind the current shapes:

- Apps/processes: **Rounded square**. Alternatives: _Simple Gear_, _Hexagon_
  - Common association now w/ mobile apps etc.
- Containers: **Hexagon**. Alternatives: _Rounded Square_
  - npolys have an association w/ the container world thanks to k8s.
  - Squares have harsh sharp edges, rounding them would confuse them with Apps. (We could choose another shape for Apps...).
  - Vertical Hexagon gives visual illusion of a 3D cube
- Hosts: **Circle**
  - It is the graph super shape. Circles are the base of most graphviz.
  - Host is the super node in the system, nothing beats host, maybe data-center.

foot:
I mocked up proc:gears and container:squares, unfortunately squares feel very natural as containers.

alt shapes

foot:
Re: diagonal stacks, I think they give too much of a 3D feel to our otherwise very flat UI.

screen shot 2016-02-10 at 17 45 55

foot:
Nicer, more subtle cog shape, evokes associations w/ ports/connections i think.

screen shot 2016-02-10 at 18 00 21

davkal:
Re diagonal shapes: I'd keep them as top perspective. you could try having the vanishing point on the right (instead of the bottom), like in the detail cards

Re cogs for processes, I think they are overloaded with "system" or "settings" semantics.

foot:
nice thought re: perspective, will try.

Cool. So initial node shape choices seem like a good starting point for everyone it seems. Maybe we can merge this in and see what complaints come in.

foot:
I like the gentle perspective!

Vertical feels like looking down onto the topo and makes the highlight shape more natural. 

screen shot 2016-02-11 at 21 48 15

foot:
@pidster check it out and have a play when you get a chance! Great to get your feedback on this one.

pidster:
Tom's suggestion for the stacks showing as 1x, 2x and all >2x as 3 layers is a good idea.
An aggregate that indicates multiple but only has one member will be confusing.

pidster:
The vertical perspective stack looks great.

So, where does it stand now: Processes are squircles, Containers are hexagons (or cogs?) and Hosts are circles?

Are the line weights the same in all of the shapes, or is that just a side-effect of the cog rendering?

foot:
The variants, in a classic rustic saddle brown.

screen shot 2016-02-17 at 09 27 47

davkal:
> An aggregate that indicates multiple but only has one member will be confusing.

@foot while I like the gray disks a lot, I concur with @pidster 
Could we get rid of the gray ones?

PR #888, opened 2016-01-29
Don't need sudo when DOCKER_HOST is tcp.

tomwilkie:


peterbourgon:
Does _anyone_ need sudo anymore? How about if you need sudo you just `sudo make`?

tomwilkie:
I think a default install of docker on linux needs sudo to run things.  Doing a sudo make will result in stuff in you directory being owned by root.

peterbourgon:
Okeedokey.

paulbellamy:
I don't need the `SUDO` bit, so unless we need it in CI, I say scrap it.

tomwilkie:
As discussed, we're going to keep the SUDO.

Issue #881, opened 2016-01-29
Deadlock in probe reporter


2opremio:
I started the [ECS demo](http://www.weave.works/guides/service-discovery-and-load-balancing-with-weave-on-amazon-ecs-2/) with the 0.12 candidate and let it run overnight. When I checked its state in the morning all the Scope UIs were empty and the probes were all consuming a lot of memory.

The demo uses 3 hosts with Scope running in Standalone mode. Unfortunately, the probes weren't run with `--probe.http.listen` so I don't have Golang memory consumption details. This is the Virtual/RSS memory consumption of the probe running on each host:

* Probe A:  1108MB/597MB
* Probe B: 592MB/218MB
* Probe C: 657MB/356MB

As I said, the Scope UIs in all hosts are empty, indicating that apart from consuming a lot of memory they are not functioning properly. Here are the logs of Probe C:

https://gist.github.com/2opremio/9414979902b50ded7245

Around 4:30 the machine starts to run out of memory and around 4:50 the probe was killed but didn't recover gracefully. Other machines have similar logs, with machine A having experienced several restart so this seems like a leak.

The restarts are not graceful since the UI is empty in all hosts

I would say there are two problems:

* The high memory consumption/leak
* The probes not recovering gracefully when they are killed.

I will leave the probes running for a while in case it's not reproducible and spin up another demo with `--probe.http.listen` to try to reproduce and get detailed memory debugging information.

2opremio:
I have added the debug output of SIGQUIT and SIGUSR1 handlers of Probe C at the end of https://gist.github.com/2opremio/9414979902b50ded7245

It's not a goroutine leak and `alloc_bytes': 354030016.000` which matches the RSS I observed

2opremio:
After reproducing in a separate cluster with `--probe.http.listen` one of the probes is at ~100MB allocations (RSS is at 255MB).

Heap profile: http://filebin.ca/2V8LhzSwijC4/pprof.localhost9999.inuse_objects.inuse_space.003.pb.gz

![probe_heap_profile](https://cloud.githubusercontent.com/assets/2362916/12677475/3dd33476-c691-11e5-980e-6b9b36aa51c0.png)

2opremio:
I am not super familiar with this part of the code, but it seems like [c.pendingStats](https://github.com/weaveworks/scope/blob/6b5647576642555d91ff42f3d15b500ff818b308/probe/docker/container.go#L213) is growing out of control.

[...]

PR #889, opened 2016-02-01
Taking a read lock twice only works most of the time.

tomwilkie:
This explains #881 - if we try and take a write lock during this window, we lock up the reporter thread, and that causes a bunch of data structures to grow linearly.

We should also stop these DSes growing linearly.

### DIFF
--- a/probe/overlay/weave.go
+++ b/probe/overlay/weave.go
@@ -207,8 +207,6 @@ func (w *Weave) Tag(r report.Report) (report.Report, error) {
 	}
 
 	// Put information from weave ps on the container nodes
-	w.mtx.RLock()
-	defer w.mtx.RUnlock()
 	for id, node := range r.Container.Nodes {
 		prefix, _ := node.Latest.Lookup(docker.ContainerID)
 		prefix = prefix[:12]


2opremio:
Looking at the sources, of `RLock` I can confirm that a goroutine acquiring the read-lock will cause a deadlock calling `RLock` again if another gorouting calls `Lock` between the initial read-lock acquisition and the `RLock` call. This is because pending write locks [have a higher priority ](https://github.com/golang/go/blob/7b767f4e521c2481e08051c843badd0382fde3b0/src/sync/rwmutex.go#L36-L37) than read locks (even if the go routing already acquired the read-lock).

This is not explicit in the documentation and not intuitive either,  in fact, the docs made me originally think that this was OK:

> The lock can be held by an arbitrary number of readers or a single writer.

so I will create a PR upstream.

But first, I will spawn this fix in ECS to see if the problem has gone away.

2opremio:
This seems to fix #881 since I cannot reproduce the deadlock (and high memory consumption of the probe) anymore. However, although there is no deadlock, the behaviour of scope is still pretty erratic in the [ECS demo](http://www.weave.works/guides/service-discovery-and-load-balancing-with-weave-on-amazon-ecs-2/). Containers appear an disappear all the time.

PR #895, opened 2016-02-01
Details panel table header looks up label in all rows

### LINKED ISSUE #891: Details panel keeps Loading... when viewing host
1. Replay this report https://gist.github.com/davkal/86847a095a3e5ec2ec96
2. click on the only host
3. BOOM

Looks like the details' `columns` include `docker_cpu_total_usage`, but the detail nodes dont have a `metrics` field. Should there have been one?

davkal:
Loop across all child nodes to find column labels in case first node does not have metrics (happens with stopped containers).

Fixes #891


foot:
Wicked.

LGTM.

PR #899, opened 2016-02-02
Increase timeout, remove double close.

### LINKED ISSUE #886: TestAppClientDetails is flaky
```
2016/01/29 21:36:15 Publish loop for http://127.0.0.1:47115 starting
2016/01/29 21:36:15 Publish loop for http://127.0.0.1:47115 exiting
--- FAIL: TestAppClientDetails (0.10s)
	app_client_internal_test.go:147: timeout
FAIL
FAIL	github.com/weaveworks/scope/probe/appclient	0.134s
```

tomwilkie:
Fix #886

### DIFF
--- a/probe/appclient/app_client_internal_test.go
+++ b/probe/appclient/app_client_internal_test.go
@@ -5,7 +5,6 @@ import (
 	"encoding/gob"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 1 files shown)


probe/appclient/app_client_internal_test.go
```diff
@@ -143,7 +143,7 @@ func TestAppClientDetails(t *testing.T) {
 
 	select {
 	case <-done:
-	case <-time.After(100 * time.Millisecond):
+	case <-time.After(200 * time.Millisecond):
```
paulbellamy: crank it up. Why not a full second?

paulbellamy: Also, why not the one above in `TestAppClientPublishInternal`?

paulbellamy:
A better solution would seem to be to make `AppClient.Publish` block until the publish has finished/failed, then you don't need polling/timeouts in the tests.

tomwilkie:
> AppClient.Publish block

That would be problematic from the POV of the code - publish is supposed to be best effort, and we don't want it to block the caller.  For instance, when publishing to two apps, do you wait for both to ACK?  If so you have the ability for a slow app to starve a faster one, for instance.  In essence it would make MultiClient's job harder.

Anyway I misdiagnosed this, it was the first few reports getting dropped.  I've increased the number and added a small sleep to give go time to spin up the right goroutines.

paulbellamy:
Then do `go AppClient.Publish()` where you don't care about the response. Not that much harder.

PR #904, opened 2016-02-02
move docker labels into their own table of the details panel

### LINKED ISSUE #802: Docker Labels should be a separate table in the details panel for containers
Raised in #752

paulbellamy:
Fixes #802

review @tomwilkie, and @davkal, please

### DIFF
--- a/client/app/scripts/components/node-details.js
+++ b/client/app/scripts/components/node-details.js
@@ -129,7 +129,7 @@ export default class NodeDetails extends React.Component {
 
   renderDetails() {
     const details = this.props.details;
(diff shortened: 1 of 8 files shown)


tomwilkie:
This seems to work, so LGTM from me.

Shame we have to hard code it in the UI/API like that.

tomwilkie:
```
./render/detailed/docker_labels.go:10:1: exported function NodeDockerLabels should have comment or be unexported
```

paulbellamy:
The flipside of "hardcoding" it is that the UI understands some of the semantics of what it's rendering, which is good.

PR #905, opened 2016-02-02
Minimize race between proc's net/tcp{,6} and fd/*

2opremio:


probe/endpoint/procspy/proc_linux.go
```diff
@@ -111,22 +118,70 @@ func walkProcPid(buf *bytes.Buffer, walker process.Walker) (map[uint64]*Proc, er
...
+		if err := fs.Stat(netNamespacePath, &statT); err != nil {
+			return
+		}
+
+		procs, ok := namespaces[statT.Ino]
+		if ok {
+			namespaces[statT.Ino] = append(procs, &p)
+		} else {
+			namespaces[statT.Ino] = []*process.Process{&p}
 		}
```
tomwilkie: You don't need the else, as the nil value for procs is a valid empty list.

tomwilkie: ie just do

``` 
procs := namespaces[statT.Ino]
namespaces[statT.Ino] = append(procs, &p)
```

probe/endpoint/procspy/proc_linux.go
```diff
@@ -68,42 +68,49 @@ func getNetNamespacePathSuffix() string {
...
-			read, _ := readFile(filepath.Join(procRoot, dirName, "/net/tcp"), buf)
-			read6, _ := readFile(filepath.Join(procRoot, dirName, "/net/tcp6"), buf)
-			hasConns = read+read6 > 0
-			namespaces[statT.Ino] = hasConns
+		read, errRead := readFile(filepath.Join(procRoot, dirName, "/net/tcp"), buf)
+		read6, errRead6 := readFile(filepath.Join(procRoot, dirName, "/net/tcp6"), buf)
+
+		if errRead != nil || errRead6 != nil {
+			// try next process
+			continue
```
tomwilkie: I don't think you need to try to read every process' `net/tcp` - if it fails I'd just bail out the entire func.

tomwilkie:
A couple of minor comments, and the tests are failing, but otherwise LGTM.

Had this had a noticeable effect?

probe/endpoint/procspy/proc_linux.go
2opremio: Thanks, will do.

probe/endpoint/procspy/proc_linux.go
2opremio: I don't think that's robust enough, particularly compared to how it was done before this PR, which implicitly tried all processes. Only trying on exited process would mean missing a whole namespace.

2opremio:
I haven't seen an obvious impact, but being a clear race it should in situations in which the file descriptors change fast. I still can't believe the kernel doesn't provide the PIDs associated to existing connections in an atomic way.

probe/endpoint/procspy/proc_linux.go
tomwilkie: > Only trying on a single exited process would mean missing a whole namespace.

Thats a good point.  I was more worried that something stopping us reading the files would causes us to try on every process, but yeah, what you said makes more sense.

PR #907, opened 2016-02-03
Periodically check for newer versions of scope.

tomwilkie:
Every 2 hours, hit https://checkpoint-api.weave.works looking for current versions of Scope.  If a new version is found, log it.

Reuses the same library as Hashicorp's products.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ the Internet.  Additionally traffic between the app and the probe is currently
 insecure and should not traverse the internet.
 
 Scope will periodically check with our servers to see if a new version is
-availible.  To disable this, run:
+available.  To disable this, run:
 
 ```
 CHECKPOINT_DISABLE=true scope launch
--- a/README.md
+++ b/README.md
@@ -56,6 +56,13 @@ such, the Scope app endpoint (port 4040) should not be made accessible on
 the Internet.  Additionally traffic between the app and the probe is currently
 insecure and should not traverse the internet.
 
+Scope will periodically check with our servers to see if a new version is
+availible.  To disable this, run:
+
+```
+CHECKPOINT_DISABLE=true scope launch
+```
+
 ## <a name="architecture"></a>Architecture
 
 Weave Scope consists of two components: the app and the probe. These two
--- a/prog/app.go
+++ b/prog/app.go
@@ -44,14 +44,13 @@ func appMain() {
 	defer log.Print("app exiting")
 
 	// Start background version checking
-	params := checkpoint.CheckParams{
+	checkpoint.CheckInterval(&checkpoint.CheckParams{
 		Product:       "scope-app",
 		Version:       app.Version,
 		SignatureFile: signatureFile,
-	}
-	checkpoint.CheckInterval(&params, versionCheckPeriod, func(r *checkpoint.CheckResponse, err error) {
+	}, versionCheckPeriod, func(r *checkpoint.CheckResponse, err error) {
 		if r.Outdated {
-			log.Printf("Scope version %s is availible; please update at %s",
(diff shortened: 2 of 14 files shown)


prog/probe.go
```diff
@@ -30,6 +31,31 @@ import (
...
+const (
+	signatureFile = "/etc/weave/signature"
+)
+
+func check() {
+	handleResponse := func(r *checkpoint.CheckResponse, err error) {
+		if err != nil {
+			log.Printf("Error checking version: %v", err)
+		} else if r.Outdated {
+			log.Printf("Scope version %s is availible; please update at %s",
```
paulbellamy: available

paulbellamy:
Logging it seems fine, but users are unlikely to see that. Would be nicer to show it when launching.

tomwilkie:
I've made a separate ticket for displaying this in the UI: https://github.com/weaveworks/scope/issues/914

I'd prefer not to show it on startup, due to the way our startup works (starting a container etc).

paulbellamy:
Also, we should document how the user can disable this.

tomwilkie:
Updated to once every 6hrs on @pidster's request.

README.md
```diff
@@ -56,6 +56,13 @@ such, the Scope app endpoint (port 4040) should not be made accessible on
 the Internet.  Additionally traffic between the app and the probe is currently
 insecure and should not traverse the internet.
 
+Scope will periodically check with our servers to see if a new version is
+availible.  To disable this, run:
```
paulbellamy: available

prog/app.go
```diff
@@ -42,6 +43,19 @@ func appMain() {
 
 	defer log.Print("app exiting")
 
+	// Start background version checking
+	params := checkpoint.CheckParams{
+		Product:       "scope-app",
+		Version:       app.Version,
+		SignatureFile: signatureFile,
+	}
+	checkpoint.CheckInterval(&params, 2*60*time.Minute, func(r *checkpoint.CheckResponse, err error) {
```
paulbellamy: `6*time.Hour`, surely

prog/app.go
```diff
@@ -42,6 +43,19 @@ func appMain() {
...
+		if r.Outdated {
+			log.Printf("Scope version %s is availible; please update at %s",
```
paulbellamy: available

scope
```diff
@@ -92,6 +92,8 @@ check_not_running() {
 launch_command() {
     echo docker run --privileged -d --name=$SCOPE_CONTAINER_NAME --net=host --pid=host \
             -v /var/run/docker.sock:/var/run/docker.sock \
+            -v /etc/weave:/etc/weave \
+            -e CHECKPOINT_DISABLE \
```
paulbellamy: Probably shouldn't create `/etc/weave` if CHECKPOINT_DISABLE is set...

scope
```diff
@@ -90,8 +90,14 @@ check_not_running() {
 }
 
 launch_command() {
+    local args="-v /etc/weave:/etc/weave"
```
2opremio: Why is this needed? It makes the container stateful and breaks reproducibility.  

For instance, I spent 30 mins going in circles, because it makes the app targets get appended between scope launches. See https://github.com/weaveworks/scope/blob/master/docker/entrypoint.sh#L168

```
echo "$MANUAL_APPS" >>/etc/weave/apps
```

That should had probably just had been `"$MANUAL_APPS" >/etc/weave/apps` from the beginning but it points out why it's a bad idea to use local volumes.

tomwilkie: Sorry about that!  We put a signature file in there, and mount it into the container to ensure the same signature is sent to the checkpoint server between runs of the scope container.

PR #911, opened 2016-02-03
Just show the untruncated label in the tooltip on children

### LINKED ISSUE #908: Tool tip of container image needs refining
crazy-tooltip-1

There's too much information - it should only show the untruncated container image name.

paulbellamy:
Fixes #908


paulbellamy:
@davkal, ptal

davkal:
Works as advertised, but we lose useful data on hover, like theadcount, parentPID, full command, etc.
So maybe we should be more careful about the metadata that is added to a child node, @paulbellamy ?

paulbellamy:
I feel like if you want that information you should click on the link to view the full node details.

davkal:
Fair enough. LGTM

PR #912, opened 2016-02-04
Rate-limit reading proc files

2opremio:
Second attempt at #814 after seeing https://github.com/weaveworks/scope/issues/812#issuecomment-179599405
Improves #812


2opremio:
@peterbourgon Can you PTAL? I will rebase before merging.

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,133 @@
...
+	return nil
+}
+
+func (br *backgroundReader) stop() error {
+	br.mtx.Lock()
+	defer br.mtx.Unlock()
+	if !br.running {
+		return fmt.Errorf("background reader already not running")
+	}
+	br.pleaseStop = true
```
peterbourgon: Hm, this probably isn't good enough. You will end up waiting for `(targetWalkTime - walkTime) + totalOperationDuration` before this is received. You probably want a signal chan + a `select` on each blocking operation in `loop`.

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,133 @@
...
+		time.Sleep(targetWalkTime - walkTime)
+	}
+}
+
+func (br *backgroundReader) getWalkedProcPid(buf *bytes.Buffer) map[uint64]*Proc {
+	br.mtx.Lock()
+	defer br.mtx.Unlock()
+
+	reader := bytes.NewReader(br.readyBuf.Bytes())
+	buf.ReadFrom(reader)
```
peterbourgon: `io.Copy(buf, br.readyBuf)`?

probe/endpoint/procspy/spy_darwin.go
```diff
@@ -13,10 +13,17 @@ const (
 	lsofBinary    = "lsof"
 )
 
+// NewConnectionScanner creates a new Darwin ConnectionScanner
+func NewConnectionScanner(_ process.Walker) ConnectionScanner {
+	return &darwinScanner{}
+}
+
+type darwinScanner struct{}
```
peterbourgon: I like this a _lot_ better than the previous cbConnections implementation. :+1:

peterbourgon:
I still think @paulbellamy should take a look for overall cohesion, but I like this approach.

probe/endpoint/procspy/background_reader_linux.go
2opremio: Thanks.

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,133 @@
...
+	br.running = true
+	go br.loop()
+	return nil
+}
+
+func (br *backgroundReader) stop() error {
+	br.mtx.Lock()
+	defer br.mtx.Unlock()
+	if !br.running {
+		return fmt.Errorf("background reader already not running")
```
paulbellamy: why fail?

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,133 @@
...
+	return br
+}
+
+// starts a rate-limited background goroutine to read the expensive files from
+// proc.
+func (br *backgroundReader) start() error {
+	br.mtx.Lock()
+	defer br.mtx.Unlock()
+	if br.running {
+		return fmt.Errorf("background reader already running")
```
paulbellamy: why fail?

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,133 @@
+package procspy
```
paulbellamy: This seems like it should be fairly consistent between linux and darwin, no?

2opremio:
Rebased to get client test fixes.

probe/endpoint/procspy/background_reader_linux.go
paulbellamy: :+1: would be nice for stop to block until it's actually stopped.

probe/endpoint/procspy/background_reader_linux.go
2opremio: Darwin doesn't have namespaces nor does it read `/proc`, which is the reason for doing the rate-limiting.

probe/endpoint/procspy/background_reader_linux.go
2opremio: Uhm, no strong opinions about this, but it stopping something twice feels like double-closing (which is not generally _allowed_ in golang)

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,133 @@
...
+			log.Warnf(
+				"background /proc reader: full pass took %s: 50%% more than expected (%s)",
+				walkTime,
+				targetWalkTime,
+			)
+		}
+
+		// Adjust rate limit to more-accurately meet the target walk time in next iteration
+		scaledRateLimitPeriod := targetWalkTimeF / walkTimeF * float64(rateLimitPeriod)
+		rateLimitPeriod = time.Duration(math.Min(scaledRateLimitPeriod, maxRateLimitPeriodF))
```
paulbellamy: no `minRateLimitPeriod`?

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		// Swap buffers
+		br.readyBuf, br.walkingBuf = br.walkingBuf, br.readyBuf
+		br.readySockets = sockets
+
+		br.mtx.Unlock()
+
+		walkTime := time.Now().Sub(start)
+		walkTimeF := float64(walkTime)
```
paulbellamy: Oof. This naming is a bit, `time_t` from C.

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,133 @@
...
+		rateLimitPeriod = time.Duration(math.Min(scaledRateLimitPeriod, maxRateLimitPeriodF))
+		log.Debugf("background /proc reader: new rate limit %s", rateLimitPeriod)
+
+		ticker.Stop()
+		ticker = time.NewTicker(rateLimitPeriod)
+
+		br.walkingBuf.Reset()
+
+		// Sleep during spare time
+		time.Sleep(targetWalkTime - walkTime)
```
paulbellamy: as peter said above, better to do a loop and select

probe/endpoint/procspy/background_reader_linux.go
paulbellamy: which means this func could take `getWalkedProcPid(w io.Writer)`.

probe/endpoint/procspy/background_reader_linux.go
2opremio: Intentionally not. If we don't meet the 10s target (to meet the 15s window of the app) then the UI will be missing information. That's why set a maximum rate limit and not a minimum.

probe/endpoint/procspy/background_reader_linux.go
2opremio: Any alternatives?

probe/endpoint/procspy/fixture.go
```diff
@@ -21,10 +13,15 @@ func (f *fixedConnIter) Next() *Connection {
...
+type FixedScanner []Connection
+
+// Connections implements ConnectionsScanner.Connections
+func (s FixedScanner) Connections(_ bool) (ConnIter, error) {
+	iter := fixedConnIter(s)
+	return &iter, nil
 }
+
+// Stop implements ConnectionsScanner.Stop (dummy since there is no background work)
+func (s FixedScanner) Stop() {}
```
paulbellamy: If this is only used in tests let's define it in a test package. Looks like it's only used in `probe/endpoint/reporter_test.go`.

paulbellamy:
Overall, it's a substantial improvement. I wonder if it could be improved with [contexts](https://blog.golang.org/context)?

probe/endpoint/procspy/background_reader_linux.go
paulbellamy: Just type-cast it where you need it as a float?

probe/endpoint/procspy/background_reader_linux.go
paulbellamy: Ok. It also means that if the system is busy and we start missing deadlines scope will work even harder (using more cpu).

probe/endpoint/procspy/background_reader_linux.go
paulbellamy: right.

probe/endpoint/procspy/background_reader_linux.go
2opremio: Uhm ... If that gets to be case, I am not sure a minimum rate-limit would help much and I wouldn't even know which one to choose.

probe/endpoint/procspy/background_reader_linux.go
2opremio: That's what I did originally, but it made the expressions more difficult to read. However, I see your point, will do.

paulbellamy: Also, do float64s have enough precision to reliably represent time durations?

2opremio:
@peterbourgon @paulbellamy please take a thorough look, I am still a bit of a newbie with Golang channels.

probe/endpoint/procspy/background_reader_linux.go
2opremio: If you meant range, they definitely have enough, considering that `Duration` is represented in ns and there is an expected walking time of tenths of seconds (~1e10 ns). 

In terms of precision, I am not too worried of the rate-limit adjustment is a bit rough.

probe/endpoint/procspy/proc_linux.go
```diff
@@ -68,51 +86,80 @@ func getNetNamespacePathSuffix() string {
...
+	var fdBlockCount uint64
+	for i, p := range namespaceProcs {
+
+		// Get the sockets for all the processes in the namespace
 		dirName := strconv.Itoa(p.PID)
 		fdBase := filepath.Join(procRoot, dirName, "fd")
 
+		if fdBlockCount > w.fdBlockSize {
+			// we surpassed the filedescriptor rate limit
+			// TODO: worth selecting on w.stopc?
```
2opremio: @peterbourgon Maybe it's overkill to select on `stopc` here too (for a higher granularity). But, either way, I am curious about how it could be done without conflicting with the `stopc` select block on `(w pidWalker) walk()`. Any ideas?

probe/endpoint/procspy/fixture.go
2opremio: It's used by `probe/endpoint/procspy/spy_darwin.go` which ... I only learned after refactoring it into a separate package ... hahaha

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return br.latestSockets, err
+}
+
+func (br *backgroundReader) loop(walker process.Walker) {
+	var (
+		begin           time.Time                      // when we started the last performWalk
+		tickc           = time.After(time.Millisecond) // fire immediately
+		walkc           chan map[uint64]*Proc          // initially nil, i.e. off
+		walkBuf         = bytes.NewBuffer(make([]byte, 0, 5000))
```
peterbourgon: I think there is no need to preallocate a buffer here. Whatever allocation costs exist will be amortized very quickly.

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,130 @@
...
+			pWalker.stop()
+			ticker.Stop()
+			return // abort
+		}
+	}
+}
+
+// Adjust rate limit for next walk and calculate how long to wait until it should be started
+func scheduleNextWalk(rateLimitPeriod time.Duration, took time.Duration) (time.Duration, time.Duration) {
+
```
peterbourgon: Extra newline

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,130 @@
...
+		case <-br.stopc:
+			pWalker.stop()
+			ticker.Stop()
+			return // abort
+		}
+	}
+}
+
+// Adjust rate limit for next walk and calculate how long to wait until it should be started
+func scheduleNextWalk(rateLimitPeriod time.Duration, took time.Duration) (time.Duration, time.Duration) {
```
peterbourgon: This is a good case for naming your return variables. What are each of these?

probe/endpoint/procspy/proc_linux.go
peterbourgon: I'm not sure what you mean by "conflicting", but if you `close(c)` then it sends a signal to _all_ goroutines receiving on `c`. And, any subsequent read from the same channel will return immediately. So a single sender (the closer) can effectively broadcast to all receivers (both pidWalker.walk and pidWalker.walkNamespace) including into the future. Does that give you want you need?

probe/endpoint/procspy/proc_linux.go
```diff
@@ -23,6 +24,23 @@ var (
 	netNamespacePathSuffix = ""
 )
 
+type pidWalker struct {
+	walker      process.Walker
+	ticker      <-chan time.Time
```
peterbourgon: I would call this `tickc` just so there's no confusion between the time.Ticker and its underlying chan.

peterbourgon: Also, it would be good to have a comment somewhere indicating this is for rate limiting of individual namespaces.

probe/endpoint/procspy/proc_linux.go
2opremio: Yes, thanks a lot. I need to read up on channels.

probe/endpoint/procspy/background_reader_linux.go
2opremio: After thinking about it a bit more, I will set the minimum rate limit to the initial rate limit which is pretty low.

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,144 @@
...
+		newRateLimitPeriod = maxRateLimitPeriod
+	} else if newRateLimitPeriod < minRateLimitPeriod {
+		newRateLimitPeriod = minRateLimitPeriod
+	}
+	log.Debugf("background /proc reader: new rate limit period %s", newRateLimitPeriod)
+
+	// when to start next walk
+	restInterval = targetWalkTime - took
+
+	return
```
peterbourgon: This is a so-called naked return, which is best avoided.

> Naked returns are okay if the function is a handful of lines. Once it's a medium-sized function, be explicit with your return values.
—https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,144 @@
...
+	fdBlockSize            = uint64(300) // Maximum number of /proc/PID/fd/* files to stat per rate-limit period
+	// (as a rule of thumb going through each block should be more expensive than reading /proc/PID/tcp{,6})
+	targetWalkTime = 10 * time.Second // Aim at walking all files in 10 seconds
+)
+
+type backgroundReader struct {
+	stopc         chan struct{}
+	mtx           sync.Mutex
+	latestBuf     *bytes.Buffer
+	latestSockets map[uint64]*Proc
```
peterbourgon: Once a `*Proc` is written to this map, is it ever modified?

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// starts a rate-limited background goroutine to read the expensive files from
+// proc.
+func newBackgroundReader(walker process.Walker) *backgroundReader {
+	br := &backgroundReader{
+		latestBuf: bytes.NewBuffer(make([]byte, 0, 5000)),
```
peterbourgon: Sorry I didn't see this before. As in loop, I think there is no reason to preallocate here.

peterbourgon: Also, you will probably want to allocate an empty `latestSockets: map[uint64]*Proc{}` here, so that someone who calls e.g. `getWalkedProcPid` before the first run doesn't get surprised by a nil map.

peterbourgon:
We could do another iteration over the `loop` body and clean it up a good bit, but I won't frustrate you any more :) LGTM after those changes.

probe/endpoint/procspy/background_reader_linux.go
2opremio: I don't think it is, but that's the format it had, (and what's expected in the output) and I don't know if it will have performance implications. I would like to avoid refactoring `procspy` more than I have in this PR.

2opremio:
@peterbourgon Please go ahead and comment, I would like to avoid refactoring other parts of the code not directly related to this change though.

probe/endpoint/procspy/background_reader_linux.go
2opremio: Re:edit. Then we are good :)

2opremio:
I should had rebased, oh well.

PR #916, opened 2016-02-04
Improve codec performance

2opremio:
Improves #854 

- [x] Fix tests
- [x] Generate encoders/decoders on the fly (or at least incorporate a make target for it)
- [x] Fix hack from (94cee66) by updating all the vendored k8s dependencies (sigh)
- [x] Rebase


2opremio:
As mentioned in https://github.com/weaveworks/scope/issues/854#issuecomment-179953727 using ffjson without code generation helps quite a bit but doesn't cut it. 

I've tried the ffjson code generator. However, the encoder generated from https://github.com/weaveworks/scope/blob/master/report/node.go breaks the UI, that is:

```
go get -u github.com/pquerna/ffjson
cd $GOPATH/github.com/weaveworks/scope/report
ffjson node.go
```

Which results in:

screen shot 2016-02-09 at 5 51 29 pm

I really don't want to debug the generated code, so I am going to try https://github.com/ugorji/go/ next

2opremio:
Using a generated JSON codec with github.com/ugorji/go/codec doesn't seem to solve the problem. Here are the profiles from the scenario described in https://github.com/weaveworks/scope/issues/854#issuecomment-179560048 https://github.com/weaveworks/scope/issues/812

Here's the app's profile:

[pprof.localhost:4040.samples.cpu.009.pb.gz](https://github.com/weaveworks/scope/files/125362/pprof.localhost.4040.samples.cpu.009.pb.gz)

app_codec

And here's the probe's profile:

[pprof.localhost:4041.samples.cpu.003.pb.gz](https://github.com/weaveworks/scope/files/125365/pprof.localhost.4041.samples.cpu.003.pb.gz)


probe_codec

As much as we would like to use json, I think it's time to move on to a faster codec. We can use part of this work for the communication with the UI (which the profile above is not taking into account).

2opremio:
I think I am going to try gob+websockets next.

2opremio:
The websocket+gob approach didn't help either (see #957 )

I replaced all the `{Marhsal,Unmarhsal}JSON` methods by [Selfer](https://godoc.org/github.com/ugorji/go/codec#Selfer)s and it helps a lot with the garbage collection because we don't need to create more intermediate buffers. Note how gzip now takes 30% of the decoding time.



App:

[pprof.localhost:4040.samples.cpu.010.pb.gz](https://github.com/weaveworks/scope/files/127065/pprof.localhost.4040.samples.cpu.010.pb.gz)

app


Probe:

[pprof.localhost:4041.samples.cpu.004.pb.gz](https://github.com/weaveworks/scope/files/127066/pprof.localhost.4041.samples.cpu.004.pb.gz)

probe

2opremio:
After moving to msgpack and updating the `ugorji/go/codec` library I get this

App: 

app

Probe:

probe


Although the profile looks the same, note how the duration has gone down considerably. 

Note that this is still measured without connecting the UI, but it brings down CPU usage of the app in the service (5 probes) to <10% compared to ~70% in master .

The next steps would be:

* Reduce the size/publish-frequency of the reports
* Tweak the gzip compression level.
* Increase GOGC? (sounds like cheating, but may be worth trading of some memory for GC CPU)
* Look into building `github.com/ugorji/go/codec/` with `-tags=unsafe` to reduce decoding allocations. I tried, but it basically broke everything. I am going to create a ticket upstream.
* Create a custom encoder/decoder (I wouldn't be up for this, but @bboreham has a lot of experience and may want to volunteer :) ).

@paulbellamy @peterbourgon Can you please take a look at this and give some feedback? If you are OK with the results/approach, I will go ahead and clean it up:

- [ ] Fix tests
- [ ] Generate encoder/decoder on the fly (or at least incorporate a make targe for it)
- [ ] Rebase

app/router.go
```diff
@@ -133,8 +133,15 @@ func RegisterReportPostHandler(a Adder, router *mux.Router) {
 
 		decoder := gob.NewDecoder(reader).Decode
 		if strings.HasPrefix(r.Header.Get("Content-Type"), "application/json") {
-			decoder = json.NewDecoder(reader).Decode
+			decoder = func(i interface{}) error {
+				return codec.NewDecoder(reader, &codec.JsonHandle{}).Decode(i)
+			}
```
paulbellamy: No need for the lambda here (and below):
`decoder = codec.NewDecoder(reader, &codec.JsonHandle{}).Decode`

app/router.go
```diff
@@ -133,8 +133,15 @@ func RegisterReportPostHandler(a Adder, router *mux.Router) {
 
 		decoder := gob.NewDecoder(reader).Decode
```
paulbellamy: why leave gob around?

probe/probe_internal_test.go
```diff
@@ -72,11 +72,27 @@ func (m mockPublisher) Stop() {
 }
 
 func TestProbe(t *testing.T) {
+	// Json marshalling->unmarshaling is not idempotent due to
+	// `json:"omitempty"` annotations So, we make DeepEqual happy by setting
+	// empty `json:"omitempty"` entries to nil
```
paulbellamy: eh?

report/controls.go
```diff
@@ -84,25 +83,20 @@ type WireNodeControls struct {
...
+	encoder.Encode(WireNodeControls{
 		Timestamp: renderTime(nc.Timestamp),
 		Controls:  nc.Controls,
 	})
-	return buf.Bytes(), err
 }
 
-// UnmarshalJSON implements json.Unmarshaler
-func (nc *NodeControls) UnmarshalJSON(input []byte) error {
+func (nc *NodeControls) CodecDecodeSelf(decoder *codec.Decoder) {
```
paulbellamy: With `MarshalJSON` and `UnmarshalJSON` removed, using `json.Decode(b, &nodeControls)` is subtly broken. Maybe we should:
```Go
func (nc *NodeControls) MarshalJSON() ([]byte, error) {
  return nil, nil
}

File: app/router.go (hunk 2 above)
Comment (2opremio) [inline app/router.go]: Backwards compatibility with 0.12. I will create a techdebt ticket to remove it later on.

File: app/router.go (hunk 1 above)
Comment (2opremio) [inline app/router.go]: Yep, it came from the migration from ffjson

File: report/controls.go (hunk 4 above)
Comment (2opremio) [inline report/controls.go]: We shouldn't use `json.Decode()` anymore. Tests-aside, I thought I had removed all references of `encoding/json`. Where did you see that?

Comment (paulbellamy) [inline report/controls.go]: > We shouldn't use json.Decode() anymore.

Exactly. Maybe we should make sure.

> Where did you see that?

Tests-aside, I didn't.

**Comment by paulbellamy** [comment]
This seems to have trouble with children's metrics.

**Comment by 2opremio** [comment]
> This seems to have trouble with children's metrics.

Thanks, I will look into that but will fix the tests first to see if that catches it.

Here's a final CPU% comparison running Scope in the service (5 probes -> one app) between 26ccb4918b15d363c50fd206421dc21feffd2c17 (master is broken on the service because of #955).

| What       | (26ccb4918b15d363c50fd206421dc21feffd2c17)           | This PR  |
| ------------- |:-------------:|:-----:|
|  Probe(s)     | 15-25%      |   20% |
| App (no UI)      | 70% | 10% |
| App (with UI)  | -      |    30% with peaks of 50% |


For reference, I have also obtained the profile of the App running with the UI attached:

app

The codec is not the culprit (i.e. unrelated to this PR).

**Comment by paulbellamy** [comment]
> Thanks, I will look into that but will fix the tests first to see if that catches it.

I suspect it is with json encoding of metrics, so it won't...

**Comment by 2opremio** [comment]
@foot @davkal This PR is breaking the controls in the UI and I am lost on what's causing it. Would you mind taking a look?

The UI is making requests to URLs like `/api/control/undefined/undefined/docker_unpause_container`. However, I don't see any missing information in `api/report` (the node ids and probe ids are there) so I think that slight changes in the decoder output may be affecting the UI.

**Comment by 2opremio** [comment]
> I suspect it is with json encoding of metrics, so it won't...

@paulbellamy What problems were you seeing with the metrics? I have noticed that the new codec outputs floats instead of ints. e.g:

```
        "metrics": {
          "docker_memory_usage": {
            "samples": [
              {
                "date": "2016-02-16T05:24:00.72055402Z",
                "value": 36864.0
              },
```

instead of:

```
        "metrics": {
          "docker_memory_usage": {
            "samples": [
              {
                "date":  "2016-02-16T05:30:54.719541452Z",
                "value": 36864
              },
```

paulbellamy:
I was just never seeing any values rendered for them on children. Didn't look further into it than that.

foot:
Sorry i clicked on the checkboxes on the first comment, not really paying attention to what I was doing!

Hope they're in the right state.

2opremio:
OK, so this is why controls are not working: https://github.com/ugorji/go/issues/141 . 

The new codec library misbehaves when embedding types, which is exactly how [our control details are defined](https://github.com/weaveworks/scope/blob/89d7ebe977f507066b3fd46325f7d114b97da3ed/render/detailed/node.go#L26-L30), sigh.

This may also be what's causing problems with the metrics.

 I think I may have a workaround though: generate selfers for all the types we serialize.

@foot No worries

paulbellamy:
> This may also be what's causing problems with the metrics.

@2opremio yep.

2opremio:
OK, I found yet another bug, https://github.com/ugorji/go/issues/142 but the controls are working and I think the metrics are alright. 

@paulbellamy PTAL, I think I will open a bottle of champaign when this gets merged.

paulbellamy:
let's ditch `GOHOSTOS` and `GOHOSTARCH` and do `unset GOOS GOARCH foo` instead, so go can just auto-detect. Less to learn for the user of the makefile.

2opremio:
> 
Why are we using your own fork of the codec, vs. the upstream?

The fork is only used to build codecgen. It's a temporary solution, see the commit comment:

> I included
  https://github.com/2opremio/go-1/tree/master/codec/codecgen instead of
  upstream until ugorji/go#139 is merged.

In fact we will be able to use the upstream one once this is fixed: https://github.com/ugorji/go/pull/140#issuecomment-185119063 , which I guess will be pretty soon judging by the maintainers responsiveness.

2opremio:
> let's ditch GOHOSTOS and GOHOSTARCH and do unset GOOS GOARCH foo instead, so go can just auto-detect

Agreed

paulbellamy:
the secondary expansion thing is terribly complex and obtuse.

2opremio:
@paulbellamy pleaser take another look

paulbellamy:
Much simpler, thanks. :) LGTM

app/router.go
2opremio: https://github.com/weaveworks/scope/issues/971

PR #920, opened 2016-02-05
minor refactor of backend metadata and metric rendering

### LINKED ISSUE #835: Refactor render/detailed/metadata.go
From #752.

`ltst` is not a great name. One proposed solution is to refactor to structs (instead of funcs), so the "type" goes at the end.

paulbellamy:
fixes #835

### DIFF
--- a/render/detailed/metadata.go
+++ b/render/detailed/metadata.go
@@ -15,42 +15,86 @@ import (
 )
 
 var (
-	processNodeMetadata = renderMetadata(
-		ltst(process.PID),
-		ltst(process.PPID),
-		ltst(process.Cmdline),
-		ltst(process.Threads),
-	)
-	containerNodeMetadata = renderMetadata(
-		ltst(docker.ContainerID),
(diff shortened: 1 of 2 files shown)


tomwilkie:
I'm not sure I like the reuse of MetadataRow here.  I think it would be better to make an interface like this:

```go
type MetadataRowRenderer interface {
  Render(report.Node) []MetadataRow
}

type Latest struct {
  ID string
}

func (l Latest) Render(report.Node) []MetadataRow {
...
}

type Metric struct {
 ID string
}

func (m Metric) Render(report.Node) []MetadataRow {
...
}
...
```

You could then see these renderers being included in the report such that plugins can add stuff to the details panel.  Make sense?

Also, this would allow you to add a field to Latest called Length, and solve #930 in a nicer way.

tomwilkie:
Could call `MetadataRowRenderer` "RowTemplate"

paulbellamy:
Applied tom's feedback verbatim, so merging.

PR #941, opened 2016-02-08
Reorder the children tables in the details panel

### LINKED ISSUE #923: Change order of tables in details panel
Option 1. Go from processes at top, to containers then container images.
Option 2. Go for an arbitrary order.  Collect data as to what is most useful, implement that

paulbellamy:
The Container images was first on hosts, but is probably the least
useful, so this moves it to the bottom.

In general, it needs more thought put into the detail panel contents,
but this is a quick fix to re-prioritize the data.

Fixes #923

### DIFF
--- a/render/detailed/node.go
+++ b/render/detailed/node.go
@@ -85,9 +85,9 @@ var (
 	}{
 		{report.Host, NodeSummaryGroup{TopologyID: "hosts", Label: "Hosts", Columns: []string{host.CPUUsage, host.MemoryUsage}}},
 		{report.Pod, NodeSummaryGroup{TopologyID: "pods", Label: "Pods", Columns: []string{}}},
-		{report.ContainerImage, NodeSummaryGroup{TopologyID: "containers-by-image", Label: "Container Images", Columns: []string{render.ContainersKey}}},
 		{report.Container, NodeSummaryGroup{TopologyID: "containers", Label: "Containers", Columns: []string{docker.CPUTotalUsage, docker.MemoryUsage}}},
 		{report.Process, NodeSummaryGroup{TopologyID: "processes", Label: "Processes", Columns: []string{process.PID, process.CPUUsage, process.MemoryUsage}}},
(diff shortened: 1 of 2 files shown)


2opremio:
LGTM code-wise. However, I would argue that having the processes on top only makes sense for containers. For a host/image you probably want to have the containers first.

Also, maybe it would make sense to bring the labels to the bottom, although I don't have a strong opinion about this. @pidster ?

paulbellamy:
After discussion, @2opremio and I have agreed that we feel that keeping the order as is (but moving container images to the end) is the most intuitive to us, instead of reversing it.

davkal:
LGTM

render/detailed/node.go
```diff
@@ -85,9 +85,9 @@ var (
 	}{
 		{report.Host, NodeSummaryGroup{TopologyID: "hosts", Label: "Hosts", Columns: []string{host.CPUUsage, host.MemoryUsage}}},
 		{report.Pod, NodeSummaryGroup{TopologyID: "pods", Label: "Pods", Columns: []string{}}},
-		{report.ContainerImage, NodeSummaryGroup{TopologyID: "containers-by-image", Label: "Container Images", Columns: []string{render.ContainersKey}}},
 		{report.Container, NodeSummaryGroup{TopologyID: "containers", Label: "Containers", Columns: []string{docker.CPUTotalUsage, docker.MemoryUsage}}},
 		{report.Process, NodeSummaryGroup{TopologyID: "processes", Label: "Processes", Columns: []string{process.PID, process.CPUUsage, process.MemoryUsage}}},
+		{report.ContainerImage, NodeSummaryGroup{TopologyID: "containers-by-image", Label: "Container Images", Columns: []string{render.ContainersKey}}},
```
2opremio: Can you also move the docker labels to the end?

paulbellamy: @2opremio see https://github.com/weaveworks/scope/issues/945

Issue #932, opened 2016-02-05
Decouple reporting/tagging and publishing


2opremio:
Following up on https://github.com/weaveworks/scope/issues/881#issuecomment-177698657:

> we should have each one [tagger/publisher] on its own goroutine, trigger a report on the groutine, and then set a timeout for receiving results from them. 

This will allow us to be more deadlock-resilient and do some analysis/detection of reporters/taggers taking longer than expected.

2opremio:
Relevant code: https://github.com/weaveworks/scope/blob/c16f0d7f40091ea42071f79f915e1fe919e6c77c/probe/probe.go#L131-L164

PR #944, opened 2016-02-08
log a warning when reporters or taggers take too long to generate

paulbellamy:
Raised in #932 to make detecting stalled reporters easier.


2opremio:
LGTM

PR #946, opened 2016-02-08
Only show key metadata by default. Hide rest in a drawer.

### LINKED ISSUE #929: Add "Show More" for metadata in details panel
Pick top n things, put everything else behind a show more.

Container:
- ID
- State
- Command

Process:
- PID
- Command

Image:
- Image ID

Host:
- Kernel Version
- Uptime

paulbellamy:
Fixes #929

To make the interface cleaner, we can hide some non-key metadata by
default.

@davkal, this needs styling, js testing, and probably a js refactor.

### DIFF
--- a/client/app/scripts/components/node-details/node-details-info.js
+++ b/client/app/scripts/components/node-details/node-details-info.js
@@ -1,10 +1,30 @@
 import React from 'react';
 
 export default class NodeDetailsInfo extends React.Component {
+
+  constructor(props, context) {
+    super(props, context);
+    this.state = {
+      expanded: false
+    };
+    this.handleClickMore = this.handleClickMore.bind(this);
+  }
+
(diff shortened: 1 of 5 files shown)


davkal:
Added styling and refactored JS.

foot:
Code looks good!

Is there any other information that can potentially be displayed on a proc node? Otherwise its a bit annoying to click the `show more` button.

screen shot 2016-02-10 at 21 10 10

paulbellamy:
We could just show everything by default on the process node. They'll never have any children anyway...

davkal:
> We could just show everything by default on the process node. There's only 4 lines, and they'll never have any children anyway...

:+1:

davkal:
Otherwise, LGTM

PR #948, opened 2016-02-09
Fixes process-by-name layout with ./foo and /foo nodes

### LINKED ISSUE #947: Empty 'processes by name' view
My "processes by name" view is empty whilst I don't think it should.

screen shot 2016-02-09 at 10 11 39

Report: https://gist.github.com/2opremio/e6a9751b1bdc7e08e7e0

foot:
lodash (used by dagre) was in effect mixing up these ids.

fixes #947

### DIFF
--- a/client/app/scripts/charts/nodes-layout.js
+++ b/client/app/scripts/charts/nodes-layout.js
@@ -18,6 +18,14 @@ const RANK_SEPARATION_FACTOR = 2.5;
 let layoutRuns = 0;
 let layoutRunsTrivial = 0;
 
+function graphNodeId(id) {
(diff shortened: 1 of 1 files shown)


foot:
This is not a super fun PR to review.

It introduces another layer of node id munging to fix the `_.has` problem mentioned in the issue.

It makes this code more prone to bugs. Longterm fix would be to try and get something push back into dagre perhaps.

davkal:
LGTM

I guess `_.has` is the same logic tha enables your favorite `_.get` :smile:

foot:
Hahah, yes, I think lodash should drop the string version of path. Explicit lists for the win.

PR #950, opened 2016-02-10
Show nice column labels when no children have metrics

### LINKED ISSUE #939: Children with no metrics, leads to ugly column ids as column titles
For example, when none of the children have any metrics (if they're all stopped).
screen shot 2016-02-08 at 10 49 40

We should either, omit columns if no children have them, or include the label in the `columns` description, so we don't have to look for the label on the children.

@davkal, wdyt?

paulbellamy:
Fixes #939

`"columns":["a", "b"]` becomes `"columns":[{"id":"a", "label":"A"}, {"id":"b", "label":"B"}]`

### DIFF
--- a/client/app/scripts/components/node-details/node-details-table.js
+++ b/client/app/scripts/components/node-details/node-details-table.js
@@ -70,18 +70,7 @@ export default class NodeDetailsTable extends React.Component {
 
   renderHeaders() {
     if (this.props.nodes && this.props.nodes.length > 0) {
-      let headers = [{id: 'label', label: this.props.label}];
-      // gather header labels from metrics and metadata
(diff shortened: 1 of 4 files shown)


foot:
Code looks good! Awesome seeing that big chuck get deleted.

Tricky to repro tho, how does one get a host w/ no containers running? Won't the probe be running in a container to see that host? Or the probe is running straight on the host, and docker is also running on the host, w/ other containers stopped.

paulbellamy:
So this was showing in containers-by-image, where all containers were stopped.

foot:
Got it! Thanks!

foot:
LGTM

Issue #874, opened 2016-01-29
Improve contrast of scope UI


tomwilkie:
Given a lot of demos, and on most projectors and TV the scope UI is hard to see as the contrast isn't high enough.

foot:
Which parts of the UI suffer the most? Don't have a projector you could take a quick photo? :). Wonder if you can simulate that on a mac screen... maybe there is a app for that.

davkal:
Yeah, which parts @tomwilkie ? Contrast is a sensitive issue. If everyone screams nobody gets heard.

tomwilkie:
Its more of just a general concern.  I don't think the grey background is helping.  I'll try and get some photos.

foot:
A little bit related but the UI behaves itself quite well w/ browser zooming.

tomwilkie:
I did have a few centring issues when switch topologies with zooming.

PR #954, opened 2016-02-11
High contrast mode switch

davkal:
* added button to the bottom right to switch to higher contrast,
* suitable for projectors
* increased contrast between background and text colors
* thicker strokes around chart elements
* ensure yellow details panel backgrounds are darker

Addresses #874

### DIFF
--- a/client/app/scripts/components/app.js
+++ b/client/app/scripts/components/app.js
@@ -51,7 +51,7 @@ export default class App extends React.Component {
     AppStore.addListener(this.onChange);
     window.addEventListener('keyup', this.onKeyPress);
 
-    getRouter().start({hashbang: true});
+    getRouter(this.props.base).start({hashbang: true});
     if (!AppStore.isRouteSet()) {
       // dont request topologies when already done via router
       getTopologies(AppStore.getActiveTopologyOptions());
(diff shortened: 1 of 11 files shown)


davkal:
Normal mode:

screen shot 2016-02-10 at 16 09 13

High contrast mode:
screen shot 2016-02-10 at 16 10 07

davkal:
Ludacris contrast:

screen shot 2016-02-10 at 16 14 52

peterbourgon:
Nice! Since this is a super-rare feature mostly for us when we give demos, what do you think of swapping the toggle from `SWITCH CONTRAST` to something smaller, for example a half-moon icon?

foot:
Code looks good, had a bit of a poke into less but couldn't see a nice way to redefine vars more dynamically. There might be reload style hacks etc. It would be cool if we could just add a class to the app and have the contrast update.

- Switching from high-contrast mode back to normal mode requires clicking the change contrast button twice atm

davkal:
After review feedback:

screen shot 2016-02-11 at 10 41 45

foot:
Nice! LGTM.

I associate this icon more classically w/ contrast, but it doesn't have a nice inverse, personal pref. http://fortawesome.github.io/Font-Awesome/icon/adjust/

davkal:
Changed the icon. Thx @foot

PR #956, opened 2016-02-11
don't render any metrics/metadata for uncontained node

### LINKED ISSUE #951: Uncontained node renders metrics/metadata of a process
screen shot 2016-02-10 at 14 00 07

paulbellamy:
Fixes #951

Also added a bit more helpful logging to the tests.

@2opremio, ptal...

### DIFF
--- a/render/mapping.go
+++ b/render/mapping.go
@@ -441,8 +441,10 @@ func MapProcess2Container(n RenderableNode, _ report.Networks) RenderableNodes {
 		id = MakeContainerID(containerID)
 		node = NewDerivedNode(id, n)
 	} else {
+		nCopy := n.Copy()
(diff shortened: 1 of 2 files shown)


render/mapping.go
```diff
@@ -441,8 +441,10 @@ func MapProcess2Container(n RenderableNode, _ report.Networks) RenderableNodes {
 		id = MakeContainerID(containerID)
 		node = NewDerivedNode(id, n)
 	} else {
+		nCopy := n.Copy()
+		nCopy.Node = nCopy.Node.WithID("").WithTopology("") // Wipe the ID so it cannot be rendered.
```
2opremio: Does this mean we won't show any information about the process? Not even the process itself? Why?

paulbellamy: It removes the red section.
screen shot 2016-02-12 at 13 28 35

2opremio: The LGTM

PR #957, opened 2016-02-11
[WIP] Transmit reports using gob with websockets ...

2opremio:
... with the intention of caching the gob engines.

Very hacky for now, just a proof of concept.

As an alternative to improve #854

### DIFF
--- a/app/router.go
+++ b/app/router.go
@@ -1,9 +1,7 @@
 package app
 
 import (
-	"compress/gzip"
 	"encoding/gob"
-	"encoding/json"
 	"net/http"
 	"net/url"
 	"strings"
@@ -11,6 +9,7 @@ import (
 	"github.com/PuerkitoBio/ghost/handlers"
 	"github.com/gorilla/mux"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/weaveworks/scope/common/hostname"
 	"github.com/weaveworks/scope/common/xfer"
 	"github.com/weaveworks/scope/report"
@@ -82,7 +81,6 @@ func TopologyHandler(c Reporter, preRoutes *mux.Router, postRoutes http.Handler)
 		gzipHandler(topologyRegistry.captureRenderer(c, handleTopology)))
 	get.HandleFunc("/api/topology/{topology}/ws",
 		topologyRegistry.captureRenderer(c, handleWs)) // NB not gzip!
-	get.HandleFunc("/api/report", gzipHandler(makeRawReportHandler(c)))
 
 	// We pull in the http.DefaultServeMux to get the pprof routes
 	preRoutes.PathPrefix("/debug/pprof").Handler(http.DefaultServeMux)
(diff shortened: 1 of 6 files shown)


2opremio:
Unfortunately this doesn't seem to help at all. When running Scope locally (with a single app and probe), it takes double as much CPU as the generated JSON decoders ( #916  ) and I get this app CPU profile:

[pprof.172.16.0.3:4040.samples.cpu.011.pb.gz](https://github.com/weaveworks/scope/files/126628/pprof.172.16.0.3.4040.samples.cpu.011.pb.gz)

app

Surprisingly, even if a single  `encoding/gob.Decoder` is used, [`compileDec()`](https://github.com/golang/go/blob/a48de745b28a39d58dcd916f410704f0f7d75d7c/src/encoding/gob/decode.go#L1080) (which createsthe gob engines) executes 40%. This isn't expected since the engines should be cached within the Decoder.

2opremio:
OK, it seems that some engines are not being cached. After adding some logging for the cache misses of [getDecEnginePtr()](https://github.com/golang/go/blob/a48de745b28a39d58dcd916f410704f0f7d75d7c/src/encoding/gob/decode.go#L1139):

```
 fmt.Printf("Decoder cache miss for type %s (remote id %d)\n", rt.Name(), remoteId)
```

I see:

```
$ docker logs  weavescope |& grep Decoder | sort | uniq -c
 182943 Decoder cache miss for type arrayType (remote id 17)
 182943 Decoder cache miss for type CommonType (remote id 18)
      1 Decoder cache miss for type Control (remote id 79)
  29517 Decoder cache miss for type EdgeMetadata (remote id 84)
 182943 Decoder cache miss for type fieldType (remote id 21)
 182943 Decoder cache miss for type gobEncoderType (remote id 24)
  29517 Decoder cache miss for type LatestEntry (remote id 86)
 182943 Decoder cache miss for type mapType (remote id 23)
      1 Decoder cache miss for type NodeControls (remote id 72)
      1 Decoder cache miss for type Node (remote id 67)
 182943 Decoder cache miss for type  (remote id 16)
      1 Decoder cache miss for type  (remote id 65)
  29517 Decoder cache miss for type  (remote id 82)
  59034 Decoder cache miss for type  (remote id 83)
  29517 Decoder cache miss for type  (remote id 85)
  29517 Decoder cache miss for type  (remote id 87)
  35357 Decoder cache miss for type  (remote id 88)
  35357 Decoder cache miss for type Sample (remote id 89)
      1 Decoder cache miss for type Sampling (remote id 81)
 182943 Decoder cache miss for type sliceType (remote id 19)
 182943 Decoder cache miss for type structType (remote id 20)
      1 Decoder cache miss for type Topology (remote id 66)
```

2opremio:
OK, this is because `counters.go`, `latest_map.go` etc ... all provide `GobDecode` which in turn creates buffers and other decoders. Let's see if I can thread the top-level decoder in.

2opremio:
Now that I think about it, this is also the case with JSON in #916

2opremio:
There isn't a  way to pass in the same encoder. Using `encoding/gob.Register` to avoid intermediate representations for `ps.Map` (11d517a) didn't fly either. I ended up getting:

```
runtime: goroutine stack exceeds 1000000000-byte limit
```

In a nutshell, gob not only sucks for request-response communication (which admittedly wasn't designed for) but also when you want to serialize intermediate representations of your types/interfaces.

2opremio:
Closing in favor of #916

2opremio:
PS: it seems like the Golang maintainers don't have any plans to improve the interface to allow more efficient serialization of custom intermediate types. https://groups.google.com/forum/#!topic/golang-nuts/BQsVtMiu0W4

PR #959, opened 2016-02-12
Overwrite /etc/weave/apps, because it might already exist

### LINKED ISSUE #937: Fix problems caused by host-mounting of `/etc/weave`
See https://github.com/weaveworks/scope/pull/907/files#r52038080

paulbellamy:
Fixes #937 

@2opremio Any other bugs caused by this?

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -5,7 +5,7 @@ usage() {
 	exit 1
 }
 
-mkdir -p /etc/weave
+mkdir -p /var/run/weave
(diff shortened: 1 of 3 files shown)


2opremio:
Not that I am aware of, but I think a more robust solution would be to isolate the parts for which we want persistence from the app configuration. i.e. a different mount-point.

paulbellamy:
Good call, how about this?

2opremio:
LGTM

PR #960, opened 2016-02-12
Don't show blank IPs metadata row for containers with no IP

paulbellamy:
Previously we were adding the blank default IP, and showing that (an empty row), but we should skip the row if there are no IPs.

@2opremio ptal

### DIFF
--- a/render/topologies.go
+++ b/render/topologies.go
@@ -137,35 +137,28 @@ func (r containerWithHostIPsRenderer) Render(rpt report.Report) RenderableNodes 
 			continue
 		}
 
-		ips, ok := c.Node.Sets.Lookup(docker.ContainerIPs)
-		if ok && len(ips) > 0 {
-			continue
-		}
-
 		h, ok := hosts[MakeHostID(report.ExtractHostID(c.Node))]
 		if !ok {
 			continue
 		}
 
-		hostNetworks, ok := h.Sets.Lookup(host.LocalNetworks)
-		if !ok {
-			continue
-		}
-
-		hostIPs := report.MakeStringSet()
+		newIPs := report.MakeStringSet()
+		hostNetworks, _ := h.Sets.Lookup(host.LocalNetworks)
 		for _, cidr := range hostNetworks {
 			if ip, _, err := net.ParseCIDR(cidr); err == nil {
-				hostIPs = hostIPs.Add(ip.String())
+				newIPs = newIPs.Add(ip.String())
 			}
 		}
 
-		c.Sets = c.Sets.Add(docker.ContainerIPs, hostIPs)
+		c.Sets = c.Sets.Add(docker.ContainerIPs, newIPs)
 		containers[id] = c
 	}
 
 	return containers
 }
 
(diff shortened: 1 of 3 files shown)


2opremio:
When does this happen? If it's due to `-net containerID` (or even `-net host`) we should try to look further instead of simply not showing it.

paulbellamy:
The specific case I was seeing is when `--net=host`. Hmm.. so maybe we should show the host IPs instead, you're saying?

2opremio:
Yes, or the referenced container IPs if it also happens with `-net
containerId`.

paulbellamy:
@2opremio wdyt?

render/topologies.go
```diff
@@ -116,6 +118,56 @@ var ContainerRenderer = MakeReduce(
...
+func (r containerWithHostIPsRenderer) Render(rpt report.Report) RenderableNodes {
+	containers := r.Renderer.Render(rpt)
+	hosts := MakeMap(
+		MapHostIdentity,
+		SelectHost,
+	).Render(rpt)
+
+	for id, c := range containers {
+		networkMode, ok := c.Node.Latest.Lookup(docker.ContainerNetworkMode)
+		if !ok || networkMode != docker.NetworkModeHost {
```
2opremio: How about extending this to show the IPs of the parent container when `-net=containerID`?

render/topologies.go
```diff
@@ -116,6 +118,56 @@ var ContainerRenderer = MakeReduce(
...
+			continue
+		}
+
+		ips, ok := c.Node.Sets.Lookup(docker.ContainerIPs)
+		if ok && len(ips) > 0 {
```
2opremio: Why is this needed? Can the IPs be non-empty when -net=host?

2opremio:
Also, it would be a good idea to add a test.

render/topologies.go
paulbellamy: a bit belt-and-braces.

render/topologies.go
paulbellamy: Could do. That is a bit more complicated, as you can do `container:weavescope`, `container:a1b2c3d4`, or `container:a1b2c3d4e5f6g7h8i9j0`, so we need to parse apart the network mode string, and lookup the container by name/id/short-id. It already sort-of works (it only shows one of the IPs). I'd say let's open a different issue for that.

2opremio: Sounds good.

render/topologies.go
2opremio: It makes it a bit confusing when reading (one may think there are actual cases in which this will happen). I would suggest removing it, but I don't have a very strong opinion.

2opremio:
Minor comment, extra issue and tests-aside LGTM

render/topologies.go
paulbellamy: https://github.com/weaveworks/scope/issues/986

paulbellamy:
Refactored a bit and added test, @2opremio wdyt?

2opremio:
LGTM, but the linter is failing: _ContainerWithHostIPsRenderer should have comment or be unexported_

PR #961, opened 2016-02-12
Gather file descriptors as process metric

### LINKED ISSUE #729: Gather number of open FDs & max number as a metric

paulbellamy:
Fixes #729 

Any other bits we should to gather? I'm not super-convinced it's actually all that useful?

@2opremio, wdyt?

See "Open Files" sparkline below.
screen shot 2016-02-12 at 17 03 39

### DIFF
--- a/probe/process/walker.go
+++ b/probe/process/walker.go
@@ -4,13 +4,13 @@ import "sync"
 
 // Process represents a single process.
 type Process struct {
-	PID, PPID       int
-	Name            string
-	Cmdline         string
-	Threads         int
-	Jiffies         uint64
-	RSSBytes        uint64
-	FileDescriptors int
+	PID, PPID      int
+	Name           string
+	Cmdline        string
+	Threads        int
+	Jiffies        uint64
+	RSSBytes       uint64
+	OpenFilesCount int
 }
 
 // Walker is something that walks the /proc directory
--- a/probe/process/walker.go
+++ b/probe/process/walker.go
@@ -4,12 +4,13 @@ import "sync"
 
 // Process represents a single process.
 type Process struct {
-	PID, PPID int
-	Name      string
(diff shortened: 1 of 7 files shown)


2opremio:
I think it's a  very useful metric, but I am not convinced it should have
such protagonism, or whether it's useful to show it it's history in a
sparkline.

tomwilkie:
Yes it would be good to have this hidden behind a show more.

paulbellamy:
Hiding a single element seems needless.

davkal:
@paulbellamy From my own debugging experience, open files gets you straight into root cause analysis, whereas cpu and memory are quite vague. so I love its prominence there (until a better metric comes along).

davkal:
Tried the branch, JS part looks good too.

probe/process/walker_linux.go
```diff
@@ -83,6 +88,11 @@ func (w *walker) Walk(f func(Process, Process)) error {
 			continue
 		}
 
+		fileDescriptors, err := readFileDescriptors(path.Join(w.procRoot, filename, "fd"))
```
2opremio: I would inline the function call. It's just two lines and the parameter is not obvious.

probe/process/walker.go
```diff
@@ -4,12 +4,13 @@ import "sync"
...
-	Threads   int
-	Jiffies   uint64
-	RSSBytes  uint64
+	PID, PPID       int
+	Name            string
+	Cmdline         string
+	Threads         int
+	Jiffies         uint64
+	RSSBytes        uint64
+	FileDescriptors int
```
2opremio: Maybe rename it to `OpenFilesCount` or similar? It's more accurate.

2opremio:
minor-comments aside, LGTM

Issue #896, opened 2016-02-01
TestAll seems flaky


tomwilkie:
```
--- FAIL: TestAll (0.25s)

	scope_test.go:77: api_topology_test.go:37: Expected status 200, got 404. Path: /api/topology/pods-by-service/container%3A5e4d3c2b1a
```

https://circleci.com/gh/weaveworks/scope/2739?utm_campaign=build-failed&utm_medium=email&utm_source=notification

tomwilkie:
https://circleci.com/gh/weaveworks/scope/2742

tomwilkie:
This is pretty easy to reproduce.  I think it has to be something to do with the merging of renderable nodes.

paulbellamy:
Based on a couple failing cases with some logging, I am wondering if it is caused by conflicting cache hits.

Reflect.Value.Pointer docs say, "the returned pointer is an underlying code pointer, but not necessarily enough to identify a single function uniquely." Whether that is causing this bug or not, it's certainly a (serious) bug.

PR #962, opened 2016-02-15
Remove pointer math (comparison) from render caching, as it is unreliable

paulbellamy:
As mentioned in https://github.com/weaveworks/scope/issues/896#issuecomment-184173816

Might fix the flakey test? Not sure. Either way `reflect.Value.Pointer` of a func is unreliable.

The tradeoff is that we have individual caches for each renderer, and have to forward `ResetCache` calls throughout the tree, which is ugly.

### DIFF
--- a/render/memoise.go
+++ b/render/memoise.go
@@ -1,33 +1,43 @@
 package render
 
 import (
+	"fmt"
+	"math/rand"
+
 	"github.com/bluele/gcache"
 
 	"github.com/weaveworks/scope/report"
 )
 
+var renderCache = gcache.New(100).LRU().Build()
+
 type memoise struct {
 	Renderer
-	cache gcache.Cache
+	id string
 }
 
 // Memoise wraps the renderer in a loving embrace of caching
-func Memoise(r Renderer) Renderer { return &memoise{r, gcache.New(10).LRU().Build()} }
+func Memoise(r Renderer) Renderer {
+	return &memoise{
+		Renderer: r,
+		id:       fmt.Sprintf("%x", rand.Int63()),
+	}
+}
 
 // Render produces a set of RenderableNodes given a Report.
 // Ideally, it just retrieves it from the cache, otherwise it calls through to
 // `r` and stores the result.
 func (m *memoise) Render(rpt report.Report) RenderableNodes {
-	if result, err := m.cache.Get(rpt.ID); err == nil {
+	key := fmt.Sprintf("%s-%s", rpt.ID, m.id)
+	if result, err := renderCache.Get(key); err == nil {
 		return result.(RenderableNodes)
 	}
 	output := m.Renderer.Render(rpt)
-	m.cache.Set(rpt.ID, output)
+	renderCache.Set(key, output)
 	return output
 }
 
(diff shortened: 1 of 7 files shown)


2opremio:
The code/approach looks fine. However, I don't know enough about this part of the code to know whether it's complete or it's missing edge cases.

paulbellamy:
I feel the same. Will have reviewed by @tomwilkie

paulbellamy:
After in-person discussion fons suggests generating a stable/random ID for each renderer, and using a single-cache against that.

2opremio:
Much cleaner :+1:

That said, it won't work against mutable renderers. Are all renderers immutable?

paulbellamy:
afaik, yes.

tomwilkie:
> Are all renderers immutable?

They are all deterministic (which I think is the property we want here).

The renderers themselves are structs which are not inherently immutable, but we don't mutate them anywhere.

tomwilkie:
> Might fix the flakey test?

I could get this to flake in a loop pretty easily.  Will test tomorrow with this fix.

render/memoise.go
```diff
@@ -13,23 +13,22 @@ var renderCache = gcache.New(100).LRU().Build()
...
 func (m *memoise) Render(rpt report.Report) RenderableNodes {
-	key := ""
-	v := reflect.ValueOf(m.Renderer)
-	switch v.Kind() {
-	case reflect.Ptr, reflect.Func:
-		key = fmt.Sprintf("%s-%x", rpt.ID, v.Pointer())
-	default:
-		return m.Renderer.Render(rpt)
-	}
+	key := fmt.Sprintf("%s-%s", rpt.ID, m.id)
```
tomwilkie: +1 much nicer.

tomwilkie:
Flakes reliably on master:
```
Toms-MacBook-Pro:app twilkie$ while ./app.test --test.run TestAll; do sleep 0.1; done
PASS
PASS
PASS
PASS
--- FAIL: TestAll (0.01s)
	scope_test.go:78: api_topology_test.go:39: Expected status 200, got 404. Path: /api/topology/containers-by-hostname/container%3A5e4d3c2b1a
FAIL
Toms-MacBook-Pro:app twilkie$ while ./app.test --test.run TestAll; do sleep 0.1; done
PASS
PASS
PASS
PASS
--- FAIL: TestAll (0.02s)
	scope_test.go:78: api_topology_test.go:39: Expected status 200, got 404. Path: /api/topology/containers-by-image/container%3A5e4d3c2b1a
FAIL
Toms-MacBook-Pro:app twilkie$ while ./app.test --test.run TestAll; do sleep 0.1; done
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
PASS
--- FAIL: TestAll (0.02s)
	scope_test.go:78: api_topology_test.go:39: Expected status 200, got 404. Path: /api/topology/containers-by-image/container_image%3Aimageid123
FAIL
```

Ran for 5mins on this branch with no flakes.

PR #965, opened 2016-02-15
Docker Labels Component

davkal:
* moved to bottom of details panel in its own section
* show all labels (no expand/collapse)
* label vs value width 1:1

screen shot 2016-02-15 at 17 11 16

### DIFF
--- a/client/app/scripts/components/node-details/node-details-labels.js
+++ b/client/app/scripts/components/node-details/node-details-labels.js
@@ -1,0 +1,24 @@
+import React from 'react';
+
+export default class NodeDetailsLabels extends React.Component {
+  render() {
+    return (
+      <div className="node-details-labels">
+        {this.props.rows.map(field => {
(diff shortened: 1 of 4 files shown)


client/app/scripts/components/node-details/node-details-labels.js
```diff
@@ -0,0 +1,24 @@
...
+    return (
+      <div className="node-details-labels">
+        {this.props.rows.map(field => {
+          return (
+            <div className="node-details-labels-field" key={field.id}>
+              <div className="node-details-labels-field-label truncate" title={field.label}>
+                {field.label}
+              </div>
+              <div className="node-details-labels-field-value" title={field.value}>
+                <div className="truncate">
```
foot: Is this `div` needed? can we move `truncate` class to parent div?

foot:
Nice one LGTM

PR #969, opened 2016-02-17
set TERM=xterm on execs to work around docker issue 9299

### LINKED ISSUE #759: Can't use top in an exec pipe - "TERM not set"
screen shot 2015-12-14 at 11 56 07

paulbellamy:
Fixes #759

`docker run -t ...` injects `TERM=xterm` into the container, but `docker exec -t ...` doesn't. We can inject it ourselves via env to work around this issue.

### DIFF
--- a/probe/docker/controls.go
+++ b/probe/docker/controls.go
@@ -99,7 +99,7 @@ func (r *registry) execContainer(containerID string, req xfer.Request) xfer.Resp
 		AttachStdout: true,
 		AttachStderr: true,
 		Tty:          true,
-		Cmd:          []string{"/bin/sh"},
+		Cmd:          []string{"/bin/sh", "-c", "TERM=xterm exec /bin/sh"},
 		Container:    containerID,
 	})
 	if err != nil {


probe/docker/controls.go
```diff
@@ -99,7 +99,7 @@ func (r *registry) execContainer(containerID string, req xfer.Request) xfer.Resp
 		AttachStdout: true,
 		AttachStderr: true,
 		Tty:          true,
-		Cmd:          []string{"/bin/sh"},
+		Cmd:          []string{"/usr/bin/env", "TERM=xterm", "/bin/sh", "-l"},
```
2opremio: Why do we want a login shell? `-l` doesn't seem to be POSIX compliant.

Also, busybox doesn't ship `/usr/bin/env` but `/bin/env`

paulbellamy: or we could run `/bin/sh -c "TERM=xterm exec /bin/sh"`, which doesn't rely on env.

2opremio: That sounds better yep!

2opremio:
LGTM

PR #979, opened 2016-02-18
Improve contrast of default view

### LINKED ISSUE #874: Improve contrast of scope UI

davkal:
* expand contrast between text styles
* remove low opacity from footer

Fixes #874 together with #954


davkal:
Before: 
screen shot 2016-02-18 at 12 37 42

After:
screen shot 2016-02-18 at 12 38 05

foot:
Nice, a subtle but definite improvement.

LGTM

PR #980, opened 2016-02-18
Remove github.com/2opremio/go-1/codec

### LINKED ISSUE #973: Remove dependency on github.com/2opremio/go-1/codec
As a followup on https://github.com/weaveworks/scope/pull/916#issuecomment-185168685

2opremio:
* Bump github.com/ugorji/go/codec, to get fixes which make
  github.com/2opremio/go-1/codec unnecessary

* Remove github.com/2opremio/go-1/codec

* Remove type embeddings to avoid the problem explained at
  https://github.com/ugorji/go/issues/141#issuecomment-185450157

Fixes #973


2opremio:
@paulbellamy @davkal (or @foot ) PTAL

client/app/scripts/components/node-details/node-details-health-item.js
```diff
@@ -10,8 +10,8 @@ class NodeDetailsHealthItem extends React.Component {
       <div className="node-details-health-item">
       <div className="node-details-health-item-value">{formatMetric(this.props.value, this.props)}</div>
         <div className="node-details-health-item-sparkline">
-          <Sparkline data={this.props.samples} max={this.props.max}
-            first={this.props.first} last={this.props.last} />
+          <Sparkline data={this.props.wire_metrics.samples} max={this.props.wire_metrics.max}
+            first={this.props.wire_metrics.first} last={this.props.wire_metrics.last} />
```
paulbellamy: We shouldn't really be changing the json for this, and certainly not exposing `wire_metrics`

davkal: Yes, this looks like an implementation detail.

2opremio: How do you suggest we do this? I guess I can create a custom serializer which obtains the values from the underlying type. Let me give it a try.

paulbellamy: That sounds fine. Otherwise, at least for metrics, we just embed it so we can add a few fields. We could just copy all the existing fields over when instantiating the Wire Metrics.

2opremio:
@paulbellamy How about now? (no more client changes)

2opremio:
On top of this, it would be good to have a lint check for detecting embedded types in structs aimed at serialization (i.e. structs with with fields including `json` tags).

paulbellamy:
LGTM, +1 for lint check

PR #981, opened 2016-02-18
Force-relayout button

### LINKED ISSUE #863: Crossing lines and immovable bobbles
screen shot 2016-01-27 at 16 40 15

My first instinct is: I need to move that bobble so I can see the lines properly. Result: the whole thing moves. Close tab.

davkal:
* new button in footer
* when clicked, forces a relayout that could help with degraded graphs
* sets a store flag that will be unset on next nodes delta update
* fixes #863

screen shot 2016-02-18 at 13 08 45


foot:
Works as advertised!

Doesn't do much most of the time when you click on it. I can see myself clicking on it a lot just to see if it makes the current layout "better" somehow.

Could disable the button when forceLayout is set to true, and move the forceLayout = false to just after the conditional that checks for it... so it only re-activates if there are new nodes. Maybe. Then I would click it once sometime during the session, and then it would be de-activated until a new node came in or I changed topo. So it would reactivate fairly quickly anyway. So this doesn't sound like a worthwhile optimization.

LGTM.

davkal:
Disabled button until new nodes appear, and added title hint.

foot:
As the button behaviour is a bit unpredictable perhaps just always have it enabled.

PR #989, opened 2016-02-19
Improve performance of docker stats obtention

2opremio:
See https://github.com/weaveworks/scope/issues/812#issuecomment-186149907

### DIFF
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -203,7 +203,9 @@ func (c *container) StartGatheringStats() error {
 		}()
 
 		var stats docker.Stats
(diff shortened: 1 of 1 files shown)


2opremio:
OK, this completely takes away the decoding load. Yay!

probe

2opremio:
Acutually, they are not gone, but it's just 2.56% (0.18s) , and there is no contention reading the http bodies.

2opremio:
@tomwilkie PTAL

tomwilkie:
LGTM!

Issue #983, opened 2016-02-18
Simplify in-field profiling 


2opremio:
Right now, to get profiling information of the probe/app you one needs to install golang. On top of that, in the case of the probe, one needs to restart the probe with `--probe.http.listen`.

We should make this simpler. We have missed a couple of chances of getting detailed performance reports from users because it's too complicated.

See: https://botbot.me/freenode/weavenetwork/2016-02-17/?msg=60310921&page=1 and https://botbot.me/freenode/weavenetwork/2016-02-18/?msg=60361117&page=1

For instance we could dump a cpu/memory profile when receiving a signal.

tomwilkie:
I would love a `scope profile` command...

2opremio:
To improve the situation and remove the in-field `go tool pprof` dependency, I've tried to get the profiles directly with curl but I've miserably failed:

```
$ curl http://172.16.0.3:4040/debug/pprof/profile?seconds=30 > profile.raw
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1072  100  1072    0     0     35      0  0:00:30  0:00:30 --:--:--   288
$ go tool pprof profile.raw
Entering interactive mode (type "help" for commands)
(pprof) top
70ms of 70ms total (  100%)
      flat  flat%   sum%        cum   cum%
      70ms   100%   100%       70ms   100%  
(pprof) 
```

(Not how the resulting profile is useless)

Maybe somebody knows better and can tell me what I'm during wrong.

2opremio:
> Maybe somebody knows better and can tell me what I'm during wrong.

It was a simple as providing the binary as an argument to pprof!

The problem is that curl doesn't symbolize the output profile: https://groups.google.com/forum/#!topic/golang-nuts/-Y1NQhexJmU

I will add this to the instructions in the README.

For completeness: `go tool pprof` obtains the profile (in the same way as curl) and then symbolizes each address using the remote `pprof/symbol` endpoint, see https://github.com/golang/go/blob/master/src/cmd/pprof/internal/symbolz/symbolz.go

bboreham:
Note the current docs for obtaining a profile are at https://github.com/weaveworks/scope/blob/da2cce36912c1bd720178f3c3f6cbafd8a174bf1/site/building.md#profiling

PR #993, opened 2016-02-22
Document how to obtain profiles without `go tool pprof`

2opremio:
Improves #983

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,3 +290,18 @@ go tool pprof http://localhost:4040/debug/pprof/heap
     ```
 go tool pprof http://localhost:4041/debug/pprof/profile
 ```
+  
+  If you don't have `go` installed, you can use a docker container instead:
+  
+  * To collect the Memory profile of the Scope App:
+
+    ```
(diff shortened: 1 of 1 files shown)


tomwilkie:
Can we roll this into a command on the scope script?

2opremio:
Not as-is. The probe needs to run with `--probe.http.listen 4041`. This is just a temporary solution until we properly implement a command for it (dumping a symbolized profile and all ...)

tomwilkie:
I have no qualms about always running the probe on :4041 if it makes this easier.

2opremio:
Sure, but when we automatize it, I'd rather do it properly (i.e. not obtaining a unsymbolized profile). My intention is to document this until somebody gets around to do it and I am working on other stuff right now.

2opremio:
Maybe it would be better to suggest using a go container instead of curl ...

2opremio:
@tomwilkie I have updated the PR to use containerized-go instead of curl, which is way simpler. PTAL

2opremio:
@tomwilkie mentioned it was OK to merge offline

PR #995, opened 2016-02-22
Add ping/pong to websocket protocol

### LINKED ISSUE #828: Add ping/pong protocol to control websocket
Reverse proxies like Nginx will close connections (including websockets) after certain application-level inactivity (i.e. tcp keepalive is not good enough). 

In order to avoid this we should add a ping/pong keepalive protocol to the control websocket.

References:

* [Websockets in Ngninx](https://blog.martinfjordvald.com/2013/02/websockets-in-nginx/):

> WebSockets are still affected by proxy_read_timeout which defaults to 60 seconds. This means that if you have an application using WebSockets but not sending any data more than once per 60 seconds you either need to increase the timeout or implement a ping message to keep the connection alive. 

* https://www.reddit.com/r/golang/comments/3bt1ej/gorilla_websockets_start_closing_after_20_clients/cspc6st

paulbellamy:
Fixes #828

Also made the websocket codec `WaitForReadError` return the error.

### DIFF
--- a/app/api_topology.go
+++ b/app/api_topology.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	log "github.com/Sirupsen/logrus"
-	"github.com/gorilla/websocket"
 	"golang.org/x/net/context"
 
 	"github.com/weaveworks/scope/common/xfer"
@@ -66,10 +65,6 @@ func handleNode(nodeID string) func(context.Context, Reporter, render.Renderer, 
 	}
 }
 
-var upgrader = websocket.Upgrader{
-	CheckOrigin: func(r *http.Request) bool { return true },
-}
-
 func handleWebsocket(
 	ctx context.Context,
 	w http.ResponseWriter,
@@ -78,12 +73,11 @@ func handleWebsocket(
 	renderer render.Renderer,
 	loop time.Duration,
 ) {
-	wsConn, err := upgrader.Upgrade(w, r, nil)
+	conn, err := xfer.Upgrade(w, r, nil)
 	if err != nil {
 		// log.Info("Upgrade:", err)
 		return
 	}
-	conn := xfer.Ping(wsConn)
 	defer conn.Close()
 
 	quit := make(chan struct{})
--- a/app/api_topology.go
+++ b/app/api_topology.go
@@ -14,8 +14,7 @@ import (
 )
 
 const (
-	websocketLoop    = 1 * time.Second
-	websocketTimeout = 10 * time.Second
+	websocketLoop = 1 * time.Second
 )
 
 // APITopology is returned by the /api/topology/{name} handler.
@@ -79,17 +78,18 @@ func handleWebsocket(
 	renderer render.Renderer,
 	loop time.Duration,
 ) {
-	conn, err := upgrader.Upgrade(w, r, nil)
+	wsConn, err := upgrader.Upgrade(w, r, nil)
 	if err != nil {
 		// log.Info("Upgrade:", err)
 		return
 	}
+	conn := xfer.Ping(wsConn)
 	defer conn.Close()
 
 	quit := make(chan struct{})
-	go func(c *websocket.Conn) {
+	go func(c xfer.Websocket) {
 		for { // just discard everything the browser sends
-			if _, _, err := c.NextReader(); err != nil {
+			if _, _, err := c.ReadMessage(); err != nil {
 				if !xfer.IsExpectedWSCloseError(err) {
 					log.Println("err:", err)
 				}
@@ -112,15 +112,10 @@ func handleWebsocket(
 		diff := render.TopoDiff(previousTopo, newTopo)
 		previousTopo = newTopo
 
-		if err := conn.SetWriteDeadline(time.Now().Add(websocketTimeout)); err != nil {
+		if err := conn.WriteJSON(diff); err != nil {
 			if !xfer.IsExpectedWSCloseError(err) {
-				log.Println("err:", err)
+				log.Errorf("cannot serialize topology diff: %s", err)
 			}
-			return
-		}
-
-		if err := xfer.WriteJSONtoWS(conn, diff); err != nil {
-			log.Errorf("cannot serialize topology diff: %s", err)
 			return
 		}
 
--- a/common/xfer/websocket.go
+++ b/common/xfer/websocket.go
@@ -2,10 +2,136 @@ package xfer
 
 import (
 	"io"
+	"sync"
+	"time"
 
 	"github.com/gorilla/websocket"
 	"github.com/ugorji/go/codec"
+
+	"github.com/weaveworks/scope/common/mtime"
 )
+
+const (
+	// Time allowed to write a message to the peer.
+	writeWait = 10 * time.Second
+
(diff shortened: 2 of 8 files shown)


tomwilkie:
Also needed for the pipes websocket, no?

And do we ever call codec.Close()?

paulbellamy:
re: pipes. probably, actually.

`codec.Close()` is called from the rpc client here: https://golang.org/src/net/rpc/client.go?s=7387:7422#L267

paulbellamy:
@tomwilkie, added ping to all probe/app websockets (I think), wdyt?

common/xfer/websocket.go
```diff
@@ -2,18 +2,110 @@ package xfer
...
+type pingingWebsocket struct {
+	pinger    *time.Ticker
+	quit      chan struct{}
+	readLock  sync.Mutex
+	writeLock sync.Mutex
+	*websocket.Conn
+}
+
+// Ping adds a periodic ping to a websocket connection.
+func Ping(c *websocket.Conn) Websocket {
```
paulbellamy: Probably want a better name for this...

paulbellamy:
Tested this by opening a websocket (via nginx) to /api/control/ws, and seeing it be terminated after 60s. With this patch (and `xfer.Ping` added on the server-side instead of the client-side), it stays open indefinitely.

app/api_topology.go
```diff
@@ -79,17 +78,18 @@ func handleWebsocket(
 	renderer render.Renderer,
 	loop time.Duration,
 ) {
-	conn, err := upgrader.Upgrade(w, r, nil)
+	wsConn, err := upgrader.Upgrade(w, r, nil)
 	if err != nil {
 		// log.Info("Upgrade:", err)
 		return
 	}
+	conn := xfer.Ping(wsConn)
```
tomwilkie: Can we provide an xfer.Upgrade which automatically makes a pinger etc?  So you don't have to do this two-step everywhere.

common/xfer/websocket.go
```diff
@@ -2,29 +2,104 @@ package xfer
...
+	p.pinger = time.AfterFunc(pingPeriod, p.ping)
+	p.ping()
+	return p
+}
+
+func (p *pingingWebsocket) ping() {
+	p.writeLock.Lock()
+	defer p.writeLock.Unlock()
+	if err := p.conn.WriteControl(websocket.PingMessage, nil, mtime.Now().Add(writeWait)); err != nil {
+		p.Close()
```
tomwilkie: Should the err be logged at all?

common/xfer/websocket.go
```diff
@@ -2,29 +2,104 @@ package xfer
...
+}
+
+// ReadMessage is a helper method for getting a reader using NextReader and
+// reading from that reader to a buffer.
+func (p *pingingWebsocket) ReadMessage() (int, []byte, error) {
+	p.readLock.Lock()
+	defer p.readLock.Unlock()
+	messageType, b, err := p.conn.ReadMessage()
+	if err == nil {
+		err = p.conn.SetReadDeadline(mtime.Now().Add(pongWait))
```
tomwilkie: Why is the read deadline set after doing the read?

common/xfer/websocket.go
```diff
@@ -2,29 +2,104 @@ package xfer
...
+// WriteJSON writes the JSON encoding of v to the connection.
+func (p *pingingWebsocket) WriteJSON(v interface{}) error {
+	p.writeLock.Lock()
+	defer p.writeLock.Unlock()
+	w, err := p.conn.NextWriter(websocket.TextMessage)
 	if err != nil {
 		return err
 	}
 	err1 := codec.NewEncoder(w, &codec.JsonHandle{}).Encode(v)
+	p.conn.SetWriteDeadline(mtime.Now().Add(writeWait))
```
tomwilkie: Shouldn't this be set before doing the write?

common/xfer/websocket.go
```diff
@@ -44,5 +121,25 @@ func ReadJSONfromWS(c *websocket.Conn, v interface{}) error {
 		// One value is expected in the message.
 		err = io.ErrUnexpectedEOF
 	}
+	if err == nil {
+		p.conn.SetReadDeadline(mtime.Now().Add(pongWait))
```
tomwilkie: Ditto

probe/appclient/app_client.go
```diff
@@ -188,13 +188,12 @@ func (c *appClient) controlConnection() (bool, error) {
...
 	url := sanitize.URL("ws://", 0, "/api/control/ws")(c.target)
-	conn, _, err := c.wsDialer.Dial(url, headers)
+	wsConn, _, err := c.wsDialer.Dial(url, headers)
 	if err != nil {
 		return false, err
 	}
-	defer func() {
-		conn.Close()
-	}()
+	conn := xfer.Ping(wsConn)
```
tomwilkie: Similarly, can we have an xfer.DialWS?

probe/appclient/app_client.go
```diff
@@ -280,6 +279,7 @@ func (c *appClient) pipeConnection(id string, pipe xfer.Pipe) (bool, error) {
 	if err != nil {
 		return false, err
 	}
+	conn := xfer.Ping(wsConn)
```
tomwilkie: Ditto.

tomwilkie:
Looking better.  If I understand this right, the probe will now ping the app and the app ping the probe?  Not problem with that I guess.

common/xfer/websocket.go
paulbellamy: It's the next time we need to see something from the other side before disconnecting.

common/xfer/websocket.go
paulbellamy: ah, probably. Was thinking the `Close()` is the write.

tomwilkie:
```
./common/xfer/websocket.go:67:1: comment on exported function DialWS should be of the form "DialWS ..."
```

common/xfer/websocket.go
```diff
@@ -44,5 +156,25 @@ func ReadJSONfromWS(c *websocket.Conn, v interface{}) error {
 		// One value is expected in the message.
 		err = io.ErrUnexpectedEOF
 	}
+	if err == nil {
+		err = p.conn.SetReadDeadline(mtime.Now().Add(pongWait))
+	}
```
tomwilkie: I think we decided to remove this, and rely on pong.

common/xfer/websocket.go
tomwilkie: decided to remove and rely on pong

common/xfer/websocket.go
```diff
@@ -2,28 +2,138 @@ package xfer
...
+	if err != nil {
+		return nil, nil, err
+	}
+	return Ping(wsConn), resp, nil
+}
+
+// Ping adds a periodic ping to a websocket connection.
+func Ping(c *websocket.Conn) Websocket {
+	p := &pingingWebsocket{conn: c}
+	p.conn.SetPongHandler(p.pong)
```
tomwilkie: set initial read timeout.

tomwilkie:
LGTM once tests pass.

PR #996, opened 2016-02-22
Multi-tenant architecture for the app

tomwilkie:
My aim is to break this into multiple PRs:
- [x] PRs which refactors the App and introduces the right interfaces for alternative, multi-tenant implementations (<s>#997, #1150</s>)
- A PR which implements various multi tenant backends (this one)
- A PR which does the right plumbing in prog/app to bring this all together (also this one)

TO DEMO:
```
$ make
$ cd experimental/multitenant
$ make
$ weave launch
$ ./run.sh
$ cd ../../
$ ./scope --probe-only
```

TODO:
- Dynamo Collector
 - [x] deal with periods spanning the hour
 - [ ] read row/column keys first, pick already decoded reports from a cache, and then send a multiget for the rest.
 - [ ] use sqs for sending shortcut reports.
- Control Router
 - [x] don't delete queues
 - [x] correctly tidy up connections
 - [ ] use consul to keep connection state
- Pipe Router
 - [x] implement delete
 - [x] make it work beyond the single replica case
 - [ ] implement garbage collection


paulbellamy:
Initially looks fine. fns (particularly consulpiperouter) are too large and need breaking down, IMO (but at least there are lots of comments).

tomwilkie:
Yeah I'm rewriting pipe router into mich smaller bits.  Thanks!

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+			}
+		})
+
+	addr := fmt.Sprintf("%s:%d", pr.advertise, privateAPIPort)
+	log.Infof("Serving private API on endpoint %s.", addr)
+	log.Infof("Private API terminated: %v", http.ListenAndServe(addr, router))
+}
+
+// Atomically modify a pipe in a callback.
+// If pipe doesn't exist you'll get nil in callback.
```
paulbellamy: wat?

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+	buf := bytes.Buffer{}
+	if err := json.NewEncoder(&buf).Encode(c); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+func (c *consulPipe) fromBytes(bs []byte) error {
+	return json.NewDecoder(bytes.NewReader(bs)).Decode(c)
+}
```
paulbellamy: naming on `toBytes` and `fromBytes` seems inconsistent with json encoding elsewhere

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+		c.ProbeEnd = addr
+	}
+}
+
+func (c *consulPipe) end(e app.End) string {
+	if e == app.UIEnd {
+		return c.UIEnd
+	}
+	return c.ProbeEnd
+}
```
paulbellamy: ??? This method seems weird.

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+	return json.NewDecoder(bytes.NewReader(bs)).Decode(c)
+}
+
+func (c *consulPipe) setEnd(e app.End, addr string) {
+	if e == app.UIEnd {
+		c.UIEnd = addr
+	} else {
+		c.ProbeEnd = addr
+	}
+}
```
paulbellamy: This is more setting the addr of the end, not the end itself, yeah?

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+
+func (c *consulPipe) decr(e app.End) int {
+	if e == app.UIEnd {
+		c.UIRef--
+		return c.UIRef
+	}
+	c.ProbeRef--
+	return c.ProbeRef
+}
```
paulbellamy: In [other place(s)](https://github.com/weaveworks/scope/blob/80c7436036845c62aec0a7b90035e0a0a1ce6062/probe/appclient/multi_client.go#L183) I think we've used `acquire` and `release`? Anywhere else we're doing reference counting?

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+
+type consulPipeRouter struct {
+	prefix    string
+	advertise string // Address of this pipe router to advertise in consul
+	client    *consul.Client
+	userIDer  UserIDer
+
+	mtx     sync.Mutex
+	cond    *sync.Cond
+	pipes   map[string]xfer.Pipe // Active pipes
```
paulbellamy: maybe `activePipes map[string]xfer.Pipe` ?

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+	bridges map[string]*bridgeConnection
+
+	// Used by Stop()
+	quit chan struct{}
+	wait sync.WaitGroup
```
paulbellamy: nice pulling these out.

common/network/interface.go
```diff
@@ -0,0 +1,29 @@
+package network
+
+import (
+	"fmt"
+	"net"
+)
+
+// GetFirstAddressOf returns the first address of the supplied interface name.
+func GetFirstAddressOf(name string) (string, error) {
```
paulbellamy: Is this always going to return the right (same) address? i.e. might it sometimes return the ipv6 addr?

paulbellamy:
The ConsulPipeRouter seems to be three things in one. There's the actual pipe routing, a consistent consul kv client, and an atomic-CASable value (backed by consul).

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+		pipe.fromBytes(kvp.Value)
+		if ok, err := f(pipe); !ok {
+			return pipe, nil
+		} else if err != nil {
+			return pipe, err
+		}
+
+		index = meta.LastIndex
+	}
+	return nil, fmt.Errorf("Timed out waiting on %s", key)
```
paulbellamy: looping and calling `Get` could miss changes while running the callback. Doesn't consul offer a `subscribe`?

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+		index = uint64(0)
+		kv    = pr.client.KV()
+	)
+	for deadline.After(mtime.Now()) {
+		// Poll waiting for the entry to get updated
+		kvp, meta, err := kv.Get(key, &consul.QueryOptions{
+			RequireConsistent: true,
+			WaitIndex:         index,
+			WaitTime:          longPollDuration,
+		})
```
paulbellamy: I'm guessing this blocks until there is some change?

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+		p.incr(e)
+		return p, false, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Next wait for other end to connect
+	// at this point we 'own' one end
+	pipe, err := pr.watch(key, mtime.Now().Add(10*longPollDuration), func(p *consulPipe) (bool, error) {
```
paulbellamy: Race between calling `cas` and calling `watch`

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+		return err
+	}
+	key := fmt.Sprintf("%s%s-%s", pr.prefix, userID, id)
+	log.Infof("Delete %s", key)
+
+	_, err = pr.cas(key, func(p *consulPipe) (*consulPipe, bool, error) {
+		if p == nil {
+			return nil, false, fmt.Errorf("Pipe %s not found", id)
+		}
+		p.DeletedAt = mtime.Now()
```
paulbellamy: races with everywhere this is checked.

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+		return p, false, nil
+	})
+	return err
+}
+
+type bridgeConnection struct {
```
paulbellamy: what is this, what is it for? comment, please.

paulbellamy: bridge, or proxy?

prog/app.go
```diff
@@ -40,14 +42,81 @@ func appMain() {
...
+		awsCreateTables = flag.Bool("aws.create.tables", false, "Create the tables in DynamoDB")
+		awsSQS          = flag.String("aws.sqs", "", "URL of SQS instance")
+		awsRegion       = flag.String("aws.region", "", "AWS Region")
+		awsID           = flag.String("aws.id", "", "AWS Account ID")
+		awsSecret       = flag.String("aws.secret", "", "AWS Account Secret")
+		awsToken        = flag.String("aws.token", "", "AWS Account Token")
+
+		consulPrefix = flag.String("consul.prefix", "", "Prefix for keys in consul")
+		consulAddr   = flag.String("consul.addr", "", "Address of consul instance")
+		consulInf    = flag.String("consul.inf", "", "The interface who's address I should advertise myself under in consul")
```
paulbellamy: adding all these flags to the open source product seems mediocre.

paulbellamy: What about consolidating them into a URL? Something like:
`-collector dynamodb://id:secret@address/databaseName?create_tables=true`
`-control.router sqs://id:secret@address`
`-pipe.router consul://consulAddr/?prefix=foo&interface=eth1`

This way there are not loads of options when the open source users eventually do `scope help`
Any other ways we could clean it up?

prog/app.go
```diff
@@ -40,14 +42,81 @@ func appMain() {
...
+		pipeRouter, err = multitenant.NewConsulPipeRouter(*consulAddr, *consulPrefix, *consulInf, userIDer)
+		if err != nil {
+			log.Fatalf("Error createing consul pipe router: %v", err)
+		}
+	} else {
+		log.Fatalf("Invalid pipe router '%s'", *pipeRouterType)
+		return
+	}
+
+	defer log.Info("app exiting")
```
paulbellamy: Do we have to do all these more inline? the main is already pretty bloated...

common/xfer/pipes.go
```diff
@@ -93,7 +93,7 @@ func (p *pipe) CopyToWebsocket(end io.ReadWriter, conn Websocket) error {
 	p.mtx.Unlock()
 	defer p.wg.Done()
 
-	errors := make(chan error, 1)
+	errors := make(chan error, 2)
```
paulbellamy: eh? why this change?

app/multitenant/sqs_control_router.go
```diff
@@ -0,0 +1,330 @@
...
+	cr.queueURL = url
+}
+
+func (cr *sqsControlRouter) getQueueURL() *string {
+	cr.mtx.Lock()
+	defer cr.mtx.Unlock()
+	return cr.queueURL
+}
+
+func (cr *sqsControlRouter) getOrCreateQueue(name string) (*string, error) {
```
paulbellamy: does the caller care if it is created here? i.e. maybe just `getQueue(name)` and a comment that it creates it idempotently.

Also, is there a race between get and create?

app/multitenant/sqs_control_router.go
```diff
@@ -0,0 +1,330 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	return createQueueRes.QueueUrl, nil
+}
+
+func (cr *sqsControlRouter) loop() {
+	for {
+		// This app has a random id and uses this as a return path for all responses from probes.
```
paulbellamy: I still think that queue-per-app is strictly worse than queue-per-client, but we've been over that several times already.

app/multitenant/sqs_control_router.go
```diff
@@ -0,0 +1,330 @@
...
+		})
+		if err != nil {
+			log.Errorf("Error recieving message from %s: %v", *cr.queueURL, err)
+			continue
+		}
+		if len(res.Messages) == 0 {
+			continue
+		}
+		cr.handleResponses(res)
+		if err := cr.deleteMessages(cr.queueURL, res.Messages); err != nil {
```
paulbellamy: Passing instance fields to a method call? That's a paddlin'.

app/multitenant/sqs_control_router.go
```diff
@@ -0,0 +1,330 @@
...
+			log.Errorf("Error decoding message: %v", err)
+			continue
+		}
+		sqsResponses = append(sqsResponses, sqsResponse)
+	}
+
+	for _, sqsResponse := range sqsResponses {
+		cr.mtx.Lock()
+		waiter, ok := cr.responses[sqsResponse.ID]
+		cr.mtx.Unlock()
```
paulbellamy: Could lock (once) around the decoding phase instead, pulling out the waiters there, which would be less locking and unlocking. But how many responses are typical in one call of this function?

Either way, this method might be slightly more readable if there's just one loop.

app/multitenant/sqs_control_router.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	probeQueueName := fmt.Sprintf("probe-%s-%s", userID, probeID)
+	probeQueueURL, err := cr.service.GetQueueUrl(&sqs.GetQueueUrlInput{
+		QueueName: aws.String(probeQueueName),
+	})
+	if err != nil {
+		return xfer.Response{}, err
+	}
+
+	// Wait for a response befor we send the request, to prevent races
```
paulbellamy: this comment seems unrelated to this block of code.

app/multitenant/sqs_control_router.go
```diff
@@ -0,0 +1,330 @@
...
+	pw, ok := cr.probeWorkers[id]
+	delete(cr.probeWorkers, id)
+	cr.mtx.Unlock()
+	if ok {
+		pw.stop()
+	}
+	return nil
+}
+
+type probeWorker struct {
```
paulbellamy: maybe a comment on what these do

app/multitenant/consul_pipe_router.go
tomwilkie: Yes thats the WaitIndex thing.

app/multitenant/consul_pipe_router.go
tomwilkie: This is a subscribe: on consul subscribe is implemented as a Get + a WaitIndex.  The semantics are "block until this thing changes past this change-index I've already seen".

app/multitenant/consul_pipe_router.go
tomwilkie: Yeah, thats not right.

app/multitenant/consul_pipe_router.go
tomwilkie: Whats the race?  Watch will always call the callback immediately with its 'current' value, which I believe will alleviate any race.

app/multitenant/consul_pipe_router.go
```diff
@@ -0,0 +1,605 @@
...
+			pr.wait.Done()
+			return
+		default:
+		}
+
+		kvps, meta, err := kv.List(pr.prefix, &consul.QueryOptions{
+			RequireConsistent: true,
+			WaitIndex:         index,
+			WaitTime:          longPollDuration,
+		})
```
paulbellamy: Will calling this in a loop miss updates while the callback is running? or is that handled by WaitIndex somehow?

app/multitenant/consul_pipe_router.go
tomwilkie: Done.

app/multitenant/consul_pipe_router.go
tomwilkie: Removed (serialization done in the new consul client.)

app/multitenant/consul_pipe_router.go
tomwilkie: Change name to addrFor(e app.End)?

app/multitenant/consul_pipe_router.go
tomwilkie: Yeah.  Done.

app/multitenant/consul_pipe_router.go
tomwilkie: Done.

app/multitenant/consul_pipe_router.go
tomwilkie: done.

app/multitenant/consul_pipe_router.go
tomwilkie: thanks!

common/network/interface.go
tomwilkie: This is only ever used on the docker bridge or on the weave interface.  But will add a check.

app/multitenant/consul_pipe_router.go
tomwilkie: Added comment to explain.

app/multitenant/consul_pipe_router.go
tomwilkie: Added comment to explain.

app/multitenant/consul_pipe_router.go
tomwilkie: In what case?  Watch always starts from version 0, so is guaranteed not to miss _any_ updates to the key.  Either way it turns out we can remove this watch.

app/multitenant/consul_pipe_router.go
tomwilkie: How so?

Remember this is CASing a shared value in an internal store.  This value is not cached locally anywhere, and a background watch thread will be notified of updates and close any resources associated with it.  Whats more, everywhere we modify a pipe, we do it in a 'CAS' transaction, where we always check to see if its closed.

app/multitenant/consul_pipe_router.go
paulbellamy: ah! that's what it's doing! please.

app/multitenant/consul_pipe_router.go
tomwilkie: Added the following comment:

```go
// A bridgeConnection represents a connection between two pipe router replicas.
// They are created & destroyed in response to events from consul, which in turn
// are triggered when UIs or Probes connect to various pipe routers.
```

app/multitenant/consul_pipe_router.go
paulbellamy: I was thinking of in the case:
```
app1.cas
app2.cas
app1.watch
app2.watch
```
maybe that's not valid?

app/multitenant/consul_pipe_router.go
paulbellamy: I was thinking we might need a read lock around calls to `p.DeletedAt.IsZero()` (etc), since we're modifying the timestamp from multiple threads concurrently. Not sure now...

prog/app.go
tomwilkie: > adding all these flags to the open source product seems mediocre.

This was the least-worse option.

> What about consolidating them into a URL?

Whilst this would be slightly nicer, it doesn't solve the real problem: the options will still exist.  I suggest we just live with it, potentially marking them as "experimental" and hiding them from `help` by default (maybe unless someone does a `-hhh` or something, like `-vvv`)

Also, I don't want to write no skanky URL parsing code...

prog/app.go
tomwilkie: Not putting them inline ends up with a bunch of factory functions (eugh) with lots of arguments.

This function is nothing compares to the probe main...

When we add help, I expect I'll factor all the args out into a struct, which would make the factory functions slightly less ugly.

prog/app.go
paulbellamy: > no URL parsing code

```net/url.Parse(...)``` ?

File: common/xfer/pipes.go (hunk 16 above)
Comment (tomwilkie) [inline common/xfer/pipes.go]: The goroutines both post their errors to the channel, which if you close() the pipe may not get read from.  Therefore it needs up to 2 slots free.

Its ugly.  I need to come back to this at some point in the future and make this all a bit more sane.

File: prog/app.go (hunk 14 above)
Comment (paulbellamy) [inline prog/app.go]: > the options will still exist [...]

I'm fine with that, just trying to minimize pollution.

The other problem is the overlap between the flags is quite implicit. e.g. `-aws.region` sets the region for dynamo *and* sqs, right? etc...

File: common/xfer/pipes.go (hunk 16 above)
Comment (tomwilkie) [inline common/xfer/pipes.go]: Added this as a comment.

File: app/multitenant/sqs_control_router.go (hunk 17 above)
Comment (tomwilkie) [inline app/multitenant/sqs_control_router.go]: > does the caller care if it is created here?

Not really, but I kinda like how the code reads when used - its very obvious what its doing.

> Also, is there a race between the get and create calls?

Good spot.  Luckily no; from http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html:

"If you provide the name of an existing queue, along with the exact names and values of all the queue's attributes, CreateQueue returns the queue URL for the existing queue."

Which of course means I can just get rid of the GetQueueURL....

File: app/multitenant/sqs_control_router.go (hunk 18 above)
Comment (tomwilkie) [inline app/multitenant/sqs_control_router.go]: So we could change to queue-per-client, but the challenge is what name to give that queue?  UIs don't have an ID right now - we could introduce one, or we could pick a random id (prefixed with the user id).

I don't think it would simplify the code at all, but its possible.

File: app/multitenant/sqs_control_router.go (hunk 19 above)
Comment (tomwilkie) [inline app/multitenant/sqs_control_router.go]: Ah yeah, but deleteMessages will delete from any queue, while cr.queueURL is the url of the queue the app is listening on for ui response (as oppose to probe requests).  So it needs to be passed about.

File: app/multitenant/sqs_control_router.go (hunk 20 above)
Comment (tomwilkie) [inline app/multitenant/sqs_control_router.go]: Yeah will go for one loop.  I didn't want to lock the whole loop, as sending to the waiter could block, but thats not the case anymore, so I'll do that.

File: app/multitenant/sqs_control_router.go (hunk 21 above)
Comment (tomwilkie) [inline app/multitenant/sqs_control_router.go]: s/Wait for a response befor/Add a response channel before/

File: app/multitenant/sqs_control_router.go (hunk 22 above)
Comment (tomwilkie) [inline app/multitenant/sqs_control_router.go]:
```
// a probeWorker encapsulates a goroutine serving a probe's websocket connection.
```

File: app/multitenant/consul_pipe_router.go (hunk 11 above)
Comment (tomwilkie) [inline app/multitenant/consul_pipe_router.go]: Its consul's job to serialise all these.  Our app1.watch can't miss  the app2.cas - if we do the watch after the cas, we will see the new value (as we start with version 0), otherwise we will see the value when the version changes.

Does that make sense?

File: app/multitenant/consul_pipe_router.go (hunk 12 above)
Comment (tomwilkie) [inline app/multitenant/consul_pipe_router.go]: The consulPipes are never cached anywhere; they just represent a parsed value read out of consul.  hence no locking - as no two thread will ever share them.    Consul here is doing the serialisation (as in linearity not parsing) for us.

File: app/multitenant/sqs_control_router.go (hunk 18 above)
Comment (paulbellamy) [inline app/multitenant/sqs_control_router.go]: We discussed f2f, and I was thinking of a control connection as being long-lived, not a rest call.

Comment (tomwilkie) [inline app/multitenant/sqs_control_router.go]: Discussed offline - queue per user would quickly become queue-per-rpc (as ui<->app is a REST HTTP call, not a websocket), which is too much.

File: app/multitenant/sqs_control_router.go (hunk 19 above)
Comment (paulbellamy) [inline app/multitenant/sqs_control_router.go]: But surely we don't need to pass it *here*, as cr.deleteMessages will have access to it.

**Comment by tomwilkie** [comment]
> The ConsulPipeRouter seems to be three things in one. There's the actual pipe routing, a consistent consul kv client, and an atomic-CASable value (backed by consul).

I've broken out the high-level consul client.  Pipe-routing and the value data type have been left in there.

File: app/multitenant/consul_pipe_router.go [hunk 24]
```diff
@@ -0,0 +1,431 @@
...
+
+			end, _ := pipe.Ends()
+			if err := pipe.CopyToWebsocket(end, conn); err != nil && !xfer.IsExpectedWSCloseError(err) {
+				log.Printf("Error copying to pipe %s websocket: %v", key, err)
+			}
+		})
+
+	addr := fmt.Sprintf("%s:%d", pr.advertise, privateAPIPort)
+	log.Infof("Serving private API on endpoint %s.", addr)
+	log.Infof("Private API terminated: %v", http.ListenAndServe(addr, router))
```
paulbellamy: Since we never close the listener, and never wait for this to exit, won't `Stop()` leak this?

tomwilkie: Yes, but we've never cared about it in the past in places like prog/app.go (these things all existing for the length of the process)

paulbellamy: Fair enough, then.

paulbellamy:
LGTM

PR #997, opened 2016-02-22
Refactor app for multitenancy

tomwilkie:
- Add interfaces to allow for alternative implementations for Collector, ControlRouter
  and PipeRouter.
- Pass contexts on http handlers to these interfaces.  Although not used by the current
  (local, in-memory) implementations, the idea is this will be used to pass headers to
  implementations which support multitenancy (by, for instance, putting an authenticating
  reverse proxy in form of the app, and then inspecting the headers of the request for
  a used id).

### DIFF
--- a/app/api_report.go
+++ b/app/api_report.go
@@ -2,12 +2,13 @@ package app
 
 import (
 	"net/http"
+
+	"golang.org/x/net/context"
 )
 
 // Raw report handler
-func makeRawReportHandler(rep Reporter) func(http.ResponseWriter, *http.Request) {
-	return func(w http.ResponseWriter, r *http.Request) {
-		// r.ParseForm()
-		respondWith(w, http.StatusOK, rep.Report())
+func makeRawReportHandler(rep Reporter) CtxHandlerFunc {
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+		respondWith(w, http.StatusOK, rep.Report(ctx))
 	}
 }
--- a/app/controls.go
+++ b/app/controls.go
@@ -1,122 +1,84 @@
 package app
 
 import (
-	"math/rand"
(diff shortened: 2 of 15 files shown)


app/api_report.go
```diff
@@ -2,12 +2,13 @@ package app
...
+	"golang.org/x/net/context"
 )
 
 // Raw report handler
-func makeRawReportHandler(rep Reporter) func(http.ResponseWriter, *http.Request) {
-	return func(w http.ResponseWriter, r *http.Request) {
-		// r.ParseForm()
-		respondWith(w, http.StatusOK, rep.Report())
+func makeRawReportHandler(rep Reporter) CtxHandlerFunc {
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
```
paulbellamy: Why pass `ctx` around everywhere when we never use it?

paulbellamy: maybe add a comment in the commit message with why.

app/controls.go
```diff
@@ -1,122 +1,84 @@
...
-		return
+func handleControl(cr ControlRouter) CtxHandlerFunc {
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+		var (
+			vars    = mux.Vars(r)
+			probeID = vars["probeID"]
+			nodeID  = vars["nodeID"]
+			control = vars["control"]
+		)
+		result, err := cr.Handle(ctx, probeID, xfer.Request{ //TODO context
```
paulbellamy: Either remove the comment or elaborate a bit to explain it.

paulbellamy:
A couple gripes about comments, but otherwise LGTM.

app/api_report.go
tomwilkie: Done

PR #1000, opened 2016-02-22
Reduce amount of objects allocated by the codec

### LINKED ISSUE #972: Compile ugorji/go/codec with -tags unsafe for higher performance
Followup of https://github.com/weaveworks/scope/pull/916 . 

Using `codecgen` with flag `-u` like we are doing may be sufficient, but it may be worth investigating if building `ugorji/go/codec` with `-tags unsafe` on top of that helps. It currently breaks encoding though.

See:
* https://github.com/ugorji/go/issues/141#issuecomment-185130876
* https://github.com/weaveworks/scope/pull/916#issuecomment-183289718

2opremio:
Closes #972 . Improves #812 and #854 

Results in the probe generating ~20% less objects. See https://github.com/ugorji/go/issues/143

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ RUN_FLAGS=-ti
 BUILD_IN_CONTAINER=true
 GO ?= env GO15VENDOREXPERIMENT=1 go
 GO_BUILD_INSTALL_DEPS=-i
-GO_BUILD_TAGS=-tags netgo
+GO_BUILD_TAGS=-tags 'netgo unsafe'
 GO_BUILD_FLAGS=$(GO_BUILD_INSTALL_DEPS) -ldflags "-extldflags \"-static\" -X main.version=$(SCOPE_VERSION)" $(GO_BUILD_TAGS)
 
 all: $(SCOPE_EXPORT)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -24,7 +24,7 @@ RUN_FLAGS=-ti
 BUILD_IN_CONTAINER=true
 GO ?= env GO15VENDOREXPERIMENT=1 go
 GO_BUILD_INSTALL_DEPS=-i
-GO_BUILD_TAGS=-tags netgo
+GO_BUILD_TAGS=-tags 'netgo unsafe'
```
2opremio: I am not too happy about providing the unsafe tag globally, but I don't know how to provide tags for the library in isolation. For instance, compiling it separately before scope, results in the library being built again without `-tags unsafe`.

paulbellamy: Even though ugorji/go is the only thing which understands `-tags unsafe` it still makes me uncomfortable.

paulbellamy:
As discussed in slack, the decoding thing will only affect gzip writers, which we might be able to solve with a `bufio.Writer` wrapping the gzip writer. the `stringView` one is a bit trickier. :|

2opremio:
Unfortunately ugorji/go#143 shows the biggest benefit when encoding. So ... should I revert it?

2opremio:
After thinking about it a bit more, we already obtain much higher performance from codecgen because we run it with `-u`, which generates code applying the same kind of tricks. So I don't think `-tags unsafe` makes the situation worse (and I certainly wouldn't consider using `codecgen` without `-u` for the reasons explained).

PR #1001, opened 2016-02-22
Make linter ignore generated files

2opremio:


tomwilkie:
Changes like this need to be made in build-tools.git and subtree'd in.

Also, see #1002 for a better way to ignore files.

PR #1005, opened 2016-02-22
Refactor Makefile a little.

tomwilkie:
This allows you to do things like `make report/report.codecgen.go` and have it correctly fork into a container.  Its also more uniform, as all the dependancies are separated out for the logic of how to build an actual target.


2opremio:
LGTM

PR #1017, opened 2016-02-23
Add connection tables to details panel

### LINKED ISSUE #378: Wrong number of TCP connections reported in the details pane
The details pane of the scope container reads *TCP connections 1*:

screen shot 2015-08-19 at 6 29 00 pm

But the number of connections is clearly higher than 1:

screen shot 2015-08-19 at 6 30 52 pm

### LINKED ISSUE #680: Consider deduping connections by client port numbers in the details panel

### LINKED ISSUE #1111: Use "connections" instead of "requests" for Inbound/Outbound node labels
Just like it says on the tin.

tomwilkie:
Part of #680, Fixes #378, Fixes #1111

This change also:
- Makes RenderableNode.Children a set of RenderableNodes, not Nodes
- <s>Puts pseudo nodes back into the container's view (makes more sense to show connections now we have shapes)</s> (This needed more work - #1123)

TODO:
- [x] make it sort by count, by default
- [x] change table ID as per davkal's request
- [x] make connection work to/from the internet node
- [x] figure out why the memoisation causes the test to fail
- [x] testing / coverage
- [x] add real nodes to the table
- [x] figure out why inbound count != outbound count for self-connections
- [x] better table for the internet node
- <s>use reverse dns info</s> (#1121)


report/id_list.go
```diff
@@ -35,6 +33,10 @@ func (a IDList) Merge(b IDList) IDList {
...
 // Contains returns true if id is in the list.
 func (a IDList) Contains(id string) bool {
-	i := sort.Search(len(a), func(i int) bool { return a[i] >= id })
-	return i < len(a) && a[i] == id
+	return StringSet(a).Contains(id)
+}
+
+// Intersection returns the intersection of a and b
+func (a IDList) Intersection(b IDList) IDList {
+	return IDList(StringSet(a).Intersection(StringSet(b)))
```
paulbellamy: nice

render/detailed/node.go
```diff
@@ -157,5 +167,140 @@ func children(n render.RenderableNode) []NodeSummaryGroup {
...
+		if !node.Adjacency.Contains(n.ID) {
+			continue
+		}
+
+		node.Children.ForEach(func(child report.Node) {
+			if child.Topology != report.Endpoint {
+				return
+			}
+
+			intersection := child.Adjacency.Intersection(myEndpoints)
```
paulbellamy: nice

render/detailed/node.go
```diff
@@ -157,5 +167,140 @@ func children(n render.RenderableNode) []NodeSummaryGroup {
...
+	remotes := map[string]int{}
+
+	for _, node := range endpoints {
+		for _, adjacent := range node.Adjacency {
+			if parts := strings.SplitN(adjacent, ":", 4); len(parts) == 4 {
+				key := parts[2] + "|" + parts[3]
+				remotes[key] = remotes[key] + 1
+			}
+		}
+	}
```
paulbellamy: Seems very similar to the above. possible to extract/combine/refactor?

davkal:
How about making connections a toplevel node field:
```
{
metadata,
metrics,
connections: [
  {
    id: "out",
    columns: [{id: "id", label: "Remote"}, {id: "port", label: "Port"}, {id: "count", label: "Label"}}],
    items:[{id: <nodeId>, port: <port>, count: <count>}, ...]
  },
  {id: in, ...}
]}
```

tomwilkie:
Sorry should have WIP marker.  Yes I hope to combined lots of this code, once I get it working.

tomwilkie:
@davkal I eventually want the rows to link to other objects in the topology (ie link to the other container talking to this one) like the other children tables.  Links will always be within the current view.  Can you update your datastructure for this?  I'm hoping links will always be possible, but you might want to make them optional.

paulbellamy:
just reading a bit for curiosity.

davkal:
Alternatively we can model connections as a nodes table, i.e. wrapping ports as metadata (typed as int to make it sortable as a number) and counts wrapped as a metric:
```
{
metadata,
metrics,
children,
connections: [{
    id: "out",
    label: "Outgoing Connections",
    columns: [{id: "label", label: "Remote"}, {id: "port", label: "Port"}, {id: "count", label: "Count"}}],
    nodes: [{
      id: <nodeId>,
      label: <nodeLabel>,
      topologyId: <topologyId>,
      metadata: [{port: <port>, type: "number"}]
      metrics: [{count: <count>}]
    }, ...]
  },
  {id: in, ...}

**Comment by davkal** [comment]
`topologyId: "endpoint-incoming"` needs to be `id: "endpoint-incoming"`. `topologyId` is something like `processes` that will be used together with a node's `id` to build a deeplink.

File: app/api_topology_test.go [hunk 4]
```diff
@@ -54,27 +54,6 @@ func TestAll(t *testing.T) {
 	}
 }
 
-func TestAPITopologyContainers(t *testing.T) {
-	ts := topologyServer()
```
tomwilkie: I dropped this test as rendered containers contain a bunch of extra special casing (image names in rank etc) and its verbose testing that doesn't add anything.

app/api_topology_test.go
```diff
@@ -112,6 +91,7 @@ func TestAPITopologyHosts(t *testing.T) {
 	defer ts.Close()
 	is404(t, ts, "/api/topology/hosts/foobar")
 	{
+		render.ResetCache() // TODO figure out why this is needed.
```
paulbellamy: Why *is* this needed? each report should have a different ID, yeah? ID collisions in the renderers, or reports?

tomwilkie: I was getting errors without it.  I don't know the root of those errors.

render/detailed/node.go
```diff
@@ -157,5 +165,144 @@ func children(n render.RenderableNode) []NodeSummaryGroup {
...
 	}
+
 	return nodeSummaryGroups
 }
+
+const (
+	remotePortKey = "Remote Port"
+	localPortKey  = "Local Port"
+	countKey      = "Count"
+	numberWang    = "number"
```
paulbellamy: wat

tomwilkie: https://www.youtube.com/watch?v=qjOZtWZ56lc

render/detailed/node.go
```diff
@@ -157,5 +165,144 @@ func children(n render.RenderableNode) []NodeSummaryGroup {
...
+
+		node.Children.ForEach(func(child render.RenderableNode) {
+			if child.Topology != report.Endpoint {
+				return
+			}
+
+			intersection := child.Adjacency.Intersection(myEndpoints)
+			if len(intersection) <= 0 {
+				return
+			}
```
paulbellamy: don't really need this, the below loop will just skip.

paulbellamy:
still quite a bit of tricky string-munging code shared between the incoming connections and outgoing connection aggregators.

paulbellamy:
`Connections Fr...` is not a great table name.
screen shot 2016-02-25 at 16 17 40

paulbellamy:
I guess no luck on converting the endpoints to their respective nodes in the relevant topology?

tomwilkie:
> I guess no luck on converting the endpoints to their respective nodes in the relevant topology?

Time constraint I'm afraid.  Will do it next sprint.

paulbellamy:
Aside from a few naming things and refactoring (if possible), this LGTM.

tomwilkie:
@paulbellamy I think I've addressed all the feedback, and got it working for the internet node.  Its a bit of a hack.  I think we should punt this form 0.13 and finish it properly.  Thoughts?

tomwilkie:
Also think the internet is going to need a custom table (for outgoing connections):

screen shot 2016-02-25 at 19 10 27

tomwilkie:
Going to hold this until #566 is merged.  Also need @pidster input on the table for the internet node(s).

app/api_topology_test.go
tomwilkie: Pretty sure its solved by #1088, will rebase when thats merged in.

render/id.go
```diff
@@ -14,6 +24,11 @@ func MakeEndpointID(hostID, addr, port string) string {
 	return makeID("endpoint", hostID, addr, port)
 }
 
+// MakePseudoEndpointID makes a pseudo endpoint node ID for rendered nodes.
+func MakePseudoEndpointID(hostID, addr, port string) string {
+	return makeID("pseudo-endpoint", hostID, addr, port)
+}
```
paulbellamy: Seems an odd differentiation between `pseudo` nodes and `pseudo-endpoint` nodes.

render/mapping.go
```diff
@@ -50,15 +51,31 @@ func theInternetNode(m RenderableNode) RenderableNode {
...
 	} else {
 		node.ID = OutgoingInternetID
 		node.LabelMajor = OutboundMajor
-		node.LabelMinor = RequestsMinor
+		node.LabelMinor = OutboundMinor
 	}
 	return node
 }
 
+// RemapEndpointIDs remaps endpoints to have the right id format.
```
paulbellamy: This comment is unhelpful. What format does it expect them to have? what is the `right` format?

render/mapping.go
```diff
@@ -118,7 +137,9 @@ func MapEndpointIdentity(m RenderableNode, local report.Networks) RenderableNode
 		minor = fmt.Sprintf("%s (%s)", minor, pid)
 	}
 
-	return RenderableNodes{id: NewRenderableNodeWith(id, major, minor, rank, m)}
+	node := NewRenderableNodeWith(id, major, minor, "", m)
+	node.Children = node.Children.Add(m)
```
paulbellamy: eh? so endpoints are children of themselves? That seems odd?

render/mapping.go
```diff
@@ -155,10 +176,9 @@ func MapContainerIdentity(m RenderableNode, _ report.Networks) RenderableNodes {
 		id       = MakeContainerID(containerID)
 		major, _ = GetRenderableContainerName(m.Node)
 		minor    = report.ExtractHostID(m.Node)
-		rank, _  = m.Latest.Lookup(docker.ImageID)
 	)
 
-	node := NewRenderableNodeWith(id, major, minor, rank, m)
```
paulbellamy: why is rank going away?

render/mapping.go
```diff
@@ -453,14 +473,9 @@ func MapEndpoint2Process(n RenderableNode, _ report.Networks) RenderableNodes {
 // It does not have enough info to do that, and the resulting graph
 // must be merged with a container graph to get that info.
 func MapProcess2Container(n RenderableNode, _ report.Networks) RenderableNodes {
-	// Propogate the internet pseudo node
-	if strings.HasSuffix(n.ID, TheInternetID) {
-		return RenderableNodes{n.ID: n}
-	}
-
-	// Don't propogate non-internet pseudo nodes
+	// Propogate pseudo node
```
paulbellamy: Propagate

render/id.go
tomwilkie: indeed, I just couldn't think of a better name.  Any suggestions?

render/detailed/connections.go
```diff
@@ -0,0 +1,197 @@
+package detailed
```
paulbellamy: This file may be easier to read if you flip the order so the primary functions (`makeIncomingConnectionsTable` and `makeOutgoingConnectionsTable`) were first, then the helpers afterwards.

paulbellamy: Also, it seems like the functions have a bit of redundancy in the names `buildThis` and `makeThat`

paulbellamy:
General approach seems fine. A couple naming and tidiness things.

render/mapping.go
tomwilkie: This is the endpoint being added as a child of the pseudo endpoint.

render/mapping.go
tomwilkie: It gets over written by containerWithImageNameRenderer anyway.

render/detailed/connections.go
tomwilkie: Done & done.

render/mapping.go
tomwilkie: Done.

render/id.go
paulbellamy: Not the naming. It seems like an odd way to slice the distinction.

render/mapping.go
paulbellamy: ah

render/id.go
tomwilkie: Well the whole point is to have the pseudo nodes have endpoints as their children.  I guess there are two alternatives:
- introduce the pseudo nodes in the process mapping instead of the endpoint mapper
- make endpoint mapper add the 'report' endpoint nodes (from the selector) as children of the endpoints & pseudo nodes, and make the connection table code look for them instead

Can you think of any other alternatives?  Which do you prefer?

paulbellamy: > introduce the pseudo nodes in the process mapping instead of the endpoint mapper

That would be more consistent with how the "uncontained" pseudo nodes are handled, no?

render/detailed/connections.go
```diff
@@ -0,0 +1,197 @@
...
+			id, label, linkable = row.remoteAddr+":"+row.port, row.remoteAddr, false
+		}
+		metadata := []MetadataRow{}
+		if includeLocal {
+			metadata = append(metadata,
+				MetadataRow{
+					ID:       "foo",
+					Value:    row.localAddr,
+					Datatype: number,
+				})
```
rade: @tomwilike as the author, and @davkal @paulbellamy as the reviewers: Can you shed any light on

1. the significance of "foo"
2. why the Datatype is 'number' when the Value is an IP address (and in more recent versions an IP address or DNS name with IP address in parentheses)? Won't this cause `parseFloat` in https://github.com/weaveworks/scope/pull/1017/files#diff-37e17af975f5417e26f6036ba3624af9R55 to return "NaN"?

paulbellamy: > The significance of MetadataRow.ID == "foo"

Just has to match the Column.ID == "foo" above.

> Won't this cause parseFloat to return "NaN".

No.

rade: > Just has to match the Column.ID == "foo" above.

Well yes, I figured that. But why is it called "foo" when all the other IDs are named sensibly?

> No.

Why not? And why do we want to set the type to 'number' here when the value *isn't* a number?

paulbellamy: > And why do we want to set the type to 'number' here when the value isn't a number?

Because we want them sorted numerically, not lexicographically.

rade: > Because that is not what parseFloat does. Try it: `parseFloat("127.0.0.1")`

I did try it before my original comment. Try: `parseFloat("cloud.weave.works (52.3.116.14)")`.

rade: Sorting IP addresses by converting them to a float?!?! You are having a laugh.
```
> parseFloat("10.0.121.1")
10
> parseFloat("10.0.118.1")
10
```

paulbellamy: > Sorting IP addresses by converting them to a float?!?!

Pull requests are welcome. Also, a better example of the behaviour is:
```
> parseFloat("192.168.12.13")
192.168
```
It is probably (almost definitely) older than the reverse dns lookup, so wouldn't have ever handled `cloud.weave.works (52.3.116.14)`.

rade: > it is probably (almost definitely) older than the reverse dns lookup

Right, I said that much: "in more recent versions"

Regarding the 'foo'... given that the other `MetadataRow.ID` values in this code are 'port' and 'count', which match the column labels, why not 'remote'?

paulbellamy: Sure, that would make sense.

rade: I have filed #1869 for the sorting problem.

rade: Looks like @foot is doing the de-foo'ing in  #1874.

foot: Yeah I had to reference the col id in the js so changed the name.

I dropped `dataType: number` too as I didn't consider the creative applications of parseFloat, I'd start introducing new dataTypes though, `number|ip|ips`

PR #1023, opened 2016-02-24
Turn off GC for builds.

tomwilkie:
Also, don't rmi the images on `make clean` - rm'ing the .uptodate files is enough to ensure we rebuild the images, and rmi'ing the images causes us to have to redownload a lot of stuff.

A complete rebuild with this PR is almost 2 minutes faster (not these both include the rmi change):

```
$ # without change
$ make clean; time make
...
real	5m38.475s
user	0m5.205s
sys	0m1.306s

$ # with change
$ make clean; time make
...
real	3m56.951s
user	0m5.190s
sys	0m1.259s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ CODECGEN_TARGETS=report/report.codecgen.go render/render.codecgen.go render/deta
 RM=--rm
 RUN_FLAGS=-ti
 BUILD_IN_CONTAINER=true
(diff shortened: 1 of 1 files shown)


paulbellamy:
I'm *assuming* doing `GOGC=off go build something` just disables GC during the build, not compiles the binary without GC, right?

tomwilkie:
I assume so too.

paulbellamy:
That assumption is correct.

PR #1026, opened 2016-02-24
Make image count sorted in the details panel

tomwilkie:


paulbellamy:
Displays a bit oddly in the details panel:
screen shot 2016-02-24 at 15 22 59

paulbellamy:
Wouldn't metadata formatted as integer still be sortable?

tomwilkie:
Metadata is sortable, yes, but not sorted by default.

@davkal if we add a format field (or sort field) to a metadata row, can we have that sorted too?

We don't need to hold 0.13 up on this.

davkal:
Yes, please specify the default sort (other than 1st metric) in the columns of a table:
```
columns: [{id, label, defaultSort: true}, ...]
```

PR #1027, opened 2016-02-24
SVG export button

### LINKED ISSUE #555: SVG export for large topologies?
Common displays aren't large enough to display all of the detail that scope can expose.  Most computers aren't powerful enough to render this detail smoothly, so:

 * .svg exports
 * a way for users to select how much they'd like their PC to be abused by scope (slider?)

davkal:
* applies CSS styles inline
* exports SVG chart node
* injects A tag to download
* based on SVG-crowbar

Fixes #555


davkal:
Current export function modifies the SVG in the main DOM which leads to sometimes disappearing borders of nodes. Would be good to copy the SVG subtree and inject the inline styles to the copy.

pidster:
LGTM (based on the demo & as soon as the tests pass).

foot:
Nice, LGTM

PR #1028, opened 2016-02-24
Make unconteinerized build work on OSX

2opremio:
@davkal PTAL


tomwilkie:
I don't understand how this even works... surely when entering the container it tries to build something that it doesn't know about?

2opremio:
What don't you understand? It simply names the binary based on the host
architecture/OS (for which it's also compiled).

It fixes trying to use a binary from the wrong architecture when mixing
builds in and out of the container (e.g. in an OSX host) which was being
experienced by @davkal.

From a conversation in Slack it seems @davkal may be having other problems
too, let's see if this helps.

davkal:
Builds in container, and on Darwin. LGTM

tomwilkie:
Pls hold off on this until I've had a chance to look at it, I have some concerns.

tomwilkie:
I get this when trying to build using docker-machine on my mac:
```
docker run --rm -ti \
		-v /Users/twilkie/Documents/src/github.com/weaveworks/scope:/go/src/github.com/weaveworks/scope \
		-v /Users/twilkie/Documents/src/github.com/weaveworks/scope/.pkg:/go/pkg \
		--net=host \
		-e GOARCH -e GOOS -e CIRCLECI -e CIRCLE_BUILD_NUM -e CIRCLE_NODE_TOTAL \
		-e CIRCLE_NODE_INDEX -e COVERDIR -e SLOW \
		weaveworks/scope-backend-build SCOPE_VERSION=18964aa GO_BUILD_INSTALL_DEPS=-i vendor/github.com/ugorji/go/codec/codecgen/bin/codecgen_darwin_amd64
make: Entering directory '/go/src/github.com/weaveworks/scope'
make: *** No rule to make target 'vendor/github.com/ugorji/go/codec/codecgen/bin/codecgen_darwin_amd64'.  Stop.
make: Leaving directory '/go/src/github.com/weaveworks/scope'
make: *** [vendor/github.com/ugorji/go/codec/codecgen/bin/codecgen_darwin_amd64] Error 2
```

2opremio:
@tomwilkie please take another look. I fixed it by moving all the dependencies to the non-containerized part.

tomwilkie:
This now required I have go installed, which I thought we'd decided against...

2opremio:
> This now requires I have go installed, which I thought we'd decided against...

No it doesn't. Have you actually tested it?

As I already said, the makefile variables are use lazily unless assigned with `:=`, although I can happily define the variable in the non-containerized-block if that's confusing.

```
vagrant@vagrant-ubuntu-vivid-64:~/scope$ echo $PATH
/home/vagrant/bin:/usr/local/go/bin:/home/vagrant/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
vagrant@vagrant-ubuntu-vivid-64:~/scope$ export PATH=/home/vagrant/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
vagrant@vagrant-ubuntu-vivid-64:~/scope$ which go
vagrant@vagrant-ubuntu-vivid-64:~/scope$ make
```

davkal:
@2opremio how did you test the darwin build? Which command do you run?

2opremio:
@davkal I have been relying on you to test the build in Darwin (I use a LinuxVM), hence the PTAL.

backend/Dockerfile
```diff
@@ -1,5 +1,4 @@
 FROM golang:1.5.3
-ENV GO15VENDOREXPERIMENT 1
```
tomwilkie: Why did you remove this?

tomwilkie: Oh I see, moved to the Makefile.

tomwilkie:
This command used to work `make report/report.codecgen.go` (on my mac, building in a container) but now does not.

Also, if I change the build container the .codecgen.go files are not rebuilt and I kinda think they should be.

In general it makes me uncomfortable to have different dependancy trees inside the container to outside.

tomwilkie:
Note both of these are minor issues.  What precisely is the problem we're trying to solve here?

tomwilkie:
I cant think of a nicer way of doing this, so assuming we still want to fix the build-in-container-then-build-on-a-mac-without-a-make-clean-inbetween problem, I'm LGTM.

2opremio:
@davkal / @foot You would still would like to have this, right?

foot:
Yes please, easier it is to build in and out of container the better!

2opremio:
> Yes please, easier it is to build in and out of container the better!

Alright, then I will rebase the changes and ask you to take another look

2opremio:
@foot Can you please test it so that we can merge it? (Ideally before it bit-rots again! :) )

foot:
`env -u` doesn't work on osx! grrr.
Linux: `env -u GOARCH -u GOOS $(GO)`
Darwin equiv: `unset GOARCH GOOS; $(GO)`

- Made a quick patch for the above.
- I also made a mess when rebasing onto master, i think i fixed it though, look okay?
- I have a new file now: `vendor/github.com/ugorji/go/codec/codecgen/codecgen`, should this be gitignored?

foot:
But otherwise it works nicely!

foot:
Rebased onto master.

foot:
But circle is giving a funny permission error now so maybe i broke something..

2opremio:
> I have a new file now: vendor/github.com/ugorji/go/codec/codecgen/codecgen, should this be gitignored?

That's because we now build `vendor/github.com/ugorji/go/codec/codecgen/codecgen_OS_ARCH` instead. Just remove the file manually (you will only need to do this once).

2opremio:
I think we were overcomplicating it. My last commit makes it much simpler (although it will only work when https://github.com/ugorji/go/pull/148 is merged and we bump the dependency).

* Get rid of GO15VENDOREXPERIMENT since it's enabled by default in Go >=1.6,
  just use go 1.6 if you want to build outside the container.
* Stop using unset/env -u, just avoid adding env variables where needed.
* Pass build flags consistently to codecgen.
* Let the experimental/ directory be.

foot:
LGTM for me!

PR #1029, opened 2016-02-24
Replace Show More buttons with carets w/ counts

davkal:
screen shot 2016-02-24 at 18 59 47

### DIFF
--- a/client/app/scripts/components/node-details.js
+++ b/client/app/scripts/components/node-details.js
@@ -130,7 +130,6 @@ export default class NodeDetails extends React.Component {
 
   renderDetails() {
     const details = this.props.details;
-    const showSummary = (details.metadata || details.metrics || details.docker_labels) !== undefined;
     const showControls = details.controls && details.controls.length > 0;
     const nodeColor = getNodeColorDark(details.rank, details.label, details.pseudo);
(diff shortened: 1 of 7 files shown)


foot:
Little issues:

1. Clicking on the minified load numbers next to the sparklines doesn't expand them
2. If there is no cpu/mem in a table the whole row doesn't highlight on mouseover.
3. With the `+42 v` off to the side the line becomes more prominent and feels "at-a-glance" to be more of a divider between sections, but a very unbalanced one (no margin-top and too much margin-bottom). But i've been playing w/ the diff combos. I think lhs/center/rhs all have strengths and weaknesses. I'm leaning a tiny bit more towards centered again but I you can arg for any of them really.

LGTM after 1 & 2!

davkal:
> If there is no cpu/mem in a table the whole row doesn't highlight on mouseover.

If there are no metrics, no `td` elements are generated, which the browser seem to need to make up a full row.

foot:
`&nbsp;` ftw. I'm not that fussed. Still LGTM i think.

pidster:
I concur, LGTM

davkal:
The styled the show moar line to be dotted. Looks more subtle, and a bit less like a divider.

PR #1043, opened 2016-02-25
Distinguish between reporting probes and controlling probes

### LINKED ISSUE #1040: k8s: Pipes/controls don't work most of the time
See video: http://recordit.co/8BTwXcjnRO

It may (or may not) be k8s-specific since it works in my local k8s cluster.

Apart from the unexpected EOF error I also got a `Bad request` in pipes and a few `Probe _ is not connected right now...`

screen shot 2016-02-25 at 17 33 22
screen shot 2016-02-25 at 17 35 28

2opremio:
Fixes #1040

### DIFF
--- a/report/report.go
+++ b/report/report.go
@@ -209,4 +209,6 @@ const (
 	HostNodeID = "host_node_id"
 	// ProbeID is the random ID of the probe which generated the specific node.
 	ProbeID = "probe_id"
+	// ControlProbeID is the random ID of the probe which controls the specific node.
+	ControlProbeID = "control_probe_id"
 )
--- a/report/report.go
+++ b/report/report.go
@@ -207,8 +207,6 @@ const (
 	// a node in the host topology. That host node is the origin host, where
 	// the node was originally detected.
 	HostNodeID = "host_node_id"
-	// ProbeID is the random ID of the probe which generated the specific node.
-	ProbeID = "probe_id"
 	// ControlProbeID is the random ID of the probe which controls the specific node.
 	ControlProbeID = "control_probe_id"
 )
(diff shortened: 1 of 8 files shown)


2opremio:
It would be good to know why `go vet` and friends didn't catch this.

2opremio:
We should also distinguish between reporting hosts and containing hosts, otherwise we get strange stuff in the UI like:

screen shot 2016-02-26 at 02 06 28

(note the three hosts on the top). I will create a separate ticket for this.

tomwilkie:
Its not go vet that catches that, its errcheck.

report/report.go
```diff
@@ -209,4 +209,6 @@ const (
 	HostNodeID = "host_node_id"
 	// ProbeID is the random ID of the probe which generated the specific node.
 	ProbeID = "probe_id"
+	// ControlProbeID is the random ID of the probe which controls the specific node.
+	ControlProbeID = "control_probe_id"
```
tomwilkie: I don't think we should add another probe id (ControlProbeID ) to the metadata.  I think we should be more specific about what ProbeID is, and make sure its added correctly.

tomwilkie:
I'd remove the extra ControlProbeID, its redundant - ProbeID is only used from controls, and it was getting added where it shouldn't - by the host tagger, on container nodes the kubernetes reported is reporting.

I would stop the host tagger from adding the ProbeID to all nodes that probe reports, and instead make the docker report add it to the nodes it controls.  And update the comment on ProbeID constant.  Does that make sense?

tomwilkie:
I'm Tom and I approve of this change.

PR #1052, opened 2016-02-27
Update go-checkpoint library and add link to README.

tomwilkie:
This fixes a connection leak in go-checkpoint library, and adds more details to the readme.


vendor/github.com/weaveworks/go-checkpoint/checkpoint.go
```diff
@@ -284,6 +291,31 @@ func checkResult(r io.Reader) (*CheckResponse, error) {
...
+// getSystemUUID returns the base64 encoded, scrypt hashed contents of
+// /sys/class/dmi/id/product_uuid (with some salt)
+func getSystemUUID() (string, error) {
+	uuid, err := ioutil.ReadFile("/sys/class/dmi/id/product_uuid")
+	if err != nil {
+		return "", err
+	}
+	if len(uuid) <= 0 {
+		return "", fmt.Errorf("Empty system uuid")
+	}
```
paulbellamy: Why is an empty uuid file an error? Shouldn't that be the same as not finding one?

vendor/github.com/weaveworks/go-checkpoint/checkpoint.go
```diff
@@ -284,6 +291,31 @@ func checkResult(r io.Reader) (*CheckResponse, error) {
...
+	hash, err := scrypt.Key(uuid, uuid, 16384, 8, 1, 32)
```
paulbellamy: this salt doesn't seem very random

paulbellamy: > The recommended parameters for interactive logins as of *2009* are N=16384, r=8, p=1.

vendor/github.com/weaveworks/go-checkpoint/checkpoint.go
```diff
@@ -284,6 +291,31 @@ func checkResult(r io.Reader) (*CheckResponse, error) {
...
+	}
+	output := bytes.Buffer{}
+	encoder := base64.NewEncoder(base64.StdEncoding, &output)
+	if _, err := encoder.Write(hash); err != nil {
+		return "", err
+	}
+	if err := encoder.Close(); err != nil {
+		return "", err
+	}
+	return output.String(), nil
```
paulbellamy:
```Go
return base64.StdEncoding.EncodeToString(hash), nil
```

paulbellamy:
How does this fix the connection leak? Was not setting a signature file opening a new connection each time?

vendor/github.com/weaveworks/go-checkpoint/checkpoint.go
tomwilkie: > this salt doesn't seem very random

(Random) salt doesn't help here, so this is equivalent to just not using salt.  I should update the comment.

> Might need increasing since 2009?

https://github.com/Tarsnap/scrypt/issues/19

tomwilkie:
> How does this fix the connection leak? Was not setting a signature file opening a new connection each time?

Good catch; this doesn't seem to include https://github.com/weaveworks/go-checkpoint/commit/eea0cb58cd03781462c52c1b14814195118320f0

tomwilkie:
> How does this fix the connection leak?

Fixed; look for a body.Close()

tomwilkie:
@paulbellamy PTAL

paulbellamy:
LGTM

PR #1054, opened 2016-02-27
circle.yml: disable GCE builds when credentials are missing

alban:
Another call to gce.sh was added in commit abebb0f197f3 ("Build new test VM
(to pickup latest Docker version)") but without the guard on
$SECRET_PASSWORD. Add the guard to be able to test on CircleCI without
Weaveworks' credentials.


tomwilkie:
Nice!

PR #1069, opened 2016-02-29
Try to figure out the most suitable shell to use

### LINKED ISSUE #1053: terminals: use bash instead of /bin/sh if available
I always find myself frustrated with the limitations of sh (no history, no edition ...) and the first thing I do is run bash.

Other options instead of prioritizing bash specifically:
* Check /etc/passwd (I am not sure this makes a lot of sense for containers)
* Somehow make it a user-configuration option

In my opinion, considering its widespread use, trying to default to bash sounds sane.

2opremio:
Fixes #1053 

Tested in busybox (no getent), alpine and ubuntu

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,4 +152,5 @@ deps:
 	$(GO) get -u -f -tags $(GO_BUILD_TAGS) \
 		github.com/FiloSottile/gvt \
 		github.com/mattn/goveralls \
-		github.com/weaveworks/github-release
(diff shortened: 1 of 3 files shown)


tomwilkie:
Nice.  I think we need an integration test for this.

2opremio:
@tomwilkie PTAL

tomwilkie:
LGTM

PR #1055, opened 2016-02-28
circle.yml: deploy on a personal hub account

alban:
I would to do some changes in Scope and test my changes in Kubernetes,
following the guide on:
https://github.com/weaveworks/scope#using-weave-scope-with-kubernetes

For that, I need a Docker image of Scope hosted on a Docker hub and
reference that Docker image in my Kubernetes resource yaml file. Instead
of building Scope on my laptop and pushing it manually on my Docker Hub
account, I am using my CircleCI account to do this automatically.

This patch adds a deployment target in circle.yml. It uses the
environment variable $DEPLOY_PRIVATE_HUB_BRANCH from my CircleCI
settings and any push on a branch with a name matching ".*master" (the
branch name must end with "master") will trigger the deployment.

This change should not impact the build of the official Scope Docker
image because $DEPLOY_PRIVATE_HUB_BRANCH is not defined in the official
CircleCI settings (https://circleci.com/gh/weaveworks/scope).

PR #1083, opened 2016-03-01
circle.yml: add variable $DOCKER_ORGANIZATION

alban:
With this patch, CircleCI will push by default to the Docker repository
$DOCKER_ORGANIZATION/scope. If $DOCKER_ORGANIZATION is not defined, it
will use $DOCKER_USER/scope.

----

See previous PR: https://github.com/weaveworks/scope/pull/1055

@tomwilkie: with this patch, adding `DOCKER_ORGANIZATION=weaveworks` in the CircleCI settings should be enough.

/cc @iaguis


iaguis:
LGTM

tomwilkie:
Thanks!

PR #1086, opened 2016-03-01
Have the probe fill in the app id, not the app

### LINKED ISSUE #1085: Pipes won't work with the service.
- Pipes are initialised by controls
- When sending a control from the app to a probe, the app fills in its unique id, so the probe knows where to send the response (or connect the newly created pipe) to.
- In the service, the apps think they have a different id to the one served under /api

To fix, have the probe will in the app id on the control request when received.

tomwilkie:
Fixes #1085

### DIFF
--- a/common/xfer/controls.go
+++ b/common/xfer/controls.go
@@ -11,7 +11,7 @@ var ErrInvalidMessage = fmt.Errorf("Invalid Message")
 
 // Request is the UI -> App -> Probe message type for control RPCs
 type Request struct {
-	AppID   string
+	AppID   string // filled in by the probe on reciept of this request
 	NodeID  string
 	Control string
 }
--- a/common/xfer/controls.go
+++ b/common/xfer/controls.go
@@ -11,7 +11,7 @@ var ErrInvalidMessage = fmt.Errorf("Invalid Message")
 
 // Request is the UI -> App -> Probe message type for control RPCs
 type Request struct {
-	AppID   string // filled in by the probe on reciept of this request
+	AppID   string // filled in by the probe on receiving this request
 	NodeID  string
 	Control string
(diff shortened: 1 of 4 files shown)


common/xfer/controls.go
```diff
@@ -11,7 +11,7 @@ var ErrInvalidMessage = fmt.Errorf("Invalid Message")
 
 // Request is the UI -> App -> Probe message type for control RPCs
 type Request struct {
-	AppID   string
+	AppID   string // filled in by the probe on reciept of this request
```
rade: receipt

rade: or, better, "on receiving this request"

2opremio:
I've done some testing and this is not enough to fix #1085 

screen shot 2016-03-02 at 7 23 07 pm

There are two additional problems:

* The first pipe request (GET) is not correctly prefixed. @davkal can you please take a look at this? Also, is there a reason make the second (websocket) request even if the first request didn't succeed? 
* The first pipe request (GET) uses the same endpoint as the second (websocket) request, I discussed this offline with @tomwilkie. This causes a conflict in the nginx configuration since websockets need special treatment (adding an upgrade header in the forwarded request etc ...) . We could have different location rules based on the request type but it's hacky http://stackoverflow.com/questions/8591600/nginx-proxy-pass-based-on-whether-request-method-is-post-put-or-delete

Finally, @tomwilkie , since this is blocking upgrading the service and requires cutting a new release (0.13.1), would you mind re-creating a pull request against a release branch forked from `v0.13.0`?

tomwilkie:
Sounds like a plan. Let's make this top priority for tomorrow.

@simon is the man you want for ui pipes.

PR #1100, opened 2016-03-03
Have the probe fill in the app id, not the app

### LINKED ISSUE #1085: Pipes won't work with the service.

tomwilkie:
Fixes #1085


2opremio:
LGTM

PR #1105, opened 2016-03-03
Metrics on canvas

### LINKED ISSUE #1074: Metrics on canvas
Placeholder.

foot:
screen shot 2016-03-03 at 13 19 01

Fixes #1074

Bugs:

- [x] Fix metric formatting, either w/ BE support from a `metric.format`, or a local mapping. It might be good to get `metric.label` here to keep the metrics consistent w/ the labels in the details panel.
- [x] Figure out what metrics are actually available to display by re-scanning `[node].metrics` onAdd/onUpdate.
- [x] Fix text size
- [x] Fix 0.0001 displaying a tiny bit of metric-area when it gets rounded away to 0.00
- [x] ~~Clarify what it means to have an agg node metric. (In metric selector perhaps. `avg`)~~ -No agg metrics for now.
- [x] ~~Do we want to show "host cpu" on the processes topo, is that interesting?~~ - No
- [x] Fixes metric keyboard sorting order.
- [x] Keep selected metric in the url state.

Features

- [x] Fade the node border color when a metric is selected to give the metric more emphasis while still (just) keeping the ability to determine a node's rank.
- [x] `<`, `>`, and `q` kb shortcuts to cycle metrics, like the `top` command.
- [x] Different color for different metrics for more easily recognising when cycling.
- [x] As soon as a metric is non-zero bump up the fill to start at effectively 10% for easier detection of 0/non-zero metrics.
- [x] Pin / unpin with by re-clicking metric.
- [x] Log scale for load (50) / open-files (100000) as these don't really have a max, though we are considering ulimit for open-files
- [x] Update metrics every 3s as the reports come in from the probe. (mocked)
- [x] Change colors away from green/red/orange.


paulbellamy:
The hover on highlight with this seems to also highlight all connected nodes...

foot:
Feature!

But will play with that. I guess its not stuff you might necessarily really be interested in, so, distracting.

paulbellamy:
Need to figure out a better way to do the pruning for the websocket diff. Maybe `RenderableNode.Diff(other RenderableNode)`?

The js seems to go a bit wonky when the websocket updates arrive. There's a metric on every node with the value `NaN` after..

foot:
- [ ] Fix metric formatting, either w/ BE support from a `metric.format`, or a local mapping.
- [ ] Keep a node's metrics in-line somehow, e.g. having details panel open shows the metrics in a diff state than to the canvas. Perhaps a more global store here.
- [ ] Figure out what metrics are actually available to display by re-scanning `[node].metrics` onAdd/onUpdate.

Questions
 - How often do we want the metrics to update? It sounded good to have every changed pushed onto the WS, but thats a lot of stuff perhaps.
 - How often does the server get a new value for a metric? every 15s or 3s?
 - We could do something similar to the details panel where we build a buffer up from the BE and stream it out slowly to give the (false) impression of "live" metrics.

paulbellamy:
> How often does the server get a new value for a metric? every 15s or 3s?

Everytime there's a new report so every 3s.

foot:
Figure out how we should handle `max`. at the mo:
- `cpu`: max is `100`. (is it 200 w/ 2 cores?)
- `load`/`memory`/`open fds`: the max value in the 15 samples the BE stores?

Does it make sense to try and figure out other values for max, e.g.
 - Total available memory on the system for memory.
 - `ulimit` for open files?

paulbellamy:
>  (is it 200 w/ 2 cores?)

yes

> Total available memory on the system for memory. `ulimit` for open files

Probably, yeah

foot:
Nice! #1149 would be pretty great.

I'm not sure about load.
 1. Take the global max of the topo
 2. Choose a very large value like `100` and use a log scale for the fill/bar.
 
A log scale could work for `open files` too as the max is probably something v.large.

paulbellamy:
Upon reflection, I'm not convinced the `PruneWithMetrics` hack will work for this. Needs revisiting.

paulbellamy:
- [ ] Figure out how to send the metric summaries in deltas and topologies updates
- [ ] Only send the last/latest value for metrics in the summaries

foot:
rfr @davkal 

Just the JS code in this branch in the end.

client/app/scripts/actions/app-actions.js
```diff
@@ -12,6 +13,42 @@ import AppStore from '../stores/app-store';
...
+    metricType: AppStore.getAvailableCanvasMetricsTypes()[nextMetric]
+  });
+  updateRoute();
+}
+
+export function pinMetric(metricId) {
+  AppDispatcher.dispatch({
+    type: ActionTypes.PIN_METRIC,
+    metricId,
+    metricType: AppStore.getAvailableCanvasMetricsTypes()[metricId]
```
davkal: Why is `metricType` computed here instead of the AppStore?

client/app/scripts/charts/nodes-chart.js
```diff
@@ -74,10 +74,10 @@ export default class NodesChart extends React.Component {
         edges: makeMap()
       });
     }
+    //
     // FIXME add PureRenderMixin, Immutables, and move the following functions to render()
-    if (nextProps.forceRelayout || nextProps.nodes !== this.props.nodes) {
-      _.assign(state, this.updateGraphState(nextProps, state));
-    }
+    _.assign(state, this.updateGraphState(nextProps, state));
+
```
davkal: This is gonna be pricey.

client/app/scripts/components/metric-selector.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { selectMetric } from '../actions/app-actions';
+import { MetricSelectorItem } from './metric-selector-item';
+
+// const CROSS = '\u274C';
+// const MINUS = '\u2212';
+// const DOT = '\u2022';
+//
```
davkal: I guess we dont need those.

client/app/scripts/charts/node-shape-circle.js
```diff
@@ -1,22 +1,41 @@
...
+  });
+  const metricStyle = {
+    fill: getMetricColor(metric)
+  };
+  const fontSize = size * CANVAS_METRIC_FONT_SIZE;
 
   return (
-    <g className="shape">
+    <g className={className}>
+      <defs>
```
davkal: Could we not render this (and the metric-fill path) when no metric is selected?

client/app/scripts/charts/nodes-chart.js
```diff
@@ -131,6 +131,13 @@ export default class NodesChart extends React.Component {
       return 1;
     };
 
+    const metric = node => {
+      const met = node.get('metrics') && node.get('metrics')
+        .filter(m => m.get('id') === this.props.selectedMetric)
+        .first();
+      return met;
+    };
+
```
foot: Pull this out and maybe pop the rest of the metrics off the nodes too, so a change in one of their values doesn't force a re-render of the chart even if its not selected.

client/app/scripts/stores/app-store.js
```diff
@@ -564,6 +608,22 @@ export class AppStore extends Store {
...
         });
 
+        availableCanvasMetrics = nodes
+          .valueSeq()
+          .flatMap(n => (n.get('metrics') || makeList()).map(m => (
+            makeMap({id: m.get('id'), label: m.get('label')})
+          )))
+          .toSet()
+          .sortBy(m => m.get('label'))
+          .toJS();
```
davkal: Why not leave it an immutable set/list? May help with purity later on.

client/app/scripts/stores/app-store.js
```diff
@@ -58,6 +59,12 @@ let controlPipes = makeOrderedMap(); // pipeId -> controlPipe
 let updatePausedAt = null; // Date
 let websocketClosed = true;
 
+let selectedMetric = null;
+let pinnedMetric = selectedMetric;
+let pinnedMetricType = null;
```
davkal: A comment would help for metric vs metricType

client/app/scripts/utils/metric-utils.js
```diff
@@ -0,0 +1,56 @@
+import _ from 'lodash';
+import d3 from 'd3';
+import { formatMetric } from './string-utils';
+import { colors } from './color-utils';
+
+
+const openFilesScale = d3.scale.log().domain([1, 100000]).range([0, 1]);
```
davkal: I was hoping for these to be more generic. As in, the backed determines the type of scale and the value range. So there would be no need to hardcode `openFiles` here.

client/app/scripts/utils/string-utils.js
```diff
@@ -4,43 +4,54 @@ import d3 from 'd3';
...
+  };
+
+  return formatters;
+}
+
+
+const formatters = makeFormatters(renderHtml);
+const svgFormatters = makeFormatters((text, unit) => `${text}${unit}`);
+
+export function formatMetric(value, opts, svg) {
```
davkal: Why didnt you export 2 formatMetric functions? You had to define 2 formatters (including the `svg` argument) anyway.

client/app/styles/main.less
```diff
@@ -387,12 +387,43 @@ h2 {
     }
   }
 
-  .shape {
+  .stack .shape .highlighted {
+    display: none;
```
davkal: Much better solution than `highlightOnly`.

client/app/styles/main.less
```diff
@@ -387,12 +387,43 @@ h2 {
...
+    .node { display: none; }
+    .highlighted { display: inline; }
+  }
+
+  .stack .shape .metric-fill {
+    display: none;
+  }
+
+  .stack .onlyMetrics .shape .metric-fill {
+    display: inline-block;
```
davkal: Why is that needed?

client/app/styles/main.less
```diff
@@ -387,12 +387,43 @@ h2 {
...
+  }
 
+  .shape {
     /* cloud paths have stroke-width set dynamically */
     &:not(.shape-cloud) .border {
       stroke-width: @node-border-stroke-width;
       fill: @background-color;
+      transition: stroke-opacity 0.5s cubic-bezier(0,0,0.21,1), fill 0.5s cubic-bezier(0,0,0.21,1);
```
davkal: What made you choose `cubic-bezier(0,0,0.21,1)`? In other places we use `ease-in-out`. Would be good to have transitions perform similarly to keep the *feel* the same. Maybe extract the transition style as a variable.

client/package.json
```diff
@@ -67,7 +68,10 @@
     "test": "jest",
     "coveralls": "cat coverage/lcov.info | coveralls",
     "lint": "eslint app",
-    "clean": "rm build/app.js"
+    "clean": "rm build/app.js",
+    "noprobe": "../scope stop && ../scope launch --no-probe --app.window 24h",
+    "loadreport": "npm run noprobe && sleep 1 && curl -X POST -H \"Content-Type: application/json\" http://$BACKEND_HOST/api/report",
+    "loadreportjson": "npm run loadreport -- -d @../k8s_report.json"
```
davkal: I'd not hardcode a report here. Either use env variables, or allow args via `npm run loadreportjson -- @report.json`.

client/webpack.local.config.js
```diff
@@ -26,7 +26,8 @@ module.exports = {
     'app': [
       './app/scripts/main',
       'webpack-dev-server/client?http://localhost:4041',
-      'webpack/hot/only-dev-server'
+      'webpack/hot/only-dev-server',
+      './app/scripts/debug'
```
davkal: Nice one. Let's keep that there.

davkal:
All works well and looks good.

My personal nit would be the fill color. I feel it's too strong. Maybe use `getNodeColorDark` for more consistent coloring and try a lower opacity for a lighter feel. Or add a new `getNodeColorLight` that corrects lightness in the manner `getNodeColorDark`, just the towards consistent lightness.

client/app/styles/main.less
foot: Was needed for metrics on stacks, which are no more!

client/app/styles/main.less
foot: I c-p'd from somewhere else... the details panel transform! But variablizing is good call, for this case its not noticeable which ease we use, perhaps more so for the details-panel.

client/app/scripts/utils/metric-utils.js
foot: openFilesScale will disappear w/ #1149.

Load will continue w/ a vague max. Current stuff in a metric object:

```js
{
  id: "process_cpu_usage_percent",
  label: "CPU",
  format: "percent",
  value: 4.16,
  samples: [],
  min: 0,
  max: 100,
  first: "2016-04-04T08:41:49.831200363Z",
  last: "2016-04-04T08:42:02.831776457Z"
}
```

`scale: "log"`, feels a bit funny in there... but i understand the gist... I'll make an issue.

client/app/scripts/utils/string-utils.js
foot: It was starting to get a bit too higher-order. But I'll give it a whirl. Would be nicer.

PR #1106, opened 2016-03-03
Pause Button

### LINKED ISSUE #1025: Simple pause button
I would like to explore a curious looking graph I can see, but it keeps evolving and I cannot keep track of the changes occurring.

davkal:
* pauses topology and details panel updates
* buffers requests
* un-pause plays back the buffered requests
* automatically un-pauses when topology is changed, or other node is
  clicked (then discards buffer)

Pause button

Fixes #1025


peterbourgon:
If state == paused, button should be play triangle, right?

errordeveloper:
This is really great!! The only thing I'm not too sure about is _resume-on-click_... When posting #1025, the use-case I had in mind was about exploring details of several nodes without having them move around, and potentially observe state of ephemeral nodes, as those naturally attract my attention.

davkal:
> If state == paused, button should be play triangle, right?

I prefer active/inactive pause. In the active state (paused) it blinks. This has the benefit of drawing attention to it in case the user wonders why nothing is updating. A blinking play button in a paused state would be misleading.

2opremio:
Awesome!

I however think we should make the Pause state more predominant, otherwise:

1. User clicks pause (intentionally or by error)
2. User forgets about it
3. User sees that the UI is frozen and gets frustrated, blaming the software.
4. User reports bug (or even worse, does a _permanent stop_ :) )

davkal:
> the use-case I had in mind was about exploring details of several nodes without having them move around

I see, essentially pausing the topology, but keeping the freedom to explore nodes. Is that more desirable than the current implementation (freezing topology and details to show them to someone)?

> and potentially observe state of ephemeral nodes, as those naturally attract my attention.

This is inherently impossible/difficult to manage in the UI because the nodes may be gone by the time you explore them at which point in time a request for details would come back empty.

davkal:
> I however think we should make the Pause state more predominant, otherwise:

> User clicks pause (intentionally or by error)
> User forgets about it
> User sees that the UI is frozen and gets frustrated, blaming the software.

The safeguards here are: 

* blinking pause button with incrementing update count
* click on a topology or node un-pauses

We could add more: (1) click on the background un-pauses, (2) show a confirmation dialog.

Could you give this a try @2opremio and see if you get stuck in a paused mode?

> User reports bug (or even worse, user does a permanent stop :) )

We certainly dont want to undermine the utility of the product as a whole. If this button is too dangerous, we should simply close this PR.

davkal:
Branch `pause-topology-only` keeps the topology fixed, but allows clicking/updating nodes.

pidster:
> This is inherently impossible/difficult to manage in the UI because the nodes may be gone by the time you explore them at which point in time a request for details would come back empty.

True until a timestamp parameter is supported & queries can be made against a history of reports...

foot:
I really like the idea of being able to completely pause and inspect the system. Though due to current limitations this would be a tricky do.
 - Grab all topologies data for every topo.
 - Do a details call for every node in every topo.
 - Store all this info.

Maybe for the next one.

Small things:

- Maybe include the label in the button here.
screen shot 2016-03-07 at 11 55 21
  - perhaps swapping "paused" for "resume" onMouseOver? Or just the tooltip.
- Set `cursor: pointer;` on all the buttons down here.
- If you have a details panel open and hit pause, the buffered chart animation keeps moving.
- The "indirect" (changin topo/clicking node) unpausing transition is a bit subtle, I didn't realise that I had unpaused the system by performing those actions.
  - Perhaps leave the "resuming" there for 3-4s, maybe w/ a darker bg that fades out.
  - Set content of new topo/details panel to be a big fat play button.

client/app/scripts/stores/app-store.js
```diff
@@ -357,7 +372,13 @@ export class AppStore extends Store {
       this.__emitChange();
       break;
 
+    case ActionTypes.CLICK_PAUSE_UPDATE:
+      updatePausedAt = new Date;
```
foot: I didn't realise parens were optional in this case! TIL. Though would still go for. `new Date()`

client/app/scripts/actions/app-actions.js
```diff
@@ -76,12 +78,20 @@ export function clickNode(nodeId, label, origin) {
     nodeId
   });
   updateRoute();
+  resetDetailsUpdateBuffer();
+  resumeUpdate();
   getNodeDetails(
```
foot: Instead of making sure we unpause in all these diff places it would be nice to have the abstraction that if we need to request fresh data to show some piece of info we have to unpause. I'm not sure where that would go, perhaps just something to keep in mind for now if we re-work the datamodel into something reduxey.

client/app/scripts/utils/update-buffer-utils.js
```diff
@@ -0,0 +1,129 @@
...
+import AppStore from '../stores/app-store';
+
+const log = debug('scope:update-buffer-utils');
+const makeList = Immutable.List;
+const feedInterval = 1000;
+const bufferLength = 100;
+
+let deltaBuffer = makeList();
+let detailsBuffer = makeList();
+let updateTimer = null;
```
foot: Again single immutable store would be awesome here! Then could pin the whole UI to some ref of that while the `HEAD` keeps moving forward w/ updates.

client/app/scripts/components/app.js
```diff
@@ -82,6 +87,13 @@ export default class App extends React.Component {
     const otherContrastModeTitle = contrastMode ? 'Switch to normal contrast' : 'Switch to high contrast';
     const forceRelayoutClassName = 'footer-label footer-label-icon';
     const forceRelayoutTitle = 'Force re-layout (might reduce edge crossings, but may shift nodes around)';
+    const isPaused = updatePaused;
+    const pauseClassName = isPaused ? 'footer-label footer-label-icon footer-label-active' : 'footer-label footer-label-icon';
+    const updateCount = getUpdateBufferSize();
+    const hasUpdates = updateCount > 0;
+    const pauseTitle = isPaused ? `Paused on ${formatDate(updatePausedAt)}` : 'Pause updates';
+    const pauseAction = isPaused ? clickResumeUpdate : clickPauseUpdate;
+    const pauseWrapperClassName = isPaused ? 'footer-active' : '';
```
foot: Maybe pull out into another cmp

client/app/scripts/utils/update-buffer-utils.js
```diff
@@ -0,0 +1,129 @@
...
+      }
+    });
+
+    // check if an updated node in first was removed in second -> remove
+    _.each(first.update, node => {
+      const removedNode = _.find(second.remove, {'id': node.id});
+      if (removedNode) {
+        toUpdate = _.reject(toUpdate, {'id': node.id});
+      }
+    });
```
foot: Maybe pull these chunks out into a "consolidateDeltas" fn, I know I'm lazy but I kind of skipped over why we were doing all this consolidation stuff on every buffered delta w/out realising this only happens if we fill up the buffer.

foot:
Being able to click around is quite nice!

Pros:
 - Solves a known use case more clearly.

Cons:
 - Some nodes will respond w/ a 404 i guess. (happens w/ both proposed solutions though).
 - Confuses the meaning of paused a little as less stuff in the UI is paused.

Its easier to forget the topo is paused if you're off exploring details, I'd make the pause state a tiny bit more prominent, maybe bold + leave the border around the button + label when activated.

I'm +1 on `pause-topology-only` though.

foot:
LGTM after TAL! Button looks slick.

Use case coverage:
 - + Can pause and look over the topo and see how transient nodes are connected.
 - - Probably can't see the details of these transient nodes as the BE won't know about them anymore after a second or two.

But I think this is a good start!

PR #1120, opened 2016-03-04
Adjust node borders

### LINKED ISSUE #1020: Stacks render badly in high-contrast mode
screen shot 2016-02-24 at 10 36 43

### LINKED ISSUE #1080: Make node outer border stroke consistent width
Single node borders are thicker than the border on a stack.
The single node should use the same smaller value for border width as the stack.

thick

thin-stack

davkal:
* increase stack node borders by 0.5px (fixes #1080)
* space out stacks in contrast mode (fixes #1020)
* fix cloud border stroke-width setting (exlude from CSS styling)
* adjust cloud border for contrast mode

### DIFF
--- a/client/app/scripts/charts/node.js
+++ b/client/app/scripts/charts/node.js
@@ -83,10 +83,6 @@ export default class Node extends React.Component {
     if (this.props.pseudo) {
       classNames.push('pseudo');
     }
-    if (this.props.shape === 'cloud') {
-      // need to identify cloud in CSS because it's stroke is not set via CSS
-      classNames.push('cloud');
-    }
 
     const classes = classNames.join(' ');
 
--- a/client/app/scripts/charts/node.js
+++ b/client/app/scripts/charts/node.js
@@ -83,6 +83,9 @@ export default class Node extends React.Component {
     if (this.props.pseudo) {
       classNames.push('pseudo');
     }
+    if (this.props.shape === 'cloud') {
+      classNames.push('cloud');
+    }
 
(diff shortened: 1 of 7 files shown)


pidster:
LGTM

client/app/scripts/utils/contrast-utils.js
```diff
@@ -0,0 +1,7 @@
+const contrastMode = window.location.pathname.indexOf('contrast') > -1;
```
foot: Maybe implement this in terms of `contrastModeUrl`

client/app/scripts/charts/node.js
```diff
@@ -83,6 +83,9 @@ export default class Node extends React.Component {
     if (this.props.pseudo) {
       classNames.push('pseudo');
     }
+    if (this.props.shape === 'cloud') {
+      classNames.push('cloud');
+    }
```
foot: Can we just pop the cloud class down in the cloud shape or needed for selectors up here?

foot:
You fixed contrast mode too! +1

LGTM

client/app/scripts/charts/node.js
davkal: This is a good one:

* clouds have strokes set on their attrs via JS because of dynamic scaling
* all shapes have strokes set their width via CSS
* CSS overrides widths set on attrs -> cloud shapes get wrong (from CSS) widths

my fix for this introduces a `.cloud` selector that can be used to address all nodes except the cloud to set widths via CSS.

Issue #538, opened 2015-10-01
Expose Kubernetes service connections


paulbellamy:
Kubernetes service connections go through kube-proxy, which shows up as an "Uncontained" node. This makes following those connections difficult for users. We should split up those connections by service (virtual IP), and render a pseudo-node for each.

The kube-proxy is either a userspace hop, or a bunch of iptables rules, depending on configuration.

Created from #441

tomwilkie:
With k8s 1.2, connections are remapped using iptables.  And we deal with them correctly.

PR #1126, opened 2016-03-04
Plugins

### LINKED ISSUE #554: RFC: Probe plugins
Aim: allow people to add nodes & metadata to the scope object model. 

Example would be embedding links and iframes from cadvisor in the UI, or adding custom metadata for filtering in the DSL.

Proposal:
- bind mount /etc/weave/scope-probe-plugins into the scope containers
- for each executable script in there, run it every probe tick and parse stdout as a json report.  Eg:
```
#!/usr/bin/python
import json

if __name__ == "__main__":
  print json.dumps({
    "Container": {
      "Nodes": {
        "<hostname>;<containerid>": {
          "metadata": {
            "some-third-party-key": "hello world"
          }
        }
      }
    }
  })
```
- add a field to the node struct called 'Links', with type []link, in such a way that plugins can add links to it, and the UI renders them in the details panel:
```
type link struct {
  label string
  url strgin
}
```
- add a field to the node struct calleds 'Embeds', with type []embed, in such a way that plugins can add links to it, and the UI renders them as iframes in the details panel:
```
type link struct {
  title string
  url string
  height, width int
}
```
- implement an example that embeds cadvisor links and iframe of graphs in the UI.
- implement an example that annotates container with something from ECS, and allows user to filter them.

paulbellamy:
Fixes #554

Commits are still super-messy, and tests are earmarked, as this is both a work-in-progress and a proof-of-concept.

HTTP Request latency is shown on the weavescope node (we can correlate it due to weavescope being in the host network).

Plugin registry turned out very similar to #809.

TODO:
- [x] figure out what the sysdig querying looks like and how flexible it is
    - [x] do some plugin data from the raw sysdig stream
      - bandwidth usage per-process, maybe
      - can we parse stuff fast enough? or filter it better?
- [ ] ~~get better data from sysdig plugin~~
    - Either need to:
      - implement the plugin in lua (possible?)
      - make the lua chisel more flexible and add an interface for the go plugin to control it
      - just add more aggregations for each type of data
- [x] show plugin list in UI
    - [x] styling
    - [ ] maybe adjust timers? or include it with report/node updates?
- [x] clean up and make plugins code sane/tested
- [x] add safe zero-values for everything in a report so json deserialization is safer on malformed inputs
- [x] all templates (not just extra, and including metrics) need to come from the probe
- [x] add deadline/timeout on plugin calls
- [x] add api versioning to plugin handshake
- [x] re-organize code to be consistent with controls
- [x] clean up todos
- [x] consider rest api instead of jsonrpc
- [x] add periodic retry on dead plugin sockets (for if they restart)
- [ ] ~~replace example plugin with something else~~ Pushed eBPF plugin into a separate PR.
  - [ ] add check to ebpf plugin for if kernel is new enough
  - [ ] port ebpf plugin to python, and do counting in kernel
  - [ ] figure out how to get pids from ebpf, so we can get more accurate counts per/process
- [ ] ~~consider an async "plugin-push" interface instead of probe polling.~~ Considered it. Delaying to a future "api version"
- [x] see about removing render/detailed/labels.go
- [x] final test fixup
- [x] documentation
- [x] update plugins when they change
  - [x] lets ditch handshake, and just assume all plugins respond to `/report`, then parse the pluginspec from that.


paulbellamy:
@davkal should we display in the UI which plugins are enabled? what would be a good way of displaying this?

davkal:
Yes, let's place it in the bottom left for now, because

* it's a status
* it may have multiple entries/rows that can stack nicely

Some thing like:
```
10 NODES (408 FILTERED)
Stopped containers hidden SHOW
System containers hidden SHOW
[plugin icon] pluginX, pluginY
```
The DS could be as simple as `[{id, label, description},...]`. The description would show on hover. If plugins support a state (on/off), we can add that too.

paulbellamy:
In order to get more/better data from the sysdig probe the best plan seems to be using the "chisel" system, which are a bunch of lua scripts sysdig can run. We could just capture the event stream, but that will likely lead to a *lot* of unneeded serialization/deserialization.

We could either write individual chisels for bits they have, and plumb them through, or we could look into writing one generic meta-chisel which would let us dynamically grab bits and do queries. We could either implement the entire plugin in lua (probably? There seems to be a jsonrpc library for lua), or we could do the barebones filtering/querying in the lua chisel, and drive that from a go plugin.

I've included a `http_txns_by_pid.lua` chisel (based on the existing `httplog` one), as an example of a chisel which would format the data for a specific query. It adds inbound/outbound http request rates to process nodes.

Some other things we could *potentially* get from the sysdig probe:
* bandwidth usage
* protocol detection/inspection
* more http metrics (latency, error rates, error codes, avg size, breakdown by endpoints/hosts/etc)
* same for other protocols (memcached, redis, etc)
* might be able to tie in with tracing?
  * could maybe even do "session capture", like chrome's network request logging
* unix socket connections
* socket queue lengths
* slow system calls / system call length

app/router.go
```diff
@@ -164,10 +164,13 @@ func RegisterReportPostHandler(a Adder, router *mux.Router) {
 	}))
 }
 
-func apiHandler(_ context.Context, w http.ResponseWriter, r *http.Request) {
-	respondWith(w, http.StatusOK, xfer.Details{
-		ID:       UniqueID,
-		Version:  Version,
-		Hostname: hostname.Get(),
-	})
+func apiHandler(rep Reporter) func(context.Context, http.ResponseWriter, *http.Request) {
```
tomwilkie: `func(context.Context, http.ResponseWriter, *http.Request)` = `CtxHandlerFunc`

paulbellamy: Thanks, I *thought* there was a type for that...

plugins/plugin_set.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/davecgh/go-spew/spew"
+	"github.com/mndrix/ps"
+	"github.com/ugorji/go/codec"
+
+	"github.com/weaveworks/scope/test/reflect"
+)
+
+// PluginSet is a set of plugins keyed on (Topology, ID). Clients must use
+// the Add method to add plugins
+type PluginSet struct {
```
tomwilkie: Is this necessary? Could you add a "plugins" topology?

prog/probe.go
```diff
@@ -221,3 +233,28 @@ func getFirstAddressOf(name string) (string, error) {
 		return "", fmt.Errorf("No address found for %s", name)
 	}
 }
+
+// TODO(paulbellamy): where should this live?
```
tomwilkie: Why not have pluginRegistry implement Reporter?

prog/probe.go
```diff
@@ -221,3 +233,28 @@ func getFirstAddressOf(name string) (string, error) {
...
+
+			pluginReport := report.MakeReport()
+			pluginReport.Plugins = pluginReport.Plugins.Add(plugin)
+			buf := &bytes.Buffer{}
+			if err := codec.NewEncoder(buf, &codec.JsonHandle{}).Encode(pluginReportJSON); err != nil {
+				log.Errorf("plugins: error pointlessly encoding/decoding report: %v", err)
+			}
+			if err := codec.NewDecoder(buf, &codec.JsonHandle{}).Decode(&pluginReport); err != nil {
+				log.Errorf("plugins: error pointlessly encoding/decoding report: %v", err)
+			}
```
tomwilkie: This is a shame...

prog/probe.go
```diff
@@ -221,3 +233,28 @@ func getFirstAddressOf(name string) (string, error) {
...
 	}
 }
+
+// TODO(paulbellamy): where should this live?
+func pluginsReporter(pluginRegistry *plugins.Registry) probe.Reporter {
+	return probe.ReporterFunc("plugins", func() (report.Report, error) {
+		rpt := report.MakeReport()
+		pluginRegistry.Implementors("reporter", func(plugin *plugins.Plugin) {
+			pluginReportJSON := map[string]interface{}{}
+			if err := plugin.Call("Report", nil, &pluginReportJSON); err != nil {
```
tomwilkie: You need some deadline on this, don't want a plugin to block the whole reporting cycle.

report/counters.go
```diff
@@ -132,9 +132,11 @@ func (c Counters) DeepEqual(d Counters) bool {
 
 func (c Counters) toIntermediate() map[string]int {
 	intermediate := map[string]int{}
-	c.psMap.ForEach(func(key string, val interface{}) {
-		intermediate[key] = val.(int)
-	})
+	if c.psMap != nil {
```
tomwilkie: nit, perhaps this would be cleaner:
```go
if c.psMap == nil {
  return map[string]int{}
}
intermediate := map[string]int{}
...
```

report/extra_metadata_templates.go
```diff
@@ -0,0 +1,33 @@
+package report
+
+type ExtraMetadataTemplates []MetadataRowTemplate
```
tomwilkie: Why "Extra"? Why not all?

Also, why a list and not a map from id -> template?

And merge does not seem entirely commutative.

tomwilkie:
First quick pass says its looking good.

I'm thinking `plugins/` should be moved into `probe/`, with some of the common datatypes moved into `report`.

prog/probe.go
paulbellamy: I wanted to keep the `plugins` package fairly separate from the concerns of scope, with an eye to extracting it into a separate package at some point.

prog/probe.go
paulbellamy: Yeah, will figure out a better way around this (that's the "make the code sane" bit on the todo list).

prog/probe.go
paulbellamy: good call

report/extra_metadata_templates.go
paulbellamy: > Why "Extra" ...

See "all templates (not just extra, and including metrics) need to come from the probe" todo note above.

> not a map

It's eventually going to be all templates, which means order will matter, unless we added a "priority" to the MetadataRowTemplate (could replace `Prime`, just have everything over priority X displayed by default).

> commutative

That concerns me too. In fact, the extra templates would be added in the order the plugins are called, which is potentially non-deterministic. I was debating adding a "sortable" priority column, which would let us (mostly?) make merge commutative. Any other alternatives?

paulbellamy:
> I'm thinking `plugins/` should be moved into `probe/`, with some of the common datatypes moved into report.

I'm not so sure on that yet. I tried it moving it into `probe/`, but it was a bit odd since it's required by `report` and `render` as well. By the time we do control and pipe plugins (and wire all that through the UI) it would probably have hooks used all over. We could split the common datatypes into `report`, but it is sort of nice to be able to re-use them between the registry and the rest of the code.

I had a vague eye to keeping the plugin registry and lifecycle-management code separate so it could be extracted into a separate library at some point.

Will see about refactoring it as it goes.

report/extra_metadata_templates.go
tomwilkie: > I was debating adding a "sortable" priority column

Yes, add a order field which defines how templates should be sorted.  Then use a map.

> commutative. Any other alternatives?

I don't think the commutative merge needs to be particularly useful on conflict, just well defined.  It might be enough to take the 'biggest' template on id conflict, where biggest is defined lexicographically on all the fields.

tomwilkie:
> We could split the common datatypes into report, but it is sort of nice to be able to re-use them between the registry and the rest of the code.

This is what I'd do - see controls; common stuff goes in common/xfer or report, probe specific stuff in probe and app specific stuff in app.

Whilst I don't feel particularly strongly about this, it would be nice if this was consistent across the whole code base, where 'vertical' features are split up like this.  It used to make a difference on build times, but now we just build on big binary.

probe/plugins/registry.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+const (
+	pluginTimeout = 500 * time.Millisecond
+	pluginRetry   = 5 * time.Second
+)
+
+// Registry maintains a list of available plugins by name.
+type Registry struct {
+	root              string
```
2opremio: rootPath? rootDirectory?

common/xfer/plugins.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/ugorji/go/codec"
+
+	"github.com/weaveworks/scope/test/reflect"
+)
+
+type PluginSpec struct {
+	ID          string   `json:"id"`
+	Label       string   `json:"label"`
+	Description string   `json:"description,omitempty"`
+	Interfaces  []string `json:"interfaces"`
```
2opremio: Can you document what Label and Interfaces are? It's not obvious.

common/xfer/plugins.go
```diff
@@ -0,0 +1,216 @@
+package xfer
```
2opremio: Shouldn't this go in report since it's part of it?

common/xfer/plugins.go
```diff
@@ -0,0 +1,216 @@
...
+	if nSize == 0 {
+		return other
+	}
+	if otherSize == 0 {
+		return n
+	}
+	result, iter := n.psMap, other.psMap
+	if nSize < otherSize {
+		result, iter = iter, result
+	}
```
2opremio: All the code above seems like an overkill early optimization.

examples/plugins/ebpf/http-parse-simple.c
```diff
@@ -0,0 +1,111 @@
...
+	struct ethernet_t *ethernet = cursor_advance(cursor, sizeof(*ethernet));
+	//filter IP packets (ethernet type = 0x0800)
+	if (!(ethernet->type == 0x0800)) {
+		goto DROP;	
+	}
+
+	struct ip_t *ip = cursor_advance(cursor, sizeof(*ip));
+	//filter TCP packets (ip next protocol = 0x06)
+	if (ip->nextp != IP_TCP) {
+		goto DROP;
```
2opremio: This won't work with Weave traffic :(

2opremio: (I am assuming this digests raw ethernet traffic as per the code above)

examples/plugins/ebpf/http-parse-simple.py
```diff
@@ -0,0 +1,108 @@
...
+
+#load eBPF program http_filter of type SOCKET_FILTER into the kernel eBPF vm
+#more info about eBPF program types
+#http://man7.org/linux/man-pages/man2/bpf.2.html
+function_http_filter = bpf.load_func("http_filter", BPF.SOCKET_FILTER)
+
+#create raw socket, bind it to eth0
+#attach bpf program to socket created
+# TODO: attach this to all interfaces
+BPF.attach_raw_socket(function_http_filter, "eth1")
```
2opremio: Is there a way to do this at a higher level? (i.e. not for a specific interface, without needing to go to layer 2)

examples/plugins/ebpf/main.go
```diff
@@ -0,0 +1,179 @@
+package main
```
2opremio: Wouldn't it make sense to write this directly in python? If we want _bcc_-people to write plugins it will make it more enticing.

examples/plugins/ebpf/main.go
```diff
@@ -0,0 +1,179 @@
...
+	sync.Mutex
+	HostID string
+	tuples []tuple
+}
+
+func (p *Plugin) Handshake(w http.ResponseWriter, r *http.Request) {
+	log.Printf("Probe %s handshake", r.FormValue("probe_id"))
+	err := json.NewEncoder(w).Encode(map[string]interface{}{
+		"name":        "ebpf",
+		"description": "Adds a graph of http requests/second to processes",
```
2opremio: Is this really true? It doesn't seem to know about PIDs.

examples/plugins/ebpf/main.go
paulbellamy: Probably yes. and would be a good example. I don't really know enough python to write it well.

examples/plugins/ebpf/main.go
paulbellamy: From the UI perspective, it adds this graph to processes. It adds the metrics onto the endpoints, which get rolled up into processes in the app, but it's not strictly true if a process has > 1 listening addr as the metrics won't be summed. Am working on this bit now, to see if there's a better way.

examples/plugins/ebpf/main.go
```diff
@@ -0,0 +1,179 @@
...
+		hostID = flag.String("hostname", hostname, "hostname of the host running this plugin")
+	)
+	flag.Parse()
+
+	log.Println("Starting...")
+
+	// Compile and run the ebpf script
+	ebpf := exec.Command("http-parse-simple.py")
+	ebpf.Stderr = os.Stderr
+	stdout, err := ebpf.StdoutPipe()
```
2opremio: I should have used this for the host ttys :) (used a buffer directly).

examples/plugins/ebpf/main.go
```diff
@@ -0,0 +1,179 @@
...
+	if err := ebpf.Start(); err != nil {
+		log.Fatal(err)
+	}
+	defer func() {
+		if ebpf.Process != nil {
+			ebpf.Process.Kill()
+		}
+	}()
+
+	os.Remove(*addr)
```
2opremio: What if another instance is already running?

2opremio: It would be helpful to have some helpers taking care of this stuff (socket creation registration, automatic handshake handle etc ..).

common/xfer/plugins.go
paulbellamy: Not sure. I moved it in here to match `common/xfer/controls.go`. It's also used in the registry and reporter. Slightly unclear where it should go.

examples/plugins/ebpf/main.go
2opremio: (i.e. it would be worth clarifying that it only works for containers because we match its IP)

common/xfer/plugins.go
paulbellamy: this datatype is a clone of `report.NodeSet`, so it got brought along with that. eugh.

examples/plugins/ebpf/http-parse-simple.c
paulbellamy: > this digests raw ethernet traffic

Correct.

Seems to work for weave traffic when it gets attached to the `weave` interface... hmm.

examples/plugins/ebpf/main.go
paulbellamy: > What if another instance is already running?

sneaky replacement happens. Convenient for dev, but probably best not left in.

> It would be helpful to have some helpers taking care of this stuff (socket creation registration, automatic handshake handle etc ..).

Yep. Pulling the socket management out into a couple helper functions would make it more obvious when/how you should do it.

examples/plugins/ebpf/main.go
paulbellamy: I'm not sure that's true, as we match on ip and port, which is process-unique.

examples/plugins/ebpf/main.go
```diff
@@ -0,0 +1,179 @@
...
+		p.tuples = p.tuples[1:]
+	}
+
+	result := make([]tuple, len(p.tuples))
+	copy(result, p.tuples)
+	p.Unlock()
+	return result
+}
+
+type tuple struct {
```
2opremio: Could use a more descriptive name :)

examples/plugins/ebpf/main.go
```diff
@@ -0,0 +1,179 @@
...
+		listener.Close()
+		os.Remove(*addr)
+	}()
+
+	log.Printf("Listening on: unix://%s", *addr)
+
+	plugin := &Plugin{HostID: *hostID}
+	go plugin.loop(stdout)
+	http.HandleFunc("/", plugin.Handshake)
+	http.HandleFunc("/report", plugin.Report)
```
2opremio: As we discussed offline, having the plugin expose a `/report` endpoint instead of doing it the other way around (the Scope probe exposing it) implies information needs to be polled. For instant updates and efficiently processing large amounts of data (i.e. large reports) you probably want a way for the plugin to convey an initial state and send deltas.

Considering the nature of reports and how we use them it seems fine to simply do polling but it's very inefficient for some cases. Here are two examples:

* TCP connection plugin (https://github.com/weaveworks/scope/issues/1168): you probably don't want to report all the connections all the time. It may be worth reporting the initial connections and then a stream of updates.
* kube-proxy status plugin (for eliding it https://github.com/weaveworks/scope/issues/538): you probably want to report the initial state of the proxy and them a stream of updates with the connections being created/destroyed.

I am thinking that leaking the report data structure (and implicitly the underlying semantics of its merge operations) in the plugins may not be a good idea. For one, it's not self contained (you need to understand what a report is and topologies are to fully understand it) plus it has the efficiency limitations I mention above. I don't have a good answer for how the plugin interface should be but we can work on it.

examples/plugins/ebpf/http-parse-simple.py
paulbellamy: Might be possible using different ebpf hooks. We could also probably do the counting in here (the in-kernel C), and expose a map of the counts up to the python program, which would be much faster, as wouldn't need to copy packets to user-space.

examples/plugins/ebpf/http-parse-simple.c
2opremio: Probably thanks to https://github.com/weaveworks/weave/issues/1577 . Sweet

examples/plugins/ebpf/main.go
2opremio: I may be wrong, but as far as I understood from @tomwilkie , when the PID is missing (i.e. conntrack) we only match ip:ports to containers and not individual processes (probably due to things like `SO_REUSEADDR`, NATing etc ...)

examples/plugins/ebpf/http-parse-simple.py
2opremio: Yes, something like https://github.com/iovisor/bcc/blob/39cc0ba649bc7c727a85bbb0d37afa2f40684ed3/tests/python/test_percpu.py#L25

I think it would also make things simpler.

examples/plugins/ebpf/main.go
2opremio: Also, how will we handle access to the socket directory (permissions etc ...)?  Will we simply trust plugins not to mess around with each other's sockets?

2opremio: It would probably good to get some input from @squaremo on this.

paulbellamy: It looks for sockets recursively, so you could do maybe do groups/permissions on folders within `/var/run/scope/plugins/`

examples/plugins/ebpf/main.go
paulbellamy: The pid is not necessarily missing, because it gets merged in from the probe.

examples/plugins/ebpf/main.go
squaremo: > It would probably good to get some input from @squaremo on this.

On anything specifically?

2opremio: (The reason I am mentioning @squaremo is that AFAIU he contributed to the definition of CNI plugins which use a similar mechanism).

examples/plugins/ebpf/http-parse-simple.py
paulbellamy: One problem with using different hooks (e.g. kprobes) is that they are not stable. http://manpages.ubuntu.com/manpages/wily/man8/kprobe.8.html#contenttoc10

examples/plugins/ebpf/main.go
squaremo: CNI just execs its plugins, rather than using any kind of socket arrangement. This works pretty well because plugins are invoked to perform operations, and state is out-of-band.

Docker plugins use HTTP over a socket to a long-running process. A downside of this is that someone (probably the user) now has to manage the lifecycle of the long-running process. But it seems appropriate enough if that process is collecting data that the system will occasionally request.

probe/plugins/registry.go
```diff
@@ -0,0 +1,271 @@
...
+		r.Close()
+		return nil, err
+	}
+	go r.loop()
+	return r, nil
+}
+
+// add recursively crawls the path provided, adding it to the watcher, and
+// looking for any existing sockets, loading them as plugins.
+func (r *Registry) addPath(ctx context.Context, path string) error {
```
2opremio: I find it confusing that the same function is used for adding the top directory and inner files generically, specially since the `remove` counterpart is only used for sockets. I think that specialized functions would be easier to read.

probe/plugins/registry.go
```diff
@@ -0,0 +1,271 @@
...
+	var statT syscall.Stat_t
+	if err := fs.Stat(path, &statT); err != nil {
```
2opremio: I am not sure how helpful this is due to the race with `Add` below.

probe/plugins/registry.go
```diff
@@ -0,0 +1,271 @@
...
+	switch statT.Mode & syscall.S_IFMT {
+	case syscall.S_IFDIR:
+		files, err := fs.ReadDir(path)
+		if err != nil {
+			log.Errorf("plugins: error loading path %s: %v", path, err)
+			return nil
+		}
+		for _, file := range files {
+			if err := r.addPath(ctx, filepath.Join(path, file.Name())); err != nil {
+				log.Errorf("plugins: error loading path %s: %v", filepath.Join(path, file.Name()), err)
```
2opremio: This should probably be a warning

probe/plugins/registry.go
```diff
@@ -0,0 +1,271 @@
...
+				return nil
```
2opremio: Why return early? You may miss out valid files.

examples/plugins/ebpf/main.go
2opremio: > CNI just execs its plugins, rather than using any kind of socket arrangement

I understood wrong then ... thanks a lot for your input :)

examples/plugins/ebpf/main.go
2opremio: > The pid is not necessarily missing, because it gets merged in from the probe.

Yes, I meant it doesn't come explicitly from the connection information.

common/xfer/plugins.go
2opremio: At least I would call it `pluginspecs` to make the separtion with `probe/plugins` more evident.

probe/plugins/registry.go
```diff
@@ -0,0 +1,271 @@
...
+
+type handshakeResponse struct {
+	Name        string   `json:"name"`
+	Description string   `json:"description,omitempty"`
+	Interfaces  []string `json:"interfaces"`
+	APIVersion  string   `json:"api_version,omitempty"`
+}
+
+// handshake periodically retries the handshake with this plugin until it succeeds.
+func (p *Plugin) handshake(ctx context.Context, expectedAPIVersion string, params url.Values, delay time.Duration) {
```
2opremio: I think we could reuse `common/backoff` for this.

probe/plugins/registry.go
paulbellamy: How's it a race? Either one will error or the other?

probe/plugins/registry.go
```diff
@@ -0,0 +1,271 @@
...
+	for _, plugin := range r.pluginsBySocket {
+		plugin.Close()
+	}
+	return r.watcher.Close()
+}
+
+type Plugin struct {
+	xfer.PluginSpec
+	context context.Context
+	socket  string
```
2opremio: I would call this `id` (since it's only used as a unique identifier) and only take the filename.

probe/plugins/registry.go
```diff
@@ -0,0 +1,271 @@
...
+func (r *Registry) loop() {
+	for {
+		select {
+		case <-r.done:
+			return
+		case evt := <-r.watcher.Events():
+			handlers := map[fsnotify.Op]func(context.Context, string) error{
+				fsnotify.Create: r.addPath,
+				fsnotify.Remove: r.removePath,
+				fsnotify.Chmod:  r.addPath,
```
2opremio: I think you also want to track `fsnotify.Rename` since the filename is used as the unique plugin identifier.

probe/plugins/registry.go
```diff
@@ -0,0 +1,271 @@
...
+			}
+			if handler, ok := handlers[evt.Op]; ok {
```
2opremio: This seems a bit convoluted to me (build a map to immediately access it and discard it). Why not a switch?

probe/plugins/registry.go
```diff
@@ -0,0 +1,271 @@
...
+				log.Errorf("plugins: event %v: no handler", evt)
+			}
+		case err := <-r.watcher.Errors():
+			log.Errorf("plugins: error: %v", err)
+		}
+	}
+}
+
+// ForEach walks through all the plugins running f for each one.
+func (r *Registry) ForEach(f func(p *Plugin)) {
```
2opremio: I would probably inline this since it's only used once (unless so clearly foresee more uses).

probe/plugins/registry.go
paulbellamy: We can't set recursive watches, and fsnotify just informs us of an event and path (not the file type), so we have to stat any newly created files, to tell *what* they are.

probe/plugins/registry.go
paulbellamy: good catch. we never actually return an error, so this code is dead.

probe/plugins/registry.go
paulbellamy: we have an `ID` as part of `PluginSpec`, which is provided by the plugin at handshake. We could pre-populate that with the `socket path`, until the handshake happens.

probe/plugins/registry.go
paulbellamy: We use it in tests also

probe/plugins/registry.go
2opremio: I am trying to say that the file can be removed between `Stat` and `Add` and thus the `Stat` is not helpful. Isn't the `Add` enough?

probe/plugins/registry.go
2opremio: I see, then it makes sense.

probe/plugins/registry.go
2opremio: And now I realize you need `Stat()` for obtaining the file type  ... sorry. It would probably be easier to read if you changed the order (assuming there is no good reason for `Stat()`ing first).

examples/plugins/ebpf/http-parse-simple.py
2opremio: There's ongoing work to allow using tracepoints which should be stable. See https://github.com/iovisor/bcc/issues/232 and https://github.com/iovisor/bcc/issues/419

2opremio: In fact it seems that a workaround was merged only two weeks ago https://github.com/iovisor/bcc/pull/439

examples/plugins/ebpf/main.go
paulbellamy: Sure. My point being that:

> we only match IPs to containers and not individual processes

is false.

davkal:
Moved the plugins to the bottom right, since they are merely information, no interaction.

screen shot 2016-04-06 at 09 37 58

paulbellamy:
@tomwilkie, @2opremio I removed the ebpf plugin from this until it's in a better state. @2opremio has been over this. PTAL and lets get this merged.

probe/plugins/reporter.go
```diff
@@ -0,0 +1,25 @@
+package plugins
```
2opremio: Maybe we could merge this with registry.go

probe/plugins/reporter.go
```diff
@@ -0,0 +1,25 @@
...
+func Reporter(pluginRegistry *Registry) probe.Reporter {
+	return probe.ReporterFunc("plugins", func() (report.Report, error) {
+		rpt := report.MakeReport()
+		pluginRegistry.Implementers("reporter", func(plugin *Plugin) {
+			pluginReport, err := plugin.Report()
+			if err != nil {
+				log.Errorf("plugins: error getting report from %s: %v", plugin.ID, err)
+				return
+			}
+			pluginReport.Plugins = pluginReport.Plugins.Add(plugin.PluginSpec)
```
2opremio: You probably want to add the plugin spec even if the plugin fails to report. Otherwise things may get more difficult to debug since, AFAIU, the plugin won't appear when querying the app's `/api` if it doesn't report anything in the last reporting window.

We may want to consider de-registering the plugin after certain amount of failures but I think that, while the plugin is registered, it should appear in `/api`. 

For debugging purposes, it may be worth adding a status field which can be be retrieved from `/api`.

probe/plugins/unix_round_tripper.go
```diff
@@ -0,0 +1,27 @@
+package plugins
+
+import (
+	"net"
+	"net/http"
+	"net/http/httputil"
+	"time"
+)
+
+type unixRoundTripper struct {
```
2opremio: Nice!

examples/plugins/iowait/main.go
```diff
@@ -0,0 +1,138 @@
...
+
+	log.Println("Starting...")
+
+	// Check we can get the iowait for the system
+	_, err := iowait()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	os.Remove(*addr)
```
2opremio: Same comment about providing a library to deal with registration and the handshake. I understand we may not want to do that now. In that case can you create an issue?

probe/plugins/registry.go
```diff
@@ -0,0 +1,293 @@
...
+		tr, err := transport(path, pluginTimeout)
+		if err != nil {
+			log.Warningf("plugins: error loading plugin %s: %v", path, err)
+			continue
+		}
+		client := &http.Client{Transport: tr, Timeout: pluginTimeout}
+		plugins[path] = NewPlugin(r.context, path, client, r.apiVersion, r.handshakeMetadata)
+	}
+	// remove plugins which weren't found
+	for path, plugin := range r.pluginsBySocket {
```
2opremio: Due to polling I don't think this will catch things such as a plugin removing the unix socket and re-recreating it between calls to `scan()`.

To solve this I think we should automatically de-register plugins when `GET /report` causes irrecoverable errors (e.g. connection errors).

examples/plugins/iowait/main.go
paulbellamy: I was expecting that most plugins would be not-in-go, so I figured an example and documentation might be more useful than a library, but I'll open an issue for it.

probe/plugins/registry.go
paulbellamy: a plugin could remove and re-create the socket without `GET /report` failing, so that doesn't really add anything.

examples/plugins/iowait/main.go
2opremio: You may be right, but who knows, Go is really strong in the ecosystem. Either way we should probably provide libraries in whatever languages are demanded.

probe/plugins/registry.go
paulbellamy: I found it quite convenient that it keeps retrying after failed `GET /report` requests when developing, as I could just restart the plugin and have test it out.

2opremio: > a plugin could remove and re-create the socket without GET /report failing, so that doesn't really add anything.

Oh then I probably misunderstand how the connections are used in the http client. Doesn't the reuse of the client across `GET /report`  invocations imply you are reusing the socket? https://github.com/weaveworks/scope/blob/plugins/probe/plugins/registry.go#L281

Maybe we can somehow control (forbid) invoking `GET /report` in a new connection before first running the handshake.

> I found it quite convenient that it keeps retrying after failed GET /report requests when developing, as I could just restart the plugin and have test it out.

This would still work due to subsequent scans. As it stands right now, that race results in a bug:

Following your example, imagine you decide to change your interface (in a future in which we support more than one) during development. You run your your plugin again, the socket is removed and Scope doesn't run a new handshake.

2opremio: > Maybe we can somehow control (forbid) invoking `GET /report` in a new socket before first running the handshake.

Can we perhaps keep track of the socket's inode?

probe/plugins/unix_round_tripper.go
```diff
@@ -0,0 +1,27 @@
...
+func makeUnixRoundTripper(address string, timeout time.Duration) (http.RoundTripper, error) {
+	return unixRoundTripper{address: address, timeout: timeout}, nil
+}
+
+func (t unixRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	conn, err := net.DialTimeout("unix", t.address, t.timeout)
+	if err != nil {
+		return nil, err
+	}
+	client := httputil.NewClientConn(conn, nil)
```
2opremio: I hadn't read this carefully. I think we could simply reuse the connection here, which would be more efficient and would let you detect removals of the unix socket without races.

probe/plugins/registry.go
paulbellamy: We do reuse the unix socket. but we open a new connection to it each request.

This was one advantage of the jsonrpc protocol, in that there was just one open (stateful) connection, and when the plugin went away it got terminated.

> inode

Maybe, yeah.

probe/plugins/unix_round_tripper.go
paulbellamy: As long as we (or the plugin) never does "Connection: Close" yeah, could do.

probe/plugins/registry.go
```diff
@@ -0,0 +1,293 @@
...
+
+func (p *Plugin) setStatus(err error) {
+	if err == nil {
+		p.Status = "ok"
+	} else {
+		p.Status = fmt.Sprintf("error: %v", err)
+	}
+}
+
+// TODO(paulbellamy): better error handling on wrong status codes
```
2opremio: I think it would be sane to start by enforcing a 200 status code. It's easier to relax restrictions in the future than the other way around.

2opremio:
I think that was all. Looking really good, thanks for your patience with all those comments @paulbellamy .

probe/plugins/unix_round_tripper.go
2opremio: Isn't that transparently handled? How does the default Transport/RoundTripper deal with this? They reuse the connections and need to handle `Connection: Close`

probe/plugins/registry.go
2opremio: > This was one advantage of the jsonrpc protocol, in that there was just one open (stateful) connection, and when the plugin went away it got terminated.

We could just use one connection as well (particularly since it's a local unix socket which should be robust) and assume that the plugin has gone away when the connection fails. A subsequent scan+handshake should be harmless anyhow.

The underlying _problem_ is that the communication is stateful due to the initial handshake. As we discussed offline, another option is to get rid of the handshake altogether and get plugins to provide context in each report response.

2opremio:
I have been working on the python ebpf plugin and, after experiencing the shittiness of python with threads, I am leaning towards Scope receiving `/report` requests in a global unix socket instead of plugins needing to serve it. 

Advantages:
* No polling, the plugin chooses when to push data.
* No request handling in the plugin: Implementing an http client is simpler than a server. Think about shell-scripts instance.
* Simpler data processing in the plugin. No need to have a background thread accumulating state and serving it whenever Scope needs it. Instead, just gather state and push it whenever necessary.

Disadvantages:
* Initialization is more complicated for the plugin (it needs to wait for the Scope plugin to be created).

For the initialization I think we should get rid of the handshake and provide context in each request, making report pushes effectively stateless.

probe/plugins/registry.go
```diff
@@ -0,0 +1,293 @@
...
+func NewPlugin(ctx context.Context, socket string, client *http.Client, expectedAPIVersion string, handshakeMetadata map[string]string) *Plugin {
+	params := url.Values{}
+	for k, v := range handshakeMetadata {
+		params.Add(k, v)
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	p := &Plugin{context: ctx, socket: socket, client: client, cancel: cancel}
+	f := p.handshake(ctx, expectedAPIVersion, params)
+	f() // try the first time synchronously
```
2opremio: Is this really necessary? I often encounter two handshake requests in the python plugin.

Also, please add error checking.

probe/plugins/registry.go
2opremio: I confirm this is needed right now for debugging plugins (my python plugin was originally failing with 404 errors and scope was logging cryptic "unexpected EOF" because it was trying to decode the body).

probe/plugins/registry.go
```diff
@@ -0,0 +1,293 @@
...
+// TODO(paulbellamy): better error handling on wrong status codes
+func (p *Plugin) get(path string, params url.Values, result interface{}) error {
+	ctx, cancel := context.WithTimeout(p.context, pluginTimeout)
+	defer cancel()
+	resp, err := ctxhttp.Get(ctx, p.client, fmt.Sprintf("unix://%s?%s", path, params.Encode()))
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	return codec.NewDecoder(resp.Body, &codec.JsonHandle{}).Decode(&result)
```
2opremio: To make plugins easier to debug it would be convenient to make decoding errors explicit (distinguish between connection and formatting errors, see my comment about "unexpected EOF" above).

```
err := NewDecoder(..).Decode()
if err != nil {
   err = fmt.Errorf("decoding error: %s", err)
}
return err
```

probe/plugins/unix_round_tripper.go
2opremio: After hitting my head against the wall for hours debugging a problem in the python plugin, it turns out that the `httputil` client was the culprit.

This commit removes it, having the nice side effect https://github.com/weaveworks/scope/commit/62d5b2530792e55a2afcddb611bbfe94479db0d3 of reusing connections and dealing with `Connection: close` transparently.

probe/plugins/registry.go
2opremio: https://github.com/weaveworks/scope/commit/62d5b2530792e55a2afcddb611bbfe94479db0d3  makes connections reusable, allowing us to automatically unregister the plugin when detecting irrecoverable errors such as the following (caused by stopping a plugin without removing its socket):

```
<probe> ERRO: 2016/04/10 02:14:03.190836 plugins: error getting report from http-requests: Get http://plugin/report: dial unix /var/run/scope/plugins/http_requests.sock: connect: connection refused
<probe> ERRO: 2016/04/10 02:14:04.191146 plugins: error getting report from http-requests: Get http://plugin/report: dial unix /var/run/scope/plugins/http_requests.sock: connect: connection refused
<probe> ERRO: 2016/04/10 02:14:05.204027 plugins: error getting report from http-requests: Get http://plugin/report: dial unix /var/run/scope/plugins/http_requests.sock: connect: no such file or directory
<probe> ERRO: 2016/04/10 02:14:06.203336 plugins: error getting report from http-requests: Get http://plugin/report: dial unix /var/run/scope/plugins/http_requests.sock: connect: no such file or directory
```

After deregistering it, scope will notice the file is still there and, try the handshake periodically.

probe/plugins/registry.go
```diff
@@ -0,0 +1,293 @@
...
+	if err == nil {
+		p.Status = "ok"
+	} else {
+		p.Status = fmt.Sprintf("error: %v", err)
+	}
+}
+
+// TODO(paulbellamy): better error handling on wrong status codes
+func (p *Plugin) get(path string, params url.Values, result interface{}) error {
+	ctx, cancel := context.WithTimeout(p.context, pluginTimeout)
```
2opremio: Why not simply use [`client.Timeout`](https://golang.org/pkg/net/http/#Client)

paulbellamy: The context also lets us manually cancel it in `Plugin.Close`

probe/plugins/registry.go
```diff
@@ -0,0 +1,293 @@
...
+	params := url.Values{}
+	for k, v := range handshakeMetadata {
+		params.Add(k, v)
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	p := &Plugin{context: ctx, socket: socket, client: client, cancel: cancel}
+	f := p.handshake(ctx, expectedAPIVersion, params)
+	f() // try the first time synchronously
+	p.backoff = backoff.New(f, "plugin handshake")
```
2opremio: I think that the `initialbackoff` and `maxbackoff` should be smaller than the default values (i.e. imagine a plugin developer waiting for 1 minute before his plugin is read again after an error).

probe/plugins/registry.go
2opremio: Actually, I now see you can cancel requests by using their `Cancel` channel https://golang.org/pkg/net/http/#Request . I guess that's the recommended way of doing it? Up to you, really.

probe/plugins/registry.go
```diff
@@ -0,0 +1,308 @@
...
+		params.Add(k, v)
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	p := &Plugin{context: ctx, socket: socket, client: client, cancel: cancel}
+	f := p.handshake(ctx, expectedAPIVersion, params)
+	// try the first time synchronously (makes testing easier)
+	if ok, err := f(); !ok {
+		log.Errorf("Error plugin handshake, backing off 10s: %v", err)
+		p.backoff = backoff.New(f, "plugin handshake")
```
2opremio: A previous comment got lost:

> The initialBackoff and maxBackoff should be smaller than the default values (e.g. imagine a plugin developer waiting for 1 minute for his plugin to reload.)

probe/plugins/registry.go
2opremio: This comment doesn't apply anymore, right? (You are now handling error)

probe/plugins/registry.go
```diff
@@ -0,0 +1,308 @@
...
+	defer cancel()
+	resp, err := ctxhttp.Get(ctx, p.client, fmt.Sprintf("unix://%s?%s", path, params.Encode()))
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("plugin returned non-200 status code: %s", resp.Status)
+	}
+	defer resp.Body.Close()
+	if err := codec.NewDecoder(resp.Body, &codec.JsonHandle{}).Decode(&result); err != nil {
```
2opremio: We should probably perform a maximum  `Content-length` check (something between 50-100MB?) before blinding parsing the response to prevent plugins from unintentionally DoS'ing the probe.

2opremio: (kudos to @peterbourgon for suggesting this on Slack :) )

peterbourgon:
```go
body := io.LimitReader(resp.Body, 50*1024*1024)
```

probe/plugins/registry.go
```diff
@@ -222,9 +226,17 @@ func NewPlugin(ctx context.Context, socket string, client *http.Client, expected
 	ctx, cancel := context.WithCancel(ctx)
 	p := &Plugin{context: ctx, socket: socket, client: client, cancel: cancel}
 	f := p.handshake(ctx, expectedAPIVersion, params)
-	f() // try the first time synchronously
-	p.backoff = backoff.New(f, "plugin handshake")
-	go p.backoff.Start()
+	// try the first time synchronously (makes testing easier)
+	if ok, err := f(); !ok {
+		log.Errorf("Error plugin handshake, backing off 10s: %v", err)
```
2opremio: It's not 10s anymore :)

probe/plugins/registry.go
```diff
@@ -0,0 +1,313 @@
...
+	for k, v := range handshakeMetadata {
+		params.Add(k, v)
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+	p := &Plugin{context: ctx, socket: socket, client: client, cancel: cancel}
+	f := p.handshake(ctx, expectedAPIVersion, params)
+	// try the first time synchronously (makes testing easier)
+	if ok, err := f(); !ok {
+		log.Errorf("Error plugin handshake, backing off 10s: %v", err)
```
2opremio: It's not 10s anymore

2opremio:
@davkal Can we get some visual feedback when the plugin status is not `"ok"`? (maybe get it breathing with the error displayed on hover-on).

probe/plugins/registry.go
```diff
@@ -0,0 +1,313 @@
...
+	defer cancel()
+	resp, err := ctxhttp.Get(ctx, p.client, fmt.Sprintf("http://plugin%s?%s", path, params.Encode()))
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("plugin returned non-200 status code: %s", resp.Status)
+	}
+	defer resp.Body.Close()
+	if err := codec.NewDecoder(io.LimitReader(resp.Body, 50*1024*1024), &codec.JsonHandle{}).Decode(&result); err != nil {
```
2opremio: I don't think it's a good idea to use  `io.LimitReader`, since it will return `EOF` when reaching the limit, resulting in a cryptic decoding error (sorry I didn't see the suggestion from @peterbourgon earlier). I would check the header.

paulbellamy: Header could be chunked encoding, or just "Connection: close", where no length is needed.

In a previous project I had to solve this by implementing a custom version of LimitReader, which returned a different error when it reached the max. Could do that here.

paulbellamy: Note, that there is a `net/http.MaxBytesReader`, but it is deeply coupled to `net/http.Server` internals. (woooooh!)

paulbellamy: Let's defer this error message being confusing to a separate issue to be solved later.

paulbellamy:
Raised separate issue for indicating broken plugins in the UI.

examples/plugins/http-requests/http-requests.py
```diff
@@ -120,21 +118,21 @@ def do_report(self):
...
-    def do_handshake(self):
-        spec = {
-            'name': 'http-requests',
-            'description': 'Adds http request metrics to processes',
-            'interfaces': ['reporter'],
-            'api_version': '1',
-        }
-        self.respond(json.dumps(spec))
-
     def respond(self, body):
```
2opremio: We should inline this (and maybe simplify the request_log handling since there's only one request type)

examples/plugins/iowait/Makefile
```diff
@@ -0,0 +1,19 @@
+.PHONY: run clean
+
+EXE=iowait
+IMAGE=weavescope-iowait-plugin
+UPTODATE=.$(EXE).uptodate
+
+run: $(UPTODATE)
+	docker run --rm -it --privileged -v /var/run/scope/plugins:/var/run/scope/plugins --name $(IMAGE) $(IMAGE) -hostname=$(shell hostname)
```
2opremio: `/var/run/scope/plugins` won't exist if Scope is not running. I am not sure we want to do something about this. The http-requests plugin has this problem too.

probe/plugins/registry.go
2opremio: Sure, can you create an issue?

examples/plugins/iowait/Makefile
paulbellamy: docker makes it!

probe/plugins/registry.go
paulbellamy: done

examples/plugins/iowait/Makefile
2opremio: Didn't know ...

2opremio:
LGTM, (I would like to have #1271 for the release though)

examples/plugins/README.md
```diff
@@ -0,0 +1,61 @@
...
+can be viewed at the `/api/report` endpoint of any scope app.
+
+In addition to any data about the topology nodes, the report returned
+from the plugin must include some information about the plugin.
+
+For example:
+
+```json
+{
+  "Processes: { ... }
```
tomwilkie: COMMA!

examples/plugins/iowait/main.go
```diff
@@ -0,0 +1,140 @@
...
+		return
+	}
+	err = json.NewEncoder(w).Encode(map[string]interface{}{
+		"Host": map[string]interface{}{
+			"nodes": map[string]interface{}{
+				p.HostID + ";<host>": map[string]interface{}{
+					"metrics": map[string]interface{}{
+						"iowait": map[string]interface{}{
+							"samples": []interface{}{
+								map[string]interface{}{
```
tomwilkie: Man, this is ugly.

paulbellamy: yes

2opremio:
Awesome job @paulbellamy! \o/

2opremio:
@alepuccetti I think you are referring to https://github.com/weaveworks/scope/pull/1837 , not this PR

PR #1138, opened 2016-03-07
Slightly thinner borders on nodes

davkal:
* epilogue to #1080
* slightly thinner borders are more in line with, ehem, other lines


foot:
Nice! LGTM

PR #1159, opened 2016-03-11
Topology Filters

### LINKED ISSUE #1116: stopped/system containers hard to spot in sea of containers
at the moment one can choose between
- showing system container or not
- showing stopped containers or not

However, if you have loads of running/non-system containers, then if you want to look at stopped/system containers they get lost in the sea of containers.

Suggestion: make the switches three-way:
- system - both - user (defaulting to 'user')
- stopped - both - running (defaulting to 'running')

davkal:
Fixes #1116 

* topology filters overhaul
* allow SHOW X, SHOW Y (which is NOT X), SHOW BOTH

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -21,14 +21,60 @@ var (
 )
 
 func init() {
-	containerFilters := map[string][]APITopologyOption{
-		"system": {
-			{"show", "System containers shown", false, render.FilterNoop},
-			{"hide", "System containers hidden", true, render.FilterSystem},
+	serviceFilters := []APITopologyOptionGroup{
+		{
+			ID:      "system",
+			Default: "application",
+			Options: []APITopologyOption{
+				{"system", "System services", render.FilterApplication},
+				{"application", "Application services", render.FilterSystem},
+				{"both", "Both", render.FilterNoop},
+			},
 		},
-		"stopped": {
-			{"show", "Stopped containers shown", false, render.FilterNoop},
-			{"hide", "Stopped containers hidden", true, render.FilterStopped},
+	}
+
+	podFilters := []APITopologyOptionGroup{
+		{
+			ID:      "system",
+			Default: "application",
+			Options: []APITopologyOption{
+				{"system", "System pods", render.FilterApplication},
+				{"application", "Application pods", render.FilterSystem},
+				{"both", "Both", render.FilterNoop},
+			},
+		},
+	}
+
+	containerFilters := []APITopologyOptionGroup{
+		{
+			ID:      "system",
+			Default: "application",
+			Options: []APITopologyOption{
+				{"system", "System containers", render.FilterApplication},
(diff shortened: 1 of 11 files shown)


client/app/scripts/stores/app-store.js
```diff
@@ -80,13 +80,11 @@ function setTopology(topologyId) {
...
   topologyList.forEach(topology => {
     let defaultOptions = makeOrderedMap();
-    if (topology.has('options')) {
-      topology.get('options').forEach((items, option) => {
-        items.forEach(item => {
-          if (item.get('default') === true) {
-            defaultOptions = defaultOptions.set(option, item.get('value'));
-          }
-        });
+    if (topology.has('options') && topology.get('options')) {
```
foot: simplified to just `topology.get('options')`

foot:
I like the aesthetic!

- Preloading positions by selecting app-containers then system-containers, and _then_ both causes bad node overlap sometimes, can be force-layed-out though. I'm not sure if there's an easy fix here? I guess you could include filtered nodes in the layout and then hide, though this could lead to sub-optimal layouts...
- Maybe give the unselected items a bg fill as they are already at 0.7opacity, looks quite nice. Otherwise a bit messy in the rare case of node/control overlap down there.
- I quite like having the two bars the same length, but it sounds like that option was tried and outvoted. :)
- Even when "both" is selected it says some nodes are still filtered, this is not a regression and i think i'm just noticing as i'm looking for little things, but maybe a tooltip? I don't even know what they are.

app/api_topologies.go
```diff
@@ -24,33 +24,69 @@ var (
...
-				{"hide", "System pods hidden", true, render.FilterSystem},
-			}},
+			Options: []APITopologyOptionGroup{
+				{
+					ID:      "system",
+					Default: "application",
+					Options: []APITopologyOption{
+						{"system", "System pods", render.FilterApplication},
+						{"application", "Application pods", render.FilterSystem},
+						{"both", "Both", render.FilterNoop},
```
tomwilkie: In https://github.com/weaveworks/scope/issues/1116#issuecomment-195478078, @rade seems to be implying that both should be in the middle.

rade: That very much depends on how this gets rendered.

tomwilkie: They get rendered in that order; @davkal should be able to provide some screenshots.

davkal:
Current implementation:

screen shot 2016-03-24 at 14 11 28

*Both* in the middle:
screen shot 2016-03-24 at 14 12 31

davkal:
@foot changed BG of unselected ones slightly and isolated topology caches, so nodes should overlap less PTAL

foot:
Topos look good!

Not sure about the BG! Looks a bit heavy now, I'd remove that actually :P.

But LGTM either way. :+1:

davkal:
Fixes #1116

PR #1161, opened 2016-03-11
strip sha256: from docker image ids

### LINKED ISSUE #1058: Docker changed the format of image IDs
screen shot 2016-02-29 at 12 04 57

Inspecting it still works, but the truncation is a bit short now.

paulbellamy:
Fixes #1058


davkal:
I'd extract the `"sha256:"` string. Otherwise code LGTM.

PR #1165, opened 2016-03-14
only check the first two segments of the kernel version

### LINKED ISSUE #1136: Error when parsing kernel version in proc background reader
```
<probe> ERRO: 2016/03/07 12:20:57.204694 getNamespacePathSuffix: cannot get kernel version: Malformed version: 4.3.5-300.fc23.x86_64
```

Full logs https://gist.github.com/janwillies/54083a50358718a4fb21

paulbellamy:
fixes #1136

@2opremio ptal


2opremio:
If we are going partly going to parse things manually, then I suggest we just ditch the `version` dependency and do something like https://github.com/ugorji/go/pull/148/files#diff-aec251e204a9b9ef2196c0937f1d8e03R1634

paulbellamy:
Good call.

2opremio:
very minor comments, otherwise LGTM.

PR #1167, opened 2016-03-14
Make app-store's topologies object immutable

davkal:
* refactoring, functionality should be the same
* extracted some functions into `utils/topology-utils.js`
* adapted tests


client/app/scripts/components/topologies.js
```diff
@@ -25,45 +24,41 @@ export default class Topologies extends React.Component {
...
         </div>
       </div>
     );
   }
 
   renderTitle(topology) {
-    return ['Nodes: ' + topology.stats.node_count,
-      'Connections: ' + topology.stats.node_count].join('\n');
+    return ['Nodes: ' + topology.get('stats').get('node_count'),
+      'Connections: ' + topology.get('stats').get('node_count')].join('\n');
```
foot: `getIn` baby

client/app/scripts/utils/topology-utils.js
```diff
@@ -0,0 +1,56 @@
...
+/* set topology.id in place on each topology */
+export function updateTopologyIds(topologies) {
+  topologies.forEach(topology => {
+    topology.id = topology.url.split('/').pop();
+    if (topology.sub_topologies) {
+      updateTopologyIds(topology.sub_topologies);
+    }
+  });
+  return topologies;
+}
```
foot: We could do a `fromJS` from anything that comes out of the network and make all these fns immutable too, for consistency!

foot:
Looks good TM

PR #1169, opened 2016-03-15
More complete node summary

paulbellamy:
Base work for #1105 and #1159

Could use more refactoring eventually, as the RenderableNode json rendering is never really used after this. We either render the NodeSummary or the detailed.Node. Might also be worth splitting `detailed` and `summary` into different packages, but that will conflict with changes in #1126.

With this change the topology views (and deltas) include the nodes' metadata and metrics (sans samples). e.g.
```js
{ nodes: [
  process:scope:1716: {
    id: "process:scope:1716",
    label: "/home/weave/weaver",
    label_minor: "scope (weave:1716)",
    rank: "/home/weave/weaver",
    shape: "square",
    linkable: true,
    metadata: [
      {
        id: "threads",
        label: "# Threads",
        value: "12",
        prime: true
      },
      ...
    ],
    metrics: [
      {
        id: "process_memory_usage_bytes",
        label: "Memory",
        format: "filesize",
        value: 46587904,
        samples: [ ],
        min: 0,
        max: 46587904,
        first: "2016-03-15T12:00:58.106201326Z",
        last: "2016-03-15T12:01:12.065987896Z"
      },
      ...
    ]
  }
]}
```

@tomwilkie, wdyt?

### DIFF
--- a/app/api_topology.go
+++ b/app/api_topology.go
@@ -104,7 +104,7 @@ func handleWebsocket(
 	}(conn)
 
 	var (
-		previousTopo render.RenderableNodes
+		previousTopo detailed.NodeSummaries
 		tick         = time.Tick(loop)
 		wait         = make(chan struct{}, 1)
 	)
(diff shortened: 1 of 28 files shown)


tomwilkie:
Superceeded by #1204

PR #1171, opened 2016-03-15
React/lodash/babel upgrades + updated linting (linted)

davkal:
* important upgrades: react to 0.14.7, react-motion 0.4, jest 0.9, lodash 4.6, babel 6.5
* linter rules upgraded, linted JS
* fixes edge highlighting on node hover


foot:
Nice one! All seems to be working!

It feels like we've introduced a bit of a perf hit. E.g. when selecting a node the animation is suffering a bit there. And dragging the canvas around a little. Perhaps the Obj.assigns? But we can work through that along w/ the pure mixins.

LGTM

PR #1173, opened 2016-03-16
Make node/edge highlighter objects immutable in app store

davkal:
* refactor
* fixed reverse edge highlighting on node hover

### DIFF
--- a/client/app/scripts/stores/app-store.js
+++ b/client/app/scripts/stores/app-store.js
@@ -35,12 +35,13 @@ function makeNode(node) {
 // Initial values
 
 let topologyOptions = makeOrderedMap(); // topologyId -> options
-let adjacentNodes = makeSet();
 let controlStatus = makeMap();
 let currentTopology = null;
 let currentTopologyId = 'containers';
 let errorUrl = null;
 let forceRelayout = false;
+let highlightedEdgeIds = makeSet();
+let highlightedNodeIds = makeSet();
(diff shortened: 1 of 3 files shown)


client/app/scripts/stores/app-store.js
```diff
@@ -433,22 +412,53 @@ export class AppStore extends Store {
...
-        mouseOverNodeId = payload.nodeId;
+        const nodeId = payload.nodeId;
+
+        // clear old highlights
+        highlightedNodeIds = highlightedNodeIds.clear();
+        highlightedEdgeIds = highlightedEdgeIds.clear();
+
+        // highlight nodes
+        highlightedNodeIds = highlightedNodeIds.add(nodeId);
+        const adjacencNodes = this.getAdjacentNodes(nodeId);
```
foot: `adjacentNodes`

foot:
Edges not going both ways but not sure if this is a regression

botways

foot:
but other than that LGTM.

davkal:
> Edges not going both ways but not sure if this is a regression

Not a regression. In this scenario `192...` has `scope-app` in its adjacency list, but not the other way around. To fix this for a given `nodeA` , we need to check the adjacency list of all nodes, to see if `nodeA` is in any of their adjacency lists. This is already done a few lines above, so I can reuse that result.

foot:
noice, lgtm

Issue #720, opened 2015-12-02
Infer Weave Net underlay topology and visualise with mesh status


awh:
Whilst investigating weaveworks/weave#1744 I produced the following diagram
![img_20151202_124014](https://cloud.githubusercontent.com/assets/1504438/11531271/8f6b17de-98f3-11e5-8969-409c166cf321.jpg)
from the output of `weave status peers` (copied here in full in case the gist disappears):

```
➜  ~  weave status

       Version: 1.3.1

       Service: router
      Protocol: weave 1..2
          Name: 66:24:da:cd:5e:9e(ubuntu)
    Encryption: disabled
 PeerDiscovery: enabled
       Targets: 2
   Connections: 10 (9 established, 1 retrying)
         Peers: 11 (with 92 established, 5 pending connections)

       Service: ipam
     Consensus: achieved
         Range: 172.16.0.0-172.16.255.255
 DefaultSubnet: 172.16.0.0/16

       Service: dns
        Domain: weave.local.
           TTL: 1
       Entries: 22

       Service: proxy
       Address: unix:///var/run/weave/weave.sock

➜  ~  weave status peers
6a:df:8b:be:fc:8e(us4)
   <- 118.70.80.24:3137     1a:16:95:88:47:58(jake-pc)       established
   <- 117.6.161.201:36481   f6:26:fb:b3:70:01(hht3)          established
   <- 118.70.80.24:24993    66:24:da:cd:5e:9e(ubuntu)        established
   <- 142.0.199.28:33132    66:73:f8:3f:fc:6e(us2)           established
   <- 142.0.199.36:54202    aa:4c:64:2e:4a:bf(us3)           established
   <- 117.6.161.201:50337   3e:98:23:98:2c:38(hht2)          established
   <- 117.6.161.201:36417   86:2f:e8:fa:fe:5a(hht6)          established
   <- 117.6.161.201:50433   76:0e:93:15:47:9d(hht1)          established
   <- 80.8.104.154:39525    9e:b6:00:c6:7c:17(androgeek-SATELLITE-L50-C) pending
   -> 142.0.199.20:6783     02:ec:03:f7:ae:45(us1)           established
86:2f:e8:fa:fe:5a(hht6)
   -> 192.168.100.109:6783  76:0e:93:15:47:9d(hht1)          established
   -> 142.0.199.20:6783     02:ec:03:f7:ae:45(us1)           established
[...]

PR #1182, opened 2016-03-17
Add the basics for a Weave-specific topology.

### LINKED ISSUE #778: Make it more obvious when there are hosts on the weave network that aren't running scope
We already know the list of weave peers, so we could easily put pseudo nodes in the hosts view for hosts that aren't running scope.

### LINKED ISSUE #1132: inclusion of overlay connections in Hosts view is confusing
Say I have three hosts which can only connect in a "chain" like...
screenshot from 2016-03-05 17 44 14

But since I am running weave, the connectivity between containers does not have to follow this constrained topology, and instead might look like...
screenshot from 2016-03-05 17 39 54

But when you do this, the Hosts view changes to...
screenshot from 2016-03-05 17 45 08

which is rather surprising and IMO confusing since the connectivity between hosts hasn't actually changed.

tomwilkie:
This is very much an early prototype.  Fixes #1132.

screen shot 2016-03-17 at 16 03 18

It uses the information from `weave report` to build a topology of nodes and edges.  It honours the edge direction from `weave report`.  Peers from `weave report` which are not running scope are shown as pseudo nodes (fixes #778).

I'd like to be able to:
- [ ] get metrics about traffic on various connections between weave routers (@awh, @rade, @bboreham?)
- [ ] show more useful info in the details panel (like what?)
- [x] ~~ensure the weave topology is shown as selected in the UI (@davkals, could you take a look?)~~ (this was a non-issue when @2opremio revived the PR)
- [x] ~~rename occurrences of overlay to weave~~ (not applicable anymore since the overlay topology includes docker nodes)
- [x] make it obvious when weave is installed on a host and scope is not (done through pseudo nodes).
- [ ] controls (forget, connect, rmpeer?)

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -147,6 +147,13 @@ func init() {
 			Name:     "Hosts",
 			Rank:     4,
 		},
+		APITopologyDesc{
+			id:       "weave",
+			parent:   "hosts",
+			renderer: render.WeaveRenderer,
+			Name:     "Weave",
+			Rank:     3,
+		},
 	)
 }
 
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -152,7 +152,6 @@ func init() {
 			parent:   "hosts",
 			renderer: render.WeaveRenderer,
 			Name:     "Weave Net",
-			Rank:     3,
 		},
 	)
 }
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -151,7 +151,7 @@ func init() {
 			id:       "weave",
 			parent:   "hosts",
 			renderer: render.WeaveRenderer,
-			Name:     "Weave",
+			Name:     "Weave Net",
 			Rank:     3,
 		},
 	)
--- a/app/api_topology.go
+++ b/app/api_topology.go
@@ -84,7 +84,7 @@ func handleWebsocket(
 		for { // just discard everything the browser sends
 			if _, _, err := c.ReadMessage(); err != nil {
 				if !xfer.IsExpectedWSCloseError(err) {
-					log.Println("err:", err)
+					log.Error("err:", err)
 				}
 				close(quit)
 				break
--- a/common/weave/client.go
+++ b/common/weave/client.go
@@ -34,8 +34,15 @@ type Status struct {
 type Router struct {
 	Name  string
 	Peers []struct {
-		Name     string
-		NickName string
+		Name        string
+		NickName    string
+		Connections []struct {
+			Name        string
+			NickName    string
+			Address     string
+			Outbound    bool
+			Established bool
+		}
 	}
 }
 
--- a/common/weave/client.go
+++ b/common/weave/client.go
@@ -25,9 +25,10 @@ type Client interface {
 
 // Status describes whats happen in the Weave Net router.
 type Status struct {
-	Router Router
-	DNS    DNS
-	IPAM   IPAM
+	Version string
+	Router  Router
+	DNS     DNS
+	IPAM    IPAM
 }
 
 // Router describes the status of the Weave Router
--- a/common/weave/client_test.go
+++ b/common/weave/client_test.go
@@ -67,6 +67,14 @@ func TestStatus(t *testing.T) {
 			Peers: []struct {
 				Name     string
 				NickName string
+				// TODO, extend
+				Connections []struct {
+					Name        string
+					NickName    string
+					Address     string
+					Outbound    bool
+					Established bool
+				}
 			}{
 				{
 					Name:     mockWeavePeerName,
--- a/common/weave/client_test.go
+++ b/common/weave/client_test.go
@@ -65,9 +65,8 @@ func TestStatus(t *testing.T) {
 	want := weave.Status{
 		Router: weave.Router{
 			Peers: []struct {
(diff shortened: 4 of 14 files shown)


rade:
>  Fixes #1132.

In #1132 I wasn't complaining about the lack of a weave view, but that the hosts view shows overlay topology rather than underlay topology.

rade:
#720 has a few ideas for what scope could show/visualise about weave net.

pidster:
Why do we want a Weave specific view?

rade:
> Why do we want a Weave specific view?

#720 provides a concrete use case.

However, I don't think providing a weave specific view should be conflated with making the existing Hosts view show underlay connections only, which is what #1132 - the apparent motivation for this PR - is about.

rade:
btw, I reckon showing the weave topology, which is what the bulk of this PR appears to do, is a reasonable first step on the way to #720.

2opremio:
Addresses https://github.com/weaveworks/scope/issues/1938

2opremio:
@tomwilkie I don't think we should be renaming the Overlay topology anymore since:

1. It contains non-weave-specific information ( https://github.com/weaveworks/scope/blob/07744961024dfa0f74a0bb07551c3aed01ac6401/probe/docker/reporter.go#L150 , which was added after this PR in https://github.com/weaveworks/scope/pull/1584 and released )
2. Backwards compatibility. The overlay topology has been providing useful info for a while (like the weave IPs), for instance (1) was released on 0.16 and there will be probes using it.

2opremio:
I've reworked the code without the renaming and applied it to current master (I saved the old code in branch [1132-weave-topology-original-tom](https://github.com/weaveworks/scope/tree/1132-weave-topology-original-tom) just in case)

bboreham:
Re controls: 
`connect` would be cool - takes an address (e.g. an IP address or DNS name).  Somewhat rare to use like this, I think, but nice for a demo.
`forget` needs to be applied to every peer in the system, so perhaps more work than it's worth.
`rmpeer` would be valid, if you can scrape the candidates from the `weave status ipam` data
`reset`, i.e. "shut down this Weave Net peer and remove it from the network" also conceptually valid, perhaps useful if you are migrating to a different machine

bboreham:
Re the details panel:
Subsetted from `weave status`:
````
        Version: git-947ef8bb7770 (up to date; next check at 2016/10/26 16:15:53)

        Service: router
       Protocol: weave 1..2
           Name: 9a:90:6c:4a:40:07(dev)
     Encryption: disabled
  PeerDiscovery: enabled
 TrustedSubnets: none

        Service: ipam
         Status: ready
          Range: 10.32.0.0/12
  DefaultSubnet: 10.32.0.0/12

        Service: dns
         Domain: weave.local.
       Upstream: 10.0.2.3
            TTL: 1
        Entries: 1

        Service: plugin
     DriverName: weave
````

then add `weave status connections` info, perhaps above all of that except the peer name:

````
-> 192.168.48.11:6783    established sleeve 46:98:1c:22:6d:b7(host1)
-> 192.168.48.12:6783    established sleeve 5a:e6:43:04:ae:40(host2)
-> 127.0.0.1:6783        failed      cannot connect to ourself, retry: never
````

The interesting parts are sleeve vs fastdp vs awsvpc, encrypted or not, error message.

**Comment by rade** [comment]
We should merge this w/o adding more stuff to the details panel. Let's do that in a separate issue/PR.

**Comment by 2opremio** [comment]
> Re controls: 

How about expose?

**Comment by 2opremio** [comment]
> We should merge this w/o adding more stuff to the details panel. Let's do that in a separate issue/PR.

Good point. Will do

**Comment by 2opremio** [comment]
Current state of the PR:

screen shot 2016-10-26 at 14 04 19

screen shot 2016-10-26 at 14 04 34

Report: [report.json.gz](https://github.com/weaveworks/scope/files/553178/report.json.gz)

(CC @davkal )

**Comment by pidster** [comment]
Q: Should we have a different icon for Weave routers?

**Comment by 2opremio** [comment]
> Q: Should we have a different icon for Weave routers? (Or, as they are containers, should the hexagon be used?)

We chose the circle because:

* Each node represents a host (they are not really containers, since they don't provide container information and each router is mapped to a host). 
* In fact, the topology is a child of the host topology (see were the "WEAVE NET" button is located)

**Comment by 2opremio** [comment]
That said, maybe we could use a different icon, I wouldn't recommend an hexagon though since it will be confused with containers.

**Comment by mjlodge** [comment]
Circles are good because they're hosts...

File: app/api_topologies.go [hunk 1]
```diff
@@ -147,6 +147,13 @@ func init() {
 			Name:     "Hosts",
 			Rank:     4,
 		},
+		APITopologyDesc{
+			id:       "weave",
+			parent:   "hosts",
+			renderer: render.WeaveRenderer,
+			Name:     "Weave Net",
+			Rank:     3,
```
paulbellamy: How did you pick this rank?

paulbellamy: I ask, because it is the same as Pods, but other sub-topologies don't seem to have a Rank at all.

2opremio: I blindly copied it from @tomwilkie's original PR. I removed the rank and works just fine.

PR #1184, opened 2016-03-17
Fix bug in container image reporting, we can't strip the prefix off until after finding the image

paulbellamy:
@tom, ptal

Bug was introduced in https://github.com/weaveworks/scope/commit/1f6652e1c3f1d10bfe09ef7226771ae57b1a1434

### DIFF
--- a/integration/115_topologies_test.sh
+++ b/integration/115_topologies_test.sh
@@ -1,0 +1,36 @@
+#! /bin/bash
+
+. ./config.sh
+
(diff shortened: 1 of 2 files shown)


tomwilkie:
Probably a separate issue, but the "tag" in the details panel includes the image version, whereas nothing else does:

screen shot 2016-03-22 at 11 53 23

tomwilkie:
Can we instead put a `trimImageID` in registry.go:233?

PR #1186, opened 2016-03-18
Performance improvements for canvas

### LINKED ISSUE #1034: Increase 100-node limit
screen shot 2016-02-25 at 14 16 39

Report: [service_report.json.gz](https://github.com/weaveworks/scope/files/146491/service_report.json.gz)

davkal:
Fixes #1034
* made some components stateless
* added pure-render mixin to pure components
* split up node chart components
* removed dynamic font sizing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ CODECGEN_EXE=$(CODECGEN_DIR)/bin/codecgen_$(shell go env GOHOSTOS)_$(shell go en
 GET_CODECGEN_DEPS=$(shell find $(1) -maxdepth 1 -type f -name '*.go' -not -name '*_test.go' -not -name '*.codecgen.go' -not -name '*.generated.go')
 CODECGEN_TARGETS=report/report.codecgen.go render/render.codecgen.go render/detailed/detailed.codecgen.go
 RM=--rm
-RUN_FLAGS=-ti
+RUN_FLAGS=-i
 BUILD_IN_CONTAINER=true
 GO_ENV=GOGC=off
 GO=env $(GO_ENV) go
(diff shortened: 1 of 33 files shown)


foot:
Wicked! So Snapppppy w/ lower node counts now.

- If you select a node, then click bg to deselect it often fails to return to the canvas due to the violation of the assumption we have max of 30 points on a line (I've seen 45 points and 61 points on a 500node layout). It would be neat to make this dynamic but I dunno how...
- Topo layout animations at high node counts are a bit janky and painful to sit through, especially at the end when they appear to have finished but are still churning tiny distances. Maybe disabling them at some point might give a nicer ux. The highlighting animations behave alright even at high node counts though.
- 998 new nodes took too long to layout on the highly connected data the debugbar generates. (2-3mins then hit pause and saw it stuck in the layout code). 500nodes took about 20s to layout. Maybe set the limit to 300-500 for now.

davkal:
> If you select a node, then click bg to deselect it often fails to return to the canvas due to the violation.. 

https://github.com/weaveworks/scope/issues/1187

> Topo layout animations at high node counts are a bit janky and painful to sit through...

Will add dynamic coords precision based on topo size in this PR.

> 998 new nodes took too long to layout on the highly connected data...

https://github.com/weaveworks/scope/issues/1188

davkal:
Found that edge points are not immutable yet. working on that in this PR.

davkal:
@foot I added the precision optimization, as well as making edge points immutable. PTAL

foot:
There might be a bit of a perf regression in the latest HEAD, dragging around the canvas is quite slow, I think it should be able to get away w/ only updating the svg element.

PR #1198, opened 2016-03-23
Fixes the terminal loading when the node doesn't have a rank.

### LINKED ISSUE #1195: UI barfs when opening a terminal on a node with a blank rank.
Ranks were blank due to missing container images (#1184). 

Error I get in the UI is `Uncaught TypeError: Cannot read property 'substr' of undefined`

After hitting this, the UI stops working altogether.

This can be reproduce by trying to open a terminal on 1f6652e1c3f1d10bfe09ef7226771ae57b1a1434.

foot:
- Safen up the color lookup, so we return a nice brown if we don't have
  anything else to work with.

Fixes #1195

### DIFF
--- a/client/app/scripts/utils/color-utils.js
+++ b/client/app/scripts/utils/color-utils.js
@@ -43,7 +43,7 @@ export function getNeutralColor() {
   return PSEUDO_COLOR;
 }
 
-export function getNodeColor(text, secondText, isPseudo = false) {
+export function getNodeColor(text = '', secondText = '', isPseudo = false) {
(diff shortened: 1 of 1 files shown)


foot:
To repro you need docker 1.10 going, which for me was:
```
brew update
brew upgrade docker docker-machine
docker-machine rm weave-1
docker-machine create --driver virtualbox weave-1
```

davkal:
Cannot repro anymore because BE fix is already merged. Code LGTM.

PR #1205, opened 2016-03-24
Clean out all the JS in the client build dir.

tomwilkie:


davkal:
LGTM, removes all built js files.

PR #1208, opened 2016-03-24
host-level ttys

### LINKED ISSUE #1016: Add terminals on the host.

2opremio:
Fixes #1016

### DIFF
--- a/common/xfer/pipes.go
+++ b/common/xfer/pipes.go
@@ -27,7 +27,16 @@ type pipe struct {
 	onClose         func()
 }
 
-// NewPipe makes a new... pipe.
+// NewPipeFromEnds makes a new pipe specifying its ends
+func NewPipeFromEnds(local io.ReadWriter, remote io.ReadWriter) Pipe {
+	return &pipe{
+		port:      local,
+		starboard: remote,
+		quit:      make(chan struct{}),
+	}
+}
+
+// NewPipe makes a new pipe
 func NewPipe() Pipe {
 	r1, w1 := io.Pipe()
 	r2, w2 := io.Pipe()
--- a/probe/host/controls.go
+++ b/probe/host/controls.go
@@ -1,0 +1,66 @@
+package host
+
+import (
+	"os/exec"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/kr/pty"
+
+	"github.com/weaveworks/scope/common/xfer"
+	"github.com/weaveworks/scope/probe/controls"
+)
+
+// Control IDs used by the host integration.
+const (
+	ExecHost = "host_exec"
+)
+
+// Controls handles controls for a hosts.
+type Controls struct {
+	pipes controls.PipeClient
+}
+
+// NewControls creates new host controls.
+func NewControls(pipes controls.PipeClient) *Controls {
+	c := &Controls{pipes: pipes}
+	controls.Register(ExecHost, c.execHost)
+	return c
+}
+
+// Stop stops the host controls.
+func (*Controls) Stop() {
+	controls.Rm(ExecHost)
+}
+
+func (c *Controls) execHost(req xfer.Request) xfer.Response {
+	cmd := exec.Command(hostShellCmd[0], hostShellCmd[1:]...)
+	cmd.Env = []string{"TERM=xterm"}
+	ptyPipe, err := pty.Start(cmd)
+	if err != nil {
+		return xfer.ResponseError(err)
+	}
+
+	id, pipe, err := controls.NewPipeFromEnds(nil, ptyPipe, c.pipes, req.AppID)
+	if err != nil {
+		return xfer.ResponseError(err)
+	}
+	pipe.OnClose(func() {
+		if err := cmd.Process.Kill(); err != nil {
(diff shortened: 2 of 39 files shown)


tomwilkie:
@2opremio would you mind squashing into 2 changesets - one for the vendoring and one for the rest?

probe/host/controls.go
```diff
@@ -0,0 +1,66 @@
...
+
+// Control IDs used by the host integration.
+const (
+	ExecHost = "host_exec"
+)
+
+// Controls handles controls for a hosts.
+type Controls struct {
+	pipes controls.PipeClient
+}
```
tomwilkie: I'd prefer to see this rolled into the host Reporter.

common/xfer/pipes.go
```diff
@@ -27,7 +27,16 @@ type pipe struct {
 	onClose         func()
 }
 
-// NewPipe makes a new... pipe.
+// NewPipeFromEnds makes a new pipe specifying its ends
+func NewPipeFromEnds(local io.ReadWriter, remote io.ReadWriter) Pipe {
```
tomwilkie: Only invocation of this has local set to `nil`.

I think I'd prefer to see a new type (that implements Pipe) but panics on `CopyToWebsocket` or something.

tomwilkie: (what I'm getting at: this type is already used in too many places for too many things, and needs to be broken down into multiple types for different uses)

2opremio:
> would you mind squashing into 2 changesets - one for the vendoring and one for the rest?

Done + rebased

probe/host/controls.go
2opremio: I thought about it after you suggested offline but:

* I am planning to extend host controls with logs next, so we will probably need to separate it from the reporter anyhow.
* Now there are architecture specific files for controls.

common/xfer/pipes.go
2opremio: > needs to be broken down into multiple types for different uses

AFAICT pipes are always created and passed around as a `Pipe` interface so I am not sure what's the benefit of creating new types if they al implement `Pipe`.

probe/host/controls.go
tomwilkie: I'd keep the separate file; its the one-field struct and the extra stuff that puts in probe's main that I object to.  And for consistency with the way docker/controls.go is done.

2opremio: OK, will do.

2opremio: Actually, in docker, the controls are associated with the registry, which doesn't have a `host` equivalent. So, there's no way to make it consistent anyways.

common/xfer/pipes.go
2opremio: After our conversation on Slack I now understand this. I need to think carefully about how to resolve this though. I am still not closely familiar with the full pipes though and I think this refactoring falls of out of the scope of this PR. I have created

probe/host/controls.go
```diff
@@ -0,0 +1,66 @@
...
+			return
+		}
+		log.Info("Host shell closed.")
+	})
+	go func() {
+		if err := cmd.Wait(); err != nil {
+			log.Errorf("Error waiting on host shell: %v", err)
+		}
+		ptyPipe.Close()
+		pipe.Close()
```
paulbellamy: Is it alright to close the pipe multiple times here?

probe/host/controls_darwin.go
```diff
@@ -0,0 +1,3 @@
+package host
+
+var hostShellCmd = []string{"/bin/bash"}
```
paulbellamy: Seems like it would be better to ask the host what the user's `$SHELL` is.

paulbellamy:
I'm *really* not convinced this is a good idea, for security reasons.

2opremio:
> I'm really not convinced this is a good idea, for security reasons.

This doesn't degrade security in any way. To prove it, just open a terminal in the Scope container and type `nsenter -t1 -m bash`

Productizing this makes it more evident as @rade pointed out offline, but that's it.

probe/host/controls.go
2opremio: Those `Closes()`s do different things (`pipe.Close()` calls `OnClose()` but doesn't close the `ptyPipe`) the pattern may be a bit odd but that's how the `Pipe` interface works (something to take into consideration for #1215)

paulbellamy: Ok, and if the ptyPipe exits, then cmd.Wait will return, and ptyPipe.Close, and pipe.Close will be called, so I'm just checking ptyPipe.Close is safe to close multiple times, etc... The cleanup seems quite convoluted.

2opremio: Would adding `ptyPipe.Close()` to `OnClose()` make it easier to understand?

2opremio: I moved `ptyPipe.Close()` to a defer right after creating it. Regarding the closing pattern, I just followed what was done in the docker controls.

paulbellamy:
Technically, looks fine to me. But I'm really not convinced it's a good idea.

pidster:
LGTM
Security in Scope is an issue should be addressed separately, as there's lots more things to consider than just this.

PR #1217, opened 2016-04-01
Fix building outside of the container

2opremio:


paulbellamy:
You probably still want to have those depend on the container in the containerized case.

2opremio:
Why? They depend on it when `BUILD_IN_CONTAINER=true`

paulbellamy:
LGTM, but would be worth having someone who uses this check it.

tomwilkie:
Not lgtm. This will break things. Please wait until I'm back from holiday
and I'll explain.

tomwilkie:
> Please wait until I'm back from holiday and I'll explain.

The dependancy between all the things built/run inside the container needs to exist so that if you change the build container (ie add a dependancy) then the executables are rebuilt inside the container.

The correct way to fix this is perhaps by clearing SCOPE_BACKEND_BUILD_UPTODATE and setting BUILD_IN_CONTAINER=false.  Although it will need more work.

davkal:
Perhaps a separate target might help, e.g., `make outside`, which could touch the container uptodate files, so those those targets wont get triggered.

tomwilkie:
@2opremio can we close this?

davkal:
FWIW I'm alright with the status quo.

PR #1225, opened 2016-04-05
Set the memory and open files maximums based on the host

### LINKED ISSUE #1149: More meaningful maximum value for metrics
raised in https://github.com/weaveworks/scope/pull/1105#issuecomment-194216241

For CPU it should be based on number of cores (as it is now), but for memory we could get system memory available, and for fds we could check ulimit.

paulbellamy:
Fixes #1149 

Now the max on Memory, and Open Files metrics are set based on the host.

However, it seems like the sparklines don't use the max value? @davkal @foot ?

### DIFF
--- a/probe/process/reporter.go
+++ b/probe/process/reporter.go
@@ -84,8 +84,8 @@ func (r *Reporter) processTopology() (report.Topology, error) {
 			node = node.WithMetric(CPUUsage, report.MakeMetric().Add(now, cpuUsage).WithMax(maxCPU))
 		}
 
-		node = node.WithMetric(MemoryUsage, report.MakeMetric().Add(now, float64(p.RSSBytes)))
-		node = node.WithMetric(OpenFilesCount, report.MakeMetric().Add(now, float64(p.OpenFilesCount)))
+		node = node.WithMetric(MemoryUsage, report.MakeMetric().Add(now, float64(p.RSSBytes)).WithMax(float64(p.RSSBytesLimit)))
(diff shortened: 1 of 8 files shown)


foot:
I think so... 2x checking.

foot:
You are right, we're taking the max of the samples I think, which might be more appropriate for sparklines? I will try w/ real max soon.

But this works great on MoC! Awesome stuff, LGTM from a behavioural POV

probe/process/reporter.go
```diff
@@ -21,20 +23,30 @@ const (
...
+	openFilesLimit OpenFilesLimit
 }
 
 // Jiffies is the type for the function used to fetch the elapsed jiffies.
 type Jiffies func() (uint64, float64, error)
 
+// MaxMemory is the type for the function used to fetch the hosts' max memory.
+type MaxMemory func() (uint64, error)
+
+// OpenFilesLimit is the type for the function used to fetch the hosts' open files limit.
```
rade: this is a *per process* limit, right?

paulbellamy: Ah, yes in fact. Have to get the information that from the procfs. Will change that.

paulbellamy:
Ok, I updated this to get per-process File limit and memory limits. Unfortunately, the memory limits are not cgroup aware (because the `/proc/<pid>/stat` is not cgroup aware). Apparently to do that you have to parse the cgroups and mount files, and dig through the cgroup file system, so I'll create a separate issue for that.

foot:
Updated js code to use `openfiles.max`.

Proc -> Container memory doesn't seem to "add up", e.g.
 `sum(proc.mem_usage) > container.mem_usage`.
But I don't think thats a regression!

LGTMTM.

PR #1229, opened 2016-04-06
Reduce CPU consumption if UI cannot connect to backend

### LINKED ISSUE #1220: browser uses lots of CPU when scope app is unreachable
My laptop fan starts spinning up whenever I stop scope while having a browser window pointing at it. Happens in both FF and Chrome. Probably some overzealous retry.

davkal:
* blank keyframes for blinking Connect animation
* increased topology API retry timer interval

Fixes #1220


foot:
Yeah, I get a drop 9% -> 5-6% usage.

Crazy it chews so much. Tried some tricks like slapping on a translate3d but no luck. Animated gifs ftw?!...

LGTM, def. improves a bit.

PR #1232, opened 2016-04-06
Multitenant DynamoDB collector should use a string for the hash key.

tomwilkie:
And propagate errors correctly.


paulbellamy:
Why?

tomwilkie:
@paulbellamy added the following to the commit: `This is because the key is of the form "<userid>-<hour bucket>", but as I was testing without a userid, I didn't notice that "-<hour bucket>" was a valid number.`

PR #1236, opened 2016-04-06
Fix terminals (broken since f34146d)

davkal:
`controlPipe` is now an immutable map, but was not treated that way by
all components.


foot:
Nice one, LGTM

PR #1237, opened 2016-04-06
Apply topology options also to sub-topologies

### LINKED ISSUE #1221: filters should be associated with views rather than sub-views
I was seeing too many / too few containers when switching between the three Container sub-views ("Container", "By DNS" and "By Image")... until I realised that my selection for "show system/application/both" and "show stopped/running/both" was being apparently being reset when switching sub-views... until I realised that it was in fact saved/restored per sub-view.

I suspect I am not the only user that finds this counter-intuitive. When I tell scope, say, to "show stopped containers", I expect it to apply that selection across all container views.

davkal:
Subtopologies inherit the applied options if the keys are the same.
The labels are still determined by the topology itself.

Fixes #1221


foot:
Nice, LGTM

PR #1239, opened 2016-04-07
Improve canvas highlighting performance.

foot:
- Untangle edge property dependencies so we don't redraw them all if a node is
  highlighted.
- Only "preparePoints" if they've changed.
- Helps smooth panning too, mouse will occasionally cross lines which can make things janky.


davkal:
Great! LGTM

PR #1241, opened 2016-04-07
Ensure k8s views are correctly hidden.

### LINKED ISSUE #1214: K8s views are always present
There is an "internet" node in the services view, which I think is causing them to display.

Seems like the unconnected internet node is not being filtered, but showing up when showing system pods/services.

The report contains no pods or services.

But `http://192.168.99.100:4040/api/topology/pods?system=true` shows:
```json
{
   "nodes" : {
      "out-theinternet" : {
         "label_minor" : "Outbound connections",
         "control_node" : "",
         "controls" : {},
         "children" : [],
         "sets" : {},
         "shape" : "cloud",
         "adjacency" : null,
         "latest" : {},
         "stack" : false,
         "edges" : {},
         "pseudo" : true,
         "id" : "out-theinternet",
         "parents" : {},
         "rank" : "",
         "counters" : {},
         "metadata" : {},
         "label_major" : "The Internet"
      }
   }
}
```

tomwilkie:
- Filter unconnected psuedo nodes from the Pods view
- Don't report these filtered nodes in stats
- Fix typo in logic for hiding views

Fixes #1214

### DIFF
--- a/render/filters.go
+++ b/render/filters.go
@@ -58,12 +58,26 @@ func ColorConnected(r Renderer) Renderer {
 // Filter removes nodes from a view based on a predicate.
 type Filter struct {
 	Renderer
-	FilterFunc func(report.Node) bool
+	FilterFunc     func(report.Node) bool
+	ReportFiltered bool // false means we don't report stats for how many are filtered
 }
 
 // MakeFilter makes a new Filter.
 func MakeFilter(f func(report.Node) bool, r Renderer) Renderer {
-	return Memoise(&Filter{r, f})
+	return Memoise(&Filter{
+		Renderer:       r,
+		FilterFunc:     f,
+		ReportFiltered: true,
+	})
+}
+
+// MakeSilentFilter makes a new Filter which does not report how many nodes it filters in Stats.
+func MakeSilentFilter(f func(report.Node) bool, r Renderer) Renderer {
+	return Memoise(&Filter{
+		Renderer:       r,
+		FilterFunc:     f,
+		ReportFiltered: false,
+	})
 }
 
 // Render implements Renderer
@@ -115,9 +129,11 @@ func (f *Filter) render(rpt report.Report) (report.Nodes, int) {
 
 // Stats implements Renderer
 func (f Filter) Stats(rpt report.Report) Stats {
-	_, filtered := f.render(rpt)
 	var upstream = f.Renderer.Stats(rpt)
-	upstream.FilteredNodes += filtered
+	if f.ReportFiltered {
+		_, filtered := f.render(rpt)
+		upstream.FilteredNodes += filtered
+	}
 	return upstream
 }
 
--- a/render/filters.go
+++ b/render/filters.go
@@ -169,6 +169,18 @@ func FilterUnconnected(r Renderer) Renderer {
 	)
 }
 
+// SilentFilterUnconnected produces a renderer that filters unconnected nodes
+// from the given renderer; nodes filtered by this are not reported in stats.
+func SilentFilterUnconnected(r Renderer) Renderer {
+	return MakeSilentFilter(
+		func(node report.Node) bool {
+			_, ok := node.Latest.Lookup(IsConnected)
+			return ok
+		},
+		ColorConnected(r),
+	)
+}
+
 // FilterNoop does nothing.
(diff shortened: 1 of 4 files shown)


render/filters.go
```diff
@@ -58,12 +58,26 @@ func ColorConnected(r Renderer) Renderer {
 // Filter removes nodes from a view based on a predicate.
 type Filter struct {
 	Renderer
-	FilterFunc func(report.Node) bool
+	FilterFunc     func(report.Node) bool
+	ReportFiltered bool // false means we don't report stats for how many are filtered
```
paulbellamy: "Report" is a bit overloaded. Maybe `Silent`?

paulbellamy: Also, do we *ever* want the stats to include pseudo nodes filtered? i.e. why not just check if the node is a pseudo before incrementing the counter?

tomwilkie: I see filtering as orthogonal to whether a node is a pseudo node or not.  As Peter might say, I think not reporting the number of filtered pseudo nodes as a category error.

render/topologies.go
```diff
@@ -224,9 +223,15 @@ var HostRenderer = MakeReduce(
...
-		ContainerRenderer,
+	MakeSilentFilter(
+		func(n report.Node) bool {
+			_, isConnected := n.Latest.Lookup(IsConnected)
+			return n.Topology != Pseudo || isConnected
+		},
+		MakeMap(
+			MapContainer2Pod,
+			ColorConnected(ContainerRenderer),
+		),
```
paulbellamy: afaict, this will still fail with containers outside pods.

eg. if container1 connects to the internet, then both will be marked as connected, but container1 will be dropped by MapContainer2Pod.

paulbellamy: If MapContainer2Pod converted non-pod-containers to a "UnPodded" pseudonode, that would probably fix that.

tomwilkie: Yeah, good catch.  I guess we don't have an equivalent of "Uncontainer" for containers that aren't part of pods... "Unpodded"? ;-)

tomwilkie: (But seriously, I'll move the colorconnected up one level to catch this)

render/filters.go
paulbellamy: From an implementation perspective sure, but from a user perspective, I think of the steps as:
1) Nodes are aggregated
2) Nodes are filtered
3) Pseudo-nodes are added

therefore, I wouldn't expect the filter count to include them, because they're an artefact of the view, not of the data.

render/topologies.go
paulbellamy: "Unmanaged"? but yeah, doing the colorconnected is fine. Maybe it's worth extracting this filter into `FilterUnconnectedPseudoNodes` or something? Only used once, though I guess.

tomwilkie:
@paulbellamy PTAL

paulbellamy:
The number of nodes filtered still seem odd.
screen shot 2016-04-07 at 12 38 23

what 32 nodes are being filtered?

paulbellamy:
Also, I still think the SilentFilter thing seems overcomplicated and error-prone. e.g. It will still count the internet node being filtered out as part of the container renderer.

tomwilkie:
> The number of nodes filtered still seem odd.

I think the stats in the bottom left lag slightly; when I do this I get `39 NODES (7 FILTERED)` (which is still wrong).

paulbellamy:
Mine's been showing `12 nodes (34 filtered)` for 28 minutes. When there are 39 nodes visible...

tomwilkie:
Try refreshing?  Otherwise its a UI bug.

paulbellamy:
Ah yeah. Now I get 39 nodes (6 filtered), which is still wrong. It occurs to me that maybe any filters used in `render/topologies.go` should be silent, and only the ones applied (at the end) by the UI widget should count towards the `(X filtered)` statistic.

tomwilkie:
@paulbellamy yup, I agree.  Have made them all silent now; which also means we need to silently filter some non-pseudo nodes.  WDYT?

paulbellamy:
Probably also want SilentFilterNonProcspied

paulbellamy:
LGTM. I take back what I said about SilentFilter not being the right abstraction. When applied to all the internal filters, it's great.

PR #1245, opened 2016-04-07
Ensure connection rows have unique IDs.

### LINKED ISSUE #1218: Inbound/Outbound connection rows/counts/ports not being calculated correctly
Seems like maybe, in particular on internet node.

Could be related to recent changes there.

tomwilkie:
Fixes #1218 

Introduces a new type for ConnectionsTable and ConnectionsRow.

TODO(@davkal)
- [x] update UI for `node_id` and field name changes.

### DIFF
--- a/client/app/scripts/components/node-details/node-details-table.js
+++ b/client/app/scripts/components/node-details/node-details-table.js
@@ -172,7 +172,7 @@ export default class NodeDetailsTable extends React.Component {
             const values = this.renderValues(node);
             const nodeId = node[nodeIdKey];
             return (
-              <tr className="node-details-table-node" key={nodeId}>
+              <tr className="node-details-table-node" key={node.id}>
                 <td className="node-details-table-node-label truncate">
                   <NodeDetailsTableNodeLink {...node} topologyId={this.props.topologyId}
                     nodeId={nodeId} />
--- a/client/app/scripts/components/node-details/node-details-table.js
+++ b/client/app/scripts/components/node-details/node-details-table.js
@@ -169,10 +169,12 @@ export default class NodeDetailsTable extends React.Component {
           <tbody>
           {nodes && nodes.map(node => {
             const values = this.renderValues(node);
+            const nodeId = node.node_id || node.id;
             return (
-              <tr className="node-details-table-node" key={node.id}>
+              <tr className="node-details-table-node" key={nodeId}>
                 <td className="node-details-table-node-label truncate">
-                  <NodeDetailsTableNodeLink topologyId={this.props.topologyId} {...node} />
+                  <NodeDetailsTableNodeLink {...node} topologyId={this.props.topologyId}
+                    nodeId={nodeId} />
                 </td>
                 {values}
               </tr>
--- a/client/app/scripts/components/node-details/node-details-table.js
+++ b/client/app/scripts/components/node-details/node-details-table.js
@@ -191,5 +191,5 @@ export default class NodeDetailsTable extends React.Component {
 }
 
 NodeDetailsTable.defaultProps = {
-  nodeIdKey: 'id'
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -20,7 +20,7 @@ CODECGEN_EXE=$(CODECGEN_DIR)/bin/codecgen_$(shell go env GOHOSTOS)_$(shell go en
 GET_CODECGEN_DEPS=$(shell find $(1) -maxdepth 1 -type f -name '*.go' -not -name '*_test.go' -not -name '*.codecgen.go' -not -name '*.generated.go')
 CODECGEN_TARGETS=report/report.codecgen.go render/render.codecgen.go render/detailed/detailed.codecgen.go
 RM=--rm
-RUN_FLAGS=-i
+RUN_FLAGS=-ti
```
paulbellamy: separate commit

tomwilkie: but mummmm.....

tomwilkie: Done; explained in commit message.

davkal:
Done UI changes. Not so pretty:

* `topologyId` got renamed to `topology_id`
* `node_id` is snake case :snake: 

The JS world is ruled by camels :camel:

tomwilkie:
Okay I'll fix this before we merge. @paulbellamy will need a review.

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -169,10 +169,12 @@ export default class NodeDetailsTable extends React.Component {
...
           {nodes && nodes.map(node => {
             const values = this.renderValues(node);
+            const nodeId = node.node_id || node.id;
             return (
-              <tr className="node-details-table-node" key={node.id}>
+              <tr className="node-details-table-node" key={nodeId}>
                 <td className="node-details-table-node-label truncate">
-                  <NodeDetailsTableNodeLink topologyId={this.props.topologyId} {...node} />
+                  <NodeDetailsTableNodeLink {...node} topologyId={this.props.topologyId}
+                    nodeId={nodeId} />
```
paulbellamy: rename `node_id` -> `nodeId` and let's not pass it in twice.

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -169,10 +169,12 @@ export default class NodeDetailsTable extends React.Component {
           <tbody>
           {nodes && nodes.map(node => {
             const values = this.renderValues(node);
+            const nodeId = node.node_id || node.id;
```
paulbellamy: this is a connection-specific thing, so it doesn't really seem like something the node details table should know about.

paulbellamy: There's probably a `table` that both node details table and connections table "inherit" from.

render/detailed/connections.go
```diff
@@ -29,13 +30,42 @@ var (
 	}
 )
 
+// ConnectionsTable is the table of connection to/form a node
+type ConnectionsTable struct {
```
paulbellamy: Having a separate type for these is good. Not convinced on the naming, as I'd like to move away from the Table/Row nomenclature (only used by Metadata/MetricRows anymore).

paulbellamy:
`ID`/`NodeID` Seems a bit confusing. At the very least a comment on why you'd use one vs the other, but I'm not sure what a better name would be.

davkal:
I'll add a field to node-details-table to tell it what row field to use for row key.

render/detailed/connections.go
tomwilkie: Roger.  Will rename to `ConnectionsSummary` and `Connection`.

foot:
Fixes report linked in issue.

PR #1252, opened 2016-04-11
EBPF plugin

2opremio:
@paulbellamy PTAL

Still a few TODOs but it's fully functional. The changes in the unix rountripper make the connections persistent and require handling errors a bit different as I mentioned in #1126


examples/plugins/ebpf/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu:wily
```
tomwilkie: Might want to consider using [tinycore python](https://github.com/weaveworks/scope/blob/master/experimental/example/app/Dockerfile#L1) if possible, ubuntu is a rather large image.

examples/plugins/ebpf/http-requests.py
```diff
@@ -0,0 +1,197 @@
...
+        #        probe to remove processes from the table when they die (this
+        #        will probably require keeping keeping track of tasks and not
+        #        just processes)
+        req_count_snapshot = dict()
+        while True:
+            time.sleep(1)
+            req_count_snapshot = self.update_http_rate_per_pid(req_count_snapshot)
+
+
+class PluginRequestHandler(BaseHTTPServer.BaseHTTPRequestHandler):
```
tomwilkie: I'd use flask for the http server routing etc; it will make this much more concise I think.

2opremio: My intention was to limit the dependencies as much as possible for now. At some point we may want to provide a library to simplify the plugins (then we can use flask or whathaveyou).

However, since we are using a container we could probably use whatever library we want already. I am not sure ...

paulbellamy: lets use flask, and install it into the plugin container. It will clarify the example with less boilerplate.

paulbellamy:
Mentioned elsewhere, but is it possible to add a test for the unix round tripper change demonstrating the changed behaviour?

paulbellamy:
Would be nice if the directory was `example/plugins/http-requests` (to match the plugin name)

examples/plugins/ebpf/http-requests.c
```diff
@@ -0,0 +1,61 @@
...
+  buffers from kernel to userspace.
+
+  skb_copy_datagram_iter() has an associated tracepoint
+  (trace_skb_copy_datagram_iovec), which would be more stable than a kprobe but
+  it lacks the offset argument.
+ */
+int kprobe__skb_copy_datagram_iter(struct pt_regs *ctx, const struct sk_buff *skb, int offset, void *unused_iovec, int len)
+{
+
+  /* Inspect the beginning of docket buffers copied to user-space to determine if they
```
paulbellamy: docker

2opremio: socket :) Thanks

examples/plugins/ebpf/http-requests.py
2opremio: OK, I have no idea about flask, so it may take a while ...

examples/plugins/ebpf/http-requests.c
```diff
@@ -0,0 +1,61 @@
...
+  */
+  if (head_len - offset < 7) {
+    return 0;
+  }
+
+  u8 data[4] = {};
+  bpf_probe_read(&data, sizeof(data), skb->data + offset);
+
+  /* TODO: support other methods and optimize lookups */
+  if ((data[0] == 'G') && (data[1] == 'E') && (data[2] == 'T')) {
```
paulbellamy: maybe `GET /` ? Any benefit in that?

paulbellamy: maybe grab the space after Get too?

2opremio: sure

examples/plugins/ebpf/http-requests.py
```diff
@@ -0,0 +1,197 @@
...
+    def update_http_rate_per_pid(self, last_req_count_snapshot):
+        new_req_count_snapshot = dict()
+        new_http_rate_per_pid = dict()
+        req_count_table = self.bpf.get_table(EBPF_TABLE_NAME)
+        for key, value in req_count_table.iteritems():
+            request_delta = value.value
+            if key.pid in last_req_count_snapshot:
+                 request_delta -= last_req_count_snapshot[key.pid]
+            if request_delta > 0:
+                new_http_rate_per_pid[key.pid] = request_delta
```
paulbellamy: doesn't this mean that unchanged rates (or decreased rates) get dropped?

2opremio: If the counts don't increase that means that there were no requests in the last second, which is an effective request rate of 0 (if the request rates don't change or decrease the count would still increment).

2opremio: (we could smooth this out with an averaging sliding window of a few seconds though).

paulbellamy: ah, of course, it's total to derivative, gotcha. I was thinking that the kernel table was rate for some reason.

2opremio:
> Mentioned elsewhere, but is it possible to add a test for the unix round tripper change demonstrating the changed behaviour?

The desired behavior (and I admit is pretty unspecific) for this pull request is getting the python plugin to work because `httputil` was misbehaving.

examples/plugins/ebpf/http-requests.py
2opremio: I see. Exactly, the kernel table stores counts, it's up to us to sample that from userspace and turn it into request rates (I wouldn't even know how to measure request rates directly in ebpf since there are no timers).

examples/plugins/ebpf/http-requests.py
2opremio: Flask doesn’t support unix sockets directly. AFAICT, everything, including the underlying WSGI library, assumes you will use standard TCP ports. The best way I found to solve it is this horror https://github.com/tomekwojcik/werkzeug_plus_unix_socket . Can you suggest any alternatives? Otherwise I think I will stick to this code and turn it into a library at some point.

2opremio: After searching for a while it seems serving on a unix socket is a pretty unusual requirement.

examples/plugins/ebpf/Dockerfile
2opremio: I've looked into it, but unfortunately bcc only provides binaries for Ubuntu/Fedora. bcc doesn't ship on pip (probably because of the C-library dependencies) and it's a bit of a pain to build. See https://github.com/iovisor/bcc/blob/master/INSTALL.md

paulbellamy:
LGTM

PR #1253, opened 2016-04-11
Try non-scaling-stroke.

foot:
Works quite nicely.
- Big boost for zooming perf.
- Nice effect when zoomed right out.
- Lose the shape a tiny bit. (hex starts to look like a circle).
- Lose "shadow-margin" (between border and shadow) unless quite zoomed
  in.

### DIFF
--- a/client/app/scripts/charts/edge.js
+++ b/client/app/scripts/charts/edge.js
@@ -3,6 +3,14 @@ import { connect } from 'react-redux';
 import classNames from 'classnames';
 
 import { enterEdge, leaveEdge } from '../actions/app-actions';
+import { featureIsEnabled } from '../utils/feature-utils';
+
+function getLinkProps() {
+  if (featureIsEnabled('non-scaling-stroke')) {
+    return { vectorEffect: 'non-scaling-stroke' };
+  }
+  return {};
+}
 
 class Edge extends React.Component {
 
@@ -21,7 +29,7 @@ class Edge extends React.Component {
         className={className} onMouseEnter={this.handleMouseEnter}
         onMouseLeave={this.handleMouseLeave} id={id}>
         <path d={path} className="shadow" />
-        <path d={path} className="link" />
+        <path d={path} className="link" {...getLinkProps()} />
       </g>
     );
   }
--- a/client/app/scripts/charts/node-shape-circle.js
+++ b/client/app/scripts/charts/node-shape-circle.js
@@ -1,8 +1,9 @@
 import React from 'react';
 import classNames from 'classnames';
+
 import { getMetricValue, getMetricColor, getClipPathDefinition } from '../utils/metric-utils';
 import { CANVAS_METRIC_FONT_SIZE } from '../constants/styles';
-
+import { getBorderProps } from './node';
 
 export default function NodeShapeCircle({id, highlighted, size, color, metric}) {
   const clipId = `mask-${id}`;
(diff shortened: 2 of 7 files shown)


davkal:
Only nit is stacks:
screen shot 2016-04-11 at 13 15 00

becomes

screen shot 2016-04-11 at 13 15 17

Here is `master` to compare:
screen shot 2016-04-11 at 13 16 36

But then again, zoomed out that far, it's hard to maintain any distinction.

foot:
I think the stacks are quite identifiable w/ their heavy shadows!

Other issues:
 - non-scaling edges start to look a bit heavy when zoomed-out.
 - having a lot of text on the canvas (if we don't truncate it when zoomed out), slows things down.

davkal:
> non-scaling edges start to look a bit heavy when zoomed-out.

Scaled edges look slicker up to a point, when they become barely perceptible. Non-scaled edges keep giving you an impression on the connectedness up to a high zoom-out, but at some point become heavier than the nodes. How about keeping non-scaled edges, but make them slightly paler on high zoom levels (e.g. via a CSS class on the chart element).

> having a lot of text on the canvas (if we don't truncate it when zoomed out), slows things down.

We could be realistic about readability and just hide them on high zoom-outs.

davkal:
Can we rewrite this using feature flags? Then we can see it in the wild.

foot:
Perhaps!

davkal:
Re-implemented with feature flag

Activate via
```
localStorage['scope-experimental:non-scaling-stroke'] = true
```

davkal:
PTAL @foot

foot:
Nice one. LGTM.

+1 this feature.

foot:
Okay it doesn't work perfectly w/ clouds right now... still LGTM but if you wanna try and fix that too..

foot:
screen shot 2016-12-28 at 13 38 52

foot:
Actually i can have a quick play now.

rade:
> it doesn't work perfectly w/ clouds

Based on the pic, that is a bit of an understatement!

foot:
`npm install svgpath`

```
var svgpath = require('svgpath');

const MAX = [1920, 1408];
const CLOUD_PATH = 'M 1920,384 Q 1920,225 1807.5,112.5 1695,0 1536,0 H 448 '
  + 'Q 263,0 131.5,131.5 0,263 0,448 0,580 71,689.5 142,799 258,853 '
  + 'q -2,28 -2,43 0,212 150,362 150,150 362,150 158,0 286.5,-88 128.5,-88 '
  + '187.5,-230 70,62 166,62 106,0 181,-75 75,-75 75,-181 0,-75 -41,-138 '
  + '129,-30 213,-134.5 84,-104.5 84,-239.5 z';

const factor = 32 / 1920;
var transformed = svgpath(CLOUD_PATH)
  .scale(factor)
  .round(3);


console.log(transformed.toString());
console.log([[0, MAX[0] * factor], [0, MAX[1] * factor]]);
```

fbarl:
So... what should we do with this PR? :)

I rebased it locally because I was curious to see how it would look like (as cloud border width was not an issue anymore), but I don't think it's possible to come up with a border good width which would be neither too thin on maximal zoom nor too thick on minimal zoom.

What I tried instead was to draw two overlapping borders, a *normal* one and a *non-scaling* one with 20% of the normal one's width. The result is that the border scales normally unless you zoom out a lot in which case a non-scaling border becomes dominant and serves as a lower-bound cap on the line thickness.

One objection might be that it would be more expensive to render two borders instead of just one, but that would only happen with the flag on and what matters more IMO is that still nothing would have to be rerendered when zooming in and out (unlike if we tried to achieve the same effect with only one border).

Is the suggestion too hacky? :D

davkal:
I suggest to revisit the premise of this PR: are non-scaling strokes an improvement? To me they bring a bit calm during zooms, which I consider good as it's less distracting. If the edge cases (particularly for the edges) are problematic, then we can simply not apply the feature in extreme zoom factors.

fbarl:
I personally wouldn't like to see non-scaling strokes applied to edges at all, as they'd take too much attention and would be hard to read on extreme zoom-out anyway.

As for the nodes, not applying them at extreme zooms might look better, but I think we'd still need to make sure the transition is smooth. @davkal would you mind if I took over this story or at least pushed the rebased version (+ some of the changes I suggested for you guys to take a look at)?

2opremio:
Should we close this?

davkal:
Yes, this will be addressed when developing other features.

PR #1261, opened 2016-04-11
Cache pan/zoom per topology

### LINKED ISSUE #1238: Layout center shared between layouts.
1) Click processes topo
2) Click containers topo
3) Pan layout
4) Back to processes
5) Note processes has also been panned.

davkal:
* remembers canvas position/zoom per topology
* makes it easy to go back to the last position in a prev topology

Fixes #1238


foot:
Nice! Works really well, might be nice to have a reset button too but I'm not sure where to put that.

LGTM

PR #1265, opened 2016-04-12
Add /api/probes endpoint

tomwilkie:


report/probes.go
```diff
@@ -0,0 +1,43 @@
...
+			result[id] = probe
+		}
+	}
+	return result
+}
+
+// Probe is the details for a single probe that generated a report.
+type Probe struct {
+	ID       string    `json:"id"`
+	LastSeen time.Time `json:"lastSeen"`
```
paulbellamy: we don't use `camelCase` anywhere else, do we?

tomwilkie: I thought that was what @davkal wanted?  Either way, this is consistent with how app mapper exposed it.

davkal: +1 for camel case in json directed towards the frontend. :camel:

PR #1266, opened 2016-04-12
Force layout topology on browser resize

### LINKED ISSUE #1096: Resizing the window and zooming in/out can confuse window size
screenshot 2016-03-02 15 45 18
screenshot 2016-03-02 15 47 56

davkal:
* prevents nodes scaling indepently of graph size

Fixes #1096

### DIFF
--- a/client/app/scripts/charts/nodes-chart.js
+++ b/client/app/scripts/charts/nodes-chart.js
@@ -80,9 +80,10 @@ export default class NodesChart extends React.Component {
       });
     }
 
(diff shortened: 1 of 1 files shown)


davkal:
Alternative: #1269

foot:
This one is pretty fun, would be neat if the center stayed the same, but I'm imaging that would be a lot of work. LGTM as is.

davkal:
Automatic layout runs may shift nodes around. That may be too surprising.
Going with the alternative #1269

PR #1272, opened 2016-04-12
Fixed column widths for node tables with few columns

### LINKED ISSUE #1076: Maximise text fields in detail panel tables.
The detail panel tables could be a little more generous with the space allowed for text fields.

The processes table truncates the last erlang process, a bit more space is available inside the table as well as from the margins.

The containers table truncates two of the fields, when there's plenty of space for them.

detail-panel-tables

davkal:
* Gives the node label column more space
* let the remaining columns fight over the rest

Fixes #1076


foot:
Nice one, LGTM

PR #1274, opened 2016-04-12
Remove load5 and load15

### LINKED ISSUE #1267: Ordering of metrics is not as expected
The order of metrics in the Host view is:
  ... Load (15m), Load (1m), Load (5m) ... 

not as expected:
  ... Load (1m), Load (5m), Load (15m) ...

metric-ordering

davkal:
* not really useful
* take up lots of real estate

Fixes #1267


peterbourgon:
👍  to the idea.

PR #1277, opened 2016-04-12
eBPF plugin improvements

2opremio:
* Check that the protocol is TCP
* Support all HTTP methods
* Clear out tasks from table when they exit (required using task ids as table
  keys). A nice side effect of this is that we can tell apart tasks receiving 0
  req/second (i.e. tasks which currently aren't receiving any requests but were
  receiving requests at some point) from tasks which don't exist anymore.
  We report metrics fro the former.
* Simplify hash table key (no need to use a struct)

@paulbellamy @tomwilkie (and @dpw if he has the time) PTAL

### DIFF
--- a/examples/plugins/http-requests/http-requests.c
+++ b/examples/plugins/http-requests/http-requests.c
@@ -1,7 +1,11 @@
 #include <linux/skbuff.h>
 #include <net/sock.h>
 
-/* Table from (Process id|Task id) to (Number of received http requests) */
+/* Table from (Task group id|Task id) to (Number of received http requests).
+   We need to gather requests per task and not only per task group (i.e. userspace pid)
+   so that entries can be cleared up independently when a task exists.
+   This implies that userspace needs to do the per-process aggregation.
+ */
 BPF_HASH(received_http_requests, u64, u64);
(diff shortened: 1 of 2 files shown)


paulbellamy:
afaict, looks fine.

dpw:
This approach has some major flaws if people try to make serious use of it .  We discussed them the other day, so I won't go into details.  I see it's in an "examples" directory, so maybe we don't intend people to use it seriously.  If so , I hope the context makes its status clear.

2opremio:
> This approach has some major flaws if people try to make serious use of it . We discussed them the other day, so I won't go into details.

Does the _Caveats_ comment in the ebpf code reflect that correctly?

dpw:
> Does the Caveats comment in the ebpf code reflect that correctly?

Yes.  But very few people will look at comments in the code.

It's a question of context.  So for example, if people have to build the examples themselves, it's fine.

2opremio:
> It's a question of context. So for example, if people have to build the examples themselves, it's fine.

Right, as you mentioned, it's in the examples directory so I think it's fine. No code comments?

PR #1280, opened 2016-04-13
Nice and red and blinky ui label for broken plugins

### LINKED ISSUE #1271: Plugins UI should have some indicator when plugins are broken
Raised in: https://github.com/weaveworks/scope/pull/1126#issuecomment-208873500

foot:
Fixes #1271


client/app/scripts/components/plugins.js
```diff
@@ -1,10 +1,20 @@
 import React from 'react';
+import classNames from 'classnames';
 
 export default class Plugins extends React.Component {
-  renderPlugin(plugin) {
+  renderPlugin({id, label, description, status}) {
+    const error = status !== 'ok';
+    const className = classNames({ error });
+    const title = error ?
+      `Status: ${status}. (Plugin description: ${description})` :
```
2opremio: Maybe we should get the same format regardless of whether there's an error.

davkal: +1, i think it's good seeing that things are ok.

foot: Yes, agree! Good suggestion!

client/app/scripts/stores/app-store.js
```diff
@@ -46,7 +46,7 @@ let highlightedEdgeIds = makeSet();
 let highlightedNodeIds = makeSet();
 let hostname = '...';
 let version = '...';
-let plugins = null;
+let plugins = [];
```
davkal: Can we use immutables here instead?

foot: yessss we cannn

client/app/styles/main.less
```diff
@@ -1007,6 +1007,11 @@ h2 {
     content: ', ';
   }
 
+  .error {
+    animation: blinking 2.0s 30 @base-ease; // blink for 1 minute;
+    color: darkred;
```
davkal: I think that's our first use of red for an error. Do we really need it?

davkal:
The blinking red is a bit subtle, and our first use of red to depict an error. The reconnect error that is displayed when the backend is not available, blinks with a circled exclamation mark. Why not just use that?

foot:
First thought is that is was very icon heavy down there so didn't actually
try it out. Will give it a spin.

davkal:
There is a `fa` icon w/o the circle if that helps.

PR #1290, opened 2016-04-13
Add control for removing stopped docker containers.

### LINKED ISSUE #1367: We should be able to control the order of controls in the details panel
Currently they are sorted alphabetically; I propose we add a rank, and sort by that.

tomwilkie:
Also fixes #1367


paulbellamy:
Maybe after the play button? Or is there a specific reason for putting it first?

paulbellamy:
Also, removing a container takes 15 seconds (so looks like the control hasn't worked), then the details panel breaks (but stays open).

tomwilkie:
> Maybe after the play button? Or is there a specific reason for putting it first?

No reason, will do.

> Also, removing a container takes 15 seconds (so looks like the control hasn't worked), then the details panel breaks (but stays open).

I have no intention of dealing with this in this PR, if thats alright.

tomwilkie:
>> Maybe after the play button? Or is there a specific reason for putting it first?

> No reason, will do.

Actually, there sorted alphabetically.  We can't control the order of them.

tomwilkie:
> Also, removing a container takes 15 seconds (so looks like the control hasn't worked), 

This will be fixed in #1363 

> then the details panel breaks (but stays open).

@davkal / @foot any ideas here?  We leave the details panel open for good reason (reconnects etc), but perhaps you could close it if you receive a Remove node diff?   But not if the ws goes away?

foot:
The details panel can handle 404s initially. Perhaps if it gets a 404 while waiting for a control to complete something bad happens. Will have a look.

davkal:
Yes, we can close the details panel on node remove (we're already handling missing mouseout events that way). Otherwise we could do control update requests to get the status on a control action.

davkal:
I propose this lifecycle-ish order:  start - restart - pause - stop - remove

davkal:
Could also be done in-band, in the control response. We just need to agree on a field. Eg. terminal controls return `{pipe}`, a remove control could return a different field. @tomwilkie

tomwilkie:
Sure, I can return a field saying along the lines of "deleted": "<nodeid>"?

tomwilkie:
I think thats all the feedback addressed - @paulbellamy PTAL?

common/xfer/controls.go
```diff
@@ -18,10 +18,15 @@ type Request struct {
...
-	RawTTY bool        `json:"raw_tty,omitempty"`
+	Value interface{} `json:"value,omitempty"`
+	Error string      `json:"error,omitempty"`
+
+	// Pipe specific fields
+	Pipe   string `json:"pipe,omitempty"`
+	RawTTY bool   `json:"raw_tty,omitempty"`
+
+	// Remove specific fields
+	RemovedNode string `json:"removedNode,omitempty"` // Set if node was removed
```
paulbellamy: adding extra fields for specific controls seems wrong.

davkal: The UI just needs to know that an action succeeded. Is that the case when there is no error? Then I think we can check the control type in the UI, and close the details panel if the control returns w/o error.

tomwilkie: Yeah no error implies success.  I don't know if hard coding specific behaviour in the UI for a given control is better than adding a field to the control response though.

davkal: Clean API is better. So I think we remove it.

paulbellamy: Now that a shortcut report is sent out when a node is removed do we need to hard-code any UI behaviour? Won't it just happen?

davkal: A node being in `remove` does not necessarily mean that it was *removed*. Correct me if I'm wrong, but it could have other reasons for short dis-/re-appearance.

tomwilkie: Yeah we decided a while back not to close the details window when nodes are removed, as nodes can come and go (with filtering, for instance).

paulbellamy: Ah, so the behaviour we want is to close the details panel on success. Perhaps some way to annotate the control as "final" or "closeOnSuccess", or something?

davkal:
I'm treating the removal in the UI only. `closeOnSuccess` is the UI's domain, the backend should not know about it.

PR #1296, opened 2016-04-13
Update to go 1.6.1

tomwilkie:
See https://groups.google.com/forum/#!msg/golang-announce/9eqIHqaWvck/kXsfO0ogLAAJ

I don't think we're affected, but might as well do it.


2opremio:
I would merge this after 0.14 (just in case)

peterbourgon:
re: stripping, are we positive this isn't going to negatively affect us in the future? Put another way: why don't they have this on by default?

paulbellamy:
Seems fine. if we fine we're missing the gdb debugging, then we could parameterize it `make DEBUG=true`, or roll it back.

tomwilkie:
+1 lets try it and roll it back if it causes problems.

2opremio:
Has anybody checked whether 'go tool pprof' works without the symbols? If it doesn't we should revert it.

PR #1298, opened 2016-04-13
Add k8s pod log control

### LINKED ISSUE #1081: Tags in header of Container view detail panel erroneously shows all hosts with container
The image shows the detail panel of a container, where the tags listed include all of the hosts in which the container image is present (or possibly just all of the hosts), instead of only the host in which this container is running.

ping-tags

### LINKED ISSUE #1093: Distinguish between reporting hosts and containing hosts
screen shot 2016-02-26 at 02 06 28

The screenshot above is from #1043 . The Container is incorrectly attributed to 3 three hosts because the probes in those 3 hosts report kubernetes information about that same container.

### LINKED ISSUE #1142: Show pods in hosts details panel
Due to the way we talk to k8s, pods don't get a host id, so we can't do this just yet.

tomwilkie:
Also:
- vendor the entire kubernetes/pkg subdir, because its easier.
- have each probe only emit local pods (so we can deterministically route controls) (fixes #1081)
- tag pods with the host id and have them included in host topology as children (fixes #1142, fixes #1093)


paulbellamy:
Looks like you've forgotten to commit your tests.

tomwilkie:
> Looks like you've forgotten to commit your tests.

Not sure this needs extra tests.  Everything it pretty well tested; this is just plumbing.

paulbellamy:
LGTM

Issue #1306, opened 2016-04-14
In OpenShift infra container image is openshift/origin-pod


errordeveloper:
<img width="997" alt="screenshot 2016-04-14 17 07 47" src="https://cloud.githubusercontent.com/assets/251467/14534979/75e3fb44-0263-11e6-8b9e-d9ab19b52165.png">

PR #1307, opened 2016-04-14
Add OpenShift infra container image name to the list of well-known system containers

errordeveloper:


2opremio:
Fixes #1306

render/filters.go
```diff
@@ -267,4 +267,6 @@ var systemImagePrefixes = map[string]struct{}{
 	"amazon/amazon-ecs-agent":             {},
 	"beta.gcr.io/google_containers/pause": {},
 	"gcr.io/google_containers/pause":      {},
+	"openshift/origin-pod":                {},
+	"docker.io/openshift/origin-pod":      {},
```
errordeveloper: I added both version of the name, as on vanilla Docker the `docker.io` prefix would quite certainly get dropped.

paulbellamy:
LGTM

PR #1308, opened 2016-04-14
Extend plugin documentation

2opremio:
* Add section to main README
* Add introduction to plugins-specific README


2opremio:
@paulbellamy @davkal PTAL

davkal:
Could we get a quick-start section in the plugins README on how to use the example plugins?

2opremio:
You only need to type `make` in the example directory but we should make it explicit. Maybe we should include a README in each example directory?

@paulbellamy Can you please take over? I won't have time for this today.

davkal:
I would add it to the central plugins readme. Let's try lowering barriers to get started. Definitely list the steps to do, and then tell them what they are expected to see if it works/doesnt work.

paulbellamy:
@davkal added that. Mind proof-reading it?

davkal:
LGTM

Issue #1278, opened 2016-04-13
Mismatching metrics on canvas and sparklines


2opremio:
It may be justified by some transformation made by the sparklines, but it should be consistent on the canvas, otherwise it's very confusing for the end user.

<img width="1749" alt="screen shot 2016-04-13 at 03 50 47" src="https://cloud.githubusercontent.com/assets/2362916/14481466/0dcd7514-012b-11e6-964c-2d59a7c4ee1e.png">

<img width="1749" alt="screen shot 2016-04-13 at 03 50 18" src="https://cloud.githubusercontent.com/assets/2362916/14481470/13daf076-012b-11e6-9169-d09c4accd5a1.png">

paulbellamy:
Could be caused by different polling intervals?

paulbellamy:
MoC are sent via websockets every second, while the sparklines are updated with the details panel (every 30s).

2opremio:
> MoC are sent via websockets every second, while the sparklines are updated with the details panel (every 30s).

Do they come from different data sources or just different polling intervals? Can we use the same data source/interval the metrics in the details panel in MoC?

foot:
Yeah, this needs reconciling the UI really.

foot:
Switch to details data source when available.

PR #1310, opened 2016-04-15
Sync Canvas metrics w/ Details Panel

foot:
Address #1278

### DIFF
--- a/client/app/scripts/charts/nodes-chart-nodes.js
+++ b/client/app/scripts/charts/nodes-chart-nodes.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PureRenderMixin from 'react-addons-pure-render-mixin';
 import reactMixin from 'react-mixin';
+import { fromJS } from 'immutable';
 
 import NodeContainer from './node-container';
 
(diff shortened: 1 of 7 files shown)


davkal:
Brainstorm:

* even though details and canvas metrics are are requested at different times, the samples' dates should be the same, there is potential to sync them
* maybe remove metrics feeder HOC, to get the sync right between details and canvas, then re-add feeder
* central metrics object for that stores metrics by node id, can be queried with a timestamp, feeder HOC feeds that timestamp
* global metrics timestamp clock, dynamically adjusted based on what sample ranges are available, keeps all metrics in sync

foot:
Mmm, I like the idea of global metrics timestamp clock!

client/app/scripts/charts/nodes-chart-nodes.js
```diff
@@ -32,11 +33,13 @@ export default class NodesChartNodes extends React.Component {
     };
 
     // TODO: think about pulling this up into the store.
-    const metric = node => (
-      node.get('metrics') && node.get('metrics')
+    const metric = node => {
+      const isHighlighted = topCardNode && topCardNode.details && topCardNode.id === node.get('id');
+      const sourceNode = isHighlighted ? fromJS(topCardNode.details) : node;
```
davkal: Sneaky 🐍

davkal:
LGTM

PR #1313, opened 2016-04-15
Add local_networks to weave Overlay nodes

### LINKED ISSUE #1162: Weave Net connections missing on Scope Service
If I launch the [ECS demo](https://www.weave.works/guides/service-discovery-and-load-balancing-with-weave-on-amazon-ecs-2/) with Scope in Standalone mode and repeatedly connect to one of the httpservers, everything works/looks as expected:

standalone_expected

and the containers show their Weave IP as expected:

standalone_details

Here's the report: [report_standalone_expected.json.gz](https://github.com/weaveworks/scope/files/169795/report_standalone_expected.json.gz)

However, if I launch it in Cloud Service mode, I miss the edges between the httpserver and the data-producers:

screen shot 2016-03-11 at 8 08 16 pm

but, unlike in https://github.com/weaveworks/scope/issues/643, the containers do show their Weave IPs

screen shot 2016-03-11 at 9 03 58 pm

Report: [report_service_unexpected.json.gz](https://github.com/weaveworks/scope/files/169802/report_service_unexpected.json.gz)

The logs from the `weavescope` container in one of the hosts don't reveal anything unusal: https://gist.github.com/2opremio/d231744001f69b51472d

Here are the arguments passed to the scope probe: 
```
[ec2-user@ip-172-31-0-8 ~]$ ps aux | grep scope-probe
root      2775  3.5  4.4 483840 44900 ?        Sl   19:04   0:39 scope-probe probe -docker=true -token=omitted scope.weave.works:443
```
 (it only adds `-token=omitted scope.weave.works:443` compared to the standalone demo)

@tomwilkie Any ideas? This looks remarkably similar to https://github.com/weaveworks/scope/issues/643 but:
* The probes seem to be talking to weave (see `Success collecting weave info` in the logs), plus I don't see how launching the probes in Service mode would make a difference in this respect.
* Restarting the probes to ensure they boot after Weave Net is running doesn't help (although it shouldn't matter since #510 is now closed).

Note: the screenshots were taken during different runs of the demo while investigating, so don't expect IPs and other metrics to be consistent across them.

paulbellamy:
So we can track weave without an exposed weave ip (e.g. in probe-only mode).

Fixes #1162

### DIFF
--- a/probe/overlay/weave.go
+++ b/probe/overlay/weave.go
@@ -8,6 +8,7 @@ import (
 	"github.com/weaveworks/scope/common/backoff"
 	"github.com/weaveworks/scope/common/weave"
 	"github.com/weaveworks/scope/probe/docker"
+	"github.com/weaveworks/scope/probe/host"
 	"github.com/weaveworks/scope/report"
 )
 
@@ -177,6 +178,14 @@ func (w *Weave) Report() (report.Report, error) {
 			WeavePeerName:     peer.Name,
 			WeavePeerNickName: peer.NickName,
 		}))
+	}
+	if w.statusCache.IPAM.DefaultSubnet != "" {
+		r.Overlay.AddNode(
+			report.MakeOverlayNodeID(w.statusCache.Router.Name),
+			report.MakeNode().WithSets(
+				report.MakeSets().Add(host.LocalNetworks, report.MakeStringSet(w.statusCache.IPAM.DefaultSubnet)),
+			),
+		)
 	}
 	return r, nil
 }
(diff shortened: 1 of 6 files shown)


probe/overlay/weave.go
```diff
@@ -178,5 +179,13 @@ func (w *Weave) Report() (report.Report, error) {
 			WeavePeerNickName: peer.NickName,
 		}))
 	}
+	if w.statusCache.IPAM.DefaultSubnet != "" {
+		r.Overlay.AddNode(
```
tomwilkie: Does it make sense to instead add this to the node in the host topology?

PR #1315, opened 2016-04-15
Removing report.Node.WithID

### LINKED ISSUE #836: Remove usage of report.Node.WithID and report.Node.WithTopology
From #752.

Big refactor, so pushed to 0.13 release.

Ideally something like:
```Go
t := report.MakeTopology("topologyName") // topologies have a name
node := report.MakeNode("nodeID") // nodes require an id
t = t.AddNode(node) // AddNode uses node.ID, and sets node.Topology
```

paulbellamy:
for after 0.14 release

Fixes #836

### DIFF
--- a/app/collector_test.go
+++ b/app/collector_test.go
@@ -19,10 +19,10 @@ func TestCollector(t *testing.T) {
 	c := app.NewCollector(window)
 
 	r1 := report.MakeReport()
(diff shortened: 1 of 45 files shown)


tomwilkie:
9885o

tomwilkie:
I think the test failure might be legit, @paulbellamy

PR #1318, opened 2016-04-18
Ensure we create a new terminal instance for each new pipe.

### LINKED ISSUE #1317: Terminal widget begin reused?
To reproduce:
- Attach to a container (ie get the logs)
- Without closing the terminal, Exec into the same container

Expected outcome:
- A new terminal replaces the existing one

Actual outcome:
- The shell seems to continue after the logs:

screen shot 2016-04-15 at 16 34 10

foot:
- Prevents new session from continuing in the same terminal as an old
  one.
- This is implemented using react's key behaviour at the moment: When
  the key changes, react unmounts and remounts the component and all
  term.js stuff inside of it. Could probably do this more explicitly.

Fixes #1317


davkal:
Works. LGTM

PR #1321, opened 2016-04-18
Better error message when plugin responses are too large

### LINKED ISSUE #1270: Plugin response body too large error message is confusing
Raised in: https://github.com/weaveworks/scope/pull/1126#discussion_r59366109

paulbellamy:
Fixes #1270 

There is a behavioural change, where plugins appear by socket filename until they have completed a successful report, and set their id/description/etc. This is so we can report errors better. Previously they would not have been shown.

### DIFF
--- a/probe/plugins/max_bytes_reader.go
+++ b/probe/plugins/max_bytes_reader.go
@@ -1,0 +1,43 @@
+package plugins
+
+import (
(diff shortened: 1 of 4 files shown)


tomwilkie:
LGTM.

PR #1327, opened 2016-04-18
Add explicit group node summariser instead of doing it in the other summaries

### LINKED ISSUE #1213: Consider adding generic "group" summariser
Raised from #1204

The summariser for containers-by-dns name should not be a special case of the container summariser; similarly, process-by-name should not be a special case of the process summariser.

And consider in the not-to-distant future we want to be able to group by anything.

So we should think about how we might summarise arbitrary groupings.

paulbellamy:
Fixes #1213

### DIFF
--- a/render/detailed/summary.go
+++ b/render/detailed/summary.go
@@ -15,30 +15,12 @@ import (
 
 // Shapes that are allowed
 const (
-	Circle   = "circle"
-	Square   = "square"
-	Heptagon = "heptagon"
-	Hexagon  = "hexagon"
-	Cloud    = "cloud"
-
 	ImageNameNone = "<none>"
 
 	// Keys we use to render container names
 	AmazonECSContainerNameLabel  = "com.amazonaws.ecs.container-name"
 	KubernetesContainerNameLabel = "io.kubernetes.container.name"
 	MarathonAppIDEnv             = "MARATHON_APP_ID"
-)
-
-var (
-	shapesByTopology = map[string]string{
-		report.Host:           Circle,
-		report.Process:        Square,
-		report.Pod:            Heptagon,
-		report.Service:        Heptagon,
-		report.Container:      Hexagon,
-		report.ContainerImage: Hexagon,
-		render.Pseudo:         Circle,
-	}
 )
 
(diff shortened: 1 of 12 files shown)


render/detailed/summary.go
```diff
@@ -29,6 +29,18 @@ const (
...
+	shapesByTopology = map[string]string{
+		report.Host:           Circle,
+		report.Process:        Square,
+		report.Pod:            Heptagon,
+		report.Service:        Heptagon,
+		report.Container:      Hexagon,
+		report.ContainerImage: Hexagon,
+		render.Pseudo:         Cloud,
+	}
+)
```
tomwilkie: I'm not a fan of hard coded things like this; perhaps we should add a `shape` field to the Topology struct?

render/detailed/summary.go
```diff
@@ -165,6 +183,7 @@ func pseudoNodeSummary(base NodeSummary, n report.Node) (NodeSummary, bool) {
 	// try rendering it as an endpoint
 	if addr, ok := n.Latest.Lookup(endpoint.Addr); ok {
 		base.Label = addr
+		base.Shape = Circle
```
tomwilkie: I feel like default should be Circle, and Cloud should be a special case.

render/detailed/summary.go
```diff
@@ -311,16 +293,38 @@ func hostNodeSummary(base NodeSummary, n report.Node) (NodeSummary, bool) {
...
+	base.Label, base.Rank = label, label
+
+	if count, ok := n.Counters.Lookup(parts[1]); ok {
+		base.LabelMinor = fmt.Sprintf("%d %s", count, parts[1])
+		if count != 1 {
+			if strings.HasSuffix(parts[1], "s") {
+				base.LabelMinor += "es"
+			} else {
+				base.LabelMinor += "s"
+			}
```
tomwilkie: This is generally frowned upon (programatically modifying text) as it localises very poorly.

I'd go for a lookup table (or a humanLabel & humanLabelPlural on the topology?)

tomwilkie:
LGTM with some minor comments.

PR #1328, opened 2016-04-18
Fix panning isolation

### LINKED ISSUE #1238: Layout center shared between layouts.

davkal:
* panning values were reused on when visiting a topology for the first
      time
    
 Fixes #1238


foot:
Nice! LGTM

Issue #1066, opened 2016-02-29
Scope errors with "websocket: close sent" and web UI not refreshed


alban:
I created a Kubernetes cluster based on `cluster.yaml` from https://coreos.com/kubernetes/docs/latest/kubernetes-on-aws.html#kube-aws-quickstart:

Then, I install Scope on Kubernetes with yaml files inspired by [weavescope](https://github.com/TheNewNormal/kube-charts/tree/master/weavescope/manifests)
- [scope-app-rc.yaml](https://gist.github.com/alban/4755128c677bd1f42f75)
- [scope-app-svc.yaml](https://gist.github.com/alban/995babe4c17bdf63ea40)
- [scope-probe-ds.yaml](https://gist.github.com/alban/f2d402c2a288571db640)

I am also installing some pods based on [guestbook-all-in-one.yaml](https://github.com/kubernetes/kubernetes/blob/master/examples/guestbook/all-in-one/guestbook-all-in-one.yaml).

I am using `ssh -L 4042:10.3.0.131:4040` and `http://localhost:4042/` to connect to scope app service from my laptop. I can see my containers in the web UI but after a couple of clicks to explore, the web UI is not refreshed: the "load in progress" animated icon is spinning. And I can see the following logs in Kubernetes:

```
$ kubectl --kubeconfig=clusters/demo-kubecon-1/kubeconfig logs weave-scope-app-rdiez
<app> INFO: 2016/02/29 17:11:49.412306 app starting, version 3e611f8, ID 7ad8e3ccecaa8e9e
<app> INFO: 2016/02/29 17:11:49.454402 listening on :4040
<app> ERRO: 2016/02/29 17:11:49.454679 Error updating weaveDNS, backing off 20s: Error running weave expose: exit status 1
<app> ERRO: 2016/02/29 17:12:09.494532 Error updating weaveDNS, backing off 40s: Error running weave expose: exit status 1
<app> ERRO: 2016/02/29 17:12:49.544389 Error updating weaveDNS, backing off 1m0s: Error running weave expose: exit status 1
<app> ERRO: 2016/02/29 17:15:23.538895 cannot serialize topology diff: websocket: close sent
```

I also check the logs on probes:

```
$ kubectl --kubeconfig=clusters/demo-kubecon-1/kubeconfig logs weave-scope-probe-ape8m
<probe> INFO: 2016/02/29 17:12:51.231727 probe starting, version 3e611f8, ID 9bb50c53ed8315a
[...]

PR #1330, opened 2016-04-18
Show k8s service IP in details panel

tomwilkie:
As part of #1066, it occurs to me that showing the service's virtual IP might be useful.


paulbellamy:
On top of having a `ClusterIP`, if they are a `NodePort` or `LoadBalancer` type service, they may have other IPs as well... Maybe call it `Cluster IP`? or `Internal IP`?

PR #1332, opened 2016-04-18
Support docker rename events

### LINKED ISSUE #1091: `docker rename` not reflected in the UI
1. launch scope
2. `docker run --name=foo -dti ubuntu` -> 'foo' shows up in scope
3. `docker rename foo bar` -> no change, even after waiting for a few minutes, clicking redraw, switching views etc

Note that the host details panel also shows the container as 'foo' instead of 'bar':
screenshot from 2016-03-01 23 24 37

but it's definitely (now) called 'bar':
```
$ docker ps -a
8f1fc8133014        ubuntu                    "/bin/bash"              5 minutes ago       Up 5 minutes                            bar
d6ef255fe439        weaveworks/scope:latest   "/home/weave/entrypoi"   5 minutes ago       Up 5 minutes                            weavescope
```

Restarting scope didn't solve the problem either! However, reloading the page (or just clicking 'redraw') afterwards did.

paulbellamy:
Fixes #1091 

@foot, doesn't propagate through to the canvas, just the details panel.

### DIFF
--- a/client/app/scripts/charts/nodes-chart.js
+++ b/client/app/scripts/charts/nodes-chart.js
@@ -333,9 +333,6 @@ export default class NodesChart extends React.Component {
 
     const stateNodes = this.initNodes(props.nodes, state.nodes);
     const stateEdges = this.initEdges(props.nodes, stateNodes);
-    const nodeMetrics = stateNodes.map(node => makeMap({
-      metrics: node.get('metrics')
-    }));
     const nodeScale = this.getNodeScale(props.nodes, state.width, state.height);
     const nextState = { nodeScale };
 
@@ -356,7 +353,7 @@ export default class NodesChart extends React.Component {
 
     // inject metrics and save coordinates for restore
     const layoutNodes = graph.nodes
-      .mergeDeep(nodeMetrics)
+      .mergeDeep(stateNodes)
       .map(node => node.merge({
         px: node.get('x'),
         py: node.get('y')
--- a/client/app/scripts/charts/nodes-chart.js
+++ b/client/app/scripts/charts/nodes-chart.js
@@ -352,12 +352,15 @@ export default class NodesChart extends React.Component {
     log(`graph layout took ${timedLayouter.time}ms`);
 
     // inject metrics and save coordinates for restore
-    const layoutNodes = graph.nodes
-      .mergeDeep(stateNodes)
+    let layoutNodes = graph.nodes
       .map(node => node.merge({
         px: node.get('x'),
         py: node.get('y')
       }));
+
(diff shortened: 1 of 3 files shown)


tomwilkie:
Backend LGTM, although am a little worries that it doesn't come through to the canvas.  Is the backend sending the right data to the frontend @paulbellamy ?

paulbellamy:
It shows the right data in the details panel, and when you refresh, but it doesn't in-place-update the canvas

tomwilkie:
Details panel is polled though, whereas the canvas is a push ws from the app.

paulbellamy:
The websocket is sending out the updated node label.

foot:
Cool will get the canvas behaving.

foot:
@davkal rfr not sure if this is the best way to do it?

client/app/scripts/charts/nodes-chart.js
```diff
@@ -356,7 +353,7 @@ export default class NodesChart extends React.Component {
 
     // inject metrics and save coordinates for restore
     const layoutNodes = graph.nodes
-      .mergeDeep(nodeMetrics)
+      .mergeDeep(stateNodes)
```
davkal: This needs to be done after the coords have been backed up (swap map and mergeDeep), because state nodes can have circle coords (just try select and deselect a node to see the error).

client/app/scripts/charts/nodes-chart.js
```diff
@@ -356,7 +353,7 @@ export default class NodesChart extends React.Component {
 
     // inject metrics and save coordinates for restore
```
davkal: Needs comment on why we're re-applying the state nodes, a la "reapply in case layout runner's node cache applied stale node metadata".

davkal:
Looks like `nodes-layout:copyLayoutProperties()` overwrites the nodes with data from the node cache. 😞 
Re-applying the `stateNodes` after the layout run works, but only after the node locations has been backed up.

I added #1353 as a techdebt ticket to rewrite that part in `nodes-layout`.

foot:
FE is ready to merge @paulbellamy

rade:
@foot What is the intended f/e behaviour? I've played with this and renaming does not update the UI. Even switching between views has no effect. However, the UI *does* update when hitting the 'redraw button'.

foot:
Intended behaviour is immediate instantaneous UI updates! So fast it changes before you can alt-tab back from the terminal and check that it changed!

Which is a bum, clear cache? I'll give it a quick test in Firefox now.

#1354 Should help w/ the caching issues if it turns out to be that.

rade:
> clear cache

That was it!

PR #1337, opened 2016-04-18
Fixes nodes w/ a slash under a path-prefixed scope.

### LINKED ISSUE #1335: Node IDs with / leads to redirect loop when scope is mounted under a path with slash redirect
Components you need are:

1) A load balancer in front of scope, which mounts it on a subdirectory, and redirects to add a trailing slash to scope's root url.
2) A node with a / in it's id (or label?) `docker run -dit -e MARATHON_APP_ID=a/foo --name foo alpine /bin/sh`
3) open a terminal on that node
4) close the terminal
5) observe redirect loop

foot:
E.g. localhost:4043/scoped/{state:{label:"/zing"...

Double encode is solution!
 - https://github.com/visionmedia/page.js/issues/187

Fixes #1335

### DIFF
--- a/client/app/scripts/utils/router-utils.js
+++ b/client/app/scripts/utils/router-utils.js
@@ -3,6 +3,12 @@ import page from 'page';
 import { route } from '../actions/app-actions';
 import AppStore from '../stores/app-store';
 
+//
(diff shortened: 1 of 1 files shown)


davkal:
LGTM

PR #1346, opened 2016-04-19
Change url encoding to handle multiple slashes in node names

### LINKED ISSUE #1335: Node IDs with / leads to redirect loop when scope is mounted under a path with slash redirect

foot:
Fixes #1335

### DIFF
--- a/client/app/scripts/utils/router-utils.js
+++ b/client/app/scripts/utils/router-utils.js
@@ -7,7 +7,16 @@ import AppStore from '../stores/app-store';
 // page.js won't match the routes below if ":state" has a slash in it, so replace those before we
 // load the state into the URL.
 //
+const SLASH = '/';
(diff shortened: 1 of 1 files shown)


foot:
To test create a "bad" node w/:

```
docker run -dit -e MARATHON_APP_ID=/foo/zing --name foo2 alpine /bin/sh
```

Then select it in a scope so its details are displayed, check the state in the url, then hit reload to make sure it doesn't redirect.

davkal:
LGTM

PR #1350, opened 2016-04-19
Add a 'Unmanaged' node to k8s views which included non-k8s containers.

### LINKED ISSUE #1246: K8s views have an uncontained node
Should we drop it for this view?

screen shot 2016-04-07 at 15 45 49

tomwilkie:
Fixes #1246

### DIFF
--- a/render/detailed/summary.go
+++ b/render/detailed/summary.go
@@ -162,6 +162,15 @@ func pseudoNodeSummary(base NodeSummary, n report.Node) (NodeSummary, bool) {
 		return base, true
 	}
 
+	// try rendering it as an unmanaged node
+	if strings.HasPrefix(n.ID, render.MakePseudoNodeID(render.UnmanagedID)) {
+		base.Label = render.UnmanagedMajor
+		base.Shape = Square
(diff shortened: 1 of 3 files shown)


render/detailed/summary.go
```diff
@@ -162,6 +162,15 @@ func pseudoNodeSummary(base NodeSummary, n report.Node) (NodeSummary, bool) {
 		return base, true
 	}
 
+	// try rendering it as an unmanaged node
+	if strings.HasPrefix(n.ID, render.MakePseudoNodeID(render.UnmanagedID)) {
+		base.Label = render.UnmanagedMajor
+		base.Shape = Square
```
paulbellamy: These are a stack of containers, no?

paulbellamy:
So, with this there will be both "Unmanaged" and "Uncontained"

tomwilkie:
> So, with this there will be both "Unmanaged" and "Uncontained", correct?

With this, the "Uncontained" processes are merged into "Unmanaged" too.  I went for a square but it could equally be the shape for a container.  WDYT?

PR #1354, opened 2016-04-20
Asset fingerprinting

davkal:
* adds `?hash` to JS and CSS bundle URLs against caching in production
* uses `html-webpack-plugin` to generate html dynamically
* html based on template in `app/html/index.html`
* per-app html files no longer needed
* `build/` dir is now completely deleteable, and fully gitignored
* top-level `make` now requires `client/app` to pick up on html and
  style changes
* cleaned up `server.js` code for production bundle testing
* added readme entry about testing production bundles


foot:
contrast mode on a production build doesn't appear to be working, but I can't immediately spot why...

foot:
terminal is working fine.

foot:
Awesommee! Works really well

LGTM

Issue #1355, opened 2016-04-20
Add Swarm to system container filters


pidster:
The Swarm master/agents are probably not interesting, and should be listed as system containers.

PR #1356, opened 2016-04-20
Add swarm-agent, swarm-agent master to sys container filters

pidster:
A small pull request, solving https://github.com/weaveworks/scope/issues/1355.

### DIFF
--- a/render/filters.go
+++ b/render/filters.go
@@ -250,12 +250,14 @@ func FilterApplication(r Renderer) Renderer {
 }
 
 var systemContainerNames = map[string]struct{}{
-	"weavescope": {},
-	"weavedns":   {},
-	"weave":      {},
-	"weaveproxy": {},
-	"weaveexec":  {},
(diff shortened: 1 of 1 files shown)


errordeveloper:
I'd also add it to `systemImagePrefixes`, as some folks might run with a different name, and it doesn't cost us anything to have it both places, don't you think?

tomwilkie:
> I'd also add it to systemImagePrefixes, as some folks might run with a different name, and it doesn't cost us anything to have it both places, don't you think?

In fact, I'd only add it to the images section - the swarm guide tell people to use different names to these for their containers:
https://docs.docker.com/swarm/install-manual/#step-2-create-your-instances

tomwilkie:
Also, pls try and compile it first:
```
./render/filters.go:259:23: missing ',' in composite literal
./render/filters.go:260:23: expected ';', found '{'
```

PR #1359, opened 2016-04-20
Use reverse-resolved DNS info in the connections table.

### LINKED ISSUE #1121: Use reverse DNS info
We could
- name pseudo nodes based on it
- use it to identify well know services and break them out of the internet node (#1082)
- in the connections details panel for the internet nodes (see #1017)

tomwilkie:
Fixes #1121 

screen shot 2016-04-20 at 17 12 21

<s>@davkal / @foot I added the `truncate` class to the details table cells, but I don't see to get tooltips.  Mind having a look?</s>

### DIFF
--- a/render/detailed/connections.go
+++ b/render/detailed/connections.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 	"strconv"
 
+	"github.com/weaveworks/scope/probe/endpoint"
 	"github.com/weaveworks/scope/render"
 	"github.com/weaveworks/scope/report"
 )
@@ -94,6 +95,8 @@ func incomingConnectionsSummary(topologyID string, r report.Report, n report.Nod
 					port:       port,
 				}
 				if isInternetNode(n) {
+					endpointNode := r.Endpoint.Nodes[localEndpointID]
+					key.localNode = &endpointNode
 					key.localAddr = localAddr
 				}
 				counts[key] = counts[key] + 1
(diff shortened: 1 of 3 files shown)


render/detailed/connections.go
```diff
@@ -206,10 +211,15 @@ func connectionRows(r report.Report, in map[connection]int, includeLocal bool) [
 			connection.Linkable = false
 		}
 		if includeLocal {
+			// Does localNode have a DNS record in it?
+			label := row.localAddr
+			if set, ok := row.localNode.Sets.Lookup(endpoint.ReverseDNSNames); ok && len(set) > 0 {
+				label = set[0]
+			}
```
paulbellamy: Just picking the first one seems a bit random. any better heuristics we could use? Shortest? Longest?

render/detailed/connections.go
```diff
@@ -144,6 +147,8 @@ func outgoingConnectionsSummary(topologyID string, r report.Report, n report.Nod
 					port:       port,
 				}
 				if isInternetNode(n) {
+					endpointNode := r.Endpoint.Nodes[remoteEndpointID]
+					key.localNode = &endpointNode
```
paulbellamy: let's inline these

tomwilkie: Can't take a pointer to a map entry unfortunately.

render/detailed/connections.go
tomwilkie: This is what we used to do.

I was tempter to instead do some string formatting, so it came out as "<hostname> (<ipadd>)".  WDYT?

paulbellamy: Could be quite long, but worth a shot.

render/detailed/connections.go
paulbellamy: eugh

tomwilkie:
PTAL @paulbellamy

PR #1360, opened 2016-04-20
Reorganise the render/ package

tomwilkie:
Aim here is for no code changes, but to break down mapping.go and topology.go into separate files for each topology, which container both the mapping functions and the topology renderers.


test/fixture/report_fixture.go
```diff
@@ -381,6 +382,7 @@ var (
 						kubernetes.PodState:        "running",
 						kubernetes.PodContainerIDs: ServerContainerID,
 						kubernetes.ServiceIDs:      ServiceID,
+						report.HostNodeID:          ServerHostNodeID,
```
paulbellamy: Why did these have to change?

paulbellamy:
The files in this, seem structured to match the probes (docker, kubernetes, process). Thoughts about having one-file-per-target? i.e. making them match `app/api_topologies.go` instead?

test/fixture/report_fixture.go
tomwilkie: Oh it didn't; I added it for completeness (all the other nodes have the host ID).

paulbellamy:
Discussed elsewhere, to rename the files according to "top-level" renderer (i.e. "process", "container", "pod", and "host"), then each file includes any sub-views (so ContainerHostname view goes in "container").

tomwilkie:
@paulbellamy PTAL.

PR #1365, opened 2016-04-21
Add a button to download the report as JSON

paulbellamy:
screen shot 2016-04-21 at 15 04 19

Tooltip says: "Save report as JSON"

Thoughts @davkal ? I'm not sure about the icon, but didn't know which would be a better one. Maybe it would be better to have one button that saves both report and svg?


davkal:
+1 on the icon, because it suggests debugging. Not sure about the title. "Report" is scope jargon. What's the user goal, maybe a better title can spring from that?

paulbellamy:
The user goal is for me, really. To save a report so I can attach it to an issue.

davkal:
I prefer "Save raw data as JSON" for now. Lets add it and see how often we ask to use it.

PR #1368, opened 2016-04-21
Add pod delete controls

tomwilkie:
Didn't add service, as it looks like [kubectl is responsible to tidying up pods...](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/delete.go)

### DIFF
--- a/probe/appclient/app_client.go
+++ b/probe/appclient/app_client.go
@@ -75,7 +75,7 @@ func NewAppClient(pc ProbeConfig, hostname, target string, control xfer.ControlH
 			TLSClientConfig: httpTransport.TLSClientConfig,
 		},
 		conns:   map[string]xfer.Websocket{},
-		readers: make(chan io.Reader),
+		readers: make(chan io.Reader, 2),
 		control: control,
 	}, nil
 }
@@ -273,6 +273,7 @@ func (c *appClient) Publish(r io.Reader) error {
 	select {
 	case c.readers <- r:
 	default:
+		log.Errorf("Dropping report to %s", c.target)
 	}
 	return nil
 }
--- a/render/pod.go
+++ b/render/pod.go
@@ -17,7 +17,7 @@ const (
 // graph by merging the container graph and the pods topology.
 var PodRenderer = MakeFilter(
 	func(n report.Node) bool {
-		// Drop deleted containers
+		// Drop deleted or empty pods
 		state, ok := n.Latest.Lookup(kubernetes.PodState)
 		return HasChildren(report.Container)(n) && (!ok || state != kubernetes.StateDeleted)
 	},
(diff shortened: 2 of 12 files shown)


tomwilkie:
@paulbellamy do you think we need to the filter/state=deleted trick here so pods go away instantly?  Its a lot more work that with docker.

paulbellamy:
We should. If we do it on containers, but not pods, then pods just seem broken.

report/id.go
```diff
@@ -175,6 +175,15 @@ func ParsePodNodeID(podNodeID string) (namespaceID, podID string, ok bool) {
...
 
+// ParseServiceNodeID produces the namespace ID and service ID from an service node ID.
+func ParseServiceNodeID(serviceNodeID string) (namespaceID, serviceID string, ok bool) {
+	fields := strings.SplitN(serviceNodeID, ScopeDelim, 2)
+	if len(fields) != 2 {
+		return "", "", false
+	}
+	return fields[0], fields[1], true
+}
+
```
paulbellamy: This is probably just in for consistency, but it's not actually used anywhere...

tomwilkie:
@paulbellamy I've added in the state / tombstone / shortcut reports stuff, and it kinda works.  Problem is multiple shortcut reports get generated very quickly (as when you delete pods, the RC recreates them very quickly), and this causes reports to get dropped.  I think we need to either add a buffer for reports, or start coalescing reports we might otherwise drop.  Thoughts?

paulbellamy:
merging instead of dropping sounds reasonable.

tomwilkie:
@paulbellamy I think this is ready to go in now.

probe/appclient/app_client.go
```diff
@@ -75,7 +75,7 @@ func NewAppClient(pc ProbeConfig, hostname, target string, control xfer.ControlH
 			TLSClientConfig: httpTransport.TLSClientConfig,
 		},
 		conns:   map[string]xfer.Websocket{},
-		readers: make(chan io.Reader),
```
paulbellamy: why 2?

render/pod.go
```diff
@@ -15,7 +15,12 @@ const (
 
 // PodRenderer is a Renderer which produces a renderable kubernetes
 // graph by merging the container graph and the pods topology.
-var PodRenderer = FilterEmpty(report.Container,
+var PodRenderer = MakeFilter(
+	func(n report.Node) bool {
+		// Drop deleted containers
```
paulbellamy: pods

probe/appclient/app_client.go
tomwilkie: Had to pick a number, and two allows for a pod deleted followed by a pod created event to be queued up.

report/id.go
tomwilkie: Yeah just for consistency.  Gonna get used real soon I think.

paulbellamy: I hope not. let's add it when we need it.

PR #1371, opened 2016-04-22
Fixing grouped node count for filtered children nodes

### LINKED ISSUE #815: Container labels are merged into the container image node when rendering topology
```sh
$ docker run -dit --name alpine1 alpine /bin/sh
$ docker run -dit --name alpine2 --label=works.weave.role=system alpine /bin/sh
```

Then go to container-by-image, and you will see that the "alpine" container image is considered a `system` image, because all metadata (including labels) has been merged up from the container nodes.

### LINKED ISSUE #1130: 'BY DNS NAME' and 'BY IMAGE' views do not filter out stopped containers
normal container view - there are 3 running containers
screenshot from 2016-03-05 10 30 51
showing stopped containers - there are 5 containers in total
screenshot from 2016-03-05 10 31 09
switching to 'BY IMAGE' view, not showing stopped containers - claims there are 5 containers in total, but in reality there are only 3
screenshot from 2016-03-05 10 33 09

### LINKED ISSUE #1219: Grouped node's counts include filtered nodes
e.g. on a container image node, it will include "Stopped" containers in the count, even though it is stopped, and lists them, even when they are filtered.

Related to #815

paulbellamy:
Fixes #1219, Fixes #1130

Does *not* fix #815 

Filtered children are still shown in the details panel, because otherwise one of them stopping could make the detail panel go empty, when the node disappears, which is slightly odd. But this means that a container image could say "1 container" (if it has 4 stopped and one running), then show all 5 containers in the detail panel, which is also slightly odd.

I'm not really sure about the naming of `APITopologyDesc.filteredRenderer`, or about having two different fields for this. I was hoping to have just one field, so all the renderers would have the same signature. But, since only some of them need the filters injected, it feels odd to force that requirement on the others.

This will (at the moment) also break memoisation for some of the rendering pipelines. There is an assumption that the map/reduce pipelines survive between requests, but this is no longer the case for all pipelines.

Will squash stuff before merging.

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -26,9 +26,9 @@ func init() {
 			ID:      "system",
 			Default: "application",
 			Options: []APITopologyOption{
-				{"system", "System services", render.FilterApplication},
-				{"application", "Application services", render.FilterSystem},
-				{"both", "Both", render.FilterNoop},
+				{"system", "System services", render.IsSystem},
+				{"application", "Application services", render.IsApplication},
+				{"both", "Both", render.Noop},
 			},
 		},
 	}
@@ -38,9 +38,9 @@ func init() {
 			ID:      "system",
 			Default: "application",
 			Options: []APITopologyOption{
-				{"system", "System pods", render.FilterApplication},
-				{"application", "Application pods", render.FilterSystem},
-				{"both", "Both", render.FilterNoop},
+				{"system", "System pods", render.IsSystem},
+				{"application", "Application pods", render.IsApplication},
+				{"both", "Both", render.Noop},
 			},
 		},
 	}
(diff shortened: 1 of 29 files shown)


paulbellamy:
@tomwilkie, PTAL

tomwilkie:
> But this means that a container image could say "1 container" (if it has 4 stopped and one running), then show all 5 containers in the detail panel, which is also slightly odd.

Perhaps we should include the containers state in the details panel child table?

paulbellamy:
state, and whether or not it's a system container...

tomwilkie:
Seem to have lots the stats on the number of containers filtered

tomwilkie:
And needs a good squash.  Otherwise, LGTM.

tomwilkie:
LGTM filter changes.

paulbellamy:
Squashed, will wait for circle, before merging

PR #1374, opened 2016-04-25
Don't show container labels on container images

### LINKED ISSUE #815: Container labels are merged into the container image node when rendering topology

paulbellamy:
Fixes #815

Images have their own labels. We can still use the container labels for
filtering, as the filters are "System Containers", not "System Images".


tomwilkie:
LGTM

PR #1375, opened 2016-04-25
Alternative approach to decorators.

tomwilkie:


paulbellamy:
Good:
* Renderers are more homogenous.
* Passing in the decorators as part of the Render(...) call seems simpler, as then all the params for a call are passed in in one location. But the decorators are still "curried" in in `renderedForRequest`, so that's not really the case.

Bad:
* Makes it more complex to predict where/when decorators are applied. The flow of decorators is hidden.
* All renderers take a "dead" argument, even when they don't need it.
* More types, methods, and abstractions.

paulbellamy:
Also, this has subtly broken memoisation. You need to include the decorators for the call as part of the cache key.

tomwilkie:
> All renderers take a "dead" argument, even when they don't need it.

Have fixed this (as discussed last night) - instead of capturing the filters in the render, we now pass them to the handlers.  This has actually simplified the node handler, as we can ignore the filters here and pass FilterNoop.

paulbellamy:
And this meant we can get rid of `WithDecorators`, which is nice. My remaining concerns are that the flow of the decorators is obscured (but less so now that it is passed with the report), and that this breaks memoisation. If we can get memoisation working, I think I'd be in favour of this.

tomwilkie:
Yeah, I'll try and get memoisation working.

tomwilkie:
Okay this should now pass (just skipping memoisation for render calls with decorators).

@paulbellamy want to take another look?

render/container_test.go
```diff
@@ -61,7 +61,7 @@ func TestContainerFilterRenderer(t *testing.T) {
 	input.Container.Nodes[fixture.ClientContainerNodeID] = input.Container.Nodes[fixture.ClientContainerNodeID].WithLatests(map[string]string{
 		docker.LabelPrefix + "works.weave.role": "system",
 	})
-	have := Prune(render.ContainerRenderer.Render(input, render.FilterSystem))
```
paulbellamy: As an aside, we should rename `ContainerWithImageNameRenderer` to just `ContainerRenderer`, as its the only one we use.

PR #1377, opened 2016-04-25
Report hostname and version in probe struct, and version in host node.

### LINKED ISSUE #1322: Reports should include the probe version and hostname
https://github.com/weaveworks/scope/issues/1311#issuecomment-211308284

tomwilkie:
Fixes #1322


tomwilkie:
@paulbellamy It occurs to me now that we could get rid of report.Probes and constitute all the information from the Host nodes.  WDYT?

paulbellamy:
So you'd add `probe_id`, `probe_version`, and `last_seen` to hosts?

tomwilkie:
This PR already adds scope_version to the host node, but yet.  Don't need to add last seen, can just use a timestamp from latest.  Or be explicit.

paulbellamy:
Sounds fine. Explicit is probably easier to work with, but slightly redundant.

Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY: all deps static clean client-lint client-test client-sync backend frontend shell lint
 
 # If you can use Docker without being root, you can `make SUDO= <target>`
-SUDO=$(shell (echo "$$DOCKER_HOST" | grep "tcp://" >/dev/null) || echo "sudo -E")
+SUDO=$(shell (docker info >/dev/null) || echo "sudo -E")
```
rade: Shouldn't that redirect stderr too?
```
$ docker info >/dev/null
Cannot connect to the Docker daemon. Is the docker daemon running on this host?
```
Also the inner parentheses are superfluous, aren't they?

tomwilkie:
@paulbellamy PTAL

PR #1378, opened 2016-04-25
Show the k8s load balancer IP if it is set

### LINKED ISSUE #1343: Show external IP for k8s external services

paulbellamy:
Fixes #1343 

Need to try this out on a k8s cluster which supports public load balancers


tomwilkie:
LGTM

PR #1381, opened 2016-04-26
Add generic path rewrite middleware

tomwilkie:


paulbellamy:
What about `r.URL`?

tomwilkie:
> What about r.URL?
> Edit: based on comments here, looks like we should update it based on url.Parse(r.RequestURI).

Its not needed for anything we do (the proxies only rely on RequestURI because of the broken golang url parsing).

And RequestURI is not a URL, so url.Parse won't work.

PR #1388, opened 2016-04-27
Migrate from Flux to Redux

davkal:
* better state visibility
* pure state changes
* state debug panel (show: crtl-h, move: ctrl-w)


foot:
Awesome stuff. Would be nice to have dev.html and index merged but can look into that later maybe.

LGTM!

PR #1397, opened 2016-04-28
Pass nil for the decorator in the rendering pipeline when possible

tomwilkie:
This improves performance as it allows us to memoize more.


paulbellamy:
Does having a nil outer decorator still calculate the "filtered" node count correctly, on e.g. the containers topology? I was of the vague impression that it might cause the "filtered" count to come from a filter "earlier" in the stack...

PR #1398, opened 2016-04-28
Don't merge nodes in the rendering pipeline

### LINKED ISSUE #1212: Whitelist metadata/metrics let through MapX2Y converters
Raised from #1204

tomwilkie:
Fixes #1212 

This make an almost 50% reduction in the cost of the rendering pipeline, reduces memory usage by 20% and makes a big difference to garbage load.

Unfortunately is breaks everything.  So we need to fix that....

### DIFF
--- a/render/id.go
+++ b/render/id.go
@@ -18,10 +18,11 @@ func MakeGroupNodeTopology(originalTopology, key string) string {
 
 // NewDerivedNode makes a node based on node, but with a new ID
 func NewDerivedNode(id string, node report.Node) report.Node {
-	return node.WithID(id).WithChildren(report.MakeNodeSet(node)).PruneParents()
+	return report.MakeNode(id).WithChildren(node.Children.Add(node))
 }
 
 // NewDerivedPseudoNode makes a new pseudo node with the node as a child
 func NewDerivedPseudoNode(id string, node report.Node) report.Node {
-	return NewDerivedNode(id, node).WithTopology(Pseudo)
+	output := NewDerivedNode(id, node).WithTopology(Pseudo)
+	return output
 }
--- a/render/container.go
+++ b/render/container.go
@@ -39,7 +39,7 @@ var ContainerRenderer = MakeFilter(
 			},
 			MakeMap(
 				MapProcess2Container,
-				ColorConnected(ProcessRenderer),
+				ProcessRenderer,
 			),
 		),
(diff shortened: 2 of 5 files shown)


tomwilkie:
@paulbellamy I think this should all work now.  PTAL

render/id.go
```diff
@@ -18,10 +18,19 @@ func MakeGroupNodeTopology(originalTopology, key string) string {
...
-	return NewDerivedNode(id, node).WithTopology(Pseudo)
+	output := NewDerivedNode(id, node).WithTopology(Pseudo)
+
+	// We propagate some metadata here, in a whitelist fashion.
+	if isConnected, timestamp, ok := node.Latest.LookupEntry(IsConnected); ok {
+		output.Latest = output.Latest.Set(IsConnected, timestamp, isConnected)
+	}
+	if hostNodeID, timestamp, ok := node.Latest.LookupEntry(report.HostNodeID); ok {
+		output.Latest = output.Latest.Set(report.HostNodeID, timestamp, hostNodeID)
+	}
```
paulbellamy: Maybe these should be propagated in `NewDerivedNode`, so they always carry over, not just on pseudonodes? wdyt?

tomwilkie: I put them here as we only use these propagated fields in pseudo nodes (AFAICT).  If we start using them in real nodes, then yes, we can move them.

paulbellamy:
Having read over it in more depth, and understood stuff a bit better, I think I prefer having the propagation of `HostNodeID` in `MapContainer2Pod` (the only place it's needed), with a comment saying why. Same for `IsConnected` in `MapProcess2Container` (if that's the only place it is needed?).

That way it's more explicit, and "obvious" where/why we're doing this.

paulbellamy:
Everywhere we use `ProcessRenderer` we wrap it in a `ColorConnected(ProcessRenderer)`. With this change, we might be able to move the `ColorConnected` into the definition of `ProcessRenderer`, so it is always included.

We depend on it in [render/detailed/summary.go:196](https://github.com/weaveworks/scope/blob/master/render/detailed/summary.go#L196), to determine if a process is "linkable" so (if possible) it would be very good to have consistency there, and make sure we don't "miss" it anywhere.

tomwilkie:
@paulbellamy PTAL?

PR #1401, opened 2016-04-28
Propagate network info for containers sharing network namespaces

### LINKED ISSUE #1394: Containers using other containers net namespace don't have IP addresses
This particularly affects k8s, where every container uses a pause containers net namespace.

The IP should show up in the details panel, and we should ignore pause containers for the purpose of rendering.

tomwilkie:
Fixes #1394

### DIFF
--- a/probe/docker/reporter.go
+++ b/probe/docker/reporter.go
@@ -49,6 +49,57 @@ var (
 	ContainerImageTableTemplates = report.TableTemplates{
 		ImageLabelPrefix: {ID: ImageLabelPrefix, Label: "Docker Labels", Prefix: ImageLabelPrefix},
 	}
+
+	ContainerControls = []report.Control{
+		{
+			ID:    AttachContainer,
+			Human: "Attach",
+			Icon:  "fa-desktop",
+			Rank:  1,
+		},
+		{
+			ID:    ExecContainer,
+			Human: "Exec shell",
+			Icon:  "fa-terminal",
+			Rank:  2,
+		},
+		{
+			ID:    StartContainer,
+			Human: "Start",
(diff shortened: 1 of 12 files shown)


paulbellamy:
I'm curious why the difference from `containerWithHostIPsRenderer`? It would be nice to do both the same way (either way is fine).

tomwilkie:
Yeah I was thinking of pull the `--net:host` stuff into here as well.  Will do.

paulbellamy:
Have also realized this will not handle `--net=container:containerName`

tomwilkie:
> Have also realized this will not handle --net=container:containerName

As discuss, this is a misfeature that we've decided not to support.

probe/docker/reporter.go
```diff
@@ -96,66 +147,73 @@ func (r *Reporter) Report() (report.Report, error) {
...
+		hostNetworkInfo := report.EmptySets
+		if hostIPs, err := getLocalIPs(); err == nil {
+			hostIPsWithScopes := addScopeToIPs(r.hostID, hostIPs)
+			hostNetworkInfo = hostNetworkInfo.
+				Add(ContainerIPs, report.MakeStringSet(hostIPs...)).
+				Add(ContainerIPsWithScopes, report.MakeStringSet(hostIPsWithScopes...))
+		}
+
+		var NetworkInfo func(prefix string) report.Sets
+		NetworkInfo = func(prefix string) report.Sets {
```
paulbellamy: Why is this capital (and defined on 2 lines?)

paulbellamy:
Quibble about naming, but otherwise, LGTM.

Issue #1404, opened 2016-04-28
k8s: Cross-host pod connections not properly rendered or missing


2opremio:
Pod connections (no services nor kube-proxy involved) across hosts in kubernetes are not properly reflected in the UI. 

Following a modified example from https://github.com/kelseyhightower/craft-kubernetes-workshop , I am using an nginx `frontend` pod (IP 10.152.1.6) which uses persistent connections to loadbalance requests to some `hello` pod backends with IPs (10.152.1.5, 10.152.0.3 and 10.152.4.5). 

nginx is using this configuration:

```
upstream hello {                                                                                                                                            
    server 10.152.1.5;                                                                                                                                      
    server 10.152.0.3;                                                                                                                                      
    server 10.152.4.5;                                                                                                                                      
    keepalive 100;                                                                                                                                            
}                                                                                                                                                           
                                                                                                                                                            
upstream auth {                                                                                                                                             
    server auth.default.svc.cluster.local;                                                                                                                  
    keepalive 2;                                                                                                                                            
}                                                                                                                                                           
                                                                                                                                                            
server {                                                                                                                                                    
    listen 443;                                                                                                                                             
    ssl    on;                                                                                                                                              
                                                                                                                                                            
    ssl_certificate     /etc/tls/cert.pem;                                                                                                                  
    ssl_certificate_key /etc/tls/key.pem;                                                                                                                   
                                                                                                                                                            
    proxy_http_version 1.1;                                                                                                                                 
    proxy_set_header Connection "";                                                                                                                         
                                                                                                                                                            
    location / {                                                                                                                                            
        proxy_pass http://hello;                                                                                                                            
    }                                                                                                                                                       
                                                                                                                                                            
    location /login {                                                                                                                                       
        proxy_pass http://auth;                                                                                                                             
    }                                                                                                                                                       
}
```

The frontend shows connections to all the pods:

```
$ netstat -tpn | grep ESTABLISHED  | grep 10.152.0.3                                                                                                         
tcp        0      0 10.152.1.6:60463        10.152.0.3:80           ESTABLISHED -                                                                           
$ netstat -tpn | grep ESTABLISHED  | grep 10.152.1.5                                                                                                         
tcp        0      0 10.152.1.6:42595        10.152.1.5:80           ESTABLISHED -                                                                           
$ netstat -tpn | grep ESTABLISHED  | grep 10.152.4.5                                                                                                         
[...]

PR #1410, opened 2016-04-29
Correctly attribute DNAT-ed short-lived connections

### LINKED ISSUE #1408: k8s: Short-lived connections not being tracked
In a local single-node k8s cluster, I create a frontend nginx-based Service, loadbalancing to requests to a single service called `hello`:

```
# create local kubernetes cluster
(cd service && infra/local-k8s up )

git clone git@github.com:kelseyhightower/craft-kubernetes-workshop.git 
cd craft-kubernetes-workshop/kubernetes
git checkout ea9a3ad47f80d7a4f1f0bdeb0ca7d26fd1bdd741

kubectl create secret generic tls-certs --from-file=tls
kubectl create configmap nginx-frontend-conf --from-file=nginx/frontend.conf
kubectl create -f deployments/hello.yaml
kubectl create -f services/hello.yaml
kubectl create -f deployments/auth.yaml
kubectl create -f services/auth.yaml 
kubectl create -f deployments/frontend.yaml
kubectl create -f services/frontend.yaml 

# add load to the loop
while true; do curl --insecure "https://$(kubectl get service  frontend  --template '{{.spec.clusterIP}}')"; sleep 1; done
```

* The resulting `frontend` service gets virtual IP 10.0.0.6 mapped to a single pod with IP 172.17.0.105.
* The resulting `hello` service gets virtual IP 10.0.0.242 and mapped to a single pod with IP 172.17.0.76.

I would expect:

* An edge between an `Uncontained` node (curl)  to the ngnix container
* An edge between the `nginx` container and the `hello` container.

But I don't see any of those edges:

screen shot 2016-04-29 at 01 35 02

The edges appear if I use persistent connections though (i.e. adapting the repro of #1404 applied to the local single-node cluster).

However, I do get an outbound `Internet` connection from the `/pause` container of the frontend (but not from the `nginx` container itself due to https://github.com/weaveworks/scope/issues/1394 ):

screen shot 2016-04-29 at 05 50 50

Using kube-proxy in userspace mode (`--proxy-mode=userspace`) does **not** help. It doesn't cause any new edges to appear (not even with kube-proxy rendered in-between, see https://github.com/weaveworks/scope/issues/1032).

If we look at the report ([report_k8s_missing_short_lived_conns.json.gz](https://github.com/weaveworks/scope/files/241658/report_k8s_missing_short_lived_conns.json.gz)) we see that conntrack gets connections between the service and  pod IPs. Here are the endpoints between the frontend pod and the hello service virtual IP:

screen shot 2016-04-29 at 02 04 36

screen shot 2016-04-29 at 02 25 09

The virtual IP doesn't have any edges and there is no endpoint for `172.17.0.76;80` (the destination).

Note that the containers are not running in the host networking namespace so they shouldn't be affected by #1260 .

2opremio:
Fixes #1408

### DIFF
--- a/probe/endpoint/reporter.go
+++ b/probe/endpoint/reporter.go
@@ -106,7 +106,7 @@ func (r *Reporter) Report() (report.Report, error) {
 	rpt := report.MakeReport()
 	seenTuples := map[string]fourTuple{}
 
-	// Consult the flowWalker for short-live connections
+	// Consult the flowWalker for short-lived connections
 	{
 		extraNodeInfo := map[string]string{
 			Conntracked: "true",
@@ -120,6 +120,25 @@ func (r *Reporter) Report() (report.Report, error) {
 			}
 			seenTuples[tuple.key()] = tuple
 			r.addConnection(&rpt, tuple, extraNodeInfo, extraNodeInfo)
+
+			// Handle DNAT-ed short-lived connections.
+			// The NAT mapper won't help since it only runs periodically,
+			// missing the short-lived connections.
+			if f.Original.Layer3.DstIP != f.Reply.Layer3.SrcIP {
+				reply_tuple := fourTuple{
+					f.Reply.Layer3.DstIP,
+					f.Reply.Layer3.SrcIP,
+					uint16(f.Reply.Layer4.DstPort),
+					uint16(f.Reply.Layer4.SrcPort),
+				}
+				// FIXME: For DNAT-ed connections the
+				// f.Original.Layer3.SrcIP -> f.Original.Layer3.DstIP connection above
+				// results in a unattributed dangling-edge to the Internet node.
+				// It could be enough to simply not add that connection, but
+				// it may not be correct in general.
+				r.addConnection(&rpt, reply_tuple, extraNodeInfo, extraNodeInfo)
+			}
+
 		})
 	}
 
--- a/probe/endpoint/reporter.go
+++ b/probe/endpoint/reporter.go
@@ -118,27 +118,20 @@ func (r *Reporter) Report() (report.Report, error) {
 				uint16(f.Original.Layer4.SrcPort),
 				uint16(f.Original.Layer4.DstPort),
 			}
-			seenTuples[tuple.key()] = tuple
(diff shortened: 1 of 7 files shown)


probe/endpoint/reporter.go
```diff
@@ -120,6 +120,25 @@ func (r *Reporter) Report() (report.Report, error) {
...
+			// The NAT mapper won't help since it only runs periodically,
+			// missing the short-lived connections.
+			if f.Original.Layer3.DstIP != f.Reply.Layer3.SrcIP {
+				reply_tuple := fourTuple{
+					f.Reply.Layer3.DstIP,
+					f.Reply.Layer3.SrcIP,
+					uint16(f.Reply.Layer4.DstPort),
+					uint16(f.Reply.Layer4.SrcPort),
+				}
+				// FIXME: For DNAT-ed connections the
```
2opremio: This is the pending issue to resolve in this PR.

2opremio:
This fixes the problem before rebasing ( https://github.com/weaveworks/scope/tree/save-dnat-shortlived-connections ), where I could see edges between `/pause` containers. However, after rebasing (which included https://github.com/weaveworks/scope/pull/1401 and https://github.com/weaveworks/scope/pull/1415 ) I don't get an edge between the application containers.

Here's a report:  [report_pause_maybe_not_discarded.json.gz](https://github.com/weaveworks/scope/files/249060/report_pause_maybe_not_discarded.json.gz)

I would expect an edge between the hello and the nginx containers (IPS 10.35.2.5 and 10.36.0.4) (Container IDs: fddd54c00c50c769b596accc23ce5867630813eba371f917c9183f8bee835347 and 5684fae9fec5ad02174af0939ea7c60d374b05af8dd5013124810e3f91833114)

2opremio:
@tomwilkie Mind taking a look?

tomwilkie:
I think we should add a minimal test covering this specific case, similar to short_lived_connections_test.go

Otherwise LGTM

2opremio:
It also fixes #1404

tomwilkie:
LGTM (assuming its green)

PR #1414, opened 2016-04-29
Make Scope work on Docker for Mac

### LINKED ISSUE #1411: `scope launch` stopped working agains latest Docker for Mac (beta9)
The new beta9 release has only so called "VPN-compatible" mode. In this mode container ports are accessible on `localhost`, but no access to host network is provided (as it bypasses Docker port publishing machinery).

```
0 %> ~/Code/scope/scope launch
ec952734f2df481f8a220024a79a5c712e2d45aa5476a762ee17fa49d95c335d
  * http://192.168.65.2:4040/
  * http://172.19.0.1:4040/

0 %> curl http://192.168.65.2:4040/
^C

1 %> curl http://172.19.0.1:4040/
^C

1 %> curl http://localhost:4040/
curl: (7) Failed to connect to localhost port 4040: Connection refused

7 %> ping -c1 docker.local
ping: cannot resolve docker.local: Unknown host

68 %> docker run -ti --net=host centos:7 curl localhost:4040
<!doctype html>
<html class="no-js">
  <head>
    <meta charset="utf-8">
    <title>Weave Scope</title>
  <link href="style-app.css?22d50f02be7313bbc64a" rel="stylesheet"></head>
  <body>
    <!--[if lt IE 10]>
      <p class="browsehappy">You are using an <strong>outdated</strong> browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
    <![endif]-->
    <div class="wrap">
      <div id="app"></div>
    </div>
  <script src="vendors.js?22d50f02be7313bbc64a"></script><script src="app.js?22d50f02be7313bbc64a"></script></body>
</html>
```

Yet, the app is fully accessible inside of the host. My assumption is that splitting app and probe and running the app in it's own network namespace should work, but I haven't tested it yet.

errordeveloper:
This fixes #1411.

### DIFF
--- a/scope
+++ b/scope
@@ -2,6 +2,8 @@
 
 set -eu
 
+ARGS="$@"
+
 usage() {
     echo "Usage:"
     echo "scope launch [<peer> ...]"
@@ -22,6 +24,7 @@ IMAGE_VERSION=${VERSION:-$IMAGE_VERSION}
 SCOPE_IMAGE_NAME=weaveworks/scope
 SCOPE_IMAGE=$SCOPE_IMAGE_NAME:$IMAGE_VERSION
 SCOPE_CONTAINER_NAME=weavescope
+SCOPE_APP_CONTAINER_NAME=weavescope-app
 IP_REGEXP="[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}"
 IP_ADDR_CMD="find /sys/class/net -type l | xargs -n1 basename | grep -vE 'docker|veth|lo' | \
     xargs -n1 ip addr show | grep inet | awk '{ print \$2 }' | grep -oE '$IP_REGEXP'"
@@ -61,6 +64,17 @@ check_docker_version() {
     fi
 }
 
+check_probe_only() {
+  echo "${ARGS}" | grep -q -E "\-\-no\-app|\-\-service\-token"
+}
+
+check_docker_for_mac() {
+  docker_for_mac_network_mode="~/Library/Containers/com.docker.docker/Data/database/com.docker.driver.amd64-linux/network"
+  [ "$(uname)" = "Darwin" ] \
+    && [ -S /var/run/docker.sock ] \
+    && grep -q hybrid $docker_for_mac_network_mode 2>/dev/null
+}
+
 # Check that a container named $1 with image $2 is not running
 check_not_running() {
     case $(docker inspect --format='{{.State.Running}} {{.Config.Image}}' $1 2>/dev/null) in
@@ -97,6 +111,13 @@ launch_command() {
             $WEAVESCOPE_DOCKER_ARGS $SCOPE_IMAGE --probe.docker=true "$@"
 }
 
+launch_app_command() {
+    echo docker run -d --name=$SCOPE_APP_CONTAINER_NAME \
+            -e CHECKPOINT_DISABLE \
+            -p 0.0.0.0:4040:4040 \
+            $WEAVESCOPE_DOCKER_ARGS $SCOPE_IMAGE --no-probe "$@"
+}
+
 check_docker_version
 
 case "$COMMAND" in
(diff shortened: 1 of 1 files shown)


tomwilkie:
Can we do better and just launch an app and probe if we detect docker 4 mac?

errordeveloper:
Supposed we could, but this may be a temporary fix, yet we don't know. I just wanted to make smallest possible change, didn't want to handle all different case. For example the container ID we output, there would have to be two or what?

errordeveloper:
I was also debating we could spin up the app from the probe container, once we figured the mode of operation as the way it's currently running two process is very awkward anyway.

tomwilkie:
> Supposedly we could, but this may be a temporary fix, yet we don't know. I just wanted to make smallest possible change, didn't want to handle all different case. For example the container ID we output, there would have to be two or what?

I wouldn't worry about the scope script output; just output both IDs.  Its more important that it "just works" out the box.

errordeveloper:
@tomwilkie I've done as you suggested, and I also discovered that WeaveDNS can only be contacted in the host network namespace, hence `${app_ip}:4040` is now passed to the probe.

errordeveloper:
@pidster still has some concerns regarding putting this kind of fix into the release...

scope
```diff
@@ -61,6 +64,16 @@ check_docker_version() {
     fi
 }
 
+check_probe_only() {
+  echo "${ARGS}" | grep -q -E "\-\-no\-app|\-\-service\-token"
```
tomwilkie: `--probe-only` should be checked for too.

scope
```diff
@@ -61,6 +64,16 @@ check_docker_version() {
...
+}
+
+check_docker_for_mac() {
+  [ "$(uname)" = "Darwin" ] \
+    && [ -S /var/run/docker.sock ] \
+    && grep -q hybrid ~/Library/Containers/com.docker.docker/Data/database/com.docker.driver.amd64-linux/network
```
tomwilkie: pls extract this path out into a variable to make this more readable.

scope
```diff
@@ -128,18 +148,33 @@ EOF
         ;;
 
     launch)
-        check_not_running $SCOPE_CONTAINER_NAME $SCOPE_IMAGE_NAME
-        docker rm -f $SCOPE_CONTAINER_NAME >/dev/null 2>&1 || true
-
-        CONTAINER=$($(launch_command "$@"))
-        echo $CONTAINER
+        if ! check_docker_for_mac || check_probe_only ; then
```
tomwilkie: Some comments (with reference back to the issue) here would be helpful.

scope
```diff
@@ -128,18 +148,33 @@ EOF
...
+            echo $CONTAINER
+        elif check_docker_for_mac ; then
+            check_not_running $SCOPE_APP_CONTAINER_NAME $SCOPE_IMAGE_NAME
+            docker rm -f $SCOPE_APP_CONTAINER_NAME >/dev/null 2>&1 || true
+            CONTAINER=$($(launch_app_command "$@"))
+            echo $CONTAINER
+            app_ip=$(docker inspect -f '{{.NetworkSettings.IPAddress}}' "${CONTAINER}")
+            check_not_running $SCOPE_CONTAINER_NAME $SCOPE_IMAGE_NAME
+            docker rm -f $SCOPE_CONTAINER_NAME >/dev/null 2>&1 || true
+            CONTAINER=$($(launch_command --no-app "${app_ip}:4040" "$@"))
```
tomwilkie: Does `"$@"` need to come before `"${app_ip}:4040"` as `"$@"` might container flags.

tomwilkie:
Some nits, but looking better.

scope
errordeveloper: Haven't put that to consideration, as does the order matter?

scope
errordeveloper: Noted.

scope
errordeveloper: Ok, I've just moved the code which was there already, happy to add this.

scope
tomwilkie: Yes the order matters.

errordeveloper: All comments had been addressed.

tomwilkie:
Doesn't work on my machine:
```
Toms-MacBook-Pro:scope twilkie$ git status
On branch fix-1411
Your branch is up-to-date with 'origin/fix-1411'.
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	Vagrantfile
	vagrant-common.rb

nothing added to commit but untracked files present (use "git add" to track)
Toms-MacBook-Pro:scope twilkie$ ./scope launch
f62c0b804bfcb735b48cc5e0e2ad990aa4c304046a78904da15c063085b0ce01
  * http://192.168.65.2:4040/
Toms-MacBook-Pro:scope twilkie$ docker ps
f62c0b804bfc        weaveworks/scope:latest   "/home/weave/entrypoi"   2 seconds ago       Up 2 seconds                            weavescope
Toms-MacBook-Pro:scope twilkie$ 
```

I should be expecting two containers, right?

PR #1418, opened 2016-05-03
A caching, log(n) complexity report merger

tomwilkie:
Contains #1398

With this, the vast majority (7 of 8s) of CPU is spent decoding reports.

### DIFF
--- a/app/collector.go
+++ b/app/collector.go
@@ -1,11 +1,9 @@
 package app
 
 import (
-	"fmt"
 	"sync"
 	"time"
 
-	"github.com/spaolacci/murmur3"
 	"golang.org/x/net/context"
 
 	"github.com/weaveworks/scope/common/mtime"
@@ -35,10 +33,12 @@ type Collector interface {
 // Collector receives published reports from multiple producers. It yields a
 // single merged report, representing all collected reports.
 type collector struct {
-	mtx     sync.Mutex
-	reports []timestampReport
-	window  time.Duration
-	cached  *report.Report
+	mtx        sync.Mutex
+	reports    []report.Report
+	timestamps []time.Time
+	window     time.Duration
+	cached     *report.Report
+	merger     Merger
 	waitableCondition
 }
 
@@ -78,6 +78,7 @@ func NewCollector(window time.Duration) Collector {
 		waitableCondition: waitableCondition{
 			waiters: map[chan struct{}]struct{}{},
 		},
+		merger: NewSmartMerger(),
 	}
 }
 
@@ -85,8 +86,10 @@ func NewCollector(window time.Duration) Collector {
 func (c *collector) Add(_ context.Context, rpt report.Report) error {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
-	c.reports = append(c.reports, timestampReport{mtime.Now(), rpt})
-	c.reports = clean(c.reports, c.window)
+	c.reports = append(c.reports, rpt)
+	c.timestamps = append(c.timestamps, mtime.Now())
+
+	c.clean()
 	c.cached = nil
 	if rpt.Shortcut {
 		c.Broadcast()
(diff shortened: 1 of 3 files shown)


app/merger.go
```diff
@@ -0,0 +1,111 @@
...
+	left, right *node
+}
+
+type byID []*node
+
+func (ns byID) Len() int           { return len(ns) }
+func (ns byID) Swap(i, j int)      { ns[i], ns[j] = ns[j], ns[i] }
+func (ns byID) Less(i, j int) bool { return ns[i].id < ns[j].id }
+
+func hash(_1, _2 string) string {
```
peterbourgon: a, b please

app/collector.go
```diff
@@ -110,15 +110,11 @@ func (c *collector) Report(_ context.Context) (report.Report, error) {
...
 	c.reports = clean(c.reports, c.window)
 
-	rpt := report.MakeReport()
-	id := murmur3.New64()
+	rpts := []report.Report{}
 	for _, tr := range c.reports {
-		rpt = rpt.Merge(tr.report)
-		id.Write([]byte(tr.report.ID))
+		rpts = append(rpts, tr.report)
 	}
```
paulbellamy: If you split timestampReport into two arrays (one for timestamps, the other for reports, keeping them in sync), you could eliminate this loop, and copying all the reports. That might not be that expensive, though, as there should only be ~15 of them?

tomwilkie: There will be 15/3 * #hosts.  So could be lots.  I doubt the copy is expensive, but we might as well do this.

paulbellamy:
You don't really need the cache if you just hold a pointer to the root of the tree (i.e. the "current report")

tomwilkie:
> You don't really need the cache if you just hold a pointer to the root of the tree (i.e. the "current report")

How so?  The algorithm would look very different (have to build and diff two trees).  Also I don't know how that would work in the multitenant case.

I actually remove the left & right pointers from the node in the latest rev...

tomwilkie:
@paulbellamy I'm more concerned about the algorithm TBH.  I sort the reports by ID and then take each pair iteratively.  As the IDs are random, this means removing a single report will mean we have to re-merge 1/2 the reports again (as they will all 'shift up' by one).

I think a better approach might be to subdivide the address space in two each time, more top-down than bottom up.  WDYT?

tomwilkie:
@paulbellamy WDYT?

paulbellamy:
I'd be interested in seeing a benchmark of a large number of report merges before/after this change.

paulbellamy:
I'm also curious about the choice of branching factor. i.e. why a binary tree, not e.g. 32 leaf nodes?

tomwilkie:
> I'm also curious about the choice of branching factor. i.e. why a binary tree, not e.g. 32 leaf nodes?

No particular reason; the branching factor is just falls out as a constant factor in the big-o anyway.

tomwilkie:
> I'm also curious about the choice of branching factor. i.e. why a binary tree, not e.g. 32 leaf nodes?

Also merge is binary.

tomwilkie:
> I'd be interested in seeing a benchmark of a large number of report merges before/after this change.

```
Toms-MacBook-Pro:app twilkie$ go test -bench Benchmark
...                 
PASS
BenchmarkSmartMerger-8              	      20	  58676996 ns/op	22620731 B/op	  137828 allocs/op
BenchmarkSmartMergerWithoutCaching-8	      20	  66029486 ns/op	33815075 B/op	  214389 allocs/op
BenchmarkDumbMerger-8               	       3	 455741265 ns/op	198807458 B/op	 1685880 allocs/op
ok  	github.com/weaveworks/scope/app	7.617s
```

app/merger.go
```diff
@@ -0,0 +1,142 @@
...
+			id:  id,
+			rpt: r,
+		})
+	}
+	sort.Sort(byID(nodes))
+
+	// Define how to merge two nodes together.  The result of merging
+	// two reports is cached.
+	hits := 0
+	misses := 0
```
paulbellamy: probably don't need to track these anymore

app/merger_test.go
```diff
@@ -0,0 +1,126 @@
...
+	benchmarMerger(b, app.NewSmartMerger(), true)
+}
+
+func BenchmarkDumbMerger(b *testing.B) {
+	benchmarMerger(b, app.MakeDumbMerger(), false)
+}
+
+const numHosts = 15
+
+func benchmarMerger(b *testing.B, merger app.Merger, clearCache bool) {
```
paulbellamy: benchmark

app/merger_test.go
```diff
@@ -0,0 +1,126 @@
...
+		ClearCache()
+	}); ok && clearCache {
+		clearable.ClearCache()
+	}
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		b.StartTimer()
```
paulbellamy: You're not doing anything outside the timer, so you don't need to start/stop it. Either that, or you should just start it around the actual "Merge" call.

tomwilkie:
@paulbellamy PTAL

PR #1357, opened 2016-04-20
Move all the flag parsing to main.go; implement `scope help`

tomwilkie:
Fixes #1312, fixes #768, fixes #774

PR #1419, opened 2016-05-03
Honour the --service-token flag properly.

tomwilkie:
Bug from #1357


paulbellamy:
why two flags for the same thing?

tomwilkie:
> why two flags for the same thing?

legacy; --service-token used to be translated into -token on the probe.  Ie --probe.token.

Issue #1299, opened 2016-04-14
It is difficult to find nodes with a specific attribute


paulbellamy:
e.g. I'm trying to find the container with ip `172.16.0.31`, and have to click through all the containers to find it.

davkal:
I like this issue a lot. It drives us closer to a search functionality.

I can imagine a search bar that does not only check visible labels, but also matches metadata and maybe overlays the matches per node below the current labels.

tomwilkie:
Search is one solution, for sure.

A table, with multiple columns you can sort on, is another.  People are good at searching through sorted lists.

I think the problem is more general - finding a node by name is equally as difficult.

paulbellamy:
Could highlight matching nodes as you "search"... That would be GoodEnough for me in this situation

PR #1416, opened 2016-05-02
Restructured Scope Docs

abuehrle:
Here's a first draft of the docs for the .15 release.  Still to do:

- [ ] ECS AMIs
- [x] incorporate changes from https://github.com/weaveworks/scope/pull/1426
- [x] incorporate changes from https://github.com/weaveworks/scope/pull/1434

Punted for now:
- Redraw the architecture drawings, ACII doesn't look great in WP
- Add sample app that describes how to troubleshoot issues (if there is time)
- Screen captures and description of Views, filtering, and metrics <--- the new grid view here as well?
- Update kubernetes namespace filters (are they implemented as filters or as views?)
- Weave Net integration?

Please have a look and let me if the structure is ok, and any other topics you think I should add.

PR #1429, opened 2016-05-04
Search on canvas

davkal:
screen shot 2016-05-04 at 12 29 23

Features:

- [x] adds a search field next to the topologies
- [x] highlight results on canvas as you type
- [x] non-matching nodes are grayed out
- [x] "prefix:" limits search to field label
- [x] apply query as filter on Enter
- [x] match metrics, e.g. `cpu > 2` or `memory > 4MB`
- [x] store (pinned) searches in URL
- [x] sanitize search queries

Fixups:

- [x] check big topologies
- [x] tweak match overlays to be always at the top
- [x] sublables overlap labels in focus mode
- [x] removing all input fast brings previous input back


tomwilkie:
if I type "python" and press enter, I get a blob/pill.  I would like to be able to delete the blob/pill using backspace.

tomwilkie:
Also I couldn't quite figure out how to do the metric search (less than, greater than etc).  Perhaps making the hints clickable (to see next hint) would help with discoverability?

tomwilkie:
Also switching between topologies seem to have some bugs.  When I search for "python" (which is not in the label or sublabel) is highlights the command (good), but then I switch away from the containers view to the the processes view and back, and they are gone.

davkal:
> (But only briefly - they come back after a few seconds)

This is might be a backend issue. A lot of the time, the backend does not send the `metadata` field of nodes in the delta.

2opremio:
Fixes #1299

foot:
I have noticed that the BE doesn't send metadata initially (w/ `add`d nodes) too.

client/app/scripts/charts/node.js
```diff
@@ -1,10 +1,13 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { connect } from 'react-redux';
-import classNames from 'classnames';
+import cx from 'classnames';
```
foot: I have slight bias towards more explicit name here.

foot:
- [ ] z-index of labels is a bit funny (not sure if they overlapped other nodes before...). But now that they are nicely wrapper the text can end up behind other nodes when hovering. Possible solution: draw hovered node last (svg equiv z-index.

2opremio:
- [ ] Extend #1416 with how to use search (e.g. labels and filters).

client/app/scripts/charts/node.js
davkal: Yup, old habits. `cx` was the import of the old react.addons.classset that was superceded by classnames.

client/app/scripts/charts/nodes-chart.js
```diff
@@ -404,7 +401,7 @@ function mapStateToProps(state) {
   return {
     adjacentNodes: getAdjacentNodes(state),
     forceRelayout: state.get('forceRelayout'),
-    nodes: state.get('nodes'),
+    nodes: state.get('nodes').filter(node => !node.get('filtered')),
```
foot: 👍

client/app/scripts/components/search.js
```diff
@@ -0,0 +1,138 @@
...
+      const metadataField = node.get('metadata').first();
+      metadataLabel = shortenHintLabel(slugify(metadataField.get('label')))
+        .split('.').pop();
+      metadataValue = shortenHintLabel(metadataField.get('value'));
+    }
+  }
+
+  return `Try "${label}", "${metadataLabel}:${metadataValue}", or "cpu > 2%".
+   Hit enter to apply the search as a filter.`;
+}
```
foot: Neat :). Didn't realise they were that dynamic.

foot: Cool for next step of goog-suggest-esqueness.

client/app/scripts/components/search.js
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  handleBlur() {
+    this.props.blurSearch();
+  }
+
+  handleChange(ev) {
+    const inputValue = ev.target.value;
+    let value = inputValue;
+    // need to distinguish quick clear, setting null
```
foot: What is quick clear?

2opremio:
The animations of the nodes are a bit strange. When hovering a node, it jumps and then the other nodes jump. Is that intentional?



Also, when clicking on a "searched" node, the surrounding nodes get grayed out, which I guess is consistent with the view, but maybe a bit unpleasant because you cannot see the edges properly (no strong opinions about this one though).

screen shot 2016-05-09 at 1 57 34 pm

2opremio:
In contrast mode the search box is not well defined since it has the same color as the background. Maybe it could be surrounded by an edge?

screen shot 2016-05-09 at 2 12 06 pm
screen shot 2016-05-09 at 2 11 06 pm

2opremio:
Svg-generation misses the node tags (which may be intentional as well):

normal:

screen shot 2016-05-09 at 2 15 58 pm

svg:

screen shot 2016-05-09 at 2 16 07 pm

client/app/scripts/utils/router-utils.js
```diff
@@ -8,13 +8,18 @@ import { route } from '../actions/app-actions';
...
 const SLASH_REPLACEMENT = '<SLASH>';
+const PERCENT = '%';
+const PERCENT_REPLACEMENT = '<PERCENT>';
 
 function encodeURL(url) {
-  return url.replace(new RegExp(SLASH, 'g'), SLASH_REPLACEMENT);
+  return url
+    .replace(new RegExp(PERCENT, 'g'), PERCENT_REPLACEMENT)
+    .replace(new RegExp(SLASH, 'g'), SLASH_REPLACEMENT);
 }
```
foot: Oh dear, would be nice to have a generic soln for this. think think think.

paulbellamy: If only there were some sort of standard encoding you could pull off-the-shelf...

paulbellamy:
screen shot 2016-05-09 at 15 48 24

2opremio:
The search bar probably shouldn't be shown in this view

screen shot 2016-05-09 at 3 39 12 pm

paulbellamy:
Whenever you alt-tab away from a selected (but empty) search box, and alt-tab back it re-expands again, which is distracting.

paulbellamy:
This interacts strangely with filters.

If I select the `Containers` topology, then set `Both` for both filters, then select the `Containers By Image` topology, select `System` filter, and start doing a search for something which will only match application containers, it highlights the `Containers` topology link. But when I click back to the `Containers` view, it applies the filters and there are no nodes shown.

paulbellamy:
Also, getting some weird rendering glitches on the labels when hovering over non-matched nodes:
screen shot 2016-05-09 at 16 02 02

client/app/scripts/utils/search-utils.js
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+ * Returns a RegExp from a given string. If the string is not a valid regexp,
+ * it is escaped. Returned regexp is case-insensitive.
+ */
+function makeRegExp(expression, options = 'i') {
+  try {
+    return new RegExp(expression, options);
+  } catch (e) {
+    return new RegExp(_.escapeRegExp(expression), options);
```
foot: yay lodash!

foot:
Code looking good. 👍 

Some nice comments there.

2opremio:
I have deployed this on the service, and as soon as I start playing with search the app gets overloaded and the probes drop reports. For instance.

```
<probe> ERRO: 2016/05/10 11:25:56.872675 Dropping report to 10.0.205.170:4040
<probe> ERRO: 2016/05/10 11:26:02.870855 Dropping report to 10.0.205.170:4040
<probe> ERRO: 2016/05/10 11:26:14.835464 Dropping report to 10.0.205.170:4040
<probe> ERRO: 2016/05/10 11:26:23.769282 Dropping report to 10.0.205.170:4040
<probe> ERRO: 2016/05/10 11:26:32.858937 Dropping report to 10.0.205.170:4040
````

This seems to be an exacerbation of #1457 but probes don't drop seem to drop reports without search :(

foot:
More recent commits look fine too. 

LGTMTM

2opremio:
I find the hovering on not-highlighted nodes/edges surprising (It may be intentional but it feels buggy):



Also, I still see tags from different nodes overlapping

screen shot 2016-05-11 at 5 36 05 pm


Performance-wise it seems to be tolerable

Report: [report_search_review.json.gz](https://github.com/weaveworks/scope/files/259790/report_search_review.json.gz)

Other than that LGTM

PR #1434, opened 2016-05-05
Update readme to express that daemon sets won't schedule on unschedulable nodes prior to kubernetes 1.2

### LINKED ISSUE #1030: k8s: Make possible and simple to deploy a Scope probe in master node
The current k8s instructions https://github.com/weaveworks/scope#using-weave-scope-with-kubernetes result in Scope probes only deployed in the worker nodes but not on master.

CC @errordeveloper

tomwilkie:
Fixes #1030


2opremio:
LGTM (cc @abuehrle )

PR #1436, opened 2016-05-05
Deployment and Replica Set views

### LINKED ISSUE #1372: k8s: Add views for Deployments and ReplicaSets
Kubernetes 1.2 comes with [Deployments](http://kubernetes.io/docs/user-guide/deployments/) and [ReplicaSets](http://kubernetes.io/docs/user-guide/replicasets/) (similar to ReplicationControllers but enable declarative changes/rollouts and use hashing for versioning), we should add views for them.

See http://blog.kubernetes.io/2016/04/using-deployment-objects-with.html for a friendly example.

For Deployments we probably want some sort of log view to see the changes which were made to the deployment (it would require a similar view to https://github.com/weaveworks/scope/issues/1329 )

paulbellamy:
Fixes #1372

### DIFF
--- a/probe/kubernetes/deployment.go
+++ b/probe/kubernetes/deployment.go
@@ -1,0 +1,71 @@
+package kubernetes
+
+import (
+	"fmt"
+
+	"github.com/weaveworks/scope/report"
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api/unversioned"
+	"k8s.io/kubernetes/pkg/apis/extensions"
+	"k8s.io/kubernetes/pkg/labels"
+)
+
+// These constants are keys used in node metadata
+const (
+	DeploymentID                  = "kubernetes_deployment_id"
+	DeploymentName                = "kubernetes_deployment_name"
+	DeploymentCreated             = "kubernetes_deployment_created"
+	DeploymentObservedGeneration  = "kubernetes_deployment_observed_generation"
+	DeploymentDesiredReplicas     = "kubernetes_deployment_desired_replicas"
+	DeploymentReplicas            = "kubernetes_deployment_replicas"
+	DeploymentUpdatedReplicas     = "kubernetes_deployment_updated_replicas"
+	DeploymentAvailableReplicas   = "kubernetes_deployment_available_replicas"
+	DeploymentUnavailableReplicas = "kubernetes_deployment_unavailable_replicas"
+	DeploymentLabelPrefix         = "kubernetes_deployment_labels_"
+	DeploymentStrategy            = "kubernetes_deployment_strategy"
+)
+
+// Deployment represents a Kubernetes deployment
+type Deployment interface {
+	Meta
+	Selector() labels.Selector
+	GetNode(probeID string) report.Node
+}
+
+type deployment struct {
+	*extensions.Deployment
+	Meta
+	Node *api.Node
+}
+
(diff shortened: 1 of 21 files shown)


probe/kubernetes/meta.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"time"
+
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/labels"
+)
+
+// Meta represents a metadata information about a Kubernetes object
+type Meta interface {
```
tomwilkie: Nice!

tomwilkie:
I think we should either include a replication controller view, or include replication controllers in replica sets.

People still have replication controllers.  Like us.

probe/kubernetes/deployment.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/weaveworks/scope/report"
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/api/unversioned"
+	"k8s.io/kubernetes/pkg/apis/extensions"
+	"k8s.io/kubernetes/pkg/labels"
+)
+
+// These constants are keys used in node metadata
+const (
+	DeploymentID                  = "kubernetes_deployment_id"
```
tomwilkie: Given we don't merge nodes of different types anymore, is it worth have a meta.GetNode() and unifying all the metadata keys between deployments, pods, services, rs, rc etc etc?

render/pod.go
```diff
@@ -57,6 +57,34 @@ var PodServiceRenderer = ConditionalRenderer(renderKubernetesTopologies,
...
+var DeploymentRenderer = ApplyDecorators(
+	FilterEmpty(report.Pod,
+		MakeReduce(
+			MakeMap(
+				MapPod2Deployment,
+				PodRenderer,
+			),
+			SelectDeployment,
+		),
+	),
```
tomwilkie: Is this right?  I was under the impression is was pod -> replica set -> deployment?

tomwilkie:
Looks good so far!  I'll work on controls for this tomorrow.

paulbellamy:
Also, need to check this with multiple nodes, as the deployments/replicasets/services will get reported multiple times

tomwilkie:
GKE ftw

render/pod.go
paulbellamy: The selectors on a deployment match either. Technically a deployment manages a replicaset which manages the pods, but in practice deployments and replica sets are somewhat interchangeable (with the difference that they tell you not to use replica sets). However, showing the full hierarchy may be more clear.

tomwilkie:
@paulbellamy I just deployed this on GKE.

It looks really good.

probe/kubernetes/meta.go
```diff
@@ -0,0 +1,66 @@
...
+func (m meta) Created() string {
+	return m.ObjectMeta.CreationTimestamp.Format(time.RFC822)
+}
+
+func (m meta) Labels() map[string]string {
+	return m.ObjectMeta.Labels
+}
+
+// MetaNode gets the node metadata
+func MetaNode(m Meta, id string) report.Node {
```
tomwilkie: nit: why not:

```
func (m Meta) MetaNode(id string) report.Node
```

tomwilkie:
Other than the nit, LGTM

PR #1439, opened 2016-05-06
Revert "Make Scope work on Docker for Mac"

tomwilkie:
Reverts weaveworks/scope#1414

### DIFF
--- a/scope
+++ b/scope
@@ -2,8 +2,6 @@
 
 set -eu
 
-ARGS="$@"
-
 usage() {
     echo "Usage:"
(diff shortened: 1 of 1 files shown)


tomwilkie:
Turns out the check is not strong enough - it only checks if d4m is install, not if running.  If I'm running docker-machine but have d4m install, we use the d4m code path (and it doesn't work).

errordeveloper:
I'd like to fix the check.

tomwilkie:
Please do.  In the meantime master is broken, so unless the fix it imminent, pls revert.

PR #1442, opened 2016-05-06
Fix Docker for Mac test by checking if it is in use

errordeveloper:


tomwilkie:
I assume this has been superseded by #1448?

PR #1445, opened 2016-05-06
Set 'default' as the default namespace filter instead of 'all'

### LINKED ISSUE #1440: Set namespace filter to `default` instead of `all` by default
Reported by @janwillies on Slack.

paulbellamy:
Fixes #1440

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -123,6 +123,9 @@ func init() {
 func kubernetesFilters(namespaces ...string) APITopologyOptionGroup {
 	options := APITopologyOptionGroup{ID: "namespace", Default: "all"}
 	for _, namespace := range namespaces {
+		if namespace == "default" {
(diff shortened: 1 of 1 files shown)


2opremio:
LGTM

2opremio:
Actually, is the `default` ns always guaranteed to exist? Will this work if it doesn't?

paulbellamy:
hmmm. good question.

paulbellamy:
Ok, so we'll only set it to default if we have a default namespace. Need to check this on monday.

2opremio:
@tomwilkie Early trigger.

tomwilkie:
Huh?

2opremio:
There was a fix to be made which I missed before the LGTM

2opremio:
Oh, I see paul fixed it already, please ignore me.

PR #1447, opened 2016-05-06
Remove cache from SmartMerger.

### LINKED ISSUE #1441: App high memory consumption
Reported by @janwillies on Slack after testing `:latest`. He used to set a 128MB limit with which the app used to work just fine, now he claims it's consuming 2.5 GB (VM I presume ,the profile below shows 1.25 GB).

I think the environment hasn't increased in size (@janwillies, can you confirm and give details on the number of probes?).

Profile: [pb.gz](https://github.com/weaveworks/scope/files/251994/pb.gz)

app_memory

tomwilkie:
Fixes #1441 

So by my calculations, with 1 host reporting every 3 seconds, there will be 5 reports involved in every merge. So ~10 intermediate, cached nodes. Every 3 seconds, 5 of these will be invalidated, so it will take 600s for the 1000 node cache up.

I added a timeout on the cache and reduce it to 200 entries, and after 12mins my scope app was using 103.8MB.

### DIFF
--- a/app/collector.go
+++ b/app/collector.go
@@ -78,7 +78,7 @@ func NewCollector(window time.Duration) Collector {
 		waitableCondition: waitableCondition{
 			waiters: map[chan struct{}]struct{}{},
 		},
-		merger: NewSmartMerger(window),
+		merger: NewSmartMerger(),
 	}
 }
 
--- a/app/collector.go
+++ b/app/collector.go
@@ -78,7 +78,7 @@ func NewCollector(window time.Duration) Collector {
 		waitableCondition: waitableCondition{
 			waiters: map[chan struct{}]struct{}{},
 		},
-		merger: NewSmartMerger(),
+		merger: NewSmartMerger(window),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


2opremio:
As I mentioned in the issue, I think the limit should dynamically depend on the number of probes.

tomwilkie:
So based on the profile the 1k cache used ~1GB.  Given that, the fixed cache size limits this to ~200MB.  The result of 100MB shows the expiring is working to limit the size.

@paulbellamy and I discussed this and given the cache is only ~10% speed up, we were going to remove it.  I only added this fix as it was straight forward and simple.  Anything more complicated and we'll just remove the cache.

paulbellamy:
While this fix is simple, I still think that (in aggregate) the caching on this section is too involved for too little reward.

tomwilkie:
@paulbellamy in the interest of getting this fixed before the release I removed the caching.  I think we should put it back at some point, but I need to try and measure it affect to GC first.

paulbellamy:
We don't really need the `id` field (or `hash`) anymore either, right? If we assume that the IDs are already evenly distributed across the space, we only need to sort them by ID at the beginning. Which means we don't really need the `type node`, it could just work on `report.Report`s.

PR #1451, opened 2016-05-09
Add scale up/down controls on deployments, replica sets, and replication controllers

paulbellamy:
@tomwilkie, ptal

Was debating adding a test for the `CaptureResource` method, as it's a bit involved.

### DIFF
--- a/probe/kubernetes/reporter.go
+++ b/probe/kubernetes/reporter.go
@@ -65,6 +65,21 @@ var (
 	TableTemplates = report.TableTemplates{
 		LabelPrefix: {ID: LabelPrefix, Label: "Kubernetes Labels", Prefix: LabelPrefix},
 	}
+
+	ScalingControls = []report.Control{
+		{
+			ID:    ScaleDown,
+			Human: "Scale Down",
+			Icon:  "fa-minus",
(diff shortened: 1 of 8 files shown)


probe/kubernetes/reporter.go
```diff
@@ -326,3 +341,18 @@ func (r *Reporter) podTopology(services []Service, replicaSets []ReplicaSet) (re
 	})
 	return pods, err
 }
+
+func addScalingControls(t report.Topology) {
```
tomwilkie: Can we do something similar to https://github.com/weaveworks/scope/blob/master/probe/docker/reporter.go#L53?

tomwilkie:
Looks good, but I think we might be able to make this more generic by using the Meta type in CaptureResource.  Will think on it.

paulbellamy:
> using the Meta type

Ah, maybe.. The tricky bit, at the moment, is mapping some "ReplicaSet" nodes back to replication controllers. Maybe that can be handled more gracefully, which would simplify CaptureResource.

tomwilkie:
> Will think on it.

I didn't think of a better way.   LGTM

Issue #1454, opened 2016-05-09
High CPU consumption in probes


2opremio:
On the three worker machines of the service the CPU consumption of the 0.15 candidate is over the roof

<img width="1418" alt="screen shot 2016-05-09 at 3 17 35 pm" src="https://cloud.githubusercontent.com/assets/2362916/15115958/49dfaff8-15f9-11e6-9941-1c6a1f7950dd.png">

Also, reports are being dropped. Probably for the same reason:

```
<probe> WARN: 2016/05/09 14:22:16.781555 Docker reporter took longer than 1s
<probe> ERRO: 2016/05/09 14:22:23.174719 Dropping report to 10.0.26.13:4040
<probe> WARN: 2016/05/09 14:22:25.812799 Docker reporter took longer than 1s
<probe> ERRO: 2016/05/09 14:22:27.388793 Dropping report to 10.0.26.13:4040
<probe> ERRO: 2016/05/09 14:22:29.333923 Dropping report to 10.0.26.13:4040
<probe> ERRO: 2016/05/09 14:22:32.135056 Dropping report to 10.0.26.13:4040
<probe> ERRO: 2016/05/09 14:22:38.283660 Dropping report to 10.0.26.13:4040
<probe> ERRO: 2016/05/09 14:22:41.262928 Dropping report to 10.0.26.13:4040
<probe> ERRO: 2016/05/09 14:22:44.862768 Dropping report to 10.0.26.13:4040
<probe> ERRO: 2016/05/09 14:22:50.425897 Dropping report to 10.0.26.13:4040
<probe> ERRO: 2016/05/09 14:22:53.383612 Dropping report to 10.0.26.13:4040
<probe> WARN: 2016/05/09 14:22:55.870504 Docker reporter took longer than 1s
<probe> ERRO: 2016/05/09 14:22:57.934171 Dropping report to 10.0.26.13:4040
<probe> ERRO: 2016/05/09 14:22:59.411752 Dropping report to 10.0.26.13:4040
<probe> WARN: 2016/05/09 14:23:01.011165 Docker reporter took longer than 1s
<probe> WARN: 2016/05/09 14:23:02.742918 Docker reporter took longer than 1s
<probe> ERRO: 2016/05/09 14:23:03.517608 Dropping report to 10.0.26.13:4040
<probe> ERRO: 2016/05/09 14:23:06.190370 Dropping report to 10.0.26.13:4040
[...]

Issue #1406, opened 2016-04-29
Add performance tests to CI


2opremio:
We should periodically run replicable performance tests to ensure that Scope's performance is not degrading or at be aware of it with objective metrics.

2opremio:
Another good example is #1454 and #1457 which started happening due to the large amount of env variables in kubernetes after starting to show them in the UI. It took hours to figure out manually but a kubernetes performance test with a reasonably-sized deployment would have probably detected it automatically.

PR #1456, opened 2016-05-10
Precompute base of the container nodes

2opremio:
Improves #1454

The CPU consumption of the probes is now better (maybe even bearable) but not acceptable:
screen shot 2016-05-10 at 07 05 11

Profile: [pprof.localhost:4041.samples.cpu.002.pb.gz](https://github.com/weaveworks/scope/files/256468/pprof.localhost.4041.samples.cpu.002.pb.gz)
probe_cpu

We really really need #1406 to avoid surprises before releases

### DIFF
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -119,14 +119,17 @@ type container struct {
 	pendingStats [20]docker.Stats
 	numPending   int
 	hostID       string
+	baseNode     report.Node
 }
 
 // NewContainer creates a new Container
 func NewContainer(c *docker.Container, hostID string) Container {
-	return &container{
+	result := &container{
 		container: c,
 		hostID:    hostID,
 	}
+	result.baseNode = result.getBaseNode()
+	return result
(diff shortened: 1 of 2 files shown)


probe/docker/container.go
```diff
@@ -380,46 +383,49 @@ func (c *container) env() map[string]string {
 	return result
 }
 
+func (c *container) getBaseNode() report.Node {
+	result := report.MakeNodeWith(report.MakeContainerNodeID(c.ID()), map[string]string{
+		ContainerID:       c.ID(),
+		ContainerName:     strings.TrimPrefix(c.container.Name, "/"),
```
tomwilkie: Name can change (docker rename)

tomwilkie:
1 comment, otherwise LGTM (assuming tests pass)

probe/docker/container.go
2opremio: Good catch!

Issue #1201, opened 2016-03-23
Make reports smaller


tomwilkie:
I took a report from the service, and found it to be 6M uncompressed (500K compressed).  This was slightly bigger than I was expecting - although lets assume this is made of 15*3=45 different reports, which still puts compressed probe -> app reports at about 10K.

I was interested in seeing which topology was using the most space - assuming it to be endpoints.  It was not:

```
-rw-r--r--@  1 twilkie  staff   6.0M 23 Mar 12:46 k8s_report.json
-rw-r--r--   1 twilkie  staff   1.2M 23 Mar 12:50 container.json
-rw-r--r--   1 twilkie  staff    31K 23 Mar 12:51 container_image.json
-rw-r--r--   1 twilkie  staff   1.7M 23 Mar 12:50 endpoints.json
-rw-r--r--   1 twilkie  staff    19K 23 Mar 12:52 host.json
-rw-r--r--   1 twilkie  staff    55K 23 Mar 12:52 pod.json
-rw-r--r--   1 twilkie  staff   2.6M 23 Mar 12:50 process.json
-rw-r--r--   1 twilkie  staff    39K 23 Mar 12:52 service.json
```

So, perhaps an easy win would be to not report on every process; for instance, we only show ones which are doing IO, so we could easily filter out processes in the probe who's PIDs don't appear in the endpoint topology.  How this would affect things like the details panel is a different problem.

I also suspect the reason the process topology is so big is because we have metrics on it (and we don't have metrics on endpoints).  Perhaps their representation could be improved?

tomwilkie:
For instance, I suspect the way we encode times in json might be a little verbose, although (a) compression should be very effective on that and (b) we use msgpack.

2opremio:
Related: https://github.com/weaveworks/scope/issues/985

2opremio:
This is getting urgent. After seeing the pressure put on gzip on #1454 #1457 I measured the reports on the service and the are all around ~10MB. Note that this is mspack and not json.

```
<app> DEBU: 2016/05/10 07:07:19.868129 Decoded report with uncompressed size: 11039439 bytes
<app> DEBU: 2016/05/10 07:07:20.514462 Decoded report with uncompressed size: 1356204 bytes
<app> DEBU: 2016/05/10 07:07:21.075427 Decoded report with uncompressed size: 8903471 bytes
<app> DEBU: 2016/05/10 07:07:22.460466 Decoded report with uncompressed size: 11021565 bytes
[...]

Issue #985, opened 2016-02-18
Send incremental reports probe->app


2opremio:
The uncompressed report sizes are very large, ~1MB each. This puts a lot of pressure on the codecs and  causes a lot of traffic.

See https://github.com/weaveworks/scope/issues/854#issuecomment-183050569

We could send reports incrementally (i.e. full report + diffs) to improve this.

peterbourgon:
To send an incremental report, the probe must know what the app "already knows" in order to calculate a diff. That means the current (stateless) POST method of getting reports to apps is insufficient, and we'll need to maintain a persistent connection. Given the Go HTTP server keeps the connection around between POSTs currently, this is probably fine. But for the cloud offering, for example, we will end up creating a persistent connection from every node in the customer infrastructure to whatever.weave.works. Do we think that's acceptable?

I suggest WebSockets, since we're already doing something exactly like this between the app and the web UI. Upon connection, send the complete report. Thereafter, send deltas. (Maybe every ~minute, dump the complete report again, as a sync.) If the connection ever dies, throw away your state and start again. Maybe can even reuse some code.

If you do this, don't remove the existing POST functionality. It's still very useful for debugging, and the various test and fix probes.

tomwilkie:
This will also need some thought wrt multi tenancy.  Some sort of
persistent connection (is a websocket) does indeed seem the way to go.

2opremio:
> To send an incremental report, the probe must know what the app "already knows" in order to calculate a diff.

Not in general due to the CRDT semantics of reports (there are some exceptions to this like metrics we need to figure out, but we can decide not to send those incrementally for now). 

We would though need to have strict guarantees on full reports being received periodically which is much easier to obtain than keeping state about what the app knows.

> POST method of getting reports to apps is insufficient, and we'll need to maintain a persistent connection

As per the comment above I don't think that's true.

bboreham:
Done in #3677 

>  the probe must know what the app "already knows" in order to calculate a diff

As described in the PR, this turned out to be unnecessary.

PR #1458, opened 2016-05-10
Meassure report sizes

2opremio:
Helps debugging issues like https://github.com/weaveworks/scope/issues/1201 and #985


tomwilkie:
Wouldn't it be easier to read the Content-Length header?

2opremio:
> Wouldn't it be easier to read the Content-Length header?

For the compressed size it would, but we can't for the uncompressed size and I like the uniformity.

Issue #1457, opened 2016-05-10
High CPU consumption in the app


2opremio:
On the service, with 4 nodes (probes) the app is over 100% CPU

<img width="1029" alt="screen shot 2016-05-10 at 07 34 46" src="https://cloud.githubusercontent.com/assets/2362916/15137421/c5955cbe-1681-11e6-8844-0efddeffda3e.png">

Profile: [pprof.localhost:4040.samples.cpu.001.pb.gz](https://github.com/weaveworks/scope/files/256507/pprof.localhost.4040.samples.cpu.001.pb.gz)

![app_cpu](https://cloud.githubusercontent.com/assets/2362916/15137570/9eeafb0e-1682-11e6-83dc-7abd85b1f9cd.png)

The profile shows 80% CPU (maybe I was seeing garbage collection spikes?)

Perhaps the cache removed on https://github.com/weaveworks/scope/pull/1447 would help?

2opremio:
This is not really a degradation compared to what we had before (see https://github.com/weaveworks/scope/issues/854 ) but not what I would had expected after #1418 )

2opremio:
Also, the pressure of gzip is pretty high (just like on #1454 ), we could try getting a better performant compressor library or reduce the compression level until we have a solution for #1201

tomwilkie:
@2opremio what version was this?  Did it contain #1418?

2opremio:
Yep, as I mentioned both #1418 and #1447 are in

2opremio:
Reverting 2dae03501e741c6a5b841964ef13db3ed544c905 makes it even worse

<img width="1019" alt="screen shot 2016-05-10 at 09 32 15" src="https://cloud.githubusercontent.com/assets/2362916/15140266/3f2d38e8-1692-11e6-9f4c-fd5f4b73746d.png">

2opremio:
I've tried using https://github.com/klauspost/compress/gzip but it doesn't seem to be a big difference. @davkal , have the requests to `/api/topology` increased?

janwillies:
We are seeing over 100% CPU usage in the app, with 5 probes connected.

Performance seems to be ok, though

This is with :latest from yesterday

2opremio:
> Performance seems to be ok, though

You mean that the app is usable?

davkal:
> have the requests to /api/topology increased in this release?

Yes they have. Whenever you click on the search field, all topologies are fetched. But only then, no recurring timer.

[...]

PR #1465, opened 2016-05-10
Limit tables to 20 rows

2opremio:
Should help with #1457 and #1454 until we have #985

### DIFF
--- a/client/app/scripts/components/node-details.js
+++ b/client/app/scripts/components/node-details.js
@@ -14,10 +14,9 @@ import NodeDetailsRelatives from './node-details/node-details-relatives';
 import NodeDetailsTable from './node-details/node-details-table';
 import Warning from './warning';
 
-function getTruncationText(label, count) {
-  return `The section ${label} has been truncated because of too many entries.
-${count} entries are not shown.
-We are working on making this better.`;
+function getTruncationText(count) {
+  return 'This section was too long to be handled efficiently and has been truncated'
+  + ` (${count} extra entries not included). We are working to remove this limitation.`;
 }
 
 export class NodeDetails extends React.Component {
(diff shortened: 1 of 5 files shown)


2opremio:
@davkal Here's a sample report with the new trunction field in place: [report_truncation.json.gz](https://github.com/weaveworks/scope/files/257715/report_truncation.json.gz)

rade:
What gets truncated by this besides container env vars?

rade:
> What gets truncated by this besides container env vars?

AFAICT it is

- docker image labels
- docker container labels
- k8s labels

2opremio:
That is correct:
```
scope/probe$ grep -r AddTable *
docker/container.go:    result = result.AddTable(LabelPrefix, c.container.Config.Labels)
docker/container.go:    result = result.AddTable(EnvPrefix, c.env())
docker/reporter.go:             node = node.AddTable(ImageLabelPrefix, image.Labels)
kubernetes/meta.go:     }).AddTable(LabelPrefix, m.Labels())
```

2opremio:
@davkal Looking good

Two comments:

* Can we make it so that the message is shown when clicking not just hover?
* Maybe the copy could be slightly refined. 
Currently: _The section Environment Variables has been truncated because of too many entries.
714 entries are not shown. We are working on making this better._

I would propose _This section is too long and has been truncated (714 extra entries not included). We are working to remove this limitation._ 

I think it's more to the point and the section is obvious since it's name is right next to the warning sign.

2opremio:
@paul @foot Can you PTAL ?

paulbellamy:
Looks good, I'd even drop _We are working on making this better._ Maybe just _714 entries hidden_ (which is all we say about filtered unconnected processes).

paulbellamy:
backend code, LGTM.

foot:
Code LGTM!

Copy about why we're truncating is kind of nice, explaining the situation a bit. But there are probably reasons not too. Like admission of imperfection in product or something.

I like
> We are working to remove this limitation.

foot:
Should host.process-list/container-list be getting truncated? Those are 500+ on the attached report and make the UI a bit janky.

2opremio:
@foot That's tricker because ordering processes by CPU use etc ... require having them all to be useful.

2opremio:
LGTM

PR #1466, opened 2016-05-10
Ignore conntracked connections on which we never saw an update

### LINKED ISSUE #1462: Edges the wrong way around
See the screenshot:
screen shot 2016-05-10 at 14 19 10

tomwilkie:
Fixes #1462

### DIFF
--- a/probe/endpoint/conntrack.go
+++ b/probe/endpoint/conntrack.go
@@ -314,9 +314,13 @@ func (c *conntrackWalker) handleFlow(f flow, forceAdd bool) {
 			c.bufferedFlows = append(c.bufferedFlows, f)
 		}
 	case f.Type == destroyType:
-		if _, ok := c.activeFlows[f.Independent.ID]; ok {
+		if active, ok := c.activeFlows[f.Independent.ID]; ok {
 			delete(c.activeFlows, f.Independent.ID)
-			c.bufferedFlows = append(c.bufferedFlows, f)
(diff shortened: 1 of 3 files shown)


2opremio:
Shouldn't we only be paying attention to ESTABLISHED events?

tomwilkie:
> Shouldn't we only be paying attention to ESTABLISHED events?

Yeah I thought about that, but the way it works it to buffer old flows that came and went during the reporting period.  So those connections would never be ESTABLISHED.

probe/endpoint/nat.go
```diff
@@ -60,6 +60,10 @@ func (n natMapper) applyNAT(rpt report.Report, scope string) {
 		if !ok {
 			return
 		}
+		// Don't apply nat mappings to conntracked connections
+		if _, procspied := node.Latest.Lookup(Procspied); !procspied {
```
2opremio: Why not? If we do this we should move SNATing for short-lived connections to report.go

tomwilkie: Yes this was a old change (I thought this might have been one of the causes of the reversed connections).

We should do source natting in conntrack too, as this isn't going to work for most short lived connections (as this is periodic and only gets current connections).

2opremio: Great, would you mind creating a ticket for it?

2opremio:
LGTM

Issue #1291, opened 2016-04-13
Docker stats collection doesn't work with kubernetes


tomwilkie:
I think its because kubernetes runs a pause container (which has stats) and then other containers which share various namespaces:

![screen shot 2016-04-13 at 12 47 53](https://cloud.githubusercontent.com/assets/444037/14492244/f9fb5f74-0175-11e6-8262-75bd6673ce93.png)

tomwilkie:
Yeah seems to be the case:

```
Toms-MacBook-Pro:docker twilkie$ docker ps | grep scope 
f0c6ddc262cd        weaveworks/scope                                     "/home/weave/entrypoi"   28 minutes ago      Up 28 minutes                                k8s_weave-scope-probe.a8eb0e36_weave-scope-probe-o7hbj_default_6351451c-016a-11e6-a58b-aac114fc48cf_cf9eb212
9b27b53a09dd        beta.gcr.io/google_containers/pause:2.0              "/pause"                 28 minutes ago      Up 28 minutes                                k8s_POD.b67ce16c_weave-scope-probe-o7hbj_default_6351451c-016a-11e6-a58b-aac114fc48cf_8cc78d96
aadda282399c        sha256:95e0d                                         "/home/weave/entrypoi"   About an hour ago   Up About an hour                             k8s_weave-scope-app.7747597c_weave-scope-app-88si2_default_ee71d9f7-0164-11e6-a58b-aac114fc48cf_e8f6d2b5
87c998b9eb5f        beta.gcr.io/google_containers/pause:2.0              "/pause"                 About an hour ago   Up About an hour    0.0.0.0:4040->4040/tcp   k8s_POD.3704032f_weave-scope-app-88si2_default_ee71d9f7-0164-11e6-a58b-aac114fc48cf_b2e908fd

# docker stats on the actual probe container
Toms-MacBook-Pro:docker twilkie$ docker stats f0c6ddc262cd
CONTAINER           CPU %               MEM USAGE / LIMIT   MEM %               NET I/O             BLOCK I/O
f0c6ddc262cd        0.00%               0 B / 0 B           0.00%               0 B / 0 B           0 B / 0 

# docker stats on the pause container
Toms-MacBook-Pro:docker twilkie$ docker stats 9b27b53a09dd
CONTAINER           CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O
[...]

PR #1473, opened 2016-05-11
If k8s objects only have one container, show that container's metrics on them

paulbellamy:
screen shot 2016-05-11 at 11 47 14

We ignore the pause container with this, (even though it uses a bit of memory).

Not super-useful when you have pods/etc with > 1 container in them, but is a nice best-effort affordance when we can do it. Bit of a stop-gap until we have metric aggregation.

### DIFF
--- a/probe/kubernetes/reporter.go
+++ b/probe/kubernetes/reporter.go
@@ -34,6 +34,8 @@ var (
 		Created:          {ID: Created, Label: "Created", From: report.FromLatest, Priority: 6},
 	}
 
+	PodMetricTemplates = docker.ContainerMetricTemplates
+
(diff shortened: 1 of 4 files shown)


tomwilkie:
Do you not hit this? https://github.com/weaveworks/scope/issues/1291

paulbellamy:
Am running docker 1.11

tomwilkie:
> Am running docker 1.11

So am I.  I just thought this was a thing.  I even see it on GKE.

paulbellamy:
Works fine for me?

tomwilkie:
That's good I guess. 

I rather solve this will real ts aggregation, but as acts not going to happen in the next week, Lgtm

Issue #1475, opened 2016-05-11
k8s port forwarding instructions are broken


errordeveloper:
The docs say:
```
kubectl port-forward --namespace=kube-system $(kubectl get pod --namespace=kube-system --selector=name=weave-scope-app -o jsonpath={.items..metadata.name}) 4040
```
We can simplify this, and align with `kubectl create` instructions, i.e.:
```
kubectl port-forward $(kubectl get pod --selector=name=weave-scope-app -o jsonpath={.items..metadata.name}) 4040
```

Or use `--namespace=kube-system` everywhere.

errordeveloper:
@tomwilkie as you pointed this out, could you suggest the prefer solution here?

PR #1433, opened 2016-05-05
Have probe query kube dns to find the app.

tomwilkie:
Fixes #1373

PR #1479, opened 2016-05-11
Fix Kubernetes docs

errordeveloper:


site/installing.md
```diff
@@ -169,13 +169,16 @@ To enable DaemonSets in an existing cluster, add the `--runtime-config=extension
 
 **Install Scope on Your Cluster**
 
-It is recommended that you run Scope natively in your Kubernetes cluster using the manifest generator service.
+It is recommended that you run Scope natively in your Kubernetes cluster using `kubectl` with launch URL as show below.
 
 The simplest way to get the latest release of Scope deployed onto a Kubernetes cluster is the following:
 
-    kubectl create -f 'https://scope.weave.works/launch/k8s/weavescope.json' --validate=false
+    kubectl --namespace=kube-system create -f 'https://scope.weave.works/launch/k8s/weavescope.json' --validate=false
```
2opremio: I don't understand, why don't you add the `kube-system` namespace in the resources? In my experience you don't need a `--namespace=kube-system` here

errordeveloper: First of all, `kube-system` is not strictly mandatory, and might not exist, good example is OpenShift. In OpenShift Scope would get deployed into it's own namespace.

2opremio: for 0.15 scope needs to run in the `kube-system`  namespace anyways because of the fix from https://github.com/weaveworks/scope/pull/1433

2opremio: Then we have a problem, because we need to deploy in the same namespace and the kubedns add-on. Otherwise 0.15 won't work.

errordeveloper: In OpenShift you can currently do this:

```
[vagrant@localhost ~]$ oc new-project weavescope
Now using project "weavescope" on server "https://10.2.2.2:8443".

You can add applications to this project with the 'new-app' command. For example, try:

    $ oc new-app centos/ruby-22-centos7~https://github.com/openshift/ruby-hello-world.git

to build a new hello-world application in Ruby.
```
In Kubernetes terms, `oc new-project` creates a namespace and switches `kubectl` context.

Next you need to do a few things to allow Scope to run:
```
[vagrant@localhost ~]$ oc adm policy add-scc-to-user privileged system:serviceaccount:weavescope:default 
[vagrant@localhost ~]$ oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:weavescope:default
```
And now you can do (note no `--validate=false` flag required actually):
```
[vagrant@localhost ~]$ oc create -f https://scope.weave.works/launch/k8s/weavescope.json
```

If we were to do what you are saying, OpenShift user would have to say something like:
```
oc create -f 'https://scope.weave.works/launch/k8s/weavescope.json?namespace=weavescope'
```

errordeveloper: Also, I have to say, I did consider inferring the namespace already, before the OpenShift question came to mind. And actually the main argument was this.

```
kubectl create -f 'https://scope.weave.works/launch/k8s/weavescope.json'
```
Next, user will try running either `kubectl get pods` or `kubectl get evets -w` and they won't see anything about Scope. The user will come and ask us and we will have to remind them it's in `kube-system` namespace. It's much much better for this to be explicit, and also works well for OpenShift.

errordeveloper: Alternatively, I would suggest that Scope deserves it's own namespace.

2opremio: @errordeveloper Again, #1433 requires Scope to be in the same namespace as kube-dns but let's worry about that later. If  ditching `--namespace=kube-system` works for both k8s and OS just do that, let's worry about 0.15 later.

errordeveloper: @2opremio well, that's unfortunate that the implementation ended in the other repo, otherwise you would know that this doesn't depend on #1433.

errordeveloper:
One takeaway I thin is to consider that Scope simply deserves it's own namespace.

Also, do consider what if we did infer the namespace, and user specifies `--namespace=myscope-namespace`?

2opremio:
@errordeveloper For the 4th time, Scope needs to run in the same namespace as kube-dns for https://github.com/weaveworks/scope/pull/1433 to work

errordeveloper:
For the record, the service doesn't depend on DNS and simply uses:
```
$(WEAVESCOPE_APP_SERVICE_HOST):$(WEAVESCOPE_APP_SERVICE_PORT)
```

I will switch all the things default namespace, which was what I was thinking to do eventually anyway, but did this due to the lack of input on #1475.

PR #1485, opened 2016-05-12
Loading indicator

### LINKED ISSUE #1176: Show a visual indicator when loading reports is slow
When using the service to monitor an application/system the UI can be blank for several seconds, as the report seems to be slow to load (even with quite small numbers of hosts & containers).

Adding a clear visual indicator (to confirm to the user that something is happening) will prevent the assumption that Scope (or the application) is broken.

foot:
Fixes #1176


foot:
Alright! So it didn't end up being very exciting. 💍 . For now.

Do you see it for very long on your machine?

client/app/scripts/reducers/root.js
```diff
@@ -70,7 +71,7 @@ export const initialState = makeMap({
   updatePausedAt: null, // Date
   version: '...',
   versionUpdate: null,
-  websocketClosed: true,
+  websocketClosed: false,
```
davkal: Why was this flipped? The websocket is closed from the start.

foot: I guess... I read it is "it has been closed", like, past tense. I guess it could be present too though.

We interpret it as being a past tense thing in status.js and show a "Trying to reconnect..." briefly before we've established the ws during init.

davkal:
I like the entertainment, but only when I have to wait. On fast loads, it looks a bit distracting. How about firing off a timer that reevaluates the state after x seconds, and only then fires an action to display the indicator.

foot:
Yep, good call

foot:
Could you do this purely in css? Perhaps... do the settimeout onMount to
add a class...

client/app/scripts/reducers/root.js
davkal: Nah, state variables should reflect the state without having to think about it. If we want to keep it flipped, I recommend changing the field name, e.g. `websocketHasBeenClosed` to make it explicit. Otherwise someone might flip it again in the future, thinking no websocket -> websocket closed.

davkal:
CSS is worth a try, maybe `transition-delay` could help. Or you piggy-back on the animation starting out transparent and use animation delay.

rade:
Is this meant to fix #1176? What's left to do here?

foot:
Yes, this should help w/ #1176 

Left to do:
 - Delay the appearance of the indicator for a second or two as its a distracting if flashing up for ~200ms.

client/app/scripts/utils/delayed-show.js
```diff
@@ -0,0 +1,55 @@
...
+    if (props.show) {
+      this.scheduleShow();
+    }
+  }
+
+  scheduleShow() {
+    this.showTimeout = setTimeout(() => this.setState({ show: true }), this.props.delay);
+  }
+
+  cancelShow() {
```
davkal: This needs to be called from `componentWillUnmount`.

client/app/scripts/utils/delayed-show.js
```diff
@@ -0,0 +1,55 @@
...
+
+export class DelayedShow extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.state = {
+      show: false
+    };
+
+    if (props.show) {
+      this.scheduleShow();
```
davkal: I'd put this in `componentDidMount`.

davkal:
Looks great. There is still the theoretical state when the UI loads w/ no backend present, then you get the load indicator (you should get an error instead). It will become less theoretical when the UI code is served by a webserver as opposed to the scope app itself.
Other than that, LGTM

rade:
> Yes, this should *help w/* #1176

"help with" or "fix"?

foot:
This will show a loading screen until the first delta is received which I believe is what is happening in #1176. This PR will fix that. Updated the description.

PR #1499, opened 2016-05-12
Blur search when opening a terminal

davkal:


foot:
Yeah I've been playing around w/ blurring a bit, the shortcuts don't works
unless the search field is properly blurred blurred which is an `esc`. I
popped an onBlur back on the field but then alt-tabbing was blurring, I
think a fix for both these might get quite involved.

foot:
Ah hah it was a PR! Need to read PR/Issue more carefully.

foot:
Testing...

foot:
Nice! Works much better:

Only small tweak, maybe move `esc` precedence up, so terminal closes before blurring/clearing search.

1. search for something
2. find it
3. details
4. terminal
5. close terminal w/ c-d, esc
6. return to search context

foot:
LGTM

PR #1500, opened 2016-05-12
Check HOME variable before using it

### LINKED ISSUE #1497: Docker for mac check breaks scope script on ECS
```
May 12 17:22:19 ip-172-31-0-6 weave_scope_runner: error: 'scope launch' failed: /usr/local/bin/scope: line 72: HOME: unbound variable
````

2opremio:
Fixes #1497

### DIFF
--- a/scope
+++ b/scope
@@ -69,11 +69,11 @@ check_probe_only() {
 }
 
 check_docker_for_mac() {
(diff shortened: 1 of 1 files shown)


errordeveloper:
Unfortunately, the logic should be reverse. I myself never remember which way around `+x` notation works. The fix is in #1551.

Issue #1505, opened 2016-05-14
Raspberry PI - make github user configurable


dilgerma:
To build weave scope for raspberry pi, one cannot use the official github-sources, but we need some small modifications, changing base images to comply to the raspberry architecture.

In order to make this process more manageable, please allow the github user to be configured.

Currently I run  this step for every release I build for raspberry.

find . -name '*.go'  | xargs sed -i "bak" 's%github.com/weaveworks/scope%github.com/dilgerma/scope%g'
find . -name *.gobak | xargs rm

Could you provide this as a Variable like 'GITHUB_URI' or something, easily changeable?

dilgerma:
#1506

tomwilkie:
Hi dilgerma; thanks for the issue!  I don't think you can change go imports like that, but you can checkout a repo in a different place - ie checkout github.com/dilgerma/scope in $GOPATH/github.com/weaveworks/scope.

PR #1506, opened 2016-05-14
1505 - Make sourcepath configurable

dilgerma:


tomwilkie:
Hi Martin; what are you trying to achieve?  I don't think you can do that with go imports.

tomwilkie:
I think this was in relation to #1505; this approach isn't feasible, so let continue the discussion on the issue.

PR #1531, opened 2016-05-19
Use image name without version as id.

### LINKED ISSUE #1521: Duplicate stack in "by image" view
screenshot from 2016-05-18 13-16-19

Why are there two stacks of weaveworks/hello-world?

tomwilkie:
Fixes #1521 

As part of this, I made it so parent links don't have to be real nodes; to simplify, I also removed the container image parent for processes.

### DIFF
--- a/probe/docker/tagger.go
+++ b/probe/docker/tagger.go
@@ -82,11 +82,21 @@ func (t *Tagger) tag(tree process.Tree, topology *report.Topology) {
 			continue
 		}
 
-		topology.AddNode(report.MakeNodeWith(nodeID, map[string]string{
+		node := report.MakeNodeWith(nodeID, map[string]string{
 			ContainerID: c.ID(),
 		}).WithParents(report.EmptySets.
 			Add(report.Container, report.MakeStringSet(report.MakeContainerNodeID(c.ID()))),
-		))
+		)
 
+		// If we can work out the image name, add a parent tag for it
+		image, ok := t.registry.GetContainerImage(c.Image())
+		if ok && len(image.RepoTags) > 0 {
+			imageName := ImageNameWithoutVersion(image.RepoTags[0])
+			node = node.WithParents(report.EmptySets.
+				Add(report.ContainerImage, report.MakeStringSet(report.MakeContainerImageNodeID(imageName))),
+			)
+		}
+
+		topology.AddNode(node)
 	}
 }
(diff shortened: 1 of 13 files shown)


tomwilkie:
Before merging, we should consider how this will behave with old probes.

probe/docker/tagger.go
```diff
@@ -85,8 +85,7 @@ func (t *Tagger) tag(tree process.Tree, topology *report.Topology) {
 		topology.AddNode(report.MakeNodeWith(nodeID, map[string]string{
 			ContainerID: c.ID(),
 		}).WithParents(report.EmptySets.
-			Add(report.Container, report.MakeStringSet(report.MakeContainerNodeID(c.ID()))).
-			Add(report.ContainerImage, report.MakeStringSet(report.MakeContainerImageNodeID(c.Image()))),
```
2opremio: This effectively means users won't be able to reach the image from a container details panel? I have reservations about removing this.

render/container.go
```diff
@@ -79,9 +79,19 @@ func (r containerWithImageNameRenderer) Render(rpt report.Report, dct Decorator)
...
+			continue
+		}
+		imageNameWithoutVersion := docker.ImageNameWithoutVersion(imageName)
+		imageNodeID := report.MakeContainerImageNodeID(imageNameWithoutVersion)
+
 		output := c.Copy()
 		output = propagateLatest(docker.ImageName, image, output)
 		output = propagateLatest(docker.ImageLabelPrefix+"works.weave.role", image, output)
+		output.Parents = output.Parents.
+			Delete(report.ContainerImage).
```
2opremio: This is for compatibility with old probes? Regardless of whether I think it's a good idea to remove the image parents, I think a comment is in order. We should probably keep track of changes done for backwards compatibility so that we can potentially remove them after the support life since they clutter the code.

Also, have you tested this with old probes?

tomwilkie: > This is for compatibility with old probes

This is rewriting the parent, not removing it.

probe/docker/tagger.go
tomwilkie: Okay will see if I can add it back in.

tomwilkie: (Not this was from the process details panel, not the container details panel)

render/container.go
tomwilkie: But yes, it was for backwards compatibility - containers are reported with a parent of image id, and here we change them to be image name (the point of this PR is to key images by name in the rendering pipeline).

tomwilkie:
@2opremio PTAL

2opremio:
LGTM (as long as it was tested against probes without this change)

tomwilkie:
Works nicely with 0.15 probe.

PR #1537, opened 2016-05-24
[do-not-merge] traffic control from Scope

alban:
This is the code I am using to [demo traffic control](https://github.com/kinvolk/demo) in Scope. This is not ready for merge: this should be reworked to use the [Scope plugins system](https://github.com/weaveworks/scope/blob/master/examples/plugins/README.md) instead.

This adds 3 buttons to configure the traffic control of containers to 3 predefined settings: slow, medium and fast with different egress latency. 

/cc @alepuccetti


alban:
This is superseded by the traffic control plugin in https://github.com/weaveworks/scope/pull/1837

PR #1538, opened 2016-05-24
Allow user to specify table name and queue prefix.

tomwilkie:


prog/app.go
```diff
@@ -84,7 +84,7 @@ func collectorFactory(userIDer multitenant.UserIDer, collectorURL string, window
 	}
 
 	if parsed.Scheme == "dynamodb" {
-		dynamoCollector := multitenant.NewDynamoDBCollector(awsConfigFromURL(parsed), userIDer)
+		dynamoCollector := multitenant.NewDynamoDBCollector(awsConfigFromURL(parsed), userIDer, parsed.Path)
```
2opremio: I guess you don't expect slashes in the path. I would suggest to verify it.

2opremio:
Minor comment, otherwise LGTM. Also, I am assuming that there will be corresponding changes in the service (i.e. adding the prefixes)

tomwilkie:
@2opremio good catch!  Just hit this in dev.  Fixed

PR #1540, opened 2016-05-25
Add docker hub badge

tomwilkie:


2opremio:
I would separate the code-quality stats from the slack and download stats (e.g. move the go-report badge one place to the left)

PR #1541, opened 2016-05-25
Don't every store NEW conntrack flows (only every store updates).

tomwilkie:
This closes a small window where we might produce reports which contain flows that are NEW but have never seen an UPDATE, which can potentially be invalid.

### DIFF
--- a/probe/endpoint/conntrack.go
+++ b/probe/endpoint/conntrack.go
@@ -305,8 +305,10 @@ func (c *conntrackWalker) handleFlow(f flow, forceAdd bool) {
 	c.Lock()
 	defer c.Unlock()
 
+	// Ignore flows for which we never saw an update; they are likely
+	// incomplete or wrong.  See #1462.
 	switch {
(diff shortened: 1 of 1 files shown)


2opremio:
> which can potentially be invalid.

When would this happen? Aren't there any situations in which we would want to track NEW flows with no UPDATES?

tomwilkie:
> When would this happen? Aren't there any situations in which we would want to track NEW flows with no UPDATES?

TBH I don't know; I just observed in #1466 that all invalid flows fit the criteria of "they never saw an update".  This PR is an extension of #1466, to catch a small window where we might include flows which never saw an update.

I don't know for sure that all NEW flows which have never seen an UPDATE are invalid.  But this certainly made a big difference to the topology.

tomwilkie:
Worth adding that all 'good' flows I've observed see an UPDATE almost immediately after a NEW, so I'm not particularly worries about missing flow.

2opremio:
OK, then LGTM

PR #1545, opened 2016-05-27
Add caching to dynamodb collector

tomwilkie:


paulbellamy:
I would suggest that, as soon as you have several replicas, the cache hit rate will be very low.

tomwilkie:
As discussed, I've put reports into s3 now.

> I would suggest that, as soon as you have several replicas, the cache hit rate will be very low.

Agreed.  I'll add an (optional) external, shared memcache for reports.  I'd prefer to do this in the next PR as there a quite a few outstanding changes.  WDYT?

Also, any recommendations re:flags?  It feels a bit clunky right now, esp. given we're about to add memcache too.

app/multitenant/dynamo_collector.go
```diff
@@ -49,19 +50,33 @@ var (
...
-		Name:      "dynamo_report_size_bytes",
-		Help:      "Size of reports read / written from dynamodb.",
+		Name:      "dynamo_value_size_bytes",
+		Help:      "Size of data read / written from dynamodb.",
 	}, []string{"method"})
+
+	reportSize = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "scope",
+		Name:      "report_size_bytes",
+		Help:      "Size of reports written to s3.",
```
paulbellamy: The name and help don't match.. Either it's just the size of reports written to s3, or it's the size of all reports (s3 or no)

app/multitenant/dynamo_collector.go
```diff
@@ -71,22 +86,26 @@ type DynamoDBCollector interface {
...
+	tableName  string
+	bucketName string
+	merger     app.Merger
+	cache      gcache.Cache
 }
 
 // NewDynamoDBCollector the reaper of souls
 // https://github.com/aws/aws-sdk-go/wiki/common-examples
-func NewDynamoDBCollector(config *aws.Config, userIDer UserIDer, tableName string) DynamoDBCollector {
+func NewDynamoDBCollector(dynamoDBConfig, s3Config *aws.Config, userIDer UserIDer, tableName, bucketName string) DynamoDBCollector {
```
paulbellamy: It's not really a `DynamoCollector` anymore, is it. It's more of an `AWSCollector`, but...

app/multitenant/dynamo_collector.go
```diff
@@ -346,30 +341,50 @@ func (c *dynamoDBCollector) Add(ctx context.Context, rep report.Report) error {
...
-				S: aws.String(rowKey),
-			},
-			tsField: {
-				N: aws.String(strconv.FormatInt(now.UnixNano(), 10)),
-			},
-			reportField: {
-				B: bytes,
+	colKey := strconv.FormatInt(now.UnixNano(), 10)
+	s3Key := fmt.Sprintf("%s/%s", rowKey, colKey)
+	err = timeRequest("Put", s3RequestDuration, func() error {
```
paulbellamy: Let's check that the rowKey gets hashed by s3, and that we don't need to do that, to ensure performance.

prog/app.go
```diff
@@ -84,8 +87,22 @@ func collectorFactory(userIDer multitenant.UserIDer, collectorURL string, window
 	}
 
 	if parsed.Scheme == "dynamodb" {
+		s3, err := url.Parse(s3URL)
```
paulbellamy: Do we show a nice error when this is missing (since it's required for dynamo)?

tomwilkie: We do now.

app/multitenant/dynamo_collector.go
tomwilkie: Will rename later to minimise diff.

tomwilkie:
@paulbellamy I can haz LGTM?

paulbellamy:
LGTM

Issue #1550, opened 2016-06-03
Scope/kubernetes/CNI doesn't get IPs for containers and therefore conntracked edges are missing


tomwilkie:

tomwilkie:
What was happening: Containers started via CNI have --net=none, so the docker reporter does not know about IP addresses; hence the rendering join-on-ip for conntracked connections fails.

Also, I have --probe.weave.addr=, which disables the probe running `weave ps`.  Enabling that restore the edges.

tomwilkie:
We could consider decoupling the weavedns lookups from `--probe.weave.addr=`, and also consider using the golang `weave ps` implementation.

rade:
> We could consider decoupling the weavedns lookups from `--probe.weave.addr=`

These days there's a `weaveEnabled` flag. There isn't a clean way to *only* disable the dns registration / lookup; I am fixing that in #2445.

> use the golang `weave ps` implementation.

We are calling `weave.PS()` i.e. using the weave client library. ATM that does end up executing `weave --local ps`, which in turn invokes `weaveutil container-addrs ...` with the result of  `docker ps -q`. `container-addrs` in turn calls the docker client API to do an `inspect` on each of the containers.

So there's quite a lot going on, and some optimisations are certainly possible, but these should all be hidden behind `weave.PS()`, so there's no work to do in the scope code base.

The rest of this issue then is a feature request: show edges between containers w/o proc scraping. This should be possible in principle since we can obtain the container IPs from docker regardless of the network implementation.

PR #1552, opened 2016-06-04
Ignore IPv6 addresses in Docker reporter

tomwilkie:
I found this useful why tracking down #1550

### DIFF
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -277,12 +277,15 @@ func (c *container) ports(localAddrs []net.IP) report.StringSet {
 			continue
 		}
 		for _, b := range bindings {
-			if b.HostIP == "0.0.0.0" {
-				for _, ip := range localAddrs {
+			if b.HostIP != "0.0.0.0" {
+				ports = append(ports, fmt.Sprintf("%s:%s->%s", b.HostIP, b.HostPort, port))
+				continue
+			}
+
+			for _, ip := range localAddrs {
+				if ip.To4() != nil {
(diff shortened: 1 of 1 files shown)


probe/docker/container.go
```diff
@@ -277,12 +277,15 @@ func (c *container) ports(localAddrs []net.IP) report.StringSet {
 			continue
 		}
 		for _, b := range bindings {
-			if b.HostIP == "0.0.0.0" {
-				for _, ip := range localAddrs {
+			if b.HostIP != "0.0.0.0" {
```
2opremio: An address different than `"0.0.0.0"` indicates IPv6? I think a comment is in order.

tomwilkie: 0.0.0.0 means any host address; hence if its not 0.0.0.0 we can use the address directly.  Otherwise we have to replace 0.0.0.0 with the host addresses.

Will add a comment,

2opremio: Otherwise, shouldn't we check the hostIP as well for IPv6?

2opremio: Oh, I see, thanks.

2opremio: Will it also be "0.0.0.0" if the host is using IPv6?

tomwilkie: > Will it also be "0.0.0.0" if the host is using IPv6?

I've no idea. This PR is not really about making it work with IPv6, but making the reports smaller and easier to understand when you are using ipv4.  If you're using IPv6, you will get undefined behaviour.  This PR doesn't change that.

2opremio: Fair enough

2opremio:
LGTM

PR #1554, opened 2016-06-07
Fetch non-cached reports in parallel

jml:
Currently, if a bunch of requested reports aren't in the cache, we fetch them one at a time inline with the request. This patch updates the code to fetch reports in parallel instead.

While making the change, I noticed that we handled S3 errors differently from gzip & decoder errors. I've updated the code to handle all errors in the same way (i.e. no special logging, just return).

Currently, `getNonCached` returns only the first error it finds. It should probably aggregate all of the errors somehow so we aren't silently ignoring important conditions. I don't know how best to do that.

Along the way, I extracted the logic for fetching a single report (to make the actual parallelism logic more clear).

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1554)

### DIFF
--- a/app/multitenant/dynamo_collector.go
+++ b/app/multitenant/dynamo_collector.go
@@ -240,21 +240,38 @@ func (c *dynamoDBCollector) getCached(reportKeys []string) ([]report.Report, []s
 	return foundReports, missingReports
 }
 
+// Fetch multiple reports in parallel from S3.
 func (c *dynamoDBCollector) getNonCached(reportKeys []string) ([]report.Report, error) {
-	reports := []report.Report{}
+	type result struct {
+		key    string
+		report *report.Report
+		err    error
+	}
+
+	ch := make(chan result, len(reportKeys))
+
 	for _, reportKey := range reportKeys {
-		rep, err := c.getNonCachedReport(reportKey)
-		if err != nil {
-			return nil, err
+		go func(reportKey string) {
+			r := result{key: reportKey}
+			r.report, r.err = c.getNonCachedReport(reportKey)
+			ch <- r
+		}(reportKey)
+	}
+
+	reports := []report.Report{}
+	for range reportKeys {
+		r := <-ch
+		if r.err != nil {
+			return nil, r.err
 		}
-		reports = append(reports, rep)
-		c.cache.Set(reportKey, rep)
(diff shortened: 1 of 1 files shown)


app/multitenant/dynamo_collector.go
```diff
@@ -240,37 +240,61 @@ func (c *dynamoDBCollector) getCached(reportKeys []string) ([]report.Report, []s
...
+			r.report, r.err = c.getNonCachedReport(reportKey)
+			ch <- r
+		}(reportKey)
+	}
+
+	reports := []report.Report{}
+	for range reportKeys {
+		r := <-ch
+		if r.err != nil {
+			return nil, r.err
```
tomwilkie: I think this will leave orphaned goroutines if the first response contains an error.

tomwilkie: Actually maybe not.

tomwilkie:
Did you see any performance improvement running this on your laptop?

jml:
Haven't actually run it on my laptop, so will at least do that before landing.

I'll also at least attempt to write unit tests (but might end up deciding they're too hard).

app/multitenant/dynamo_collector.go
```diff
@@ -240,37 +240,61 @@ func (c *dynamoDBCollector) getCached(reportKeys []string) ([]report.Report, []s
...
 		}
-		reports = append(reports, rep)
-		c.cache.Set(reportKey, rep)
+		reports = append(reports, *r.report)
+		c.cache.Set(r.key, *r.report)
```
jml: Thanks, I wasn't sure so erred on the side of caution. Out of curiosity, how could I have inferred that it was safe without asking you?

tomwilkie: Read the docs: https://godoc.org/github.com/bluele/gcache

jml:
Ran this locally & verified that it doesn't break. No metrics. Won't bother with unit tests now.

@tomwilkie OK to merge

jml:
@tomwilkie asked for metrics:

### Before

20160608-143026_3196x1717_scrot

### After

20160608-144305_3196x1717_scrot

Running on my laptop locally.

tomwilkie:
LGTM

Issue #1539, opened 2016-05-24
k8s views show all nodes with same colour


rade:
I've seen a number of examples of k8s views now which show all nodes with the same(ish) colour, even though the pods/whatever represented were doing radically different things.

#783 is possibly related.

rade:
I am pretty sure this is a dup of #783. Closing.

Issue #783, opened 2015-12-17
Repeated colors for containers with different images


2opremio:
<img width="1958" alt="screen shot 2015-12-17 at 19 17 28" src="https://cloud.githubusercontent.com/assets/2362916/11879377/dd946bda-a4f2-11e5-8db8-5608d3ddcc2e.png">

ecs-agent and httpserver containers use different images and yet they are represented with the same color

The same applies to weave, weaveproxy and weavescope containers. They all use different images and yet they are represented using the same color.

Here is the "by image" view:

<img width="1978" alt="screen shot 2015-12-17 at 19 16 20" src="https://cloud.githubusercontent.com/assets/2362916/11879356/c013c196-a4f2-11e5-9a11-30bf845fd10b.png">

Here's the report: https://gist.github.com/2opremio/0fa26c14f5340bd87d87

davkal:
@2opremio Your intuition was that nodes of same color are based on the same image. That seems reasonable. In fact color is determined by a node field named `rank`. The backend render engine decides what to put there. In your report it was `2opremio/weaveecsdemo` for all containers. I produced a screenshot that prints the rank to show it:

![screen shot 2015-12-30 at 13 02 05](https://cloud.githubusercontent.com/assets/859729/12051042/a95c23fa-aef5-11e5-9575-853a4898064c.png)

So the color engine works correctly. But I see 2 issues here:

1. The ranking/grouping was not based on what your intuition would tell you (container image).
2. It's unclear to the user what the color is based on, since we dont display the rank.

The quickfix for (1) would be to make rank more meaningful for container topologies. To fix (2) we need to associate the color with a meta data field that it is based on. The backend could mark the metafield that informs the rank with a flag, so the frontend know which one it is. Unfortunately ``2opremio/weaveecsdemo` does not appear in the node's metadata:

```json
[
  {
    "id": "docker_container_id",
    "label": "ID",
    "value": "31f5360b12522404ce9e7e499a775b77ebb4e44f214b925fee9730ce5017b620"
  },
  {
    "id": "docker_image_id",
    "label": "Image ID",
    "value": "a4e0640a29e126734c124298e97dec0e0ff3461487939dcee89a1c55b6e70f02"
  },
  {
    "id": "docker_container_created",
    "label": "Created",
    "value": "17 Dec 15 19:00 UTC"
  },
  {
    "id": "docker_container_command",
    "label": "Command",
[...]

PR #1567, opened 2016-06-08
[RFC] Separate node colors

davkal:
Node colors are procedurally generated by a node's rank. This can be a
string like `tnsfork/lb-1`. Now if all the nodes share the same
prefix in their rank, e.g. `tnsfork/` they will all get the same
color.

screen shot 2016-06-08 at 15 56 50

An approach to separate colors a bit more would be to remove the longest
common rank prefix and set that as the color key.

screen shot 2016-06-08 at 15 56 03

Pros:
- colors are better separation for homogeneous rank values

Cons:
- if 99 nodes share a prefix and one does not, the 99 will be colored
  similarly (which is what happens now)
- if that one node appears and dissappears, the 99 nodes' colors will
  switch based on their new LCP
- Loading details panels from other topologies may yield a different
  color because that node is loaded in isolation (LCP of that
  topology's nodes is undetermined)

### DIFF
--- a/client/app/scripts/charts/node.js
+++ b/client/app/scripts/charts/node.js
@@ -76,13 +76,13 @@ class Node extends React.Component {
   }
 
   render() {
-    const { blurred, focused, highlighted, label, matches = makeMap(), networks,
-      pseudo, rank, subLabel, scaleFactor, transform, zoomScale, exportingGraph,
+    const { blurred, colorKey, focused, highlighted, label, matches = makeMap(), networks,
+      pseudo, subLabel, scaleFactor, transform, zoomScale, exportingGraph,
       showingNetworks, stack } = this.props;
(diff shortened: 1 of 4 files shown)


rade:
The 99-1 problem is nasty; and presumably the node appearance issue would also arise for 2-1, which is also nasty.

Perhaps ranking should operate on words rather than characters, e.g. `tnsfork_lb2_1` would result in three words - 'tnsfork_', 'lb2_', '1', which you could hash to get a series of number in MSB...LSB order for determining the colour.

rade:
See #783 and #1539 for the issue this is meant to fix.

Network colouring already use the prefix removal algo above, and it turns out to run into the 99-1 problem for trivial examples. So back to the drawing board.

davkal:
Closing this for inactivity.

PR #1568, opened 2016-06-08
Use NATS for shortcut reports in the service.

tomwilkie:
Current issue is NATS client fails on creation if it can't reach the NATS server.  Am investigating a fix.


app/multitenant/dynamo_collector.go
```diff
@@ -87,6 +97,11 @@ type DynamoDBCollector interface {
 	CreateTables() error
 }
 
+type watchKey struct {
```
jml: I'd really welcome a brief comment explaining what this is & what's it's used for.

app/multitenant/dynamo_collector.go
```diff
@@ -419,9 +434,63 @@ func (c *dynamoDBCollector) Add(ctx context.Context, rep report.Report) error {
...
+	if err != nil {
+		log.Errorf("Error getting user id in WaitOn: %v", err)
+		return
+	}
+
+	c.waitersLock.Lock()
+	key := watchKey{userid, waiter}
+	sub := c.waiters[key]
+	delete(c.waiters, key)
+	c.waitersLock.Unlock()
```
jml: I feel like there should already be a structure for a lock-protected dictionary. But I guess not having generics means that it's easier to just write it ourselves?

prog/main.go
```diff
@@ -176,6 +177,7 @@ func main() {
 	flag.StringVar(&flags.app.s3URL, "app.collector.s3", "local", "S3 URL to use (when collector is dynamodb)")
 	flag.StringVar(&flags.app.controlRouterURL, "app.control.router", "local", "Control router to use (local or sqs)")
 	flag.StringVar(&flags.app.pipeRouterURL, "app.pipe.router", "local", "Pipe router to use (local)")
+	flag.StringVar(&flags.app.natsHostname, "app.nats", "", "Hostname for NATS")
```
jml: What's the app behavior if `natsHostname` is not provided? Probably should document that here in the option.

jml:
Very cool! 

I've made a few comments. Please address them as you see fit & merge when green.

app/multitenant/dynamo_collector.go
tomwilkie: Yeah, welcome to go.

tomwilkie:
@jml PTAL

app/multitenant/dynamo_collector.go
```diff
@@ -485,10 +485,21 @@ func (c *dynamoDBCollector) WaitOn(ctx context.Context, waiter chan struct{}) {
...
+			natsRequests.WithLabelValues("NextMsg", errorCode(err)).Add(1)
+			if err != nil {
+				log.Debugf("NextMsg error: %v", err)
+				return
+			}
+			select {
+			case waiter <- struct{}{}:
+			default:
+			}
+		}
```
jml: Optional: Since you made this change because you discovered a bug, it'd be nice if this could have some sort of test that fails with the previous version & passes with this one.  

I can see why that might be a bit of a faff though.

prog/main.go
```diff
@@ -177,7 +177,7 @@ func main() {
 	flag.StringVar(&flags.app.s3URL, "app.collector.s3", "local", "S3 URL to use (when collector is dynamodb)")
 	flag.StringVar(&flags.app.controlRouterURL, "app.control.router", "local", "Control router to use (local or sqs)")
 	flag.StringVar(&flags.app.pipeRouterURL, "app.pipe.router", "local", "Pipe router to use (local)")
-	flag.StringVar(&flags.app.natsHostname, "app.nats", "", "Hostname for NATS")
+	flag.StringVar(&flags.app.natsHostname, "app.nats", "", "Hostname for NATS service to use for shortcut reports.  It empty, shortcut reporting will be disabled.")
```
jml: s/It empty/If empty/

jml:
LGTM, modulo a typo & optional additional test.

app/multitenant/dynamo_collector.go
tomwilkie: We should add a test, yes.  Too much work for today though.  Will file a ticket.

tomwilkie: #1571

PR #1570, opened 2016-06-09
If we don't get a path name from the router, make one up from the url.

tomwilkie:
To stop the monitoring showing "unamed_path" everywhere.

### DIFF
--- a/common/middleware/instrument.go
+++ b/common/middleware/instrument.go
@@ -40,19 +40,29 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
 	})
 }
 
-var invalidChars = regexp.MustCompile(`[^a-zA-Z0-9]+`)
-
+// Return a name identifier for ths request.  There are three options:
+//   1. The request matches a gorilla mux route, with a name.  Use that.
+//   2. The request matches an unamed gorilla mux router.  Munge the path
+//      template such that templates like '/api/{org}/foo' come out as
+//      'api_org_foo'.
+//   3. The request doesn't match a mux route.  Munge the Path in the same
+//      manner as (2).
+// We do all this as we do not wish to emit high cardinality labels to
+// prometheus.
 func (i Instrument) getRouteName(r *http.Request) string {
 	var routeMatch mux.RouteMatch
-	if !i.RouteMatcher.Match(r, &routeMatch) {
-		return MakeLabelValue(r.URL.Path)
+	if i.RouteMatcher.Match(r, &routeMatch) {
+		if name := routeMatch.Route.GetName(); name != "" {
+			return name
+		}
+		if tmpl, err := routeMatch.Route.GetPathTemplate(); err != nil {
+			return MakeLabelValue(tmpl)
+		}
 	}
-	name := routeMatch.Route.GetName()
-	if name == "" {
-		return MakeLabelValue(r.URL.Path)
-	}
-	return name
+	return MakeLabelValue(r.URL.Path)
 }
+
+var invalidChars = regexp.MustCompile(`[^a-zA-Z0-9]+`)
 
 // MakeLabelValue converts a Gorilla mux path to a string suitable for use in
 // a Prometheus label value.
@@ -62,6 +72,9 @@ func MakeLabelValue(path string) string {
 
 	// Trim leading and trailing underscores.
 	result = strings.Trim(result, "_")
+
+	// Make it all lowercase
+	result = strings.ToLower(result)
 
 	// Special case.
 	if result == "" {
--- a/common/middleware/instrument.go
+++ b/common/middleware/instrument.go
@@ -2,6 +2,7 @@ package middleware
 
 import (
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


common/middleware/instrument.go
```diff
@@ -39,14 +40,32 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
...
 	if name == "" {
-		return "unnamed_path"
+		return MakeLabelValue(r.URL.Path)
 	}
 	return name
 }
+
+// MakeLabelValue converts a Gorilla mux path to a string suitable for use in
+// a Prometheus label value.
+func MakeLabelValue(path string) string {
```
jml: Why is this public when `getRouteName` is not? (I don't think it's wrong that it should be, but curious as to why -- is it so we can write tests for it?)

common/middleware/instrument.go
```diff
@@ -39,14 +40,32 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
...
+	// Convert non-alnums to underscores.
+	result := invalidChars.ReplaceAllString(path, "_")
+
+	// Trim leading and trailing underscores.
+	result = strings.Trim(result, "_")
+
+	// Special case.
+	if result == "" {
+		result = "root"
```
jml: How will this be distinguished from a literal path "root"?

common/middleware/instrument.go
```diff
@@ -39,14 +40,32 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
 	})
 }
 
+var invalidChars = regexp.MustCompile(`[^a-zA-Z0-9]+`)
```
jml: Are there other things that specify, e.g., what is a valid org name?

common/middleware/instrument_test.go
```diff
@@ -0,0 +1,29 @@
...
+		"/foo":                   "foo",
+		"foo/":                   "foo",
+		"/foo/":                  "foo",
+		"/foo/bar":               "foo_bar",
+		"foo/bar/":               "foo_bar",
+		"/foo/bar/":              "foo_bar",
+		"/foo/{orgName}/Bar":     "foo_orgname_bar",
+		"/foo/{org_name}/Bar":    "foo_org_name_bar",
+		"/foo/{org__name}/Bar":   "foo_org_name_bar",
+		"/foo/{org___name}/_Bar": "foo_org_name_bar",
```
jml: Maybe also add a case or two for other invalid values? e.g. `/foo.bar/baz.qux/` -> `foo_bar_baz_qux`

jml:
Mostly looks good but I've got a couple of questions. Thanks!

common/middleware/instrument.go
tomwilkie: for testing; we have a weird convention when tests are in a different package, so things you want to test need to be to be public.  We should change that, but not today.

common/middleware/instrument.go
tomwilkie: It will not be - this code is copied from the service repo btw.

common/middleware/instrument.go
tomwilkie: Hmm looking into this, there appears to be no restrictions on label values.  @peterbourgon do you remember why this existed to begin with?

tomwilkie:
@jml you're comments prompted me to give this some more thought.  Have added a comment explaining what I'm trying to achieve in https://github.com/weaveworks/scope/pull/1570/commits/ae39936a571783ba066042a7357495d6ad7f0a2e.

Would you mind sanity checking this?  Also /cc @peterbourgon

common/middleware/instrument_test.go
tomwilkie: +1 done.

common/middleware/instrument.go
tomwilkie: Have tried to reverse engineer reasoning here; put a comment in https://github.com/weaveworks/scope/pull/1570/commits/ae39936a571783ba066042a7357495d6ad7f0a2e.  Does this sound about right?

common/middleware/instrument.go
```diff
@@ -40,20 +40,30 @@ func (i Instrument) Wrap(next http.Handler) http.Handler {
...
-
+// Return a name identifier for ths request.  There are three options:
+//   1. The request matches a gorilla mux route, with a name.  Use that.
+//   2. The request matches an unamed gorilla mux router.  Munge the path
+//      template such that templates like '/api/{org}/foo' come out as
+//      'api_org_foo'.
+//   3. The request doesn't match a mux route.  Munge the Path in the same
+//      manner as (2).
+// We do all this as we do not wish to emit high cardinality labels to
+// prometheus.
```
jml: +1 (informative). Would read again.

common/middleware/instrument.go
jml: Ah, I didn't know that. I guess a future PR should make it a library or something so we're not maintaining two copies?

jml:
Looks good to me. I very much appreciate the comment explaining the intent.

Punting to @peterbourgon for approval.

common/middleware/instrument.go
tomwilkie: Indeed, in a future PR I'll update the vendor in service and remove the duplication.

jml:
Looks good to me. I very much appreciate the comment explaining the intent.

Punting to @peterbourgon for approval.

tomwilkie:
Merging w/o @peterbourgon as I'd like to deploy this today, along with the NATS stuff.

common/middleware/instrument.go
peterbourgon: If MakeLabelValue will work for all the routes, can we remove options 1 and 2 (including removing named routes and all that supporting code) and just use 3?

PR #1572, opened 2016-06-09
Tag scope images on docker hub as we do in service

tomwilkie:


squaremo:
LGTM. We may want to add the same -WIP failsafes as service -- once I have those working!

PR #1575, opened 2016-06-09
[WIP] Join container to pods based on name with k8s1.1

### LINKED ISSUE #1558: Edges broken with k8s < 1.2
It would seem the containers aren't being associated with pods correctly:

screen shot 2016-06-08 at 10 49 52

tomwilkie:
Fixes #1558


tomwilkie:
Right this approach won't work as connects to pseudo nodes get dropped.  Back to drawing board.

PR #1578, opened 2016-06-10
Gather IPs from all the container networks

### LINKED ISSUE #1573: Container in multiple networks not showing all connections
We have service which is connected to two networks, so has two IPs (as seen here via docker inspect)
screen shot 2016-06-09 at 14 52 32

On Scope, it is only showing one IP:
screen shot 2016-06-09 at 14 53 52
 
Attaching reports file. Affected service is "front-end"
[report_09_06_json.zip](https://github.com/weaveworks/scope/files/307018/report_09_06_json.zip)

2opremio:
Fixes #1573

### DIFF
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -311,18 +311,21 @@ func (c *container) NetworkInfo(localAddrs []net.IP) report.Sets {
 	c.RLock()
 	defer c.RUnlock()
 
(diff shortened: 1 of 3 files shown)


jml:
This looks good to me. Is it possible to write a test?

paulbellamy:
LGTM

PR #1580, opened 2016-06-13
Helper for creating default Router

jml:
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1580)

### DIFF
--- a/common/middleware/middleware.go
+++ b/common/middleware/middleware.go
@@ -2,6 +2,8 @@ package middleware
 
 import (
 	"net/http"
+
+	"github.com/gorilla/mux"
 )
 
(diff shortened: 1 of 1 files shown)


common/middleware/middleware.go
```diff
@@ -31,3 +33,12 @@ func Merge(middlesware ...Interface) Interface {
 		return next
 	})
 }
+
+// Gorilla Router with sensible defaults, namely:
```
2opremio: I think this will make the linter fail (the comment needs to start with `NewRouter()`). Did you get to run `make lint` locally?

jml:
Decided that this isn't middleware and that it's not worthy of coming up with a new name for a new module.

PR #1581, opened 2016-06-14
Add filters for pseudo nodes.

### LINKED ISSUE #1129: "uncontained" nodes clutter container view
screenshot from 2016-03-05 10 07 32

If I was to draw the structure of that service on the whiteboard, I certainly would not be including any of the "uncontained" nodes. I might include the Internet, if my service was accessible from / talking to that. And I might include some uncontained processes if they are interacting with my service.

Possible fix: add a show/hide for Uncontained.

tomwilkie:
Fixes #1129

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -21,6 +21,14 @@ var (
 	topologyRegistry = &registry{
 		items: map[string]APITopologyDesc{},
 	}
+	k8sPseudoFilter = APITopologyOptionGroup{
+		ID:      "pseudo",
+		Default: "show",
+		Options: []APITopologyOption{
+			{"show", "Show Unmanaged", nil, false},
+			{"hide", "Hide Unmanaged", render.IsNotPseudo, true},
+		},
+	}
 )
 
 func init() {
@@ -29,18 +37,26 @@ func init() {
 			ID:      "system",
 			Default: "application",
 			Options: []APITopologyOption{
-				{"system", "System containers", render.IsSystem},
-				{"application", "Application containers", render.IsApplication},
-				{"both", "Both", nil},
+				{"system", "System containers", render.IsSystem, false},
+				{"application", "Application containers", render.IsApplication, false},
+				{"both", "Both", nil, false},
 			},
 		},
 		{
 			ID:      "stopped",
 			Default: "running",
 			Options: []APITopologyOption{
-				{"stopped", "Stopped containers", render.IsStopped},
-				{"running", "Running containers", render.IsRunning},
-				{"both", "Both", nil},
+				{"stopped", "Stopped containers", render.IsStopped, false},
+				{"running", "Running containers", render.IsRunning, false},
+				{"both", "Both", nil, false},
+			},
+		},
+		{
+			ID:      "pseudo",
+			Default: "show",
+			Options: []APITopologyOption{
+				{"show", "Show Uncontained", nil, false},
+				{"hide", "Hide Uncontained", render.IsNotPseudo, true},
 			},
 		},
 	}
@@ -52,7 +68,7 @@ func init() {
 			Options: []APITopologyOption{
 				// Show the user why there are filtered nodes in this view.
 				// Don't give them the option to show those nodes.
-				{"hide", "Unconnected nodes hidden", nil},
+				{"hide", "Unconnected nodes hidden", nil, false},
 			},
 		},
 	}
@@ -142,9 +158,11 @@ func kubernetesFilters(namespaces ...string) APITopologyOptionGroup {
 		if namespace == "default" {
 			options.Default = namespace
 		}
-		options.Options = append(options.Options, APITopologyOption{namespace, namespace, render.IsNamespace(namespace)})
+		options.Options = append(options.Options, APITopologyOption{
+			namespace, namespace, render.IsNamespace(namespace), false,
+		})
 	}
-	options.Options = append(options.Options, APITopologyOption{"all", "All Namespaces", nil})
+	options.Options = append(options.Options, APITopologyOption{"all", "All Namespaces", nil, false})
 	return options
 }
 
@@ -169,7 +187,9 @@ func updateFilters(rpt report.Report, topologies []APITopologyDesc) []APITopolog
 	sort.Strings(ns)
 	for i, t := range topologies {
 		if t.id == "pods" || t.id == "services" || t.id == "deployments" || t.id == "replica-sets" {
-			topologies[i] = updateTopologyFilters(t, []APITopologyOptionGroup{kubernetesFilters(ns...)})
+			topologies[i] = updateTopologyFilters(t, []APITopologyOptionGroup{
+				kubernetesFilters(ns...), k8sPseudoFilter,
+			})
 		}
 	}
 	return topologies
@@ -224,7 +244,8 @@ type APITopologyOption struct {
 	Value string `json:"value"`
 	Label string `json:"label"`
 
-	filter render.FilterFunc
+	filter       render.FilterFunc
+	filterPseudo bool
 }
 
 type topologyStats struct {
@@ -329,7 +350,7 @@ func (r *registry) rendererForTopology(topologyID string, values url.Values, rpt
 	}
 	topology = updateFilters(rpt, []APITopologyDesc{topology})[0]
 
-	var filters []render.FilterFunc
+	var decorators []render.Decorator
 	for _, group := range topology.Options {
 		value := values.Get(group.ID)
 		for _, opt := range group.Options {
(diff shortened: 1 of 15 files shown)


app/api_topologies.go
```diff
@@ -21,6 +21,14 @@ var (
 	topologyRegistry = &registry{
 		items: map[string]APITopologyDesc{},
 	}
+	pseudoFilter = APITopologyOptionGroup{
+		ID:      "pseudo",
+		Default: "show",
+		Options: []APITopologyOption{
+			{"show", "Show pseudo nodes", nil, false},
+			{"hide", "Hide pseudo nodes", render.IsNotPseudo, true},
```
2opremio: While perfectly valid as an internal name, I am not sure _pseudo_ will tell much to the end user. Maybe _Uncontained_ if it's accurate enough?

tomwilkie: Agree its a shitty term for the UI.

Its _uncontained_ on the containers view, _unmanaged_ on the k8s views.  And also the internet nodes.

2opremio: Also, the filtering area is already pretty crowded, #1129 is not about psuedo nodes per se but about showing groups of pseudonodes unconnected from the rest of the topology. See https://github.com/weaveworks/scope/issues/1129#issuecomment-192688819

2opremio:
I'm not fully convinced this is the best way to solve #1129 . Instead of adding another knob I think we should filter out pseudo-nodes not connected to the _meaningful_ nodes of the view being displayed (e.g. container nodes).

rade:
> I'm not fully convinced this is the best way to solve #1129

It certainly isn't, but the more interesting question is whether it is a net improvement...

I imagine it will be quite common to have both proper and pseudo nodes connected to the internet, and only wanting to see the former nodes&connections but not the latter. Combined with the "extra knob that really shouldn't be needed", that swings the answer slightly in the 'non' direction for me.

tomwilkie:
Feedback from actual user that inspired this PR contradict your views:

> Give me the option to hide/filter anything that is not a Pod, service, or container. Some of the items on the screen don't make sense to show in most contexts.

davkal:
Maybe this is a naming thing. "pseudo" is Scope jargon. What do those nodes have in common? Are they *ancillary*, *supporting*, *external*, *bogus*, *extra*, *helper*, *secondary*.

rade:
> Feedback from actual user that inspired this PR contradict your views

It does not. Bear in mind that it was me who raised the issue and proposed this very solution as a quick fix. Three months ago. Since then the UI has moved on a lot. Yes, for the users that have no pseudo-nodes connected to their proper nodes the quick fix is an improvement. For them, not showing pseudo nodes at all would be even ever. For everybody else it isn't an improvement *and* it burdens *everyone* with an additional knob.

If it takes little effort to review this then I am willing to let it through under one condition: we raise a separate issue for the proper fix.

rade:
> I imagine it will be quite common to have both proper and pseudo nodes connected to the internet, and only wanting to see the former nodes&connections but not the latter.

It occurs to me that this could be addressed by not including the internet nodes in this filter.

Which would then also address David's concern, since the filter could say 'Uncontained' or 'Unmanaged', which makes it very obvious what this is hiding since it corresponds to labels in the UI.

Question: if we did exclude internet nodes from the filter, would the internet nodes be hidden if the filter is on and only unmanaged/uncontained nodes are connected to them?

davkal:
> Question: if we did exclude internet nodes from the filter, would the internet nodes be hidden if the filter is on and only unmanaged/uncontained nodes are connected to them?

I think always showing the internet node (exclude from the filter) would be fine by me. Then it would be independent of the knob.

rade:
My point was that ideally it wouldn't show the internet nodes if they don't have edges to any other shown nodes. So I was wondering whether perhaps that would be happening anyway in the current algorithm.

davkal:
> My point was that ideally it wouldn't show the internet nodes if they don't have edges to any other shown nodes. So I was wondering whether perhaps that would be happening anyway in the current algorithm.

Ah yes, that's what I meant too. I had that algo in mind (not showing internet nodes that dont have edges to them)

tomwilkie:
This is a one line change.

tomwilkie:
@2opremio @davkal @rade PTAL

rade:
can't/won't comment on the code, but if it does what I suggested then that SGTM.

render/filters.go
```diff
@@ -95,7 +122,10 @@ func (f *Filter) render(rpt report.Report, dct Decorator) (report.Nodes, int) {
 	inDegrees := map[string]int{}
 	filtered := 0
 	for id, node := range f.Renderer.Render(rpt, dct) {
-		if node.Topology == Pseudo || f.FilterFunc(node) {
+		if (f.filterPseudo || node.Topology != Pseudo) && f.FilterFunc(node) {
+			output[id] = node
+			inDegrees[id] = 0
+		} else if !f.filterPseudo && node.Topology == Pseudo {
```
2opremio: My head hurts a bit when reading this, isn't this equivalent to the following?

```golang
if !f.filterPseudo && node.Topology == Pseudo || f.FilterFunc(node) {
	output[id] = node
	inDegrees[id] = 0
} else {
	filtered++
}
```

2opremio: Yep, I think so:

Tried `((a | b) & c) || (~((a | b) & c) & (~a & ~b)) ` at http://www.boolengine.com/exp2exp.php

app/api_topologies.go
```diff
@@ -224,7 +244,8 @@ type APITopologyOption struct {
 	Value string `json:"value"`
 	Label string `json:"label"`
 
-	filter render.FilterFunc
+	filter       render.FilterFunc
+	filterPseudo bool
```
2opremio: Do we really need to `filterPseudo` field? Can't we do it including it implicitly in `render.FilterFunc`? (through function composition maybe)

render/filters.go
tomwilkie: Aww man, I was trying to work out the simplification...

render/filters.go
```diff
@@ -73,25 +73,24 @@ func ComposeFilterFuncs(fs ...FilterFunc) FilterFunc {
 // Filter removes nodes from a view based on a predicate.
 type Filter struct {
 	Renderer
-	FilterFunc   FilterFunc
-	filterPseudo bool
```
2opremio: Can't `filterPseudo` also be removed on `APITopologyOption`?

tomwilkie: Okay so I put this together and its a much bigger change.  I'm not super keen on the changing this, but have a look yourself and let me know:
[diff.patch.zip](https://github.com/weaveworks/scope/files/316907/diff.patch.zip)

2opremio: My idea was to add an `And` combinator (like `Complement`) and combine it with  `IsNotPseudo`  in the APITopologyOptions, instead of embedding `IgnorePseudo` in the filter definitions.

Something like

```go
func And(f FilterFunc, g FilterFunc) FilterFunc {
	 return func(node report.Node) bool { return f(node) && g(node)}
}

option :=  APITopologyOption{"stopped", "Stopped containers", render.And(render.IsStopped, render.IsNotPseudo)}

```

Would that make the code simpler? I think that could make it cleaner but feel free to disregard it if you think it's too complex.

tomwilkie: Interesting idea. Will give it a go tomorrow morn.

A lot of these functions are only used in tests and can be moved there
making the surface area smaller.

app/api_topologies.go
```diff
@@ -21,6 +21,14 @@ var (
 	topologyRegistry = &registry{
 		items: map[string]APITopologyDesc{},
 	}
+	k8sPseudoFilter = APITopologyOptionGroup{
+		ID:      "pseudo",
+		Default: "show",
+		Options: []APITopologyOption{
+			{"show", "Show unmanaged nodes", nil, false},
+			{"hide", "Hide unmanaged nodes", render.IsNotPseudo, true},
```
rade: not sure about the "nodes". Drop it? @davkal ?

app/api_topologies.go
```diff
@@ -29,18 +37,26 @@ func init() {
...
+				{"running", "Running containers", render.IsRunning, false},
+				{"both", "Both", nil, false},
+			},
+		},
+		{
+			ID:      "pseudo",
+			Default: "show",
+			Options: []APITopologyOption{
+				{"show", "Show uncontained nodes", nil, false},
+				{"hide", "Hide uncontained nodes", render.IsNotPseudo, true},
```
rade: not sure about the "nodes". Drop it? @davkal ?

render/filters.go
tomwilkie: @2opremio had a go and its equally pervasive.  The problem is changing the default filter to not ignore pseudo nodes pushes that into every calling site, for the sake of these few places where we want to not ignore pseudo node.  

I did manage to move some code into tests though.

PR #1583, opened 2016-06-14
Fix tall search box in Firefox

davkal:
* FF has an issue with flexbox and set width
* setting flex only when focused
* downside is that there is a small jump to 2 rows when expanding the box in FF

### DIFF
--- a/client/app/styles/main.less
+++ b/client/app/styles/main.less
@@ -1271,8 +1271,7 @@ h2 {
       border-radius: 0;
       background: transparent;
       color: @text-color;
-      flex: 1;
-      width: 60px;
+      width: 100px;
 
       &:focus {
         outline: none;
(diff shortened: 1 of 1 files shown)


foot:
You do get a "flash-of-tallness" on focus. Which is a little distracting but still better than before.

Removing the flex-wrap fixes this too but might be breaking another edge case?

```diff
diff --git a/client/app/styles/main.less b/client/app/styles/main.less
index da5f5d8..417947d 100644
--- a/client/app/styles/main.less
+++ b/client/app/styles/main.less
@@ -1260,7 +1260,6 @@ h2 {
     border: @search-border-width solid @search-border-color;
     padding: 2px 4px;
     text-align: left;
-    flex-wrap: wrap;

     &-field {
       font-size: 0.8rem;
```

davkal:
Here is the edge case: 

screen shot 2016-06-14 at 17 27 50

Pinned searches are supposed to wrap.

PR #1586, opened 2016-06-14
updated Weave Cloud and clarified setup

abuehrle:
@tomwilkie  made a few tweaks to the wording on Weave Cloud.  Maybe include a section on how to sign up, but I'm not sure that's really needed.

### DIFF
--- a/site/installing.md
+++ b/site/installing.md
@@ -9,9 +9,10 @@ Weave Scope consists of three parts: the probe, the app and the user interface. 
 The following topics are discussed:
 
  * [Installing Scope on Docker](#docker)
-   * [Using Weave Net](#docker-weave)
+   * [Enabling Weave Cloud](#docker-weave)
+   * [Weave Net and Scope](#net-scope)
    * [Using Docker Compose](#docker-compose)
-   * [Using Docker Compose in Weave Cloud](#docker-compose-sercice)
+   * [Using Docker Compose in Weave Cloud](#docker-compose-cloud)
  * [Installing Scope on Kubernetes in Standalone Mode or in Weave Cloud](#k8s)
  * [Installing Scope on Amazon ECS](#ecs)
  * [Installing Scope on minimesos](#minimesos)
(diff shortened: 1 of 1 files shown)


site/installing.md
```diff
@@ -4,23 +4,22 @@ menu_order: 20
...
 
 The following topics are discussed:
 
  * [Installing Scope on Docker](#docker)
-   * [In Service Mode](#docker-service)
    * [Using Weave Net](#docker-weave)
    * [Using Docker Compose](#docker-compose)
-   * [Using Docker Compose in Service Mode](#docker-compose-sercice)
- * [Installing Scope on Kubernetes](#k8s)
+   * [Using Docker Compose in Weave Cloud](#docker-compose-sercice)
```
rade: That link goes nowhere. Not only is it misspelled (always was), but there is no corresponding anchor; instead we have two 'docker-compose' anchors.

rade: Actually, this whole sub-heading list doesn't match the document at all, which has, in order:

- Enabling Weave Cloud (`#docker-service`)
- Using Weave Net (`#docker-weave`)
- Using Docker Compose (`#docker-compose`)
- Using Docker Compose with Weave Cloud (`#docker-compose`)

site/installing.md
```diff
@@ -172,22 +178,22 @@ To enable DaemonSets in an existing cluster, add the `--runtime-config=extension
...
-> The `--validate=false` flag is currently required due to a bug in Kubernetes (see
-[kubernetes/kubernetes#24089](https://github.com/kubernetes/kubernetes/issues/24089) for more details
+    kubectl create -f 'https://cloud.weave.works/launch/k8s/weavescope.yaml' --validate=false
 
 Allowable parameters for the launcher URL:
 
-- `v` - Weave Scope version or tag, e.g. `latest` or `0.15.0`, current release is the default
-- `service-token` - Weave Scope Cloud Service token
+- `v` - Weave Scope version or tag, e.g. `latest` current release is the default
+- `service-token` - Weave Cloud Service token (necessary if you want to run Weave Cloud)
```
2opremio: Can you maybe elaborate on how to find the token? Also, it may be helpful to include a ` kubectl create` example using the with the `service-token` parameter

davkal: Created #1596 for that.

PR #1590, opened 2016-06-15
Get route name before munging request

jml:
This ought to correct a problem where we incorrectly report route names
for admin routes.

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1590)


tomwilkie:
LGMT when green

PR #1591, opened 2016-06-15
Add all docker subnets to overlay topology

### LINKED ISSUE #1563: Missing edges when using Docker's IPAM driver
Scope is not displaying some connections while we are running continuous load through our application.

Specifically, the front-end service is only shown as having one outbound connection, to the catalogue service. In fact it is communicating with the login, orders, accounts, and carts services as well. I have verified this via our log files.

Please ping me if you need any additional information.

[weave_report_and_svg.zip](https://github.com/weaveworks/scope/files/304731/weave_report_and_svg.zip)

2opremio:
Fixes #1563


tomwilkie:
LGTM

Issue #1592, opened 2016-06-16
repeated colours for networks


rade:
Networks in the new network overlay are often non-distinct, i.e. many networks have same(ish) colour.

Similar to #783.

davkal:
Fixed by #1593. Better solution sought in #1567

PR #1593, opened 2016-06-16
Network colors from a scale

davkal:
* network-name independent
* all colors are unique and separated enough
* only 10 colors are available
* contains red and green

### DIFF
--- a/client/app/scripts/charts/node-networks-overlay.js
+++ b/client/app/scripts/charts/node-networks-overlay.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import d3 from 'd3';
 import { List as makeList } from 'immutable';
-import { getNodeColor } from '../utils/color-utils';
+import { getNetworkColor } from '../utils/color-utils';
 import { isContrastMode } from '../utils/contrast-utils';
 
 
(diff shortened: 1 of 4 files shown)


rade:
Is this meant to fix #1592?

> network-name independent

Does that mean the same network may be coloured differently, depending on context?

davkal:
It's a hack that fixes #1592 for network counts <= 10.

The colouring depends on the order of which we see the networks for the first time. That means it's stable while you keep the browser window open. It may be different if you reload or visit scope at a later point.

foot:
I think its a good, safe solution for now, will work well for demos etc.

I'll give it a test.

rade:
> network counts <= 10

what happens when you have more? Colours get re-used? Or does it explode?

foot:
Repeats. category20 gives us 20 tho...

http://jsbin.com/kedozadozu/edit?html,js,output

davkal:
cat20 is not distinct enough. It's 10 hues, with 2 different shades each:

https://github.com/d3/d3/wiki/Ordinal-Scales#categorical-colors

We could brew our own color scale...

rade:
stick to cat10. merge it. move on to #783.

PR #1595, opened 2016-06-16
Updates to vendoring document

jml:
Fixes for spelling & correctness. Also provide a link to `gvt`, which has surprisingly bad Google juice.

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1595)

### DIFF
--- a/vendor/README.md
+++ b/vendor/README.md
@@ -4,14 +4,21 @@ These operations result in uncommitted changes to your
 branch; you will need to commit them as normal. Execute them in the
 root of your checkout.
 
-## Adding a New Dependency
+For these changes to take effect, you'll have to `make clean` before running
+`make`.
 
-    ~/scope$ gvt fetch example.com/organisation/module vendor/example.com/organisation/module
+## Installing gvt
 
-## Updateing a Specific Dependancy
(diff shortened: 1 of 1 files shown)


vendor/README.md
```diff
@@ -4,14 +4,21 @@ These operations result in uncommitted changes to your
...
-## Updateing a Specific Dependancy
+    $ go get -u github.com/FileSottile/gvt
 
-    ~/scope$ gvt update example.com/organisation/module vendor/example.com/organisation/module
+## Adding a Dependency
 
-## Remove a Dependency
+    ~/service$ gvt fetch example.com/organisation/module
 
-    ~/scope$ gvt delete example.com/organisation/module vendor/example.com/organisation/module
```
errordeveloper: I don't think that even worked...

errordeveloper:
LGTM, but I haven't used `gvt` very extensively, so perhaps someone else should take a second look.

vendor/README.md
jml: Correct. I tried it initially and it failed. The above was the command that actually made things work for me.

Issue #971, opened 2016-02-17
Remove gob encoding support


2opremio:
After https://github.com/weaveworks/scope/pull/916  the probe<->app communication is done in msgpack but the plan is to leave gob around until at least 0.14 for backwards compatibility.

See https://github.com/weaveworks/scope/pull/916#discussion_r52733650

2opremio:
This may already be fixed, but there still might be some traces of gob encoding in the code.

PR #1600, opened 2016-06-17
Helper for reading & writing from binary

jml:
I found a couple of places where we had said in code that reports are serialized as compress msgpacks. This patch formalizes that convention by providing an API.

I'm working on another PR where I re-use this API in a different place.

There are a bunch of places in tests that could probably also be changed.

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1600)

### DIFF
--- a/app/router.go
+++ b/app/router.go
@@ -1,8 +1,6 @@
 package app
 
 import (
-	"compress/gzip"
-	"encoding/gob"
 	"io"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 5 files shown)


tomwilkie:
Should we factor out https://github.com/weaveworks/scope/blob/master/app/router.go#L121 as well?  Would need to support JSON.

Otherwise LGTM.

jml:
I did the refactoring. Basically I make `ReadBinary` take more options and leave `MakeFromBinary` being the default gzipped msgpack. I'm not so fond of the names any more, but I think it's an incremental improvement.

report/marshal.go
```diff
@@ -0,0 +1,53 @@
...
+		return err
+	}
+	if err = codec.NewEncoder(gzwriter, &codec.MsgpackHandle{}).Encode(&rep); err != nil {
+		return err
+	}
+	gzwriter.Close() // otherwise the content won't get flushed to the output stream
+	return nil
+}
+
+// ReadBinary reads into a Report from a gzipped msgpack.
```
rade: this no longer matches the functionality

report/marshal.go
```diff
@@ -0,0 +1,53 @@
...
+//
+// Will decompress the binary if gzipped is true, and will use the given
+// codecHandle to decode it. If codecHandle is nil, will decode as a gob.
```
rade: I find the gob hack in this really jarring, but cannot think of a neat way of avoiding it.

report/marshal.go
```diff
@@ -0,0 +1,53 @@
...
+func (rep *Report) ReadBinary(r io.Reader, gzipped bool, codecHandle codec.Handle) error {
+	var err error
+	if gzipped {
+		r, err = gzip.NewReader(r)
+		if err != nil {
+			return err
+		}
+	}
+	decoder := gob.NewDecoder(r).Decode
```
rade: I know it was like this before, but... we are wasting memory allocation here if codecHandle != nil. I suggest paying the price of an extra 'var' and 'else' to avoid that.

app/router.go
```diff
@@ -122,43 +120,22 @@ func RegisterReportPostHandler(a Adder, router *mux.Router) {
...
+		if err := rpt.ReadBinary(reader, gzipped, handle); err != nil {
 			respondWith(w, http.StatusBadRequest, err)
 			return
 		}
-		log.Debugf(
-			"Received report sizes: compressed %d bytes, uncompressed %d bytes (%.2f%%)",
-			compressedSize,
-			uncompressedSize,
-			float32(compressedSize)/float32(uncompressedSize)*100,
-		)
```
rade: We are losing this debug output. Do we care?

report/marshal.go
jml: Done.

report/marshal.go
2opremio: I think we can get rid of gob altogether. It's been a long time since probes provide msgpack.

app/router.go
jml: Apparently. I've restored the logic & added a comment saying why we think it's worth having the extra complexity.

jml:
PTAL.

report/marshal.go
jml: Dropped.

report/marshal.go
```diff
@@ -56,13 +55,7 @@ func (rep *Report) ReadBinary(r io.Reader, gzipped bool, codecHandle codec.Handl
 	if log.GetLevel() == log.DebugLevel {
 		r = byteCounter{next: r, count: &uncompressedSize}
 	}
-	var decoder func(interface{}) error
-	if codecHandle != nil {
-		decoder = codec.NewDecoder(r, codecHandle).Decode
-	} else {
-		decoder = gob.NewDecoder(r).Decode
```
2opremio: Awesome step forward for https://github.com/weaveworks/scope/issues/971. It's a larger task but it would be good to remove all the GobEncode/GobDecode methods  (probably in a separate PR).

app/router.go
```diff
@@ -102,63 +99,27 @@ func RegisterTopologyRoutes(router *mux.Router, r Reporter) {
...
 		if strings.HasPrefix(r.Header.Get("Content-Type"), "application/json") {
-			decoder = codec.NewDecoder(reader, &codec.JsonHandle{}).Decode
+			handle = &codec.JsonHandle{}
 		} else if strings.HasPrefix(r.Header.Get("Content-Type"), "application/msgpack") {
-			decoder = codec.NewDecoder(reader, &codec.MsgpackHandle{}).Decode
+			handle = &codec.MsgpackHandle{}
 		}
 
-		if err := decoder(&rpt); err != nil {
+		if err := rpt.ReadBinary(reader, gzipped, handle); err != nil {
```
rade: `handle` must be non-null here, right? Otherwise things explode, probably with a rather obscure error. Perhaps we should enforce the non-nullness here by adding another branch to the preceding if/else. Or, even better, turn that into a `switch` with a default case that complains suitably.

jml: Done.

jml:
@tomwilkie @rade PTAL.

tomwilkie:
LGTM but the build is failing for a reason that seems related...

jml:
The test was for gob support, which this branch removes.

PR #1602, opened 2016-06-20
Query memcached from dynamo db collector

jml:
* Extract the report (de)serialization logic to a set of common functions
* Add options for using memcached to find & store reports
* Vendor gomemcache client
* Update `timeRequest` to parametrize the error -> status behaviour

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1602)

### DIFF
--- a/app/multitenant/common.go
+++ b/app/multitenant/common.go
@@ -13,10 +13,28 @@ func errorCode(err error) string {
 	return "500"
 }
 
+// timeRequest runs 'f' and records how long it took in the given Prometheus
+// metric. If 'f' returns successfully, record a "200". Otherwise, record
+// "500".
+//
+// If you want more complicated logic for translating errors into statuses,
+// use 'timeRequestStatus'.
 func timeRequest(method string, metric *prometheus.SummaryVec, f func() error) error {
+	return timeRequestStatus(method, metric, errorCode, f)
+}
+
+// timeRequestStatus runs 'f' and records how long it took in the given
+// Prometheus metric.
+//
+// toStatusCode is a function that translates errors returned by 'f' into
+// HTTP-like status codes.
+func timeRequestStatus(method string, metric *prometheus.SummaryVec, toStatusCode func(error) string, f func() error) error {
+	if toStatusCode == nil {
+		toStatusCode = errorCode
+	}
 	startTime := time.Now()
 	err := f()
 	duration := time.Now().Sub(startTime)
-	metric.WithLabelValues(method, errorCode(err)).Observe(float64(duration.Nanoseconds()))
+	metric.WithLabelValues(method, toStatusCode(err)).Observe(float64(duration.Nanoseconds()))
 	return err
 }
--- a/app/multitenant/dynamo_collector.go
+++ b/app/multitenant/dynamo_collector.go
@@ -15,6 +15,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/bluele/gcache"
+	"github.com/bradfitz/gomemcache/memcache"
 	"github.com/nats-io/nats"
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/net/context"
@@ -24,12 +25,16 @@ import (
 )
 
 const (
-	hourField             = "hour"
-	tsField               = "ts"
-	reportField           = "report"
-	reportCacheSize       = (15 / 3) * 10 * 5 // (window size * report rate) * number of hosts per user * number of users
-	reportCacheExpiration = 15 * time.Second
-	natsTimeout           = 10 * time.Second
+	hourField              = "hour"
+	tsField                = "ts"
+	reportField            = "report"
+	reportCacheSize        = (15 / 3) * 10 * 5 // (window size * report rate) * number of hosts per user * number of users
+	reportCacheExpiration  = 15 * time.Second
+	memcacheExpiration     = 15 // seconds
+	memcacheUpdateInterval = 1 * time.Minute
+	natsTimeout            = 10 * time.Second
+	hitLabel               = "hit"
+	missLabel              = "miss"
 )
 
 var (
@@ -38,16 +43,6 @@ var (
 		Name:      "dynamo_request_duration_nanoseconds",
 		Help:      "Time spent doing DynamoDB requests.",
 	}, []string{"method", "status_code"})
-	dynamoCacheHits = prometheus.NewCounter(prometheus.CounterOpts{
-		Namespace: "scope",
-		Name:      "dynamo_cache_hits",
-		Help:      "Reports fetches that hit local cache.",
-	})
-	dynamoCacheMiss = prometheus.NewCounter(prometheus.CounterOpts{
-		Namespace: "scope",
-		Name:      "dynamo_cache_miss",
-		Help:      "Reports fetches that miss local cache.",
-	})
 	dynamoConsumedCapacity = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "scope",
 		Name:      "dynamo_consumed_capacity",
@@ -58,6 +53,12 @@ var (
 		Name:      "dynamo_value_size_bytes",
 		Help:      "Size of data read / written from dynamodb.",
 	}, []string{"method"})
+
+	inProcessCacheCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "scope",
+		Name:      "in_process_cache",
+		Help:      "Reports fetches that hit in-process cache.",
+	}, []string{"result"})
 
 	reportSize = prometheus.NewCounter(prometheus.CounterOpts{
 		Namespace: "scope",
@@ -80,10 +81,9 @@ var (
 
 func init() {
 	prometheus.MustRegister(dynamoRequestDuration)
-	prometheus.MustRegister(dynamoCacheHits)
-	prometheus.MustRegister(dynamoCacheMiss)
 	prometheus.MustRegister(dynamoConsumedCapacity)
 	prometheus.MustRegister(dynamoValueSize)
+	prometheus.MustRegister(inProcessCacheCounter)
 	prometheus.MustRegister(reportSize)
 	prometheus.MustRegister(s3RequestDuration)
 	prometheus.MustRegister(natsRequests)
@@ -95,6 +95,11 @@ type DynamoDBCollector interface {
 	CreateTables() error
 }
 
+// ReportStore is a thing that we can get reports from.
+type ReportStore interface {
+	FetchReports([]string) ([]report.Report, []string, error)
+}
+
 type dynamoDBCollector struct {
 	userIDer   UserIDer
 	db         *dynamodb.DynamoDB
@@ -102,7 +107,8 @@ type dynamoDBCollector struct {
 	tableName  string
 	bucketName string
 	merger     app.Merger
-	cache      gcache.Cache
+	inProcess  inProcessStore
+	memcache   *MemcacheClient
 
 	nats        *nats.Conn
 	waitersLock sync.Mutex
@@ -128,13 +134,30 @@ type watchKey struct {
 func NewDynamoDBCollector(
 	userIDer UserIDer,
 	dynamoDBConfig, s3Config *aws.Config,
-	tableName, bucketName, natsHost string,
+	tableName, bucketName, natsHost, memcachedHost string,
+	memcachedTimeout time.Duration, memcachedService string,
 ) (DynamoDBCollector, error) {
 	var nc *nats.Conn
 	if natsHost != "" {
 		var err error
 		nc, err = nats.Connect(natsHost)
 		if err != nil {
+			return nil, err
+		}
+	}
+
+	var memcacheClient *MemcacheClient
+	if memcachedHost != "" {
+		var err error
+		memcacheClient, err = NewMemcacheClient(memcachedHost, memcachedTimeout, memcachedService, memcacheUpdateInterval, memcacheExpiration)
+		if err != nil {
+			// TODO(jml): Ideally, we wouldn't abort here, we would instead
+			// log errors when we try to use the memcache & fail to do so, as
+			// aborting here introduces ordering dependencies into our
+			// deployment.
+			//
+			// Note: this error only happens when either the memcachedHost or
+			// any of the SRV records that it points to fail to resolve.
 			return nil, err
 		}
 	}
@@ -146,7 +169,8 @@ func NewDynamoDBCollector(
 		tableName:  tableName,
 		bucketName: bucketName,
 		merger:     app.NewSmartMerger(),
-		cache:      gcache.New(reportCacheSize).LRU().Expiration(reportCacheExpiration).Build(),
+		inProcess:  newInProcessStore(reportCacheSize, reportCacheExpiration),
+		memcache:   memcacheClient,
 		nats:       nc,
 		waiters:    map[watchKey]*nats.Subscription{},
 	}, nil
@@ -252,20 +276,6 @@ func (c *dynamoDBCollector) getReportKeys(rowKey string, start, end time.Time) (
 	return result, nil
 }
 
-func (c *dynamoDBCollector) getCached(reportKeys []string) ([]report.Report, []string) {
-	foundReports := []report.Report{}
-	missingReports := []string{}
-	for _, reportKey := range reportKeys {
-		rpt, err := c.cache.Get(reportKey)
-		if err == nil {
-			foundReports = append(foundReports, rpt.(report.Report))
-		} else {
-			missingReports = append(missingReports, reportKey)
-		}
-	}
-	return foundReports, missingReports
-}
-
 // Fetch multiple reports in parallel from S3.
 func (c *dynamoDBCollector) getNonCached(reportKeys []string) ([]report.Report, error) {
 	type result struct {
@@ -291,7 +301,7 @@ func (c *dynamoDBCollector) getNonCached(reportKeys []string) ([]report.Report, 
 			return nil, r.err
 		}
 		reports = append(reports, *r.report)
-		c.cache.Set(r.key, *r.report)
+		c.inProcess.StoreReport(r.key, *r.report)
 	}
 	return reports, nil
 }
@@ -315,24 +325,46 @@ func (c *dynamoDBCollector) getNonCachedReport(reportKey string) (*report.Report
 
 func (c *dynamoDBCollector) getReports(userid string, row int64, start, end time.Time) ([]report.Report, error) {
 	rowKey := fmt.Sprintf("%s-%s", userid, strconv.FormatInt(row, 10))
-	reportKeys, err := c.getReportKeys(rowKey, start, end)
+	missing, err := c.getReportKeys(rowKey, start, end)
 	if err != nil {
 		return nil, err
 	}
 
-	cachedReports, missing := c.getCached(reportKeys)
-	dynamoCacheHits.Add(float64(len(cachedReports)))
-	dynamoCacheMiss.Add(float64(len(missing)))
-	if len(missing) == 0 {
-		return cachedReports, nil
+	var reports []report.Report
+	for _, store := range []ReportStore{c.inProcess, c.memcache} {
+		if store == nil {
+			continue
+		}
+		found, missing, err := store.FetchReports(missing)
+		if err != nil {
+			log.Warningf("Error fetching from cache: %v", err)
+		}
+		reports = append(reports, found...)
+		if len(missing) == 0 {
+			return reports, nil
+		}
 	}
 
-	fetchedReport, err := c.getNonCached(missing)
+	fetchedReports, err := c.getNonCached(missing)
 	if err != nil {
 		return nil, err
 	}
 
-	return append(cachedReports, fetchedReport...), nil
+	return append(reports, fetchedReports...), nil
+}
+
+func memcacheStatusCode(err error) string {
+	// See https://godoc.org/github.com/bradfitz/gomemcache/memcache#pkg-variables
+	switch err {
+	case nil:
+		return "200"
+	case memcache.ErrCacheMiss:
+		return "404"
+	case memcache.ErrMalformedKey:
+		return "400"
+	default:
+		return "500"
+	}
 }
 
 func (c *dynamoDBCollector) Report(ctx context.Context) (report.Report, error) {
@@ -402,7 +434,20 @@ func (c *dynamoDBCollector) Add(ctx context.Context, rep report.Report) error {
 		return err
 	}
 
-	// third, put the key in dynamodb
+	// third, put it in memcache
+	if c.memcache != nil {
+		err = timeRequestStatus("Put", memcacheRequestDuration, memcacheStatusCode, func() error {
+			return c.memcache.StoreBytes(s3Key, buf.Bytes())
+		})
+		if err != nil {
+			// NOTE: We don't abort here because failing to store in memcache
+			// doesn't actually break anything else -- it's just an
+			// optimization.
+			log.Warningf("Could not store %v in memcache: %v", s3Key, err)
+		}
+	}
+
+	// fourth, put the key in dynamodb
 	dynamoValueSize.WithLabelValues("PutItem").
 		Add(float64(len(s3Key)))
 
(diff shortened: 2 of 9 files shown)


app/multitenant/common.go
```diff
@@ -13,10 +13,19 @@ func errorCode(err error) string {
...
+func timeRequest(method string, metric *prometheus.SummaryVec, toStatusCode func(error) string, f func() error) error {
+	if toStatusCode == nil {
+		toStatusCode = errorCode
+	}
 	startTime := time.Now()
 	err := f()
 	duration := time.Now().Sub(startTime)
-	metric.WithLabelValues(method, errorCode(err)).Observe(float64(duration.Nanoseconds()))
+	metric.WithLabelValues(method, toStatusCode(err)).Observe(float64(duration.Nanoseconds()))
 	return err
```
tomwilkie: is errCode no longer used?  if so delete it.

app/multitenant/dynamo_collector.go
```diff
@@ -78,6 +81,24 @@ var (
...
+		Namespace: "scope",
+		Name:      "memcache_hits",
+		Help:      "Reports that missed our in-memory cache but went to our memcache",
+	})
+
+	memcacheMiss = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "scope",
+		Name:      "memcache_miss",
+		Help:      "Reports that missed both our in-memory cache and our memcache",
+	})
```
tomwilkie: A better way of doing this is to have a single memcache counter, with a label for hit or miss.  Makes writing the queries easier.  You could also fix up the other hit/miss counter too.

app/multitenant/dynamo_collector.go
```diff
@@ -78,6 +81,24 @@ var (
...
+
+	memcacheRequestDuration = prometheus.NewSummaryVec(prometheus.SummaryOpts{
+		Namespace: "scope",
+		Name:      "memcache_request_duration_nanoseconds",
+		Help:      "Time spent doing memcache requests.",
+	}, []string{"method", "status_code"})
```
tomwilkie: The summaryVecs export a _count metric, making the hit/miss metrics redundant above (if you do the right labelling).

app/multitenant/dynamo_collector.go
```diff
@@ -130,7 +155,8 @@ type watchKey struct {
 func NewDynamoDBCollector(
 	userIDer UserIDer,
 	dynamoDBConfig, s3Config *aws.Config,
-	tableName, bucketName, natsHost string,
+	tableName, bucketName, natsHost, memcachedHost string,
+	memcachedTimeout time.Duration, memcachedService string,
```
tomwilkie: I think its probably time to have a Options struct?

app/multitenant/common.go
tomwilkie: Oh I see.  Ignore me.

app/multitenant/dynamo_collector.go
```diff
@@ -330,13 +386,35 @@ func (c *dynamoDBCollector) getReports(userid string, row int64, start, end time
...
+	if c.memcache != nil {
+		var memcachedReports []report.Report
+		memcachedReports, missing, err = c.fetchFromMemcache(missing)
+		memcacheHits.Add(float64(len(memcachedReports)))
+		memcacheMiss.Add(float64(len(missing)))
+		if err != nil {
+			// REVIEWER: jml is unclear whether we should abort in this case or
+			// just carry on. Aborting is easier to reason about for us, but
+			// suppressing is probably OK since failure just means we fetch
+			// from S3.
```
tomwilkie: In what circumstances will it error?

I guess it makes sense to just log this (errors should show up in memcache monitoring, right?) and continue in this case.  Failure to get the report from S3 (the source of truth) should error.

app/multitenant/dynamo_collector.go
jml: Agreed. Mind if I do that in a follow-up PR?

app/multitenant/dynamo_collector.go
jml: * fail to successfully talk to memcache: dodgy network, dodgy server. Monitoring should catch this
* corrupt report in the memcache: we have a warning for this but no monitoring

I'll delete the comment and leave it erroring for now.

app/multitenant/dynamo_collector.go
jml: Given that we use `GetMulti` for our memcache queries, would it work to label each request with num hits & num total? (i.e. we can't simply have a boolean hit-or-miss label)

jml:
Done some more local testing -- everything works. Note: after scaling replicas, SRV resolution seems to fail until I delete & recreate the k8s memcached-svc

app/multitenant/dynamo_collector.go
tomwilkie: Good point; lets just have a single memcache hit/miss counter then.

app/multitenant/dynamo_collector.go
tomwilkie: Not particularly.

app/multitenant/dynamo_collector.go
```diff
@@ -141,6 +156,25 @@ func NewDynamoDBCollector(
...
 
+	var memcacheClient *memcache.Client
+	if memcachedHost != "" {
+		var memcacheServers memcache.ServerList
+		memcacheClient = memcache.NewFromSelector(&memcacheServers)
+		memcacheClient.Timeout = memcachedTimeout
+		err := updateMemcacheServers(memcachedHost, memcachedService, &memcacheServers)
+		if err != nil {
+			// REVIEWER: Undecided whether this should exit or not. Thoughts?
+			log.Errorf("Could not set memcache servers: %v", err)
```
tomwilkie: Whats the behaviour if you run this thing and it can't talk to the servers?

What we want: a client that can be instantiated without throwing errors if it cannot resolve/contact the servers, but which then throws errors on individual operations.

app/multitenant/dynamo_collector.go
```diff
@@ -141,6 +156,25 @@ func NewDynamoDBCollector(
...
+		}
+
+		ticker := time.NewTicker(memcacheUpdateInterval)
+		go func() {
+			for range ticker.C {
+				updateMemcacheServers(memcachedHost, memcachedService, &memcacheServers)
+			}
+		}()
```
tomwilkie: I'd like to encapsulate this behaviour in a separate file / type please, much like we do with the consul client.

app/multitenant/dynamo_collector.go
```diff
@@ -149,11 +183,30 @@ func NewDynamoDBCollector(
...
+	var servers []string
+	for _, srv := range addrs {
+		servers = append(servers, fmt.Sprintf("%s:%d", srv.Target, srv.Port))
+	}
+	// ServerList deterministically maps keys to _index_ of the server list.
+	// Since DNS returns records in different order each time, we sort to
+	// guarantee best possible match between nodes.
+	sort.Strings(servers)
+	return serverList.SetServers(servers...)
+}
```
tomwilkie: I'd like to encapsulate this behaviour in a separate file / type please, much like we do with the consul client.
 Add a line note

app/multitenant/common.go
```diff
@@ -13,10 +13,19 @@ func errorCode(err error) string {
...
+// timeRequest runs 'f' and records how long it took in the given Prometheus
+// metric.
+//
+// toStatusCode is a function that translates errors returned by 'f' into
+// HTTP-like status codes. If 'nil', uses a default function that returns
+// "500" for any error and "200" otherwise.
+func timeRequest(method string, metric *prometheus.SummaryVec, toStatusCode func(error) string, f func() error) error {
+	if toStatusCode == nil {
+		toStatusCode = errorCode
+	}
```
tomwilkie: I'm not a big fan of passing nil as a meaningful default.

Perhaps a second function (`timeRequest` vs `timeRequestStatus`) the first with the original semantics, the second which takes a fn -> string as a parameter)

app/multitenant/dynamo_collector.go
```diff
@@ -330,9 +375,27 @@ func (c *dynamoDBCollector) getReports(userid string, row int64, start, end time
 		return nil, err
 	}
 
+	// TODO(jml): Refactor this so that we have a standard interface
+	// IReportStore for "fetch :: [ReportKeys] -> ([Report], missing, error)"
+	// and a list of providers: in-memory cache, memcache, dynamo. Possibly
+	// also an "always everything missing" implementation to simplify logic.
```
tomwilkie: I say do this refactoring now, but in a lightweight fashion:

```go
for _, cachefn in range []func([]string) ([]report, []string, err) {
  c.getInProcessCache,
  c.getMemcache,
} {
  report, missing, err := cachefn(missing)
  ...
}
```

tomwilkie: And maybe break out the type to make it more readable.

app/multitenant/dynamo_collector.go
```diff
@@ -345,6 +408,72 @@ func (c *dynamoDBCollector) getReports(userid string, row int64, start, end time
...
+
+func (c *dynamoDBCollector) fetchFromMemcache(reportKeys []string) ([]report.Report, []string, error) {
+	var foundReports map[string]*memcache.Item
+	err := timeRequest("Get", memcacheRequestDuration, memcacheStatusCode, func() error {
+		var err error
+		foundReports, err = c.memcache.GetMulti(reportKeys)
+		if err != nil {
+			return err
+		}
+		return nil
```
tomwilkie: why u no:

```go
var err error
foundReports, err = c.memcache.GetMulti(reportKeys)
return err
```

app/multitenant/dynamo_collector.go
```diff
@@ -345,6 +408,72 @@ func (c *dynamoDBCollector) getReports(userid string, row int64, start, end time
...
+					log.Warningf("Corrupt report in memcache %v: %v", reportKey, err)
+					ch <- result{key: reportKey}
+				} else {
+					ch <- result{report: rep}
+					c.cache.Set(reportKey, *rep)
+				}
+			}(reportKey)
+		} else {
+			missing = append(missing, reportKey)
+		}
```
tomwilkie: (Sorry to be so nit picky!)

This loop seems fussy.  How about:

```go
for _, reportKey := range reportKeys {
    item, ok := foundReports[reportKey]
    if !ok {
        missing = append(missing, reportKey)
        continue
    }
    
    go func(reportKey string) {
        ...
        }(reportKey)
}
```

app/multitenant/dynamo_collector.go
```diff
@@ -345,6 +408,72 @@ func (c *dynamoDBCollector) getReports(userid string, row int64, start, end time
...
+	if err != nil {
+		return nil, reportKeys, err
+	}
+
+	// Decode all the reports in parallel.
+	type result struct {
+		key    string
+		report *report.Report
+	}
+	ch := make(chan result, len(reportKeys))
```
tomwilkie: I would be tempted to have two channels, one for missing keys and one for reports, and unify the error handling and the missing handling in the loop below, something like:

```
missing := make(chan string, len(...))
reports := make(chan report, len(...))
for _, key := range reportKeys {
    item, ok := foundReports[key]
    go func(key string, buf []byte, bool found) {
        if !ok {
            missing <- key
            return
        }        
        
        rep, err := report.MakeFromBinary(bytes.NewReader(item.Value))
        if err != nil {
            log...
            missing <- key
            return
        }
        
        reports <- rep
    }(reportKey, item, ok)
}
```

WDYT?

app/multitenant/dynamo_collector.go
```diff
@@ -451,6 +573,22 @@ func (c *dynamoDBCollector) Add(ctx context.Context, rep report.Report) error {
 		return err
 	}
 
+	// fourth, put it in memcache
+	if c.memcache != nil {
+		// REVIEWER: Should we call this "Put" to be consistent with S3, or
+		// "Set" to be consistent with memcache?
```
tomwilkie: Ha I have no idea.  I don't think it really matters does it?  It will just come out in the graphs.

app/multitenant/dynamo_collector.go
```diff
@@ -345,6 +408,72 @@ func (c *dynamoDBCollector) getReports(userid string, row int64, start, end time
...
+		item, ok := foundReports[reportKey]
+		if ok {
+			go func(reportKey string) {
+				rep, err := report.MakeFromBinary(bytes.NewReader(item.Value))
+				if err != nil {
+					log.Warningf("Corrupt report in memcache %v: %v", reportKey, err)
+					ch <- result{key: reportKey}
+				} else {
+					ch <- result{report: rep}
+					c.cache.Set(reportKey, *rep)
```
tomwilkie: I feel this cache set should be moved to the outer function, although I'm not sure.

app/multitenant/dynamo_collector.go
```diff
@@ -451,6 +573,22 @@ func (c *dynamoDBCollector) Add(ctx context.Context, rep report.Report) error {
 		return err
 	}
 
+	// fourth, put it in memcache
+	if c.memcache != nil {
```
tomwilkie: Can we do this before we put into dynamodb pls? In fact, can we go in this order: memcache, dynamodb, s3?  That will allow reports to appear earlier in the UI.  OTOH the order needs to be s3, dyanamo when there is no memcache...

tomwilkie:
@jml I've left quite a few comments, let me know if you want to discuss.  Am going to test now, but will assign back to you for the time being to address these comments.

app/multitenant/dynamo_collector.go
tomwilkie: I see you've done that too.  Thanks!

app/multitenant/common.go
jml: Done.

app/multitenant/dynamo_collector.go
jml: Legacy from printf instrumentation. Fixed.

app/multitenant/dynamo_collector.go
jml: I agree.

app/multitenant/dynamo_collector.go
jml: ISTR you expressing an opinion on the subject. Leaving as "Put" for now & deleting the comment.

jml: Updated to: s3, memcache, dynamo.

app/multitenant/dynamo_collector.go
jml: @tomwilkie By "break out the type", do you mean define a named struct type & use that rather than a tuple? (I think that's a good idea)

app/multitenant/dynamo_collector.go
jml: I tried and couldn't figure out how to do this without adding a bunch of extra book-keeping to track how many things we'd sent to the `missing` channel.

Have updated to use your preferred early-exit style.

app/multitenant/dynamo_collector.go
jml: Done.

app/multitenant/dynamo_collector.go
jml: I've documented this. I agree with the desired behaviour, but I think it should wait for a follow-up, as this PR is already big and I'm keen to make incremental improvements.

app/multitenant/dynamo_collector.go
jml: Done.

app/multitenant/dynamo_collector.go
jml: I've done this refactoring, but I haven't broken out the type.

jml:
@tomwilkie Thanks for the thorough review. I've responded to all of your comments & have followed your suggestions on most. PTAL.

app/multitenant/dynamo_collector.go
```diff
@@ -59,6 +54,12 @@ var (
 		Help:      "Size of data read / written from dynamodb.",
 	}, []string{"method"})
 
+	inProcessCacheCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "scope",
+		Name:      "dynamo_cache", // TODO(jml): Rename this monitoring variable.
```
tomwilkie: As its renamed from dynamo_cache(_hits/_miss) already, might as well go the whole hog.

app/multitenant/dynamo_collector.go
```diff
@@ -315,24 +324,46 @@ func (c *dynamoDBCollector) getNonCachedReport(reportKey string) (*report.Report
...
-	if len(missing) == 0 {
-		return cachedReports, nil
+	var reports []report.Report
+	for _, store := range []ReportStore{c.inProcess, c.memcache} {
+		if store == nil {
+			continue
+		}
+		found, missing, err := store.FetchReports(missing)
+		if err != nil {
+			log.Warningf("Could not fetch from cache: %v", err)
```
tomwilkie: nit, I would probably go with "Error fetching from cache"

app/multitenant/memcache_client.go
```diff
@@ -0,0 +1,131 @@
...
+	client.Timeout = timeout
+	err := updateMemcacheServers(host, service, &servers)
+	if err != nil {
+		return nil, err
+	}
+
+	ticker := time.NewTicker(updateInterval)
+	go func() {
+		for range ticker.C {
+			updateMemcacheServers(host, service, &servers)
```
tomwilkie: You're swallowing an error here that should probably be logged.  I wonder why errcheck didn't catch this.

app/multitenant/memcache_client.go
```diff
@@ -0,0 +1,131 @@
...
+
+type memcacheClient struct {
+	client     *memcache.Client
+	expiration int32
+}
+
+// Create a new MemcacheClient that gets its server list from SRV and updates
+// the server list on a regular basis.
+func NewMemcacheClient(host string, timeout time.Duration, service string, updateInterval time.Duration, expiration int32) (*memcacheClient, error) {
+	var servers memcache.ServerList
```
tomwilkie: I'd probably be more comfortable is this was a member of memcacheClient, and not passed through the closure like this.

app/multitenant/memcache_client.go
```diff
@@ -0,0 +1,131 @@
...
+	client := memcache.NewFromSelector(&servers)
+	client.Timeout = timeout
+	err := updateMemcacheServers(host, service, &servers)
+	if err != nil {
+		return nil, err
+	}
```
tomwilkie: Why return the error here?  Don't we want this to not error if the dependancies can't be found?

app/multitenant/memcache_client.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil, keys, err
+	}
+
+	// Decode all the reports in parallel.
+	type result struct {
+		key    string
+		report *report.Report
+	}
+	ch := make(chan result, len(keys))
+	var missing []string
```
tomwilkie: I still like the idea of making missing a channel; means you can get rid of the awkward result struct.  WDYT?

app/multitenant/memcache_client.go
tomwilkie: Is this goroutine safe?

app/multitenant/memcache_client.go
```diff
@@ -0,0 +1,131 @@
...
+		servers = append(servers, fmt.Sprintf("%s:%d", srv.Target, srv.Port))
+	}
+	// ServerList deterministically maps keys to _index_ of the server list.
+	// Since DNS returns records in different order each time, we sort to
+	// guarantee best possible match between nodes.
+	sort.Strings(servers)
+	return serverList.SetServers(servers...)
+}
+
+func (c memcacheClient) FetchReports(keys []string) ([]report.Report, []string, error) {
```
tomwilkie: Should this be `(c *memcacheClient)`?

app/multitenant/memcache_client.go
```diff
@@ -0,0 +1,131 @@
...
+			reports = append(reports, *r.report)
+		}
+	}
+
+	memcacheCounter.WithLabelValues(hitLabel).Add(float64(len(reports)))
+	memcacheCounter.WithLabelValues(missLabel).Add(float64(len(missing)))
+	return reports, missing, nil
+}
+
+func (c memcacheClient) StoreBytes(key string, content []byte) error {
```
tomwilkie: Should this be `(c *memcacheClient)`?

app/multitenant/dynamo_collector.go
```diff
@@ -95,14 +95,19 @@ type DynamoDBCollector interface {
 	CreateTables() error
 }
 
+type ReportStore interface {
+	FetchReports([]string) ([]report.Report, []string, error)
+}
```
tomwilkie: Just out of interest, why did you go for a full blow interface and not just a function type?

app/multitenant/dynamo_collector.go
```diff
@@ -315,24 +324,46 @@ func (c *dynamoDBCollector) getNonCachedReport(reportKey string) (*report.Report
 
 func (c *dynamoDBCollector) getReports(userid string, row int64, start, end time.Time) ([]report.Report, error) {
```
tomwilkie: This function looks much nicer; thanks!

app/multitenant/dynamo_collector.go
tomwilkie: Ah I see, you want the nil test for memcache.  Understood.

app/multitenant/memcache_client.go
```diff
@@ -0,0 +1,131 @@
...
+		for range ticker.C {
+			updateMemcacheServers(host, service, &servers)
+		}
+	}()
+	return &memcacheClient{client, expiration}, nil
+}
+
+// updateMemcacheServers sets a memcache server list from SRV records. SRV
+// priority & weight are ignored.
+func updateMemcacheServers(hostname, service string, serverList *memcache.ServerList) error {
```
tomwilkie: its typical to put the loop in this function, and then call `go loop()` from the constructor.

tomwilkie: should also have a Stop function to stop this loop, using a close channel and waitgroup.

app/multitenant/dynamo_collector.go
jml: Good point. Done.

app/multitenant/dynamo_collector.go
jml: Changed.

app/multitenant/memcache_client.go
jml: Updated for your comfort (although I'd like to know more about why you feel that way).

Is calling `NewMemcacheClient` safe to call from goroutines? It doesn't mutate any non-local state, so I don't see why not. Or do you mean, is the goroutine in `NewMemcacheClient` safe? Yes, it only mutates `ServerList`, which is documented as being safe to call from goroutines.

app/multitenant/memcache_client.go
jml: Returning the error here to let code elsewhere decide policy on error handling. 

There's a comment in `dynamo_collector.go` that explains what's going on.

app/multitenant/memcache_client.go
jml: I think the extra book-keeping required to keep track of how many things has been sent to `missing` would make the routine more fragile.

app/multitenant/memcache_client.go
jml: I don't know. I made it a value method because it doesn't change any state (following the advice in the [go FAQ](https://golang.org/doc/faq#methods_on_values_or_pointers)).

app/multitenant/memcache_client.go
jml: I _think_ it's OK as is (the pointer to `client` gets copied and then `Set` is called on the original), but I've changed it because thinking about things sucks.

app/multitenant/memcache_client.go
jml: Logged as a warning. I don't have errcheck running in my flycheck atm because I've yet to install it on this machine..

app/multitenant/dynamo_collector.go
jml: A pleasure. Future PR will make S3 implement the same interface.

app/multitenant/memcache_client.go
jml: Both of these done.

jml:
@tomwilkie All responded. PTAL.

tomwilkie:
LGTM once green.

tomwilkie:
And a big squash!

PR #1604, opened 2016-06-20
readded signup page

abuehrle:


abuehrle:
@tomwilkie can you merge this, so that we can update the docs on the site?  There's a section on there with "old screen captures".

Issue #1599, opened 2016-06-17
failed weave interactions are logged as errors, even though it's normal operation


jml:
e.g.

```
<probe> ERRO: 2016/06/17 12:56:35.574649 'weave ps' cmd failed: exit status 1
```

It's confusing to use "ERROR" for things that are part of normal operation. This should either be a warning, or just removed from the logs completely.

rade:
ditto for
```
<app> ERRO: 2016/06/17 12:26:35.242523 Error updating weaveDNS, backing off 20s: exit status 1
```

> This should either be a warning, or just removed from the logs completely.

Removing would be a bad idea because if you *are* running weave then these messages *are* noteworthy.

jml:
We talked a bit about this in person and agreed that:

* scope shouldn't spam the logs if `weave` isn't there
* UX is complicated by scope not having any knowledge of whether the user _intends_ to use weave

peterbourgon:
Would be good to bump the priority on this one. Basically everyone trying to troubleshoot a Scope issue hits this and gets confused.

tomwilkie:
What changed to make this fail? It never used to fail.  Would be good to understand what changed.

rade:
> It never used to fail.

Are you saying that scope never used to log an error when it cannot talk to weave / weaveDNS? That's not my recollection.

rade:
I cannot reproduce the `weave ps` error, with or without weave installed.

I *do* get the `weaveDNS` error, with or without weave installed.

I also see
```
<probe> ERRO: 2016/06/20 09:35:18.442144 Error collecting weave info, backing off 10s: Get http://127.0.0.1:6784/report: dial tcp 127.0.0.1:6784: getsockopt: connection refused
```
with our without weave installed.

tomwilkie:
There are 3 integration points with weave:
1. in the probe, we fetch the weave report and use it to build/tag the containers topology (here)[https://github.com/weaveworks/scope/blob/master/probe/overlay/weave.go#L56]
2. in the probe, we run weave ps and tag containers with their weave ips (same place as 1.)
3. in the probe, we talk to weave dns to resolve things (here)[https://github.com/weaveworks/scope/blob/master/prog/probe.go#L195]
4. in the app, we talk to weave dns to insert a record for scope.weave.works (here)[https://github.com/weaveworks/scope/blob/master/app/weave.go#L80]

[...]

PR #1607, opened 2016-06-23
Print DNS errors

### LINKED ISSUE #1577: Scope probe fails to report to some of the target apps
When mirroring scope reports from dev to the new dev-ka cluster, I noticed that two hosts were missing.

Inspecting the logs of the two host probes show evidence of publishing to frontend.dev-ka.weave.works:80 not starting at all (there are no lines like `Publish loop for [DevKAIP]:80 starting` are present in the logs)

logs: https://gist.github.com/2opremio/5d9335c84bf9f2929acd54df1cc173af

goroutine dump: https://gist.github.com/2opremio/5d9335c84bf9f2929acd54df1cc173af#file-logs-txt-L2129

2opremio:
Fixes #1577

### DIFF
--- a/probe/appclient/resolver.go
+++ b/probe/appclient/resolver.go
@@ -49,10 +49,11 @@ type Resolver interface {
 }
 
 type staticResolver struct {
-	setters []setter
-	targets []target
-	quit    chan struct{}
-	lookup  LookupIP
+	setters           []setter
+	targets           []target
+	failedResolutions map[string]struct{}
+	quit              chan struct{}
+	lookup            LookupIP
 }
 
 // LookupIP type is used for looking up IPs.
@@ -67,10 +68,11 @@ func (t target) String() string { return net.JoinHostPort(t.host, t.port) }
 // resolve to multiple IPs.  It uses the supplied DNS server name.
 func NewResolver(targets []string, lookup LookupIP, setters ...setter) Resolver {
 	r := staticResolver{
-		targets: prepare(targets),
-		setters: setters,
-		quit:    make(chan struct{}),
-		lookup:  lookup,
+		targets:           prepare(targets),
+		setters:           setters,
+		failedResolutions: map[string]struct{}{},
+		quit:              make(chan struct{}),
+		lookup:            lookup,
 	}
 	go r.loop()
 	return r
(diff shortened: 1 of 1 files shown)


tomwilkie:
By default we try and resolve scope.weave.local, which will fail if people aren't running weave (most aren't), so this will pollute the logs, no?

2opremio:
> By default we try and resolve scope.weave.local, which will fail if people aren't running weave (most aren't), so this will pollute the logs, no?

True, it's similar to https://github.com/weaveworks/scope/issues/1599#issuecomment-226777072 but I don't think the solution is to remove all errors, including legitimate ones.

How about only including the `scope.weave.local` target once weave is found?

tomwilkie:
> How about only including the scope.weave.local target once weave is found?

I don't like the hardcoding - how about we only report resolution errors once (or n times) per target and silence them after?  And similarly, when the start succeeding again, we log something.

2opremio:
> I don't like the hardcoding

It's not hardcoding, I wrote `scope.weave.local` to continue the conversation but we will use whatever is provided as `--probe.weave.hostname`

It makes sense, if weave is not running, why are we even targeting that host?

> how about we only report resolution errors once (or n times) per target and silence them after? And similarly, when the start succeeding again, we log something.

That's an option too, but I prefer the former because it won't display irrelevant errors

tomwilkie:
> It makes sense, if weave is not running, why are we even targeting that host?

We don't know if we are running weave or not.  We assume we are, and try to use it.

I think you're suggesting we only log errors after we have successfully resolved a host?  But you are also arguing we should show resolution errors?  Or would you just do this as a special case for resolution of scope's weavedns hostname?

I think it too special cased, and a more general rule (like only logging the first error) would be easier to understand.

2opremio:
> We don't know if we are running weave or not. We assume we are, and try to use it.
>
> I think you're suggesting we only log errors after we have successfully resolved a host? But you are also arguing we should show resolution errors? Or would you just do this as a special case for resolution of scope's weavedns hostname?

What I am suggesting is we only register the weave target in the probes after we have proof of weave working (e.g. after `weave ps` succeeds or the `weave status` request suceeds ...)


> I think it too special cased, and a more general rule (like only logging the first error) would be easier to understand.

We can do both. Only log the first error and only target WeaveDNS target once we know weave is successfully running.

tomwilkie:
LGTM; I see we only get/set from the single goroutine - does it matter that we create the map on a different one?

2opremio:
> does it matter that we create the map on a different one?

Not AFAIK

PR #1608, opened 2016-06-23
Add backoff to the consul client

tomwilkie:
Fixes https://github.com/weaveworks/service/issues/423

### DIFF
--- a/app/multitenant/consul_client.go
+++ b/app/multitenant/consul_client.go
@@ -133,14 +133,20 @@ func (c *consulClient) CAS(key string, out interface{}, f CASCallback) error {
 }
 
 func (c *consulClient) WatchPrefix(prefix string, out interface{}, done chan struct{}, f func(string, interface{}) bool) {
-	index := uint64(0)
+	const (
+		initialBackoff = 1 * time.Second
+		maxBackoff     = 1 * time.Minute
+	)
+	var (
+		backoff = initialBackoff / 2
+		index   = uint64(0)
+	)
 	for {
 		select {
 		case <-done:
 			return
 		default:
 		}
-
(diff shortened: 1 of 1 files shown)


2opremio:
Wouldn't it be cleaner to reuse `common/backoff` ?

tomwilkie:
`common/backoff` is for periodic tasks, this one is continuous (ie not delay between iterations).  Also `common/backoff` logging behaviour is not appropriate here.

I considered changing common/backoff to be more general purpose, but I didn't think it was the right call.  I could be persuaded otherwise, but it will make common/backoff more complicated.

app/multitenant/consul_client.go
```diff
@@ -133,21 +133,40 @@ func (c *consulClient) CAS(key string, out interface{}, f CASCallback) error {
...
+			}
+			select {
+			case <-done:
+				return
+			case <-time.After(backoff):
+				continue
+			}
+		}
+		backoff = initialBackoff
+		if index == meta.LastIndex {
```
2opremio: Duplicated if

app/multitenant/consul_client.go
```diff
@@ -133,21 +133,40 @@ func (c *consulClient) CAS(key string, out interface{}, f CASCallback) error {
...
+		index   = uint64(0)
+	)
 	for {
 		select {
 		case <-done:
 			return
 		default:
 		}
-
 		kvps, meta, err := c.kv.List(prefix, &consul.QueryOptions{
```
2opremio: I was looking for a way to cancel the request directly but there doesn't seem to be a way, meh

2opremio:
> I considered changing common/backoff to be more general purpose, but I didn't think it was the right call.

Fair enough.

2opremio:
LGTM (comment-aside)

PR #1610, opened 2016-06-24
Make it easier to disable weave integrations

### LINKED ISSUE #1008: Alarming errors on weaveDNS in deployments without weave
Now Scope periodically checks for weave being started as a result of #867 .

However, this results in somehow alarming errors in deployments without Weave Net:

```
$ docker logs weavescope |& grep weaveDNS
<app> ERRO: 2016/02/23 09:24:29.108373 Error updating weaveDNS, backing off 20s: Error running weave expose: exit status 1
<app> ERRO: 2016/02/23 09:24:49.424425 Error updating weaveDNS, backing off 40s: Error running weave expose: exit status 1
<app> ERRO: 2016/02/23 09:25:29.732988 Error updating weaveDNS, backing off 1m0s: Error running weave expose: exit status 1
``` 

I understand those errors are legitimate and that we want to make those checks by default.

You can disable this by doing `--probe.weave.router.addr ""`  but that's not very intuitive (and inconvenient). I would suggest either/or:

* Lowering the alarming tone of those errors.
* Adding an extra more-intuitive flag: `--disable.weave` ?

### LINKED ISSUE #1078: Scope runs weave ps a little aggressively...
Every 10s backing off to 1min...

Also `weave expose` invocation should be done with --local.

### LINKED ISSUE #1599: failed weave interactions are logged as errors, even though it's normal operation

tomwilkie:
Also:
- Treat `weave ps` exit code like any other error (so we silence it if its failing all the time) (fixes #1599)
- Don't run weave ps so aggressively (fixes #1078)
- Log backoff-able errors at WARN (fixes #1008)

### DIFF
--- a/prog/main.go
+++ b/prog/main.go
@@ -81,6 +81,7 @@ type probeFlags struct {
 	kubernetesAPI      string
 	kubernetesInterval time.Duration
 
+	weaveEnabled  bool
 	weaveAddr     string
 	weaveHostname string
 }
@@ -92,6 +93,7 @@ type appFlags struct {
 	logPrefix string
 	logHTTP   bool
 
+	weaveEnabled   bool
 	weaveAddr      string
 	weaveHostname  string
 	containerName  string
@@ -116,6 +118,7 @@ func main() {
 		flags         = flags{}
 		mode          string
 		debug         bool
+		weaveEnabled  bool
 		weaveHostname string
 		dryRun        bool
 	)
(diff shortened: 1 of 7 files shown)


2opremio:
From #1078 

> Also weave expose invocation should be done with --local.

I can't see this being implemented

tomwilkie:
Good catch!

PTAL

2opremio:
LGTM

prog/main.go
```diff
@@ -115,12 +117,14 @@ func main() {
...
 		mode          string
 		debug         bool
+		weaveEnabled  bool
 		weaveHostname string
 	)
 
 	// Flags that apply to both probe and app
 	flag.StringVar(&mode, "mode", "help", "For internal use.")
 	flag.BoolVar(&debug, "debug", false, "Force debug logging.")
+	flag.BoolVar(&weaveEnabled, "weave", true, "Enable Weave Net integrations.")
```
rade: Boolean flags with a default value of true are hard for users to grok, and don't allow the "--flag" shorthand.

So I suggest flipping this to "noweave" (or "no-weave" or whatever conventions you follow).

tomwilkie: Yuk.

Given that the majority of Scope users are not also using weave, perhaps if should be false by default?

tomwilkie: Okay decide we don't want to turn it off by default, and @peterbourgon also doesn't like -noweave, so we'll leave as is.

PR #1612, opened 2016-06-24
Fix nil pointer error when memcache not enabled

jml:
When no memcache options were provided, we would get stack traces indicating that we were calling `GetMulti` on a nil memcache client.

This PR fixes that problem.

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1612)


tomwilkie:
Ah I see.  Yes there is an implicit cast here.

http://stackoverflow.com/questions/29138591/hiding-nil-values-understanding-why-golang-fails-here

LGTM

PR #1616, opened 2016-06-30
Refactor caching layers in dynamo collector

jml:
I wanted to better understand what was going on, so I refactored it.

Highlights:

* S3 client code moved from `dynamo_collector` into a new file, and given same interface as `MemcacheClient`
* `NewDynamoDBCollector` now takes both a memcache client and an s3 client, rather than being responsible for constructing both
* `FetchReports` returns a map of fetched reports, making it easier to write back to the in-process cache

Lowlights:

* We write back to the in-process cache after retrieving from the in-process cache
* memcache constants now in `prog/app.go`
* despite branch name, doesn't actually instrument

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1616)

### DIFF
--- a/app/multitenant/dynamo_collector.go
+++ b/app/multitenant/dynamo_collector.go
@@ -13,7 +13,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
-	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/bluele/gcache"
 	"github.com/nats-io/nats"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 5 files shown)


app/multitenant/dynamo_collector.go
```diff
@@ -102,18 +91,17 @@ type DynamoDBCollector interface {
 
 // ReportStore is a thing that we can get reports from.
 type ReportStore interface {
-	FetchReports([]string) ([]report.Report, []string, error)
+	FetchReports([]string) (map[string]report.Report, []string, error)
 }
 
 type dynamoDBCollector struct {
```
tomwilkie: Can we start calling this the awsCollector?

app/multitenant/dynamo_collector.go
```diff
@@ -138,9 +126,10 @@ type watchKey struct {
 // https://github.com/aws/aws-sdk-go/wiki/common-examples
 func NewDynamoDBCollector(
 	userIDer UserIDer,
-	dynamoDBConfig, s3Config *aws.Config,
-	tableName, bucketName, natsHost, memcachedHost string,
-	memcachedTimeout time.Duration, memcachedService string,
+	dynamoDBConfig *aws.Config, tableName string,
+	s3Store *S3Store,
+	natsHost string,
+	memcacheClient *MemcacheClient,
```
tomwilkie: We discussed having a config struct in the previous PR.  Now seems like a good opportunity.

jml: Done.

app/multitenant/dynamo_collector.go
jml: Absolutely. Renamed file, type, and methods.

jml:
PTAL.

tomwilkie:
LGTM!

PR #1627, opened 2016-07-01
Allow disabling controls in probes

### LINKED ISSUE #1605: view-only mode
[Some users](https://weaveworks.slack.com/archives/weave-users/p1466510817000440) want to share their scope with others but don't allow them to modify anything.

2opremio:
Fixes #1605

### DIFF
--- a/probe/appclient/app_client.go
+++ b/probe/appclient/app_client.go
@@ -57,11 +57,12 @@ type appClient struct {
 	readers     chan io.Reader
 
 	// For controls
-	control xfer.ControlHandler
+	noControls bool
+	control    xfer.ControlHandler
 }
 
 // NewAppClient makes a new appClient.
-func NewAppClient(pc ProbeConfig, hostname, target string, control xfer.ControlHandler) (AppClient, error) {
+func NewAppClient(pc ProbeConfig, hostname, target string, control xfer.ControlHandler, noControls bool) (AppClient, error) {
 	httpTransport, err := pc.getHTTPTransport(hostname)
 	if err != nil {
 		return nil, err
@@ -82,9 +83,10 @@ func NewAppClient(pc ProbeConfig, hostname, target string, control xfer.ControlH
 		wsDialer: websocket.Dialer{
 			TLSClientConfig: httpTransport.TLSClientConfig,
 		},
-		conns:   map[string]xfer.Websocket{},
-		readers: make(chan io.Reader, 2),
-		control: control,
+		conns:      map[string]xfer.Websocket{},
+		readers:    make(chan io.Reader, 2),
+		noControls: noControls,
+		control:    control,
 	}, nil
 }
 
@@ -231,6 +233,9 @@ func (c *appClient) controlConnection() (bool, error) {
 }
 
 func (c *appClient) ControlConnection() {
+	if c.noControls {
+		return
+	}
 	go func() {
 		log.Infof("Control connection to %s starting", c.target)
 		defer log.Infof("Control connection to %s exiting", c.target)
--- a/probe/appclient/app_client.go
+++ b/probe/appclient/app_client.go
@@ -57,12 +57,11 @@ type appClient struct {
 	readers     chan io.Reader
 
 	// For controls
-	noControls bool
-	control    xfer.ControlHandler
+	control xfer.ControlHandler
 }
 
 // NewAppClient makes a new appClient.
-func NewAppClient(pc ProbeConfig, hostname, target string, control xfer.ControlHandler, noControls bool) (AppClient, error) {
+func NewAppClient(pc ProbeConfig, hostname, target string, control xfer.ControlHandler) (AppClient, error) {
 	httpTransport, err := pc.getHTTPTransport(hostname)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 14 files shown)


probe/appclient/app_client.go
```diff
@@ -231,6 +233,9 @@ func (c *appClient) controlConnection() (bool, error) {
 }
 
 func (c *appClient) ControlConnection() {
+	if c.noControls {
+		return
+	}
```
tomwilkie: Would it be easier to not plumb this flag through to the appClient, but rather the multiclient, and then not call ControlConnection()?

2opremio: It's not simpler, but maybe cleaner, let me check.

probe/appclient/multi_client.go
```diff
@@ -100,9 +102,7 @@ func (c *multiClient) Set(hostname string, endpoints []string) {
 	hostIDs := report.MakeIDList()
 	for tuple := range clients {
 		hostIDs = hostIDs.Add(tuple.ID)
-
-		_, ok := c.clients[tuple.ID]
-		if !ok {
```
tomwilkie: Is this safe?  Where else is c.clients used?

probe/appclient/report_publisher.go
```diff
@@ -8,18 +8,37 @@ import (
...
-func NewReportPublisher(publisher Publisher) *ReportPublisher {
+func NewReportPublisher(publisher Publisher, noControls bool) *ReportPublisher {
 	return &ReportPublisher{
-		publisher: publisher,
+		publisher:  publisher,
+		noControls: noControls,
 	}
 }
 
+func removeControls(r report.Report) report.Report {
```
tomwilkie: My initial thought was to put this logic in the individual probe modules that populate the controls (ie so nodes don't get controls to begin with), but this is much cleaner!

Is there a way to do this without maintaining the list of topologies here?  We have some helper functions for iterating over topologies in report.go.

2opremio: Yep, I thought about that too. It would be easy to forget to add new topologies here.

The exiting listing function copies the topologies, so it won't let us modify them in the report.

I will add a MapTopologies function.

probe/appclient/multi_client.go
2opremio: It's only used for Pipes, which shouldn't be triggered if there are no controls.

I could do:

```go
                if _, ok := c.clients[tuple.ID]; !ok {
                        c.clients[tuple.ID] = tuple.AppClient
                        if !c.noControls {
                                 tuple.AppClient.ControlConnection()
                        }
                }
```

But I don't think it's the right thing to do, since, if called by error, `PipeConnection()` will try to contact the app.

2opremio: Actually, I am wrong, it's also used by [`Publish()`](https://github.com/weaveworks/scope/blob/74e508671ce26156118f9b4f60aaec72c7de2303/probe/appclient/multi_client.go#L172). Fixing ...

rade:
excuse my ignorance, but does 'no controls' mean 'no logs'? Or just 'no terminals', 'no start/stop', ...?

tomwilkie:
No logs, terminals, start / stop etc.

rade:
Hmm. 'no logs' is a bit annoying.

tomwilkie:
To do this would require us to leave all the pipe and control logic
running, and would make the surface area to 'secure' much larger and make
it much easier to mess up.  This way has a high degree of certainty of
working and not rotting.

If we did allow some controls / pipes, we'd need to specify and enforce
which ones are read only, and also build a concept of a one way pipe. Not
impossible, but worth the extra few days?  can always be done in the
future.

rade:
Agreed.

rade:
PS: would be really good to get this merged asap, so I can drop it from my priority list :)

2opremio:
> PS: would be really good to get this merged asap, so I can drop it from my priority list :)

It's just pending a LGTM :)

rade:
> It's just pending a LGTM :)

I know. I am not qualified to provide that.

prog/main.go
```diff
@@ -145,6 +146,7 @@ func main() {
 	flag.DurationVar(&flags.probe.publishInterval, "probe.publish.interval", 3*time.Second, "publish (output) interval")
 	flag.DurationVar(&flags.probe.spyInterval, "probe.spy.interval", time.Second, "spy (scan) interval")
 	flag.StringVar(&flags.probe.pluginsRoot, "probe.plugins.root", "/var/run/scope/plugins", "Root directory to search for plugins")
+	flag.BoolVar(&flags.probe.noControls, "probe.no-controls", false, "Disable controls (read-only mode)")
```
tomwilkie: Is this the best name for the flag? I suspect --readonly would be more intuitive.  Also, this disables pipes (indirectly) as well.

tomwilkie:
LGTM with one minor question about flag naming, but I don't particularly mind what its called...

prog/main.go
2opremio: That's how I originally named the flag, but decided it wasn't accurate enough since we also disable "read-only" controls like logs.

2opremio: Maybe I should remove the clarification _(read-only mode)_

PR #1628, opened 2016-07-01
Fixes to memcache support

jml:
* Log an error rather than aborting when memcache doesn't resolve
* Initialize map correctly

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1628)

### DIFF
--- a/app/multitenant/memcache_client.go
+++ b/app/multitenant/memcache_client.go
@@ -169,7 +169,7 @@ func (c *MemcacheClient) FetchReports(keys []string) (map[string]report.Report, 
 		}(key)
 	}
 
-	var reports map[string]report.Report
+	reports := make(map[string]report.Report)
 	for i := 0; i < len(keys)-len(missing); i++ {
 		r := <-ch
 		if r.report == nil {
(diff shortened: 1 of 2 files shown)


prog/app.go
```diff
@@ -114,15 +114,7 @@ func collectorFactory(userIDer multitenant.UserIDer, collectorURL, s3URL, natsHo
...
-				// TODO(jml): Ideally, we wouldn't abort here, we would instead
-				// log errors when we try to use the memcache & fail to do so, as
-				// aborting here introduces ordering dependencies into our
-				// deployment.
-				//
-				// Note: this error only happens when either the memcachedHost
-				// or any of the SRV records that it points to fail to
-				// resolve.
-				return nil, err
+				log.Errorf("Error resolving memcache servers for '%v': %v", memcachedHostname, err)
```
tomwilkie: Can you move this into NewMemcacheClient and have it not return an error?

app/multitenant/memcache_client.go
```diff
@@ -169,7 +169,7 @@ func (c *MemcacheClient) FetchReports(keys []string) (map[string]report.Report,
 		}(key)
 	}
 
-	var reports map[string]report.Report
+	reports := make(map[string]report.Report)
```
tomwilkie: the "weaveworks style" is `map[string]report.Report{}`

prog/app.go
jml: Done.

app/multitenant/memcache_client.go
jml: Amended.

jml:
Have just confirmed that this works for me on local.

PR #1640, opened 2016-07-04
Bump memcache expiration

jml:
We were getting memcache misses and they are probably due to entries expiring between fetching the keys from dynamo and performing the query to memcache.

This PR adds a flag for controlling expiration length and changes the default to be twice the default window.

In addition, it adds a `MemcacheConfig` struct, and passes the `app.window` flag through to the aws_collector, thus removing a lot of duplication of the magic number 15. I've also left some instrumentation in.

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1640)

### DIFF
--- a/app/multitenant/aws_collector.go
+++ b/app/multitenant/aws_collector.go
@@ -23,12 +23,10 @@ import (
 )
 
 const (
-	hourField             = "hour"
-	tsField               = "ts"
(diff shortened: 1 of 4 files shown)


tomwilkie:
LGTM

PR #1641, opened 2016-07-05
Force some known column widths to prevent truncation of others

### LINKED ISSUE #1492: Remote IPs get truncated in the internet view
It's not terribly useful, as you can't select them. The column should probably be wide enough for an IP.
screen shot 2016-05-12 at 16 02 17

### LINKED ISSUE #1532: Connections-table columns should be able to fit an ipv4 address
Right now they are truncated, which is really frustrating for users.

I saw this myself a couple of weeks ago but wasn't sure it was a problem, now we have confirmation from a user.

Context: https://weaveworks.slack.com/archives/weave-users/p1463732653000585

foot:
Port/count won't exceed a certain width so slim those down to free up
space for the IPs.

This can be extended to CPU/memory in the future but that needs a little
more refactoring in the number renderers

Fixes #1532 
Fixes #1492


paulbellamy:
LGTM

PR #1646, opened 2016-07-05
Promote fixprobe and delete rest of experimental

jml:
We don't really need experimental, there are better ways of experimenting with code, and sometimes they fail the build, which gets in the way of making changes to scope. 

This PR deletes the experimental directory, but promotes `fixprobe`, which I'm told we use.

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1646)


tomwilkie:
Can we keep the example subdirectory?  It doesn't require building, and is very useful.

Otherwise I'm happy with this, but we should check with others.

tomwilkie:
Am still happy, but I think you should also get @paulbellamy and @2opremio's LGTM.

paulbellamy:
Maybe `s/extras/tools/` ?

paulbellamy:
Not fussy about that though, LGTM.

jml:
Really no preference at all. Will rename after @2opremio has had a chance to look, in case he has a preference.

2opremio:
LGTM

jml:
There's already a `tools` directory, a git subtree that has our core build tools. Merging as-is. 

Thanks for the reviews!

PR #1647, opened 2016-07-06
Support single/double quote key on us-intl keyboards in terminal

### LINKED ISSUE #1403: Terminals don't support quote characters from intl keyboard layouts
Users with an international keyboard layout cannot type any quote characters on the terminals (e.g. `"` `'` ``  `  `` etc ...). This makes it impossible to type any complex shell commands.

screen shot 2016-04-28 at 20 10 59

See http://www.starr.net/is/type/intlchart.html

This was reproduced on Chrome under OSX

foot:
Doesn't support the accenting capabilities of this key but gives you
single/double quotes back which is a long way. If you need accented keys
for filenames etc you'll have to use other kb tricks.

Fixes #1403

### DIFF
--- a/client/app/scripts/vendor/term.js
+++ b/client/app/scripts/vendor/term.js
@@ -2756,6 +2756,13 @@ Terminal.prototype.keyDown = function(ev) {
     case 123:
       key = '\x1b[24~';
       break;
+    // Special case for US-intl '/"
+    case 222:
+      key = '\'';
+      if (ev.shiftKey) {
+        key = '"';
+      }
+      break;
     default:
       // a-z and space
       if (ev.ctrlKey) {
--- a/client/app/scripts/vendor/term.js
+++ b/client/app/scripts/vendor/term.js
@@ -2756,14 +2756,22 @@ Terminal.prototype.keyDown = function(ev) {
     case 123:
       key = '\x1b[24~';
       break;
-    // Special case for US-intl '/"
-    case 222:
-      key = '\'';
-      if (ev.shiftKey) {
-        key = '"';
-      }
-      break;
     default:
+      // Special case for US-intl
+      // Firefox leaves ev.key empty, Chrome sets it to "Dead"
+      if (!ev.key || ev.key === 'Dead') {
+        if (ev.keyCode === 54) {
+          key = ev.shiftKey ? '^' : '6';
+          break;
+        } else if (ev.keyCode === 192) {
+          key = ev.shiftKey ? '~' : '`';
+          break;
+        } else if (ev.keyCode === 222) {
+          key = ev.shiftKey ? '"' : "'";
+          break;
+        }
+      }
+
       // a-z and space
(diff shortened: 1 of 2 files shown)


client/app/scripts/vendor/term.js
```diff
@@ -2756,6 +2756,13 @@ Terminal.prototype.keyDown = function(ev) {
     case 123:
       key = '\x1b[24~';
       break;
+    // Special case for US-intl '/"
+    case 222:
+      key = '\'';
+      if (ev.shiftKey) {
+        key = '"';
+      }
+      break;
```
2opremio: I don't know how you are handling vendoring but, How do you make sure changes made directly are not lost when updating the library?

foot: We're lucky, the library is unmaintained! :D

We'd probably do something like `patch -p1 < git log -p term.js`.

We could look into forking the lib on github which I guess is the proper
way to do it..

2opremio:
Some comments:
* This improves #1403 (thanks!) but doesn't really fix it (I would expect my keyboard map to work in the seamlessly on terminals) so we shouldn't close it.
* Backticks `` ` `` don't seem to work. For some reason it causes Evernote to open instead (backticks work just fine at places like the search bar). Even after disabling the Evernote shortcuts, typing `` ` `` doesn't show any changes in the terminal window.

screen shot 2016-07-07 at 3 35 28 pm

screen shot 2016-07-07 at 3 35 05 pm

foot:
Yes, it would be nice to have this working seamlessly like the search bar etc! There's a bit of a diff in how they work intrinsically though. term.js, the lib we use has a lot of complicated logic to handle control codes / terminal stuff (low level JS event handling), vs. just letting the `<input ...` do its thing (search bar). Unfortunately term.js doesn't support many intl kb's right now though there's work being done in that area! https://github.com/sourcelair/xterm.js/pulls

I tested out more of the dead keys and you're right quite a few of them are broken e.g. backtick and `^` too.

Going forward.
 1. Support parts of us-intl if it appears to be a particularly popular layout.
 2. (maybe) Detect us-intl and popup a notifaction suggesting they change their layout if they're having issues.
 3. Keep up to date w/ termjs improvements.

2opremio:
> I tested out more of the dead keys and you're right quite a few of them are broken e.g. backtick and ^ too.

Can you include the ones you think are missing? Backtick is more or less essential for shell-scripts. After that we can merge this and then revisit this once/if we move to xterm.js

foot:
5 basic dead keys[0]
- <kbd>'</kbd>
- <kbd>"</kbd>
- <kbd>`</kbd>
- <kbd>~</kbd>
- <kbd>^</kbd>

[0] http://dry.sailingissues.com/us-international-keyboard-layout.html

2opremio:
That would be enough to get me going.

2opremio:
Thanks @foot !

PR #1650, opened 2016-07-06
Fix ensure plugins dir exists (close #1649)

### LINKED ISSUE #1649: Scope fails to launch on latest Docker for Mac (beta18)
```
> bash ./scope launch
68d432bf5c07d17e7b42204c85bddc1a9d0ac30869324329645d094357019d0d
docker: Error response from daemon: Mounts denied: 
The path /var/run/scope/plugins
is not shared from OS X and does not belong to the system.
You can configure shared paths from Docker -> Preferences...
..
```

errordeveloper:


scope
```diff
@@ -181,6 +187,7 @@ EOF
         if ! check_docker_for_mac || check_probe_only ; then
             check_not_running $SCOPE_CONTAINER_NAME $SCOPE_IMAGE_NAME
             docker rm -f $SCOPE_CONTAINER_NAME >/dev/null 2>&1 || true
+            create_plugins_dir
```
errordeveloper: It's a good idea to do this anyway, not just for Docker for Mac, as Docker has deprecated the behaviour where it creates new directory if path didn't exist.

errordeveloper:
I think it'd great to make a patch release. @lukemarsden and I would like to publish a guide using socks shop and Scope on Docker for Mac.

tomwilkie:
Why is this needed for a patch release? Is it fixing some important
blocking issue?

tomwilkie:
Oh I see #1649. Sorry.

scope
2opremio: When did this change?

2opremio:
> I think it'd great to make a patch release. @lukemarsden and I would like to publish a guide using socks shop and Scope on Docker for Mac.

I think we are pretty close to finish the 0.17 milestone

scope
errordeveloper: Couldn't find an issue yesterday, but found this just now https://github.com/docker/docker/commit/249f45bcfe1adf4b3550306951f2768b17802217. Also, we are already handing this in Weave Net (https://github.com/weaveworks/weave/issues/1653).

errordeveloper: However, it turns out that this had been _undepricated_ more recently (https://github.com/docker/docker/issues/21652#issuecomment-203654182). I think we should still keep it here anyway, at least for consistency with the Docker for Mac code path.

2opremio:
@errordeveloper I would only explicitly create the directory in the Docker4Mac case and add a comment about why (e.g. incompatible with standard docker in which the directory is created if it doesn't exist)

errordeveloper:
> I would only explicitly create the directory in the Docker4Mac case

Perhaps, now that we learned about the behaviour being undepricated...

> incompatible with standard docker in which the directory is created if it doesn't exist

That's actually not the case, if you take a look at #1649, the reason is a bit different.

2opremio:
> That's actually not the case, if you take a look at #1649, the reason is a bit different.

Sure, I was just providing a possible comment. My point was explaining the cause and only applying the fix for Docker4mac.

errordeveloper:
@2opremio I've addressed your concerns and rebased.

2opremio:
Thanks @errordeveloper ! 

LGTM

PR #1651, opened 2016-07-06
Quote jsonpath so shell doesn't try to expand

### LINKED ISSUE #1526: Cannot port-forward app from kubernetes with command in documentation
I'm running weave scope in kubernetes and followed the [documentation steps](https://www.weave.works/docs/scope/latest/installing/#k8s) to get it running.

When trying to port forward weave scope to my localhost host using the command

```
kubectl port-forward $(kubectl get pod --selector=weavescope-component=weavescope-app -o jsonpath={.items..metadata.name}) 4040
```

It doesn't work because kubectl complains 'error: name cannot be provided when a selector is specified'
With debug output `--v=9` it looks like the problem is in helpers.go line 107. I didn't dig into the code but wanted to raise the issue with documentation.

If I remove the output formatting `-o jsonpath=...` the command works and shows me the proper name. An alternative command could also be
```
kubectl port-forward $(kubectl get pod --selector=weavescope-component=weavescope-app | grep -v '^NAME' | awk '{print $1}' | head -n 1) 4040
```
obviously not as elegant but uses kubectl pod NAME instead of parsing the json.

I'm using kubernetes 1.2.3 with kubectl 1.2.3.

rothgar:
This fixes #1526 I realized it was a problem with zsh trying to expand `items..metadata`. Single quotes fixes it for zsh and also works in bash

example in zsh
```
$ echo {items..metadata}
. a d e i m s t
$ echo '{items..metadata}'
{items..metadata}
```

### DIFF
--- a/site/installing.md
+++ b/site/installing.md
@@ -212,7 +212,7 @@ To download and read the Scope manifest run:
 
 **Open Scope in Your Browser**
 
-    kubectl port-forward $(kubectl get pod --selector=weavescope-component=weavescope-app -o jsonpath={.items..metadata.name}) 4040
+    kubectl port-forward $(kubectl get pod --selector=weavescope-component=weavescope-app -o jsonpath='{.items..metadata.name}') 4040
 
 Open http://localhost:4040 in your browser. This allows you to access the Scope UI securely, without opening it to the Internet.
 


tomwilkie:
Build failure is due to this being a PR from a non-weaveworks branch.  We should fix that, but its unrelated.

LGTM

Issue #1652, opened 2016-07-06
CI fails for PR contributions from other than weaveworks/scope...


tomwilkie:
See: https://circleci.com/gh/weaveworks/scope/4769

Its due to the image tagging.

bboreham:
Fixed by #1655.

(although we currently have "Run forked pull requests" turned off)

PR #1655, opened 2016-07-07
circle.yml: deploy master with non-upstream hub accounts

alban:
Since b4abe582f4ffe20a93c9d24f4ba699d60e2d8fe2, images are tagged on Docker Hub. But this was not working for non-upstream Docker Hub accounts. This patch should fix that.

[Symptoms](https://circleci.com/gh/kinvolk/scope/23):

> The push refers to a repository [docker.io/kinvolk/scope] (len: 1)
> error getting tags for kinvolk/scope: Tag does not exist for
> master-b89e9ec

Also, ensure that the Docker tag name does not include any '/'.

[Symptoms](https://circleci.com/gh/kinvolk/scope/24):

> docker tag weaveworks/scope weaveworks/scope:alban/hub1-ef739d9
> repository name component must match "[a-z0-9]+(?:[._-][a-z0-9]+)*"


circle.yml
```diff
@@ -66,7 +66,10 @@ deployment:
       - >
         test -z "${DOCKER_USER}" || (
           docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS &&
-          (test "${DOCKER_ORGANIZATION:-$DOCKER_USER}" == "weaveworks" || docker tag weaveworks/scope:latest ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:latest) &&
+          (test "${DOCKER_ORGANIZATION:-$DOCKER_USER}" == "weaveworks" || (
+            docker tag weaveworks/scope:latest ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:latest &&
+            docker tag weaveworks/scope:$(./image-tag) ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:$(./image-tag)
+          )) &&
           docker push ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope
```
tomwilkie: I'm curios how this works without an `&&` on the end?

tomwilkie:
Nice one @alban, this should fix #1652 too!

tomwilkie:
The failure is due to our test scheduler getting your circle runs confused with our cicrle runs.  Looks like the owner of the repo should be included in the test schedule name.

circle.yml
alban: I guess it will run the next line unconditionally, even if the previous line fails.

I am adding a `&&`, this will look more correct!

alban:
@tomwilkie where is that test schedule name defined?

tomwilkie:
https://github.com/weaveworks/scope/blob/6d527821e71a8b8550ef3c4a740985df0fe549d5/experimental/in_parallel.sh#L21 the second argument to this is the schedule name.  It should be update to include the full repo path or something.

Note experimental has gone away now.  But the same problem probably exists for other uses of the scheduler (ie unit tests).  They are run by this script https://github.com/weaveworks/build-tools/blob/master/test.

alban:
@tomwilkie thanks! This time, CircleCI passed fine.

/cc @krnowak @alepuccetti

circle.yml
```diff
@@ -67,7 +67,10 @@ deployment:
...
         test -z "${DOCKER_USER}" || (
           docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS &&
-          (test "${DOCKER_ORGANIZATION:-$DOCKER_USER}" == "weaveworks" || docker tag weaveworks/scope:latest ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:latest) &&
+          (test "${DOCKER_ORGANIZATION:-$DOCKER_USER}" == "weaveworks" || (
+            docker tag weaveworks/scope:latest ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:latest &&
+            docker tag weaveworks/scope:$(./image-tag) ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:$(./image-tag)
+          )) &&
           docker push ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope &&
           docker push ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:$(./image-tag) &&
           wcloud deploy ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:$(./image-tag)
```
tomwilkie: It occurs to me this line will fail horribly for non-weaveworks builds.  let me think for a min.

tomwilkie: Okay I think it might be worth adding a owner section for weaveworks (see https://circleci.com/docs/configuration/#deployment) and another for everyone else, which does the org move.

alban: ok, I am trying this. Another way could be to guard the wcloud command with `test -z "$SERVICE_TOKEN" || ...`.

alban:
Branch updated. CircleCI passed [here](https://circleci.com/gh/kinvolk/scope/43) but of course it didn't run the deployment for the master branch. So I tried on another repository on the master branch and it also passed [here](https://circleci.com/gh/alban/scope/57).

tomwilkie:
LGTM!

PR #1659, opened 2016-07-11
Adding a static report file mode.

paulbellamy:
Adds a "file" collector to service a static report from file, then adds a demo.json report to the dockerfile.

Can run it with: `scope launch --no-probe --weave=false --app.collector=file:///demo.json`


paulbellamy:
I'm also happy to leave out the demo.json file, as it adds 2.8M to the image.

app/collector.go
```diff
@@ -131,3 +136,56 @@ func (c *collector) clean() {
...
+type StaticCollector report.Report
+
+// Report returns a merged report over all added reports. It implements
+// Reporter.
+func (c StaticCollector) Report(context.Context) (report.Report, error) { return report.Report(c), nil }
+
+// Add adds a report to the collector's internal state. It implements Adder.
+func (c StaticCollector) Add(context.Context, report.Report) error { return nil }
+
+// WaitOn lets other conponents wait on a new report being received. It
```
jml: Nit: "components"

app/collector.go
```diff
@@ -131,3 +136,56 @@ func (c *collector) clean() {
...
+// implements Reporter.
+func (c StaticCollector) WaitOn(context.Context, chan struct{}) {}
+
+// UnWait lets other conponents stop waiting on a new report being received. It
```
jml: Here too.

app/collector.go
```diff
@@ -131,3 +136,56 @@ func (c *collector) clean() {
...
+// implements Reporter.
+func (c StaticCollector) UnWait(context.Context, chan struct{}) {}
+
+// NewFileCollector reads and json parses the given path, returning a collector
```
jml: Delete "json", since it does more that that. Alternatively, update the docstring to say that it does json or msgpack based on the filename extension.

jml:
I don't have strong feelings about whether the JSON file should stay or go.

Everything else looks good, modulo a couple of nitpick comments. Please merge after addressing them.

paulbellamy:
Thanks, discussed offline that adding the json file is terrible, but all the other options are worse.

Issue #1010, opened 2016-02-23
Reduce Garbage Collection pressure


2opremio:
We can see from #812 (specifically https://github.com/weaveworks/scope/issues/812#issuecomment-187244284) and more importantly from #854 (specifically https://github.com/weaveworks/scope/issues/854#issuecomment-187250446 ) that the current main performance bottleneck is the Garbage collector pressure:

> Note how runtime.scanobject accounts to almost ~50% of the CPU time (scanobject+mallocgc) and how ps.Cons+ps.Set generate almost 30% of the garbage.

rade:
I think I have identified a potentially quite significant improvement here...

While `LatestMap.Merge` is commutative, the order nevertheless matters - to performance. When newer entries are in the target map rather than the argument map we avoid allocation altogether.

AFAICT, both `collector` and `awsCollector` invoke the merger with a list of reports that is largely in oldest to youngest order. `SmartMerger` does some hashing, as a result of which reports will be merged in random order.

Ideally `SmartMerger` would first merge reports from the same probe in reverse timestamp order, and only then merge them together with its tree-merge.

Another optimisation: `SmartMerger` has this line:

    report.MakeReport().Merge(left.rpt).Merge(right.rpt),

but note that report merging already copies the lhs, so the above really should just be

    left.rpt.Merge(right.rpt)

Yet another optimisation: merging should be n-ary, not binary, to avoid creation and copying of intermediate data structures.

rade:
re merging reports in newest-to-oldest order... Note that it's actually the probe-side timestamps that matter here, since that is what `LatestMap.Merge` is comparing. Alas in the `awsCollector` reports are keyed on the *app* timestamp. The potential is there for successive reports from a single probe to hit different app instances and receive timestamps that effectively switches their order. Hopefully this is rare.

rade:
> Yet another, more substantial optimisation: merging should be n-ary, not binary, to avoid creation and copying of intermediate data structures.

Doing that will make `SmartMerger` superfluous. Which in turn means we can do a dumb merge in reverse report timestamp order. i.e. start with a blank report, merge in the youngest report (this will end up copying the that report), then the 2nd youngest (this will end up copying new entries and overwrite existing entries where we see a newer timestamp, which should be rare), etc.

In the optimal case this will only allocate memory equivalent to final merged report size. And we will usually get close to that optimum.

2opremio:
Another thing which could potentially make a difference: Field allocations on Node initialization.

[...]

PR #1660, opened 2016-07-11
Use a slice instead of a persistent list for temporary accumulation of lists

bboreham:
Reduce garbage created in `Merge()` and `Add()` from O(n) to O(1) by using a slice with enough capacity to hold the temporary head-portion of the list accumulated while walking the list to insert element(s).

(EDIT: except in the case where all merged data is newer, `Merge()` discards at least one of the input lists, so in that sense still creates O(n) garbage. I've reduced the internally-created garbage)

### DIFF
--- a/report/metrics.go
+++ b/report/metrics.go
@@ -109,14 +109,14 @@ func last(t1, t2 time.Time) time.Time {
 	return t2
 }
 
-// revCons appends acc to the head of curr, where acc is in reverse order.
-// acc must never be nil, curr can be.
-func revCons(acc, curr ps.List) ps.List {
+// concat returns a new list formed by adding each element of acc to curr
+// acc or curr can be nil.
+func concat(acc []interface{}, curr ps.List) ps.List {
 	if curr == nil {
-		return acc.Reverse()
+		curr = ps.NewList()
 	}
-	for !acc.IsNil() {
-		acc, curr = acc.Tail(), curr.Cons(acc.Head())
+	for i := len(acc) - 1; i >= 0; i-- {
+		curr = curr.Cons(acc[i])
 	}
 	return curr
 }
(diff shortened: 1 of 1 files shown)


tomwilkie:
Nice!  I can't see any problems with this.

2opremio:
Contributes to #1010 

Thanks @bboreham !

I have measured the impact this would have when using Scope to monitor our service, but it doesn't look very high. Here are the results of an `-alloc_objects -focus 'Metric'` heap profile of the:

* Probe: [pprof.localhost:4041.alloc_objects.alloc_space.001.pb.gz](https://github.com/weaveworks/scope/files/357307/pprof.localhost.4041.alloc_objects.alloc_space.001.pb.gz)

probe_alloc_objects_metric

* App: [pprof.localhost.alloc_objects.alloc_space.001.pb.gz](https://github.com/weaveworks/scope/files/357310/pprof.localhost.alloc_objects.alloc_space.001.pb.gz)

app_alloc_objects_metric


In what scenario did this make a considerable difference? Do you happen to have allocation profiles?

2opremio:
I've just realized that we are running Scope on the service with `--probe.processes=false` which will probably explain the difference. I will try again without this.

(No wonder I was surprised by the low CPU consumption).

2opremio:
After re-enabling the process topology and connecting the browser I get a closer behavior but not as close as I expected:

Here's the App's `-alloc_objects` heap profile (with 4 connected probes in the dev environment):

app_alloc_objects

And here's the same profile with focusing on revCons (`-focus 'revCons'`):

app_alloc_objects_revcons

So, this would reduce object allocation in ~5%.

Anyways, the change LGTM. @bboreham mentioned offline that this change was done after profiling the object allocation https://github.com/weaveworks/weavedemo , I will take a look at that next.

bboreham:
The expected reduction is about half the allocations seen through `revCons`.  We save the accumulator list, but still have to build a new list.

PR #1667, opened 2016-07-12
Fixes network bars position when a node is selected.

### LINKED ISSUE #1663: Networks bar hides the node name
This occurred to me with some zoom settings that I cannot reproduce now.

In any case, my thoughts are that the bar could appear in the top right hand corner of the details panel instead, we could also show names next to colours...

foot:
Bar position should be a function of node size!

Note: This is still an estimated position though it looks fine. Would be
nice to take into account size of labels (foreignObjects). The
transforms (for "undoing" the canvas zoom level) are a bit hectic here
and would be really nice to pull out and put in a wrapper...

Fixes #1663

### DIFF
--- a/client/app/scripts/charts/node-networks-overlay.js
+++ b/client/app/scripts/charts/node-networks-overlay.js
@@ -6,17 +6,13 @@ import { isContrastMode } from '../utils/contrast-utils';
 
 
 const barHeight = 5;
-const barMarginTop = 6;
-const labelHeight = 32;
 // Gap size between bar segments.
 const padding = 0.05;
 const rx = 1;
 const ry = rx;
 const x = d3.scale.ordinal();
 
-function NodeNetworksOverlay({labelOffsetY, size, stack, networks = makeList()}) {
-  const offset = labelOffsetY + labelHeight + barMarginTop;
-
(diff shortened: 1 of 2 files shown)


foot:
Not working for varied node sizes on the canvas (e.g. if you start w/ a very small browser window...).

Poking...

foot:
Moves bar position

screen shot 2016-07-12 at 16 11 55

foot:
Bar has moved above label as it was a bit tricky aligning it accurately in the middle of the text blob.

rade:
don't you end up with an awkward gap between node and text when *not* showing networks?

foot:
No, you get a gentle "bump" when enabling/disabling the networks overlay.

I don't think its too jarring. Animating it could improve it further.

rade:
tried the branch. The "bump" is fine. animating it would be nice but not essential. HOWEVER, on firefox the bar overlaps with the text. and on Chrome it is very close to the text, much closer than on your screen shot above...

Firefox

Chrome

foot:
Thanks for testing it. Testing out ff..

foot:
Can do with another test! :)

rade:
LGTM. Though I cannot comment on the code.

paulbellamy:
Code looks fine.

PR #1669, opened 2016-07-13
Fix Docker for Mac issue with `--service-token` (close #1649)

### LINKED ISSUE #1649: Scope fails to launch on latest Docker for Mac (beta18)

errordeveloper:


scope
```diff
@@ -195,6 +195,7 @@ EOF
         if ! check_docker_for_mac || check_probe_only ; then
             check_not_running $SCOPE_CONTAINER_NAME $SCOPE_IMAGE_NAME
             docker rm -f $SCOPE_CONTAINER_NAME >/dev/null 2>&1 || true
+            create_plugins_dir
```
errordeveloper: We could make this `check_docker_for_mac && create_plugins_dir`, but I'm not sure if we really have to...

rade: This ends up creating the plugin dir even when not on a Mac. Which I thought we tried to avoid.

rade: > We could make this `check_docker_for_mac && create_plugins_dir`

Why not put this before the 'if'? Then we don't need `create_plugins_dir` in two branches.

2opremio:
The launch code could really use a refactoring. We check for `check_docker_for_mac` at 3 different places and `check_probe_only` at 2 different places.

2opremio:
@jml Can you please take a look at the refactoring I made?

scope
```diff
@@ -192,41 +206,36 @@ EOF
...
             # - https://forums.docker.com/t/ports-in-host-network-namespace-are-not-accessible/10789
             check_not_running $SCOPE_APP_CONTAINER_NAME $SCOPE_IMAGE_NAME
             docker rm -f $SCOPE_APP_CONTAINER_NAME >/dev/null 2>&1 || true
-            CONTAINER=$($(launch_app_command "$@"))
+            CONTAINER=$($(launch_docker4mac_app_command "$@"))
             echo $CONTAINER
             app_ip=$(docker inspect -f '{{.NetworkSettings.IPAddress}}' "${CONTAINER}")
             check_not_running $SCOPE_CONTAINER_NAME $SCOPE_IMAGE_NAME
             docker rm -f $SCOPE_CONTAINER_NAME >/dev/null 2>&1 || true
             create_plugins_dir
```
errordeveloper: Is this redundant now?

scope
```diff
@@ -146,13 +146,27 @@ launch_command() {
...
 
-launch_app_command() {
+launch_docker4mac_app_command() {
     echo docker run -d --name=$SCOPE_APP_CONTAINER_NAME \
             -e CHECKPOINT_DISABLE \
             -p 0.0.0.0:4040:4040 \
             $WEAVESCOPE_DOCKER_ARGS $SCOPE_IMAGE --no-probe "$@"
 }
 
+launch_normally() {
```
jml: Please add a documentation comment: "normally" is a bit of a weasel word.

scope
2opremio: Oh, yeah! Good catch!

scope
2opremio: Would a better name help instead? `do_default_launch`?

2opremio: OK, that name is not better, but I take suggestions ...

rade: `launch`

jml: I think adding a comment describing what it's _not_ would help.

errordeveloper:
I'm looking at this now, and I see that `create_plugins_dir` and some other functions are not consistent with other function. I've spoke with @tomwilkie, and the reason `launch_command` does `echo docker run ...` and is later called from a subshell is because subshell creates new scope, and thereby prevents crazy use of global variables etc. I am wondering whether we might want to ensure consistency, once we are refactoring this...

scope
```diff
@@ -146,13 +146,27 @@ launch_command() {
...
+    check_not_running $SCOPE_CONTAINER_NAME $SCOPE_IMAGE_NAME
+    docker rm -f $SCOPE_CONTAINER_NAME >/dev/null 2>&1 || true
+    CONTAINER=$($(launch_command "$@"))
+    echo $CONTAINER    
+}
+
+print_app_endpoints() {
+    echo "Weave Scope is reachable at the following URL(s):" >&2
+    for ip in $@; do
```
jml: Quotes around `$@`, e.g. `"$@"`.

scope
2opremio: @rade suggested `launch()` let's go for that.

jml:
Thanks. Refactoring LGTM.

2opremio:
> create_plugins_dir` and some other functions are not consistent with other function

Let's not fix the world in this PR :) I only did the refactoring because it was easier than understanding the code for reviewing the PR.

errordeveloper:
Sadly, I'm getting this on Docker for Mac:
```
0 %> ./scope launch --service-token=MYTOKEN
d6fa231c17f8ce7d669d8f92015f100d7fffc16a7e42161c1b9176b68c87892e
./scope: line 213: return: can only `return' from a function or sourced script

1 %> ./scope launch
af3b8d723dab193662c60c5ea96e2e379de4342c1e5cf249ddb2781e5ea5286b
weavescope is already running.

1 %> ./scope stop

0 %> ./scope launch
c9dea490b5f8cb391168d05531e4a80ebdea08987a5858456a5fdcd92be5bf08
  * http://localhost:4040/
./scope: line 229: return: can only `return' from a function or sourced script

1 %> ./scope stop

0 %> ./scope launch --service-token=MYTOKEN
c011d7673d9fb774f4aca16c93acd506694e5c863bafe305edf124c58dca4a4e
./scope: line 213: return: can only `return' from a function or sourced script

1 %> 
```

errordeveloper:
Tested on Docker for Mac, all works well. LGTM.

Issue #921, opened 2016-02-05
Its not obvious what the "Tags" (under title in the details panel) do


tomwilkie:
Especially if you have a weirdly named host (such as 'default', which is what most docker-machine users seems to be ending up with).  Perhaps an icon or short piece of text (ie host:default, container:foo, process:bar) would help?

tomwilkie:
Node shapes!

tomwilkie:
Suggestion is to put little node shapes next to the tags to make it clear what they refer to.

pidster:
I don't think the header is necessarily the best place for this information. 
The design should communicate the meaning of it more clearly, just adding icons won't do that.

rade:
@davkal @foot suggestions?

davkal:
I'd keep it a comma-separated list for information density. If that field is unclear, a label could help. Something like 
```
CONTEXT: default, image123123
```
or
```
RELATED: default, image123123
```

I think icons could add information scent. I'd like to try it before ruling it out.

rade:
Surely there must be a more interesting (and functional) way to show this information than a comma-separated list in the details panel. Isn't the information essentially hierarchical? Then why can't we visualise it as a hierarchy?

davkal:
It's a network, not a hierarchy. A container has links to images and hosts. In the code we call them relatives. So we could render a network, but that needs a lot more space than a comma-sep list. I posit that it would not be proportionate to its importance.

rade:
>  I posit that it would not be proportionate to its importance.

I posit that as it stands most people cannot figure out what information is being conveyed there.

davkal:
> I posit that as it stands most people cannot figure out what information is being conveyed there.

Hence my proposed changes. I just cautioned for a measured response, and not a giant tag cloud.

davkal:
What this issue is missing, is that when you hover over the link `default` it will show you a tooltip "View in hosts: default". Progressive reveal.

davkal:
> I don't think the header is necessarily the best place for this information. 

The idea behind placing it there was that of breadcrumbs. The node title is colocated with its relatives for context.
In a previous discussion we concluded that putting it anywhere else makes even less sense. Or can you see a good match somewhere else @pidster ?

davkal:
> The design should communicate the meaning of it more clearly, just adding icons won't do that.

If we didnt have the hover tooltip already, I'd support the prefix that @tomwilkie suggested. We can give that a try though. I'm just concerned with label overdose.

[...]

PR #1673, opened 2016-07-13
Grid Mode

### LINKED ISSUE #1621: table mode
Allow users to list nodes in a tabular view instead of a connected topology.

Related: https://github.com/weaveworks/scope/tree/grid-view https://github.com/weaveworks/scope/issues/1614 https://github.com/weaveworks/scope/issues/1092 https://github.com/weaveworks/scope/issues/1032

foot:
Fixes #1621

- [x] Click row to open details panel
- [x] Filter table when hitting <enter> in search.
- [x] Add some rank color to the row
- [x] Remove `networks` column
- [x] Add row selection (highlighing)
- [x] Change enable/disable gridmode icons to something clearer..
- [x] Add context links to related items
  - [x] Remove hex id cols
- [x] k8s columns in agg views (check existing cols are sufficient)
- [x] Add sort column to url state
- [x] Test in Firefox!!
- [x] Display `label_minor` somewhere...
- [x] Split relatives out into other columns so you can sort by them.
- [x] Add search text matches
- [x] The column headings are mis-aligned (most systems other than osx have scrollbars).
- [x] Fix sorting null/undefined/empty str values.
- [x] Clarify default sort order onShow
- [x] Make row selection model clearer

- [x] Update row selection model, click anywhere on row to select, handle case of highlightling text somehow.
- [x] Highlight whole row blue

Pushed
- #1735 determine table columns in backend
- #1737 show-node-in-topo should perhaps scroll table to the position
- #1736 controls on grid
- #1738 consider dancing rows
- #1743 reconsider grid/viz toggle button position


rade:
when did you last rebase this branch? I am having trouble building it.

foot:
Yep, origin was in a bit of a bad way until ~2-3 mins ago.

Rebased again onto latest `master` too just now.

Gonna try building clean too as haven't done that for a bit...

rade:
re "Add context links to related items".. We need to be careful here not to repeat the mistakes of the details panel - see #921. We could have individual columns here for each of the contexts, e.g. for containers we would have host and pod and image columns. Trouble is that the kind of parents may well differ per item. Hmm. Also, what are we linking to here? Details panels?

Re k8s columns... not sure what you mean by that, but having a grid mode for all the k8s views is not a stretch goal.

Other missing bits:
- controls - so you don't have to go to the details panel for these
- highlighting interaction with details panel
  - when I click on a clickable link in a row and the details panel pops up, I expect the row to stay highlighted, just as in the topo mode the node stays highlighted
  - when navigating from the details panel to the grid using the "<-/->" button the row corresponding to the details panel item should be highlighted, just as happens in the topo mode

rade:
I am not keen on the increased row spacing. It halves the information density compared to the previous version.

rade:
> I am not keen on the increased row spacing

False alarm. Kinda. It's a firefox thing:

screenshot_2016-07-18_15-15-40

foot:
Yeah, will play w/ that wrapping, doesn't work that well in chrome w/ narrower browser window.

rade:
when I use context navigation, e.g. click on an image or host link in a container row, the details panel that pops up is lacking the "</>" button to switch to the table view for that kind of item.

foot:
Good one.

rade:
There will typically be two "Internet" rows - one for inbound and one for outbound - but since the table doesn't show sub-labels, there is no way to tell which is which.

rade:
what's the default sort order? Is there one? None of the headings have that little arrow on by default.

rade:
What is the purpose of the light blue highlighting for the name and context?

Looks like the context is just a list of links. That means I cannot sort by them, e.g. I cannot sort containers by image or host.

The aggregate process and container views (I haven't looked at k8s views yet) are pretty bare. Not much we can do about that until we have aggregate metrics. Though the process-by-name and container-by-dns-name views should at least show process/container counts.

rade:
I consider the current list of 'stretch' goals to be rather important. Of those, incremental search is the most important - I thought search wasn't working at all because I was so used to seeing the incremental matches from the topo mode. "scroll-to-position" is probably next since the UX can be really puzzling otherwise.

foot:
> What is the purpose of the light blue highlighting for the name and context?

Its the selector! Allows double-click on other table cells to copy-paste their values w/out opening the details panel. But that not be very clear.

> Looks like the context is just a list of links. That means I cannot sort by them, e.g. I cannot sort containers by image or host.

True, that might be nice.

rade:
> > What is the purpose of the light blue highlighting for the name and context?
>
> Its the selector! Allows double-click on other table cells to copy-paste their values w/out opening the details panel. But that not be very clear.

This either doesn't work or I don't understand how it's supposed to work.

For me, the blue highlight always shows for the name column when I hover over a row - it never moves to any other cells. And double clicking doesn't do anything special. Well, it pops up the details panel and immediately hides it, and selects the name.

rade:
A few other niggles...

screenshot_2016-07-22_08-55-00

- the name heading says 'PROCS' even though I am in the 'CONTAINERS' view
- the 'CONTAINERS-BY-NAME' heading is truncated and does not pop up a tooltip with the expansion
- a bunch of columns are truncated, yet there is acres of space in the first column
- the column headings are mis-aligned - notice how the 'HOSTS' and 'CREATED' headings spill over the column to their right. (somewhat excusable for 'HOSTS' since the column is very narrow, but not for 'CREATED')

I think the freeze-on-hover is too aggressive. Firstly it seems to cover the entire table area, not just the rows, so even having the mouse below the last row still freezes the table. Secondly, it looks like the freeze stops updating completely - I reckon it should only stop rows from moving, not values from updating. Tricky though if the most recent data adds/removes rows. Hmm.

A few alternative strategies to deal with updates while hovering over a row...
- fix the current row in place and move all the other rows around it to maintain sort order, i.e. effectively "scroll" the table around the current row
- lay out the table in the correctly sorted order, except for always placing the current row s.t. its position is maintained

The latter is slightly wrong but probably offers a more stable view overall.

foot:
> I think the freeze-on-hover is too aggressive

Agree

foot:
rebased on master.

rade:
>  Controls on grid

I'd be happy to see this tackled in a follow-up issue.

rade:
>  Consider dancing rows

I'd be happy to see this tackled in a follow-up issue, especially since a) the table only updates every few seconds, and b) until we add controls, nothing bad happens when you accidentally click on the wrong row.

Most of the other niggles I found have been fixed, except for

> a bunch of columns are truncated, yet there is acres of space in the first column

rade:
There's something weird going on with the sort-order when a field is absent/empty. In the following processes-sorted-by-containers, there are blank entries at the top and near the bottom, though all the non-blank entries are sorted correctly.

screenshot_2016-07-26_08-01-48

davkal:
Functionality:

- sort works great, i can see the "top k of X"
  - no pause on hover is fine, user can use pause button
- mode works seamlessly, switching back and forth between topo works great
- search works, but it's not clear what's matched, even non-visible fields are matched (maybe display them somehow?)

UX:

- why put the mode selector next to the node filters? I think it should be superior, like the topologies, or search (long term i'm assuming them to be in a menu bar)
- why doesnt clicking anywhere on the row select that node? currently i have to click on the label
  - easily clicking on the row seems greatly more important to me than to distinguish clicks on special cells like relatives (if you want to click on relatives you can do it from the details panel). Also, the underlining of relatives is weird (I know it's for consistency, but on the details panel, a click on a relative stacks another panel on top, which does not happen here) If we find that clicking on relatives is a major flow, than we can think about it again
- why isnt the whole row highlighted blue when selected? seems more in line with the whole node being selected in the graph view
- How about dynamically adding more columns if the screen gets wide enough? May need introduction of backend column ranking/weight.

Visual:

- it's hard to follow a row across the screen when column distances get bigger, maybe increase row distance? (also see note about dynamically adding columns)
- top border feels weird, because there is a shadow suggesting inset, then on the sides no inset can be seen, Escher-Grid?
- What's the purpose of the bottom border?
- the blue background over the light selected row background feels weird

rade:
I see I am not the the only person confused by the blue highlighting :)

> maybe increase row distance

Please don't. Information density is key to this mode. Alternating colours, perhaps?

foot:
Check this out we get this cool gradient where the alt. colors become more pronounced the further along the row you go because of our bg gradient! Schön oder?

screen shot 2016-07-27 at 19 37 01

rade:
Sehr schön.

Well, except for that control panel in the bottom left obscuring a whole bunch of entries. But you already know about that, and we can fix that later. And the columns that are infuriatingly truncated because they are just a few pixels too narrow. Please do fix that.

foot:
So I'll have a go moving the toggle button and double check those column widths but do you think you might have time to read through some code today @davkal ?

foot:
> How about dynamically adding more columns if the screen gets wide enough? May need introduction of backend column ranking/weight.

That would be cool.

> top border feels weird, because there is a shadow suggesting inset, then on the sides no inset can be seen, Escher-Grid?

I like the effect! the rows are disappearing underneath something. We could turn it off too though. Try it out :).

> What's the purpose of the bottom border?

A bordered area for the rows to appear out of. Bottom of the screen and they clash w/ the control icons in bottom right.

foot:
> I like the effect! the rows are disappearing underneath something. 

I've seen this effect done elsewhere but perhaps the shadow only appears when you start scrolling.

foot:
> I've seen this effect done elsewhere but perhaps the shadow only appears when you start scrolling.

E.g. gmail!

client/app/scripts/actions/app-actions.js
```diff
@@ -32,6 +34,18 @@ export function toggleHelp() {
   };
 }
 
+
+export function sortOrderChanged(sortBy, sortedDesc) {
```
davkal: Why no component/feature prefix? Is there a global sort order?

rade:
'Processes' table isn't working anymore. i.e. when I click on 'Table' when in 'Processes' view nothing happens. And from then on the UI no longer responds - buttons still get highlighted when hovering over them but clicking does nothing.

client/app/scripts/actions/app-actions.js
```diff
@@ -217,6 +232,22 @@ export function clickForceRelayout() {
   };
 }
 
+export function toggleGridMode(enabledArgument) {
```
davkal: I would have made that a string option to allow for more modes than 2.

client/app/scripts/charts/nodes-chart.js
```diff
@@ -402,7 +394,6 @@ function mapStateToProps(state) {
   return {
     adjacentNodes: getAdjacentNodes(state),
     forceRelayout: state.get('forceRelayout'),
-    nodes: state.get('nodes').filter(node => !node.get('filtered')),
```
davkal: Why not keep it there? The parent component does not need `nodes`. `layoutPrecision` can be calculated here too.

client/app/scripts/charts/nodes-grid.js
```diff
@@ -0,0 +1,147 @@
...
+import { connect } from 'react-redux';
+import { List as makeList, Map as makeMap } from 'immutable';
+import NodeDetailsTable from '../components/node-details/node-details-table';
+import { clickNode, sortOrderChanged, clickPauseUpdate,
+  clickResumeUpdate } from '../actions/app-actions';
+
+import { getNodeColor } from '../utils/color-utils';
+
+
+const IGNORED_COLUMNS = ['docker_container_ports', 'docker_container_id', 'docker_image_id',
```
davkal: What's important and what can be ignored, should be determined by the backend. The current metadata field `priority` is not sufficient, though. Looks like we need another field.

client/app/scripts/charts/nodes-grid.js
```diff
@@ -0,0 +1,147 @@
...
+    .flatMap(n => {
+      const metadata = (n.get('parents') || makeList())
+        .map(m => makeMap({ id: m.get('topologyId'), label: m.get('topologyId') }));
+      return metadata;
+    })
+    .toSet()
+    .toList()
+    .sortBy(m => m.get('label'));
+
+  return relativesColumns.concat(metadataColumns.concat(metricColumns)).toJS();
```
davkal: I'd prefer `l1.concat(l2, l3).toJS()`

client/app/scripts/charts/nodes-grid.js
```diff
@@ -0,0 +1,147 @@
...
+          />}
+      </div>
+    );
+  }
+}
+
+
+export default connect(
+  () => ({}),
+  { clickNode, sortOrderChanged, clickPauseUpdate, clickResumeUpdate }
```
davkal: `clickPauseUpdate` and `clickResumeUpdate` are not used.

client/app/scripts/components/grid-mode-selector.js
```diff
@@ -0,0 +1,61 @@
...
+    );
+  }
+
+  render() {
+    const { gridMode } = this.props;
+
+    return (
+      <div className="grid-mode-selector">
+        <div className="grid-mode-selector-wrapper">
+          {this.renderItem('fa fa-share-alt', 'Visualization', !gridMode, this.disableGridMode)}
```
davkal: "Visualization" is pretty broad. How about "Graph"?

client/app/scripts/components/node-details.js
```diff
@@ -48,7 +47,7 @@ export class NodeDetails extends React.Component {
   }
 
   renderTools() {
-    const showSwitchTopology = this.props.index > 0;
+    const showSwitchTopology = this.props.nodeId !== this.props.selectedNodeId;
```
davkal: 👍

client/app/scripts/components/node-details/node-details-table-row.js
```diff
@@ -0,0 +1,143 @@
...
+    return <td className="node-details-table-node-value" style={style} key={id} />;
+  });
+}
+
+
+export default class NodeDetailsTableRow extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+
+    this.mouseDragOrigin = [0, 0];
```
davkal: Comment for what it's used would be nice.

client/app/scripts/components/node-details/node-details-table-row.js
```diff
@@ -0,0 +1,143 @@
...
+
+  onMouseUp(ev) {
+    const [originX, originY] = this.mouseDragOrigin;
+    const { pageX, pageY } = ev;
+    const thresholdPx = 2;
+    const movedTheMouseTooMuch = (
+      Math.abs(originX - pageX) > thresholdPx ||
+      Math.abs(originY - pageY) > thresholdPx
+    );
+    if (movedTheMouseTooMuch) {
```
davkal: 👍

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -13,22 +13,128 @@ function isNumberField(field) {
 
 const COLUMN_WIDTHS = {
```
davkal: maybe we should do width types, like XS, S, M, L, and assign width via CSS classes

client/app/scripts/components/nodes.js
```diff
@@ -63,19 +64,33 @@ class Nodes extends React.Component {
...
-          hasSelectedNode={hasSelectedNode}
-        />
+        {gridMode ?
+          <NodesGrid {...this.state}
+            nodeSize="24"
+            nodes={nodes}
+            topology={this.props.currentTopology}
+            topologyId={this.props.currentTopologyId}
+            margins={CANVAS_MARGINS}
+            layoutPrecision={layoutPrecision}
```
davkal: Not needed.

client/app/scripts/components/nodes.js
```diff
@@ -94,9 +109,16 @@ class Nodes extends React.Component {
...
+    gridSortBy: state.get('gridSortBy'),
+    gridSortedDesc: state.get('gridSortedDesc'),
+    nodes: state.get('nodes').filter(node => !node.get('filtered')),
+    currentTopology: state.get('currentTopology'),
+    currentTopologyId: state.get('currentTopologyId'),
     topologyEmpty: isTopologyEmpty(state),
+    searchNodeMatches: state.getIn(['searchNodeMatches', state.get('currentTopologyId')]),
+    searchQuery: state.get('searchQuery'),
+    selectedNodeId: state.get('selectedNodeId')
   };
```
davkal: The props needed only for gridview, should be put in the grid's `mapStateToProps`.

client/app/scripts/utils/router-utils.js
```diff
@@ -39,11 +39,14 @@ export function getUrlState(state) {
 
   const urlState = {
     controlPipe: cp ? cp.toJS() : null,
+    mode: state.get('gridMode') ? 'grid' : 'topo',
```
davkal: `mode` is very broad. How about `topologyViewMode`?

client/app/scripts/utils/update-buffer-utils.js
```diff
@@ -32,7 +32,7 @@ function maybeUpdate(getState) {
       receiveNodesDelta(delta);
     }
     if (deltaBuffer.size > 0) {
-      updateTimer = setTimeout(maybeUpdate, feedInterval);
+      updateTimer = setTimeout(() => maybeUpdate(getState), feedInterval);
```
davkal: Nice find.

client/app/styles/main.less
```diff
@@ -15,6 +15,7 @@
 /* weave company colours */
 @weave-gray-blue: rgb(85,105,145);
 @weave-blue: rgb(0,210,255);
+@weave-blue-transparent: rgb(0,210,255, 0.1);
```
davkal: `rgba`?

davkal:
Commented on JS/CSS code. Minor comments, rest looks good

foot:
Continued in
- #1735 - get BE more involved in table column visibility
- #1736 - add controls to table
- #1737 - scroll table on show-node-in-topo

foot:
@rade please have a quick look at the columns when you get a chance. The label columns are now as balanced as the other variable columns which should reduce the acreage. Please post screenshot if I'm mis-understanding the issue.

davkal:
Alternative selector positions:

above search: as the paramount view selector
screen shot 2016-08-01 at 17 26 39

between search and topos (all needs shift to the right): seeing mode and search as 2 dimensions
screen shot 2016-08-01 at 17 29 32

search above: search as overarching element that is affecting all below
screen shot 2016-08-01 at 17 41 17

rade:
Visually this LGTM now. I suggest filing a separate issue for the selector positioning.

rade:
Have just noticed that the name doesn't have a tool tip. Which is especially annoying when it is truncated.

rade:
no longer LGTM. i am sitll running into https://github.com/weaveworks/scope/pull/1673#issuecomment-235975345.

rade:
js console shows
```
vendors.js?4b45a9c…:1 Uncaught Invariant Violation: Minified React error #121; visit http://facebook.github.io/react/docs/error-decoder.html?invariant=121&args[]=93&args[]=78 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
```

foot:
Cheers, thats a new error message for me! Investigating.

foot:
I'm trying to repro, not got it yet. Can you post a report if you get the error consistently.

rade:
> Can you post a report if you get the error consistently.

[report.json.gz](https://github.com/weaveworks/scope/files/396192/report.json.gz)

foot:
Cheers, good catch w/ label tooltip too.

rade:
process view fix is working for me. and tooltip too.

Other than filing an issue for the selector position (re)location, this LGTM.

PR #1675, opened 2016-07-14
Make sure we generate a proper docker tag

krnowak:
I have a branch named like <user>/<feature>, which becomes a part of
the tag name. Docker complains about it, because / is not a valid
character in the tag.

This commit replaces all sequences of invalid chars with a single dash
and makes sure that the tag name never exceeds the 127 chars limit.

### DIFF
--- a/image-tag
+++ b/image-tag
@@ -4,9 +4,21 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-WORKING_SUFFIX=$(if ! git diff --exit-code --quiet HEAD >&2; \
-                 then echo "-WIP"; \
-                 else echo ""; \
-                 fi)
-BRANCH_PREFIX=$(git rev-parse --abbrev-ref HEAD)
-echo "${BRANCH_PREFIX//\//-}-$(git rev-parse --short HEAD)$WORKING_SUFFIX"
+if [[ $# -gt 0 ]]; then
+    BRANCH_SUFFIX=""
+    BRANCH_PREFIX="${1}"
+else
+    WORKING_SUFFIX=$(if ! git diff --exit-code --quiet HEAD >&2; \
+                     then echo "-WIP"; \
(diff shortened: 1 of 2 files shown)


alban:
This is similar to https://github.com/weaveworks/scope/pull/1655

And probably failing for the same reason.

tomwilkie:
Hi @krnowak; can you rebase this on top of master?  It should pass now.

krnowak:
Updated, lint seems to fail with some typos. For lint fixes please see https://github.com/weaveworks/scope/pull/1751.

image-tag
```diff
@@ -8,5 +8,10 @@ WORKING_SUFFIX=$(if ! git diff --exit-code --quiet HEAD >&2; \
                  then echo "-WIP"; \
                  else echo ""; \
                  fi)
-BRANCH_PREFIX=$(git rev-parse --abbrev-ref HEAD)
-echo "${BRANCH_PREFIX//\//-}-$(git rev-parse --short HEAD)$WORKING_SUFFIX"
```
alban: For consistency, should `${CIRCLE_BRANCH//\//-}` be changed in `circle.yml` as well?

krnowak:
Updated, so we can be sure that tag names in circleCI are also valid.

2opremio:
@krnowak / @alban Do we still need this?

krnowak:
Up to you, really. But now I guess this PR would need to be split in two - one for image-tag to be filed in build-tools repository and one for scope importing the new build-tools version and modifying circle.yml.

My issue was the same as the one @alban had - I had / in my branch name and docker did not like it. I just made extra sure that the tag we generate based on a branch name is always valid (no invalid chars and shorter than 128 chars).

tomwilkie:
Already an image-tag in build-tools: https://github.com/weaveworks/build-tools/blob/master/image-tag

PR #1681, opened 2016-07-14
Add flag for block profiling

### LINKED ISSUE #1672: Add flag to enable Go blocking profiling
We'd like to understand whether Go is blocking much on our services. Go's block profiling would be useful for this. We should add a flag to enable it.

c.f. https://golang.org/pkg/runtime/#SetBlockProfileRate

jml:
Fixes #1672

Test by adding `--app.block.profile.rate=1`, running scope in
multi-tenant mode and browsing to http://<scope-url>/debug/pprof/

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1681)

### DIFF
--- a/site/building.md
+++ b/site/building.md
@@ -76,7 +76,7 @@ To collect a blocking profile of the Scope App, make sure you have launched
 Scope with `--app.block.profile.rate=N` (where `N` is the number of
 nanoseconds between samples) and then:
 
-    go tool pprof http://localhost:4040/debug/pprof/
+    go tool pprof http://localhost:4040/debug/pprof/block
 
 If you don't have `go` installed, you can use a Docker container instead:
 
--- a/site/building.md
+++ b/site/building.md
@@ -72,6 +72,12 @@ To collect the CPU profile of the Scope Probe:
 
     go tool pprof http://localhost:4041/debug/pprof/profile
 
+To collect a blocking profile of the Scope App, make sure you have launched
+Scope with `--app.block.profile.rate=N` (where `N` is the number of
+nanoseconds between samples) and then:
+
+    go tool pprof http://localhost:4040/debug/pprof/
+
 If you don't have `go` installed, you can use a Docker container instead:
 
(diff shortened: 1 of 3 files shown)


2opremio:
> Test by adding --app.block.profile.rate=1

Isn't this flag a bit fine-grained? How about simply enabling block profiling when supplying --debug?

Also, could you please update `site/building.md`?

2opremio:
Another option is adding something like `--debug.block-profiling-rate=n` which would affect both in the app and the probe.

(these are just suggestions, it's ultimately up to you)

2opremio:
Thinking about it a bit more, in the documentation it seems we separate debugging (showing more info in the logs) from profiling so feel free to ignore my comments about the flag names.

jml:
> Isn't this flag a bit fine-grained? How about simply enabling block profiling with a default value when supplying --debug

I anticipate that we'll get the default value wrong & will want to iterate quickly. Flags are the best way we currently have of doing that.

> Another option is adding something like --debug.block-profiling-rate=n which would affect both in the app and the probe.

Having one flag for both app & probe seems like a good idea. Does your other comment about separating debug from profiling preclude this as a flag name?

2opremio:
> Does your other comment about separating debug from profiling preclude this as a flag name?

Probably, at least we separate them in the docs.

site/building.md
```diff
@@ -72,6 +72,12 @@ To collect the CPU profile of the Scope Probe:
 
     go tool pprof http://localhost:4041/debug/pprof/profile
 
+To collect a blocking profile of the Scope App, make sure you have launched
+Scope with `--app.block.profile.rate=N` (where `N` is the number of
+nanoseconds between samples) and then:
+
+    go tool pprof http://localhost:4040/debug/pprof/
```
2opremio: Shouldn't this be

`go tool pprof http://localhost:4040/debug/pprof/`**`block`** ?

2opremio:
LGTM (provided you get a green from circle)

PR #1682, opened 2016-07-14
RFC: forwarding control requests to plugins

krnowak:
This is a tentative PR that implements forwarding control requests to plugins.

~~Problems so far:~~

- ~~I made iowait example plugin a controller (adds a button that switches between showing iowait and idle), but:~~
  - ~~GUI shows plugin's statistics in a separate node, while it used to be in the same one~~
  - ~~GUI shows no button from plugin~~


krnowak:
Updated the PR. There are still some more problems though:

- NodeControls' not-really-merging behaviour of its Merge function causes races between plugins adding controls to the same node - only the plugin that sends a report with newest timestamp will have its controls displayed.
  - Probably needs a LatestMap style handling of the NodeControls, because my current hack causes the obsolete controls to be still shown for several seconds
- Shows obsolete statistics
  - Likely the fix could be also as the LatestMap style handling.
- Slow reaction to controls
  - The plugin gets the request quickly
  - The new statistics are shown after several seconds though

examples/plugins/iowait/main.go
```diff
@@ -52,70 +53,188 @@ func main() {
...
+	expectedControlID, _, _ := p.controlDetails()
+	if expectedControlID != xreq.Control {
+		log.Printf("Bad control, expected %q, got %q", expectedControlID, xreq.Control)
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	p.iowaitMode = !p.iowaitMode
+	w.WriteHeader(http.StatusOK)
+	fmt.Fprint(w, "{}")
+}
```
paulbellamy: As this is an example, maybe it's worth writing it in such as way that the plugin could handle multiple different controls? Maybe not worth the effort?

examples/plugins/iowait/main.go
```diff
@@ -52,70 +53,188 @@ func main() {
...
+			"icon": %q,
+			"rank": 1
+		}
+`, id, id, human, icon)
+	nodeControl := fmt.Sprintf(`
+		"timestamp": %q,
+		"controls": [%q]
+`, nowISO, id)
+	return topologyControl, nodeControl
+}
```
paulbellamy: This might be nicer with a struct and json.Marshal.

paulbellamy: Also metrics snippets above.

probe/controls/controls.go
```diff
@@ -6,33 +6,109 @@ import (
...
+	// Register a new control handler under a given id.
+	Register(control string, f xfer.ControlHandlerFunc)
+	// Rm deletes the handler for a given name.
+	Rm(control string)
+	// Handler gets the handler for the given id.
+	Handler(control string) (xfer.ControlHandlerFunc, bool)
+	// RegisterBatch adds many controls in one go
+}
+
+type realHandlerRegistry map[string]xfer.ControlHandlerFunc
```
paulbellamy: maybe just `handlerRegistry` instead.

probe/controls/controls.go
```diff
@@ -6,33 +6,109 @@ import (
...
+	handler, ok := r[control]
+	return handler, ok
+}
+
 var (
-	mtx      = sync.Mutex{}
-	handlers = map[string]xfer.ControlHandlerFunc{}
+	mtx                          = sync.Mutex{}
+	realRegistry                 = realHandlerRegistry{}
+	registry     HandlerRegistry = realRegistry
```
paulbellamy: If we can, lets avoid singletons. Also, the mutex should probably just be part of the registry itself.

krnowak: The mutex needs to be outside the registry, because we lock it also when we change the registry with `SetHandlerRegistry` or `ResetHandlerRegistry`.

probe/plugins/registry_internal_test.go
```diff
@@ -413,6 +453,16 @@ func TestRegistryRejectsErroneousPluginResponses(t *testing.T) {
 			Name:    "nonJSONResponseBody",
 			Handler: stringHandler(http.StatusOK, `notJSON`),
 		}.file(),
+		mockPlugin{
+			t:       t,
+			Name:    "changedId",
```
paulbellamy: I believe we generally use `ID` instead of `Id`.

alban:
Could you update [examples/plugins/README.md](https://github.com/weaveworks/scope/blob/master/examples/plugins/README.md)?

probe/controls/controls.go
paulbellamy: Ok. But if registry is not a singleton, then we don't need that.

examples/plugins/README.md
```diff
@@ -69,16 +86,131 @@ For example:
...
+Add the "controller" string to the `interfaces` field in the plugin
+specification.
+
+#### Control
+
+The `POST` requests will have a JSON-encoded body with the following contents:
+
+```json
+{
+  "AppID": "some ID of an app"
```
alban: missing commas at the end of the lines

alban: What is an app in this context? Can a plugin have several "app"?

examples/plugins/README.md
```diff
@@ -69,16 +86,131 @@ For example:
...
+  "NodeID": "an ID of the node that had the control activated"
+  "Control": "the name of the activated control"
+}
+```
+
+The body of the response should also be a JSON-encoded data. Usually
+the body would be an empty JSON object (so, "{}" after
+serialization). If some error happens during handling the control,
+then the plugin can send a response with an `error ` field set, for
```
alban: extra space after `error`.

examples/plugins/README.md
```diff
@@ -69,16 +86,131 @@ For example:
...
+      }
+    },
+    "nodes": {
+      "host1": {
+        "latestControls": {
+          "latest": {
+            "ctrl-one": {
+              "timestamp": "2016-07-20T15:51:05Z01:00",
+              "value": {
+                "dead": false
```
alban: What is "dead"? Does it mean the button is disabled? Or is it an arbitrary string displayed to the user?

examples/plugins/README.md
```diff
@@ -69,16 +86,131 @@ For example:
...
+    }
+  }
+}
+```
+
+When control "ctrl-one" is activated, the plugin will receive a request like:
+
+```json
+{
+  "AppID": "some ID of an app"
```
alban: missing commas at the end of the lines

examples/plugins/README.md
krnowak: I guess it is an ID of the scope app. No idea about second question.

examples/plugins/README.md
krnowak: Means the button is not shown. I probably came up with this word, because I was thinking about tombstones. ;)

probe/controls/controls.go
krnowak: I'm not sure I understand. This registry was always a singleton, I just kept it like that and added some more functions to the package.

The registry could be a separate object, but I think this could be done in a separate PR.

krnowak:
Updated. There are still some failures, likely because of changing `report.LatestEntry.Value` from `string` to `interface{}`. I would appreciate some help/hints how to fix them.

examples/plugins/README.md
```diff
@@ -25,21 +25,38 @@ of the `docker run ...` command.
 If the running plugin was picked up by Scope, you will see it in the list of `PLUGINS`
 in the bottom right of the UI.
 
+## Plugin ID
 
-## <a id="protocol"></a>Protocol
+Each plugin should have an unique ID. It is forbidden to change it
+during plugin's lifetime. The scope probe will get the plugin's ID
```
paulbellamy: `during a plugin's lifetime` or `during the plugin's lifetime`

examples/plugins/README.md
```diff
@@ -25,21 +25,38 @@ of the `docker run ...` command.
...
+from the plugin's socket filename. So from the socket named
```
paulbellamy: For example, the socket named ...

examples/plugins/README.md
```diff
@@ -25,21 +25,38 @@ of the `docker run ...` command.
...
+my-plugin.sock, the scope probe will deduce ID as "my-plugin". ID can
```
paulbellamy: can use back-ticks to quote these bits: `my-plugin.sock`, and `my-plugin`

examples/plugins/README.md
```diff
@@ -25,21 +25,38 @@ of the `docker run ...` command.
...
+contain only alphanumeric sequences optionally separated with a dash.
```
paulbellamy: IDs can only contain alphanumeric sequences, separated with a dash.

examples/plugins/README.md
```diff
@@ -25,21 +25,38 @@ of the `docker run ...` command.
...
 also valid to put the plugin unix socket in a sub-directory, in case you want
 to apply some permissions, or store other information with the socket.
 
-When a new plugin is detected, the scope probe will begin requesting
-reports from it via `GET /report`.
+## Protocol
+
+There are several interfaces a plugin may (or must) implement. Usually
+implementing an interface means handling specific requests. What
+requests these are is described below.
```
paulbellamy: These requests are described below.

examples/plugins/README.md
```diff
@@ -25,21 +25,38 @@ of the `docker run ...` command.
...
+
+### Reporter interface
 
-All plugin endpoints are expected to respond within 500ms, and respond in the JSON format.
+The reporter interface is an interface that every plugin _must_
```
paulbellamy: Every plugin _must_ implement the reporter interface.

examples/plugins/README.md
```diff
@@ -69,16 +86,131 @@ For example:
...
+You may notice a small chicken and egg problem - the plugin reports to
+the scope probe what interfaces it supports, but the scope probe can
+learn that only by doing a `GET /report` request which will be handled
+by the plugin if it implements the "reporter" interface. This is
+solved (or worked around) by requiring the plugin to always implements the
+"reporter" interface.
+
+### Controller interface
+
+Implementing the controller interface means that the plugin can react
```
paulbellamy: Maybe add: Plugins _may_ implement the controller interface.

examples/plugins/README.md
```diff
@@ -69,16 +86,131 @@ For example:
...
+        "id": "ctrl-two",
+        "human": "Ctrl Two",
+        "icon": "fa-beer",
+        "rank": 2
+      }
+    },
+    "nodes": {
+      "host1": {
+        "latestControls": {
+          "latest": {
```
paulbellamy: It seems a bit odd to have the nested `"lastestControls": { "latest": { ...`. The inner "latest" seems unnecessary?

examples/plugins/README.md
```diff
@@ -69,16 +86,131 @@ For example:
...
+```json
+{
+  "AppID": "some ID of an app",
+  "NodeID": "host1",
+  "Control": "ctrl-one"
+}
+```
 
-Currently the only interface a plugin can fulfill is `reporter`.
+A short note about the "icon" field of the topology control - the
```
paulbellamy: nice, thanks.

probe/controls/controls.go
```diff
@@ -6,33 +6,109 @@ import (
...
+	handler, ok := r[control]
+	return handler, ok
+}
+
 var (
-	mtx      = sync.Mutex{}
-	handlers = map[string]xfer.ControlHandlerFunc{}
+	mtx                          = sync.Mutex{}
+	realRegistry                 = handlerRegistry{}
+	registry     HandlerRegistry = realRegistry
```
paulbellamy: I still think we should get rid of the global state here, and move the lock into the `handlerRegistry` implementation.

probe/plugins/registry.go
```diff
@@ -180,23 +210,140 @@ func (r *Registry) Name() string { return "plugins" }
...
+func realPluginAndControlID(fakeID string) (string, string) {
+	parts := strings.SplitN(fakeID, "~", 2)
+	if len(parts) != 2 {
+		return "", fakeID
+	}
+	return parts[0], parts[1]
+}
+
+func fakeControlID(pluginID, controlID string) string {
+	return fmt.Sprintf("%s~%s", pluginID, controlID)
```
paulbellamy: In other places (report/id.go) we use `;` for node IDs, and `|` for edge IDs. I suppose, it's good to have a different one, so they can't conflict. I was going to suggest a different character, but actually, I like that `~` is url-safe. TL;DR would be nice if we didn't have to munge IDs around, but looks fine, and seems reasonably tidy in this case.

paulbellamy: I'm assuming there is a check for control IDs not including `~`? (or should there be?)

report/latest_map.go
```diff
@@ -90,27 +90,42 @@ func (m LatestMap) Lookup(key string) (string, bool) {
...
 	if m.Map == nil {
 		m = EmptyLatestMap
 	}
 	return LatestMap{m.Map.Set(key, LatestEntry{timestamp, value})}
 }
 
+// Set the value for the given key.
+func (m LatestMap) Set(key string, timestamp time.Time, value string) LatestMap {
+	return m.SetExt(key, timestamp, value)
+}
```
paulbellamy: SetExt was a bit unclear to me. What do you think about:
```
func (m LatestMap) Set(key string, timestamp time.Time, value interface{}) LatestMap
func (m LatestMap) SetString(key string, timestamp time.Time, value string) LatestMap
```

i.e. making the interface{} one the "default", then having different names for the more-specifically-typed variants...

paulbellamy: Same for ForEach/ForEachExt, as well...

paulbellamy:
This *looks* like it's all backwards-compatibility with the existing controls? But wanted to check the interaction between them. For example, what happens if you have both?

probe/plugins/registry.go
krnowak: No, there is only a check for plugin IDs not including `~`. Control IDs can contain `~` - we split the fake ID only against the first tilde.

paulbellamy: ok.

examples/plugins/README.md
krnowak: Hmm, yeah. `latestControls` is a separate struct that has the `report.LatestMap` member. I did it that way in case we want to add something more to `latestControls`… Would you prefer to merge them or a better name would suffice?

paulbellamy: What is latestNodeControls embedded, or aliased LatestMap?

report/latest_map.go
krnowak: I did it that way, because I didn't want to change the LatestMap API, so I opted for leaving `Set` as is and adding a `SetExt`. I could for example rename it to `SetInterface` instead…

But I can break it if you want, no problem.

krnowak:
> This _looks_ like it's all backwards-compatibility with the existing controls? But wanted to check the interaction between them. For example, what happens if you have both?

I was wondering how to handle it. For now I just ignore the old `nodeControls`. I see the following options here:
- either remove old `nodeControls` altogether
- convert `nodeControls` to `latestControls`.
- ignore and warn.

But this problem is not really severe. All the internal probe plugins are already converted to `latestControls`. And `nodeControls` weren't really usable from plugins before this PR.

probe/controls/controls.go
krnowak: Alright.

report/latest_map.go
paulbellamy: > I didn't want to change the LatestMap API

Why? It's only used internally, no?

paulbellamy:
My initial reaction is to convert any found `nodeControls` to `latestControls`, so it all keeps working.

The probe plugins are converted for *newly* deployed probes. But it would break existing probes already deployed.

krnowak:
Partially updated, the circleci failure looks unrelated.

I took different approach to latest map thingy - LatestMap is now a "base class" which has the implementation for lookups, marshalling/unmarshalling and so on, and we use a go:generator to generate thin types on top of LatestMap like StringLatestMap and NodeControlDataLatestMap, so we get a type safety and proper marshalling/unmarshalling (unmarshalling a string to interface{} required a hack for mappack codec).

This also removes one level from latestControls (so it is `"latestControls": { "ctrl1": {...} }` instead of `"latestControls": {"latest": {"ctrl1": {...}}}`.

I still need to write more info in plugins readme (for example, about node naming).

I haven't yet addressed the backward compatilibity issue. For now I'm not sure about feasibility of supporting every combination of a control plugin/{old,new} probe/{old,new} app. But I'll give more thought to it.

From smaller new changes I made the differ a bit more verbose, so it may be easier to find the cause of DeepEqual failures, where the differ was showing an empty string. Also, I added a possibility of running only some tests with `make tests` by running `TESTDIRS="./reports ./probe/plugins" make tests`.

krnowak:
So, about new/old plugin/probe/app interactions - we have 2^3 interactions here. Let's talk about plugin/probe first (in order of increasing troublesomeness):

- old plugin, old probe - no problem, controls in plugins were not supported
- new plugin, new probe - no problem, controls are supported
- old plugin, new probe - no problem either, old plugins can't expect that their controls will work, since they don't implement the controller interface
- new plugin, old probe - the controls will be simply ignored, which is the only thing we can do - old probe does not implement control request forwarding

So, old/new probe/app is more problematic:

- old probe, old app - alright, both are going to send/consume the old nodecontrols data in the report
- new probe, new app - fine and dandy, both are going to send/consume the new latestcontrols data in the report
- old probe, new app - old probe has no control forwarding, thus there are no problems with control appearance races, so the app could translate nodecontrols received from old probe to latestcontrols
- new probe, old app - the new probe can add a backward-compatible Controls instance to each node based on the node's LatestControls member.

So, what I did is:

- the app does adds new LatestControls based the old Controls in its `collector.Report()` function. This means that the final form of a report (as returned by `GET /api/report`) is "upgraded" and cached.
- the probe does the conversion right before sending it to the app in its `probe.drainAndPublish()` function. This means that the final report from all the plugins (or the shortcutted report only) is converted before sending.

krnowak:
CircleCI lint failure is unrelated to this PR I believe. I need to address the coverage issue.

krnowak:
Updated, lint failures are unrelated (lint finds misspelled "daed" word in "e937db5a3120c83115dc47e20c9b960503f612848daed7c6e0739b10372fb5db"…)

paulbellamy:
Ok, that shouldn't be failing on master. Please rebase/merge/whatever, so we get a clean build.

krnowak:
It already is rebased on current master.

krnowak:
Has something changed in linting recently? Because the line behind the `app/multitenant/sqs_control_router.go:336:21:found "recieving" a misspelling of "receiving"` error comes from March, git blame:
```
98857064 (Tom Wilkie   2016-03-09 13:47:02 +0000 336)                   log.Errorf("Error recieving message: %v", err)
```

Makefile
```diff
@@ -52,27 +52,28 @@ $(RUNSVINIT): vendor/runsvinit/*.go
 
 $(SCOPE_EXE): $(shell find ./ -path ./vendor -prune -o -type f -name *.go) prog/static.go $(CODECGEN_TARGETS)
 
-report/report.codecgen.go: $(call GET_CODECGEN_DEPS,report/)
+report/latest_map_impls_generated.go: report/latest_map_impls.go tools/generate_latest_map
+report/report.codecgen.go: $(call GET_CODECGEN_DEPS,report/) report/latest_map_impls_generated.go
```
paulbellamy: Not sure why we need these dep changes? Regardless, adding the `report/latest_map_impls_generated.go` as a dependency looks to be done in the wrong place.

app/collector.go
```diff
@@ -118,7 +118,9 @@ func (c *collector) Report(_ context.Context) (report.Report, error) {
 	}
 
 	c.clean()
-	return c.merger.Merge(c.reports), nil
+	rpt := c.merger.Merge(c.reports).Upgrade()
```
paulbellamy: nice

probe/controls/controls.go
```diff
@@ -6,33 +6,138 @@ import (
...
+	defer r.backend.Unlock()
+	r.RegisterLocked(control, f)
+}
+
+// Rm deletes the handler for a given name.
+func (r *HandlerRegistry) Rm(control string) {
+	r.backend.Lock()
+	defer r.backend.Unlock()
+	r.RmLocked(control)
+}
```
paulbellamy: Why do we need the separation between the HandlerRegistry, and the HandlerRegistryBackend? They look almost the same to me?

probe/kubernetes/controls.go
```diff
@@ -144,15 +144,19 @@ func (r *Reporter) ScaleDown(req xfer.Request, resource, namespace, id string) x
...
-	controls.Register(GetLogs, r.CapturePod(r.GetLogs))
-	controls.Register(DeletePod, r.CapturePod(r.deletePod))
-	controls.Register(ScaleUp, r.CaptureResource(r.ScaleUp))
-	controls.Register(ScaleDown, r.CaptureResource(r.ScaleDown))
+	r.handlerRegistry.Lock()
+	defer r.handlerRegistry.Unlock()
+	r.handlerRegistry.RegisterLocked(GetLogs, r.CapturePod(r.GetLogs))
+	r.handlerRegistry.RegisterLocked(DeletePod, r.CapturePod(r.deletePod))
+	r.handlerRegistry.RegisterLocked(ScaleUp, r.CaptureResource(r.ScaleUp))
+	r.handlerRegistry.RegisterLocked(ScaleDown, r.CaptureResource(r.ScaleDown))
```
paulbellamy: In vague pseudo-code, here I'd either, just call:
```go
r.handlerRegistry.Register(GetLogs, r.CapturePod(r.GetLogs))
r.handlerRegistry.Register(DeletePod, r.CapturePod(r.deletPod))
...
```
or:
```go
r.handlerRegistry.Register(map[string]Handler{
  GetLogs: r.CapturePod(r.GetLogs),
  DeletePod: r.CapturePod(r.deletePod),
})
```

probe/plugins/registry.go
```diff
@@ -180,23 +212,140 @@ func (r *Registry) Name() string { return "plugins" }
...
+func (r *Registry) updateAndRegisterControlsInTopology(pluginID string, topology *report.Topology) {
+	var pluginControls []string
+	newControls := report.Controls{}
+	for controlID, control := range topology.Controls {
+		fakeID := fakeControlID(pluginID, controlID)
+		log.Debugf("plugins: replacing control %s with %s", controlID, fakeID)
+		control.ID = fakeID
+		newControls.AddControl(control)
+		pluginControls = append(pluginControls, controlID)
+		r.handlerRegistry.RegisterLocked(fakeID, r.pluginControlHandler)
```
paulbellamy: What's wrong with just calling `r.handlerRegistry.Register(fakeID, r.pluginControlHandler)` here?

report/report.go
```diff
@@ -254,6 +255,57 @@ func (r Report) Validate() error {
 	return nil
 }
 
+// Upgrade returns a new report based on a report received from the old probe.
```
paulbellamy: These are really nice.

tools/generate_latest_map
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env bash
```
paulbellamy: This comment text is wrong, but in general, since we are only generating this for two types, I'd probably just include them as static files, and not do a `go generate` step.

Makefile
krnowak: I had some errors during building `report.codecgen.go` about unknown type `StringLatestMap`, so I added the dep to make sure that this file with this type is generated before.

probe/controls/controls.go
krnowak: registry backend was for storing, and retrieving handlers and registry was for actually calling the handler and having a consistent error reporting.

paulbellamy: Ok, I'd suggest just merging them, as there's only ever one implementation of a backend right now.

Makefile
paulbellamy: As discussed elsewhere, ditching the go generate step, so this shouldn't be an issue.

probe/kubernetes/controls.go
krnowak: Yeah, that's ugly. Maybe I should just add a batch function like:
```go
func (r *HandlerRegistry) Batch(toRemove []string, toAdd map[string]xfer.ControlHandlerFunc)
{ ... }
```
That would probably be much better than explicit locking and unlocking.

probe/plugins/registry.go
krnowak: I think I wanted to avoid races between a goroutine that removes all the handlers of a specific plugin and the goroutine handling the control request from the plugin` like in the following scenario:

- goroutine 1:
  - plugin registry receives the report from the plugin `X`
  - plugin registry removes all the control handlers specific to the plugin `X`
  - gets preempted
- goroutine2:
  - got a control request with control ID `X~foo`
  - tries gets a handler for this control, but all the handlers were removed
  - control request fails
  - gets preempted
- goroutine 1:
  - add new handlers to the handler registry for the plugin `X` controls.

So I see two solutions here:

- Batch method for handler registry like you and I suggested in the other comment.
- Do a symmetric diff of the set of already registered handlers and the set of handlers to be registered (or do two relative complements). Then remove handlers that should be removed and add handlers that should be added.

I think I'd prefer the first solution, it's simpler.

tools/generate_latest_map
krnowak: Will do.

report/report.go
krnowak: Thanks.

probe/controls/controls.go
krnowak: Also, with the batch function, they won't be so similar anymore.

app/collector.go
krnowak: Oh yeah, that one. I noticed that `c.cached` was never set to anything but `nil`. I guess it was a bug…

probe/controls/controls.go
krnowak: There is also one implementation in tests to make sure that the proper handlers were registered.

probe/plugins/registry.go
paulbellamy: agreed

krnowak:
Updated. I added the generated file to git. I replaced the `Lock`, `Unlock`, `RegisterLocked` and `RmLocked` functions in handler registry with the `Batch` function. I kept the handler registry and the backend separate as we have the default implementation for registry and a test implementation used in plugins tests. I also made some small fixes in plugins registry for bugs I spotted when updating it to use the `Batch` function - namely to make sure we remove control request handlers for plugins that are closed/removed.

For lint fixes, please see https://github.com/weaveworks/scope/pull/1751.

krnowak:
Updated, CircleCI is finally green.

krnowak:
@paulbellamy: Is there anything else I should address in this PR for it to be merged?

paulbellamy:
@krnowak I was just going to ask you the same. If you're done, I will give it a final review tomorrow (Aug 12).

krnowak:
Cool. I'm done, so I'm waiting for a review.

krnowak:
Updated to fix the merge conflict.

paulbellamy:
LGTM

PR #1684, opened 2016-07-15
Add options for storing memcached reports with different compression levels

### LINKED ISSUE #1671: Add flag to control whether we store memcache reports as compressed or not
We have a suspicion that decompressing reports is a large source of latency in our production scope. We should try to falsify this suspicion by storing decompressed reports. To control this, we should have a flag.

jml:
Also factor out functions for calculating the keys we use for storage. Not intrinsically related to this change, but helped me understand what was going on.

Note that now if we have the same compression level between memcache & s3, we'll duplicate work.

Fixes #1671

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1684)

### DIFF
--- a/app/multitenant/aws_collector.go
+++ b/app/multitenant/aws_collector.go
@@ -2,6 +2,7 @@ package multitenant
 
 import (
 	"bytes"
+	"compress/gzip"
 	"crypto/md5"
 	"fmt"
 	"io"
@@ -346,7 +347,7 @@ func (c *awsCollector) Add(ctx context.Context, rep report.Report) error {
 
 	// first, encode the report into a buffer and record its size
 	var buf bytes.Buffer
-	rep.WriteBinary(&buf)
+	rep.WriteBinary(&buf, gzip.BestCompression)
 	reportSizeHistogram.Observe(float64(buf.Len()))
 
 	// second, put the report on s3
--- a/app/multitenant/aws_collector.go
+++ b/app/multitenant/aws_collector.go
@@ -368,30 +368,30 @@ func (c *awsCollector) Add(ctx context.Context, rep report.Report) error {
 
 	// second, put the report on s3
 	rowKey, colKey := calculateDynamoKeys(userid, time.Now())
-	s3Key, err := calculateReportKey(rowKey, colKey)
+	reportKey, err := calculateReportKey(rowKey, colKey)
 	if err != nil {
 		return err
 	}
 
-	err = c.s3.StoreBytes(s3Key, buf.Bytes())
+	err = c.s3.StoreBytes(reportKey, buf.Bytes())
 	if err != nil {
 		return err
 	}
 
 	// third, put it in memcache
 	if c.memcache != nil {
-		err = c.memcache.StoreBytes(s3Key, buf.Bytes())
+		err = c.memcache.StoreBytes(reportKey, buf.Bytes())
 		if err != nil {
 			// NOTE: We don't abort here because failing to store in memcache
 			// doesn't actually break anything else -- it's just an
 			// optimization.
-			log.Warningf("Could not store %v in memcache: %v", s3Key, err)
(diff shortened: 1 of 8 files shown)


app/multitenant/aws_collector.go
```diff
@@ -338,45 +337,55 @@ func (c *awsCollector) Report(ctx context.Context) (report.Report, error) {
...
 
+// calculateDynamoKeys generates the row & column keys for Dynamo.
+func calculateDynamoKeys(userid string, now time.Time) (string, string) {
+	rowKey := fmt.Sprintf("%s-%s", userid, strconv.FormatInt(now.UnixNano()/time.Hour.Nanoseconds(), 10))
+	colKey := strconv.FormatInt(now.UnixNano(), 10)
+	return rowKey, colKey
+}
+
+// calculateReportKey determines the key we should use for a report.
+func calculateReportKey(rowKey, colKey string) (string, error) {
```
2opremio: Why not call it `calculateS3Key` ?

2opremio: (It would had helped me understand it)

2opremio: `calculateReportS3Key` if you will ...

report/marshal_test.go
```diff
@@ -0,0 +1,49 @@
+package report_test
```
2opremio: Nice!

app/multitenant/s3_client.go
```diff
@@ -84,6 +85,16 @@ func (store *S3Store) fetchReport(key string) (*report.Report, error) {
...
+func (store *S3Store) StoreReport(key string, report *report.Report) (int, error) {
+	var buf bytes.Buffer
+	report.WriteBinary(&buf, gzip.BestCompression)
+	err := store.StoreBytes(key, buf.Bytes())
+	return buf.Len(), err
+}
+
 // StoreBytes stores a report in S3, expecting the report to be serialized
 // already.
 func (store *S3Store) StoreBytes(key string, content []byte) error {
```
2opremio: Is this used somewhere other than StoreReport? Otherwise I would inline StoreBytes in StoreReport.

app/multitenant/memcache_client.go
```diff
@@ -201,6 +204,14 @@ func (c *MemcacheClient) FetchReports(keys []string) (map[string]report.Report,
...
+// StoreReport serializes and stores a report.
+func (c *MemcacheClient) StoreReport(key string, report *report.Report) (int, error) {
+	var buf bytes.Buffer
+	report.WriteBinary(&buf, c.compressionLevel)
+	err := c.StoreBytes(key, buf.Bytes())
+	return buf.Len(), err
+}
+
 // StoreBytes stores a report, expecting the report to be serialized already.
 func (c *MemcacheClient) StoreBytes(key string, content []byte) error {
```
2opremio: Same here

app/multitenant/aws_collector.go
jml: Because everywhere else, we call it a "report key". Further, the key is used for memcache as well as S3. 

If still you think "s3 key" would be better after this explanation, then let me know & I'll change it.

report/marshal_test.go
jml: Thanks.

app/multitenant/s3_client.go
jml: Fixed.

app/multitenant/memcache_client.go
jml: Fixed.

jml:
Thanks for the review @2opremio! PTAL.

app/multitenant/aws_collector.go
2opremio: Fair enough

2opremio:
LGTM

PR #1685, opened 2016-07-15
Update version of build tools

jml:
Should result in slightly faster lint checks and less noisy test output.

---
This change is [Reviewable](https://reviewable.io/reviews/weaveworks/scope/1685)


2opremio:
LGTM (provided green). It might be good to get @tomwilkie to check if the subtree-stuff is done properly though, since I am not very familiar with it.

tomwilkie:
Yeah looks good.

PR #1693, opened 2016-07-17
reduce horizontal gap between nodes in topology views

### LINKED ISSUE #1688: topology layout consumes too much horizontal space
I've looked at various systems with scope in the past few days and in many of them not all the nodes fit on the screen at a readable zoom level, due to too much horizontal space being consumed. https://cloud.weave.works/demo is one example.

To my eyes, there is too much horizontal space being wasted - there is too much space between nodes, more than in the vertical dimension. I reckon it could be reduced by ~30% to about the same as the vertical spacing.

rade:
Fixes #1688

### DIFF
--- a/client/app/scripts/charts/node.js
+++ b/client/app/scripts/charts/node.js
@@ -85,7 +85,7 @@ class Node extends React.Component {
     const color = getNodeColor(rank, label, pseudo);
     const truncate = !focused && !hovered;
     const labelTransform = focused ? `scale(${1 / zoomScale})` : '';
-    const labelWidth = nodeScale(scaleFactor * 4);
+    const labelWidth = nodeScale(scaleFactor * 3);
     const labelOffsetX = -labelWidth / 2;
     const labelDy = (showingNetworks && networks) ? 0.75 : 0.60;
     const labelOffsetY = focused ? nodeScale(labelDy) : nodeScale(labelDy * scaleFactor);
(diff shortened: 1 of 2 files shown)


rade:
btw, I don't pretend to know what I am doing here. And the fact that I had to adjust labelWidth to prevent overlapping makes me uneasy - I would have expected that to be adjusted automatically with the node separation factor.

davkal:
It all comes down to the search results being displayed. Sadly, browser support for controlling their layout (`foreignObject` in SVG) is limited. So you end up with blocks that are wider than they need to be, and assume full width. At that point they are too close together with the neighbour nodes, see lower row here:

screen shot 2016-07-20 at 09 22 46

For the row separation, the more-matches text is blocked:

screen shot 2016-07-20 at 09 26 21

Here is the report I'm using: https://gist.github.com/davkal/b02d9e6c90b73fd39e7d9fbaec11815e

rade:
>  At that point they are too close together with the neighbour nodes

I had put a screen shot just like that in #1688. I don't see it as a problem. Yes, more separation would be better, but increasing horizontal node space to do that is worse.

> For the row separation, the more-matches text is blocked:

That's about vertical spacing, surely. How does that have anything to do with this change?

rade:
> That's about vertical spacing, surely. How does that have anything to do with this change?

And the answer is: nothing. I get the same (and more) overlap on our dev admin scope.

davkal:
You're right. I thought rank separation was changed here too, but it wasnt.
The node separation LGTM.

foot:
This LGTM, works nicely on the wide layouts.

PR #1694, opened 2016-07-17
hide uncontained/unmanaged by default

### LINKED ISSUE #1689: unmangaged/uncontained clutter view by default
By default we show uncontained/unmanaged containers in the topology. In many cases this clutters the view and makes for a bad first impression. See [our public demo](https://cloud.weave.works/demo/) for an example.

Hence I reckon we should hide unmanaged/uncontained by default.

rade:
They are of no interest to most users and affect the initial user experience.

Fixes #1689.


tomwilkie:
LGTM

PR #1708, opened 2016-07-21
show names of stopped containers

### LINKED ISSUE #1691: stopped containers not shown with their names
Run

    docker create --name=foo -ti alpine /bin/sh

and look at the stopped container in Scope. It is shown without the name.

This used to work not so long ago since I did my dockercon demo with stopped containers as the starting point.

rade:
This got broken in b78c503

Fixes #1691


2opremio:
Thanks for the fix! LGTM (provided circle gets green)

PR #1723, opened 2016-07-26
Name our routes, so /metrics gives more sensible aggregations

paulbellamy:
Previously was showing stuff like:
```
{route="api_pipe_pipe_8999861204239855606_probe"}   2
{route="api_pipe_pipe_2001043336351237495_check"}   1
{route="api_pipe_pipe_5613831799035630055"}
```

### DIFF
--- a/app/controls.go
+++ b/app/controls.go
@@ -14,9 +14,14 @@ import (
 
 // RegisterControlRoutes registers the various control routes with a http mux.
 func RegisterControlRoutes(router *mux.Router, cr ControlRouter) {
-	router.Methods("GET").Path("/api/control/ws").
+	router.
+		Methods("GET").
+		Path("/api/control/ws").
 		HandlerFunc(requestContextDecorator(handleProbeWS(cr)))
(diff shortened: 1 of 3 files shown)


app/controls.go
```diff
@@ -14,9 +14,14 @@ import (
 
 // RegisterControlRoutes registers the various control routes with a http mux.
 func RegisterControlRoutes(router *mux.Router, cr ControlRouter) {
-	router.Methods("GET").Path("/api/control/ws").
+	router.
+		Methods("GET").
+		Path("/api/control/ws").
 		HandlerFunc(requestContextDecorator(handleProbeWS(cr)))
```
jml: Should this also have a `Name` call?

jml:
One question, otherwise LGTM.

app/controls.go
paulbellamy: It has no dynamic segments to the path, so the instrumentation will infer it automatically.

jml: Thanks.

jml:
LGTM.

PR #1728, opened 2016-07-28
don't Copy() self on Merge()

rade:
Merge() is always returning a copy, so there is no need to Copy() struct fields first before calling Merge() on them.

This reduces GC pressure (#1010) and helps overall app performance (#1457).

### DIFF
--- a/report/metadata_template.go
+++ b/report/metadata_template.go
@@ -124,11 +124,14 @@ func (e MetadataTemplates) Copy() MetadataTemplates {
 // Merge merges two sets of MetadataTemplates so far just ignores based
 // on duplicate id key
 func (e MetadataTemplates) Merge(other MetadataTemplates) MetadataTemplates {
-	result := e.Copy()
+	if e == nil && other == nil {
+		return nil
+	}
+	result := make(MetadataTemplates, len(e))
+	for k, v := range e {
+		result[k] = v
+	}
 	for k, v := range other {
-		if result == nil {
-			result = MetadataTemplates{}
-		}
 		if existing, ok := result[k]; !ok || existing.Priority < v.Priority {
 			result[k] = v
 		}
--- a/report/metric_template.go
+++ b/report/metric_template.go
@@ -67,11 +67,14 @@ func (e MetricTemplates) Copy() MetricTemplates {
 // Merge merges two sets of MetricTemplates so far just ignores based
 // on duplicate id key
 func (e MetricTemplates) Merge(other MetricTemplates) MetricTemplates {
-	result := e.Copy()
+	if e == nil && other == nil {
+		return nil
+	}
+	result := make(MetricTemplates, len(e))
+	for k, v := range e {
+		result[k] = v
+	}
 	for k, v := range other {
-		if result == nil {
-			result = MetricTemplates{}
-		}
 		if existing, ok := result[k]; !ok || existing.Priority < v.Priority {
 			result[k] = v
 		}
--- a/report/node.go
+++ b/report/node.go
@@ -194,23 +194,27 @@ func (n Node) Copy() Node {
 // Merge mergses the individual components of a node and returns a
 // fresh node.
 func (n Node) Merge(other Node) Node {
-	cp := n.Copy()
-	if cp.ID == "" {
-		cp.ID = other.ID
+	id := n.ID
+	if id == "" {
+		id = other.ID
 	}
-	if cp.Topology == "" {
-		cp.Topology = other.Topology
-	} else if other.Topology != "" && cp.Topology != other.Topology {
-		panic("Cannot merge nodes with different topology types: " + cp.Topology + " != " + other.Topology)
+	topology := n.Topology
+	if topology == "" {
+		topology = other.Topology
+	} else if other.Topology != "" && topology != other.Topology {
+		panic("Cannot merge nodes with different topology types: " + topology + " != " + other.Topology)
 	}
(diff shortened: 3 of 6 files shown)


rade:
@2opremio I *think* my interpretation of the intended (and actual) Merge() semantics is correct, but I defer to your superior knowledge of scope internals.

I have tested this by `scope launch`ing the app on my desktop (~300 processes, though few with connections), while having ~20 connected containers, pointing my browser at the containers view, and monitoring the app's CPU usage with pidstat. This showed a reduction of about 20%.

rade:
Nodes, Metrics, EdgeMetadata, MetadataTemplates could potentially also benefit from the same treatment, but are less straightforward than Report and Node.

rade:
Is Merge() required to create deep copies or is it simply required not to modify self or other?

My code in this PR, in common with a bunch of other the Merge() code I have come across, meets the former criteria. But there is also code which doesn't do that. For example Nodes.Merge() only copies self but not other (in particular elements from other that are not present in self are not copied), and so only meets the latter criteria.

If only the latter criteria is required, a whole bunch of easy, high-value optimisations are possible.

rade:
> Is Merge() required to create deep copies or is it simply required not to modify self or other?

@paulbellamy says it's the second. Which is just as well since otherwise some of the existing code would be wrong :)

2opremio:
Thanks for this @rade! Calling Copy() inside Merge() at each level of the data structure means that objects were being copied `depth - 1` times. (Not completely accurate since some levels like Topology didn't do this).

> Is Merge() required to create deep copies or is it simply required not to modify self or other?

Unless I am missing something important, I believe Merge() is simply required not to modify self or other. In fact, as I mentioned above `Topology` already 

Also, I believe that the benefit of this would be lower if all the data-structures where immutable internally (e.g. Metrics is still mutable) making copies essentially for free.

> Nodes, Metrics, EdgeMetadata, MetadataTemplates could potentially also benefit from the same treatment, but are less straightforward than Report and Node.

I don't immediately see how you can avoid copies with things like Metrics (based on a mutable map).

2opremio:
> @paulbellamy says it's the second. Which is just as well since otherwise some of the existing code would be wrong :)

Great, then we agree.

rade:
> I don't immediately see how you can avoid copies with things like Metrics (based on a mutable map).

It's the *deep* copy that is avoidable.

2opremio:
> It's the deep copy that is avoidable.

But Metrics is a leaf of the data structure. The only optimization I see is using `make` to preallocate objects.

rade:
> But Metrics is a leaf of the data structure.

Sorry, I meant MetricTemplates.

2opremio:
> Sorry, I meant MetricTemplates.

Oh, I see. All good.

report/metadata_template.go
```diff
@@ -124,11 +124,14 @@ func (e MetadataTemplates) Copy() MetadataTemplates {
 // Merge merges two sets of MetadataTemplates so far just ignores based
 // on duplicate id key
 func (e MetadataTemplates) Merge(other MetadataTemplates) MetadataTemplates {
-	result := e.Copy()
+	if e == nil && other == nil {
+		return nil
+	}
+	result := make(MetadataTemplates, len(e))
```
2opremio: A similar trick can be applied in Metrics. Also, I think we can afford to do `len(e)+len(other)` (or at least the maximum)

report/metadata_template.go
```diff
@@ -124,11 +124,14 @@ func (e MetadataTemplates) Copy() MetadataTemplates {
...
+	for k, v := range e {
+		result[k] = v
+	}
 	for k, v := range other {
```
2opremio: As a further optimization, this iteration should be done over the shortest map.

2opremio: (It will reduce the number of inner checks)

rade: I don't think that is worth the extra code.

report/metadata_template.go
rade: I don't think + would be a good idea - won't the keys be the same or nearly the same in most cases? As for max... if only golang didn't make me write my own function for that. Also this mini optimisation would apply in more places than the ones I changed. So I'd rather leave it for now.

2opremio: Then, this initialization and loop below are equivalent to `result := e.Copy` after adding `make(MetadataTemplates, len(e))` to `Copy()`.

report/metadata_template.go
2opremio: You pretty much get it for free if the other optimization gets implement it :)

report/metadata_template.go
rade: > Then, this initialization and loop below are equivalent to `result := e.Copy()`

No, they are note, because `Copy()` does a *deep* copy.

2opremio: True, sorry.

paulbellamy: Doesn't `len(e)+len(other)` assume there is no overlap?

2opremio: > Doesn't len(e)+len(other) assume there is no overlap?

Yes, so it will consume unnecessary memory if there is overlap but it will be optimal in terms of objects allocated (reducing garbage collector pressure).

2opremio:
Note (discussed offline) See if we can get rid of all the `Copy()` functions.

2opremio:
LGTM

PR #1734, opened 2016-08-01
Dont show weavescope logo when running in a frame

davkal:
The logo wont show when running in a service.

Approach A: Detect iframe (`window.top` is not `window`).

Approach B: Check if hostname is `cloud.weave.works`.


foot:
LGTM!

PR #1741, opened 2016-08-02
make StringSet.Add/Remove copy instead of modifying target

2opremio:


2opremio:
LGTM

rade:
I am concerned about the performance impact of this. AFAICT, the existing uses of `Add()` populate freshly created StringSets. Introducing copying here could prove expensive.

rade:
I have convinced myself that all the direct and indirect call sites of `StringSet.Add()` are safe.

This is not obvious, and it would be very easy to accidentally introduce an unsafe call.

So the intention of this PR is laudable. But there are definitely some call sites which are performance critical, so I am (still) worried that introducing copying in `Add()` will noticeably hurt us.

Issue #1735, opened 2016-08-01
Table-mode columns should be determined by backend.


foot:
> What's important and what can be ignored, should be determined by the backend. The current metadata field priority is not sufficient, though. Looks like we need another field.

bboreham:
What is the "priority" for?  I thought it would control the order of columns, but by inspection at #3090 the columns are in alphabetical order.

PR #1748, opened 2016-08-03
Make the port column a little wider to stop truncation

### LINKED ISSUE #1711: details panel truncates port to four digits
ports can be five digits, so truncating at four is just mean

screenshot_2016-07-21_19-47-26

foot:
For firefox+linux which renders slightly tighter it seems

Fixes #1711

### DIFF
--- a/client/app/scripts/components/node-details/node-details-table.js
+++ b/client/app/scripts/components/node-details/node-details-table.js
@@ -11,27 +11,30 @@ function isNumberField(field) {
 }
 
 const CW = {
-  S: '44px',
+  S: '50px',
   M: '80px',
-  L: '140px',
-  XL: '170px',
+  L: '120px',
+  XL: '140px',
+  XXL: '170px',
 };
 
 const COLUMN_WIDTHS = {
   count: '70px',
-  docker_container_created: CW.L,
+  docker_container_created: CW.XL,
   docker_container_restart_count: CW.M,
-  docker_container_state_human: CW.XL,
+  docker_container_state_human: CW.XXL,
   docker_container_uptime: '85px',
   docker_cpu_total_usage: CW.M,
   docker_memory_usage: CW.M,
   open_files_count: CW.M,
   pid: CW.M,
-  port: '44px',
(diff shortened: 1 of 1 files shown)


rade:
replicasets and deployments need the same treatment. And hosts.

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -27,11 +28,15 @@ const COLUMN_WIDTHS = {
...
   pid: CW.M,
-  port: '44px',
+  port: CW.S,
   ppid: CW.M,
   process_cpu_usage_percent: CW.M,
   process_memory_usage_bytes: CW.M,
   threads: CW.M,
+  //
+  // pods > services > details panel.
+  kubernetes_ip: '120px',
```
davkal: Why does this get a special value?

davkal:
Bar comment, code LGTM, and looks good on Mac/Chrome

@rade could you confirm the port width fix viewed on Linux?

rade:
> @rade could you confirm the port width fix viewed on Linux?

See https://github.com/weaveworks/scope/issues/1711#issuecomment-237305358 - i.e. I am no longer seeing that problem even without that change - looking at the images, it seems that the column width increased since I reported the problem.

foot:
re: replicasets, hosts and deployments this fixes it for those too, bad commit message.

rade:
> re: replicasets, hosts and deployments this fixes it for those too, bad commit message.

And comment in the code!

Are you sure this fixes *hosts*? What does that have to do with `kubernetes_ip`?

foot:
Yes! and in the code.

Fixes hosts in the sense that the only place I could find an ip in the hosts panel was in the k8s section.

rade:
> Fixes hosts in the sense that the only place I could find an ip in the hosts panel was in the k8s section.

Ah yes, good point. What an odd choice of columns - listing the IPs of pods but not containers.

foot:
More long term solution could be covered in #1735

foot:
Squashed

PR #1753, opened 2016-08-04
Move grid selector to top menu

### LINKED ISSUE #1743: Table-mode: Move selector position
Probably up to the top w/ search and topologies.

davkal:
* move to top right, to bookend topologies
* hierarchically similar to search, so similar positioning

Fixes #1743

screen shot 2016-08-04 at 14 47 12

### DIFF
--- a/client/app/scripts/components/app.js
+++ b/client/app/scripts/components/app.js
@@ -124,6 +124,7 @@ class App extends React.Component {
           </div>
           <Search />
           <Topologies />
+          <GridModeSelector />
         </div>
 
         <Nodes />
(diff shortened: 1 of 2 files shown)


rade:
visually/functionally this LGTM. Can't comment on the code.

foot:
Noticed
- obscured by details panel
- suffers a bit on small windows:
  - Nice to have a right margin
  - Style so it handles being squished into a vertical nicely or prevent that.

As we're moving to a new menu anyway and most ppl have reasonably sized windows I think this stuff is minor. 

screen shot 2016-08-04 at 16 05 23

foot:
Code LGTM

davkal:
> obscured by details panel

I think this is ok, because that does not seem to disturb a known workflow.

> suffers a bit on small windows:

Yes, it gets bad below 1024px width. But I'd punt that.

PR #1756, opened 2016-08-04
Tighter node bounding boxes

### LINKED ISSUE #1520: Highlighting is unpredictable
The hit boxes seem to be a bit off -- sometimes things will highlight when the pointer is nowhere near them, or not highlight when the pointer is right on them.

screenshot from 2016-05-18 13-16-19 (an artist's impression of the pointer position has been added)

This is an accidentally zoomed-in view, but the same problem is present when not zoomed.

Firefox 46, Ubuntu 15.10.

foot:
Fixes #1520

### DIFF
--- a/client/app/styles/main.less
+++ b/client/app/styles/main.less
@@ -165,6 +165,8 @@ h2 {
 }
 
 .header {
+  pointer-events: none;
+
   position: absolute;
   top: 32px;
   width: 100%;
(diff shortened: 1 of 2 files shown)


rade:
This does seem indeed seem to work.

client/app/styles/main.less
```diff
@@ -165,6 +165,8 @@ h2 {
 }
 
 .header {
+  pointer-events: none;
```
davkal: good one.

davkal:
LGTM

PR #1757, opened 2016-08-04
Docker for Mac 1.12 has only one network mode (close #1755)

### LINKED ISSUE #1755: Scope fails to launch on fresh Docker for Mac installs
We had a check whether Docker for Mac runs in hybrid network mode or not, it's no longer applicable and fails with the stable version.

The fix is to stop checking `$HOM/Library/Containers/com.docker.docker/Data/database/com.docker.driver.amd64-linux/network`.

errordeveloper:


errordeveloper:
I got a little confused at first, as it didn't quite fail for me, but that was because plugins dir existed in the running VM.

2opremio:
@errordeveloper So, you managed to reproduce after all?

errordeveloper:
@2opremio yes indeed.

2opremio:
@errordeveloper Can you recreate and test the PR onto the `release-0.16` branch? (we are going to cut a release just with this extra change) Thanks

errordeveloper:
@2opremio done https://github.com/weaveworks/scope/pull/1764

PR #1758, opened 2016-08-04
[WIP] Implement LatestMap as a slice of slab strings

2opremio:
- [ ] Fix remaining bugs which have been introduced e.g.:

screen shot 2016-08-04 at 20 08 25

(note the missing labels)

Attempts to improve #1010 and #1457

### DIFF
--- a/report/latest_map.go
+++ b/report/latest_map.go
@@ -5,83 +5,272 @@ import (
 	"fmt"
 	"sort"
 	"time"
+	"unsafe"
 
 	"github.com/ugorji/go/codec"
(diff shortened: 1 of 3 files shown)


2opremio:
Some other ideas:

* Implement n-way merges (in that way we only need to allocate only LatestMap for each node in the final report). This is applicable in general, regardless of the latestMap implementation.
* Adjust the slabSize to something smaller than 128.

rade:
> Implement n-way merges

See https://github.com/weaveworks/scope/issues/1010#issuecomment-233560728

PR #1764, opened 2016-08-05
Docker for Mac 1.12 has only one network mode (close #1755)

### LINKED ISSUE #1755: Scope fails to launch on fresh Docker for Mac installs

errordeveloper:


2opremio:
LGTM

Issue #1760, opened 2016-08-05
labels overlap with nodes


mindfulmonk:
Hi All,

When running more than 50 containers the graph view becomes quite unusable. The container names start overlapping and the connections between containers are hard to see.

Here is an example:
![scope_err_too_many](https://cloud.githubusercontent.com/assets/602167/17430633/33cb779a-5aed-11e6-8dea-97fc60e224e8.png)

Possible solutions to this would be the ability to tweak the fonts/icon sizes or have something more like http://gource.io/ for the graph view.

rade:
The current graph view struggles with high node/edge counts. See #1092 and #1614. That was actually one motivation for developing the table mode. For now  that is your best option.

I am not seeing the overlap in my scopes. What browser is this? And would you mind attaching the report ("</>" in bottom right)?

rade:
I reckon this is related to #1692.

rade:
#1701 is also possibly related.

rade:
@foot @davkal I know in #1773 you said that it didn't fix this issue, but have you actually been able to reproduce the above problem since then?

foot:
proud-wind w/ both app and sys containers + browser size: 1440x800 gives something similar.

There is a bit of quick fix here in allowing the fontsize to become smaller. It looks tidier but you can't read them anymore as they're too small. Then you zoom in I guess..

rade:
> There is a bit of quick fix here in allowing the fontsize to become smaller.

Sounds reasonable. @davkal what do you think?

davkal:
Longterm we should go for semantic zoom (with nodes spacing dynamically, and hiding sub labels when zoomed out, and all labels when a complete overview fills the screen). 

This particular screenshot shows overlap in the non-connected nodes group, this is a separate layout engine, maybe we can tweak the spacing for that (wont solve this issue for 100 nodes though, need the longterm fix for that). We should not make labels unreadable, rather hide them. If they're hidden, they cant overlap :)

rade:
@foot proposed the unreadable labels as a *quick fix*. And it *is* better than overlapping labels.

Are any of the proposed alternatives just as quick to implement?

davkal:
> Are any of the proposed alternatives just as quick to implement?

Yes: node spacing for unconnected nodes.

rade:
> Yes: node spacing for unconnected nodes.

[...]

PR #1773, opened 2016-08-09
Scale node labels with the node's size.

### LINKED ISSUE #1692: increasing vertical window size decreases node label font size
Showing exactly the same topology, at exactly the same zoom level / node size / node space, but with different vertical window size, changes the font size on node labels - the taller the window the smaller (paradoxically) the font size gets.

The following results were obtained by opening fresh browser windows, sizing them, and then entering the local scope url.

short window
screenshot_2016-07-17_15-58-03
tall window
screenshot from 2016-07-17 15-58-29

Happens in both Chrome and FF.

### LINKED ISSUE #1763: Details panel does not spawn at node location on click
This used to work before #1756. Now it spawns the below the screen (seen in Chrome and Firefox).
Weirdly, the origin's x coordinate still works.

foot:
- Within certain bounds. Still have min-label size, mainly effects nodes
  that get really big.
- Set a max size on nodes too, really big ones lose their border.

Fixes #1692 
Fixes #1763

### DIFF
--- a/client/app/scripts/charts/node.js
+++ b/client/app/scripts/charts/node.js
@@ -128,7 +128,7 @@ class Node extends React.Component {
           <foreignObject style={{pointerEvents: 'none'}} x={labelOffsetX} y={labelOffsetY}
             width={labelWidth} height="100em">
             <div className="node-label-wrapper"
-              style={{pointerEvents: 'all', fontSize, lineHeight: '150%', maxWidth: labelWidth}}
+              style={{pointerEvents: 'all', fontSize, maxWidth: labelWidth}}
               {...mouseEvents}>
               <div className={labelClassName}>
                 <MatchedText text={label} match={matches.get('label')} />
--- a/client/app/scripts/charts/node.js
+++ b/client/app/scripts/charts/node.js
@@ -62,6 +62,7 @@ class Node extends React.Component {
     this.handleMouseClick = this.handleMouseClick.bind(this);
     this.handleMouseEnter = this.handleMouseEnter.bind(this);
     this.handleMouseLeave = this.handleMouseLeave.bind(this);
+    this.saveShapeRef = this.saveShapeRef.bind(this);
     this.state = {
       hovered: false,
       matched: false
@@ -139,7 +140,7 @@ class Node extends React.Component {
             </div>
           </foreignObject>}
 
-        <g {...mouseEvents}>
+        <g {...mouseEvents} ref={this.saveShapeRef}>
           <NodeShapeType
             size={size}
             color={color}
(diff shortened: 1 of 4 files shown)


foot:
Nodes prior to max size restriction

localhost-4042-

foot:
Max size:

localhost-4042- 1

foot:
That is for the base zoom level.

Zooming you can decrease/increase still.

foot:
Hint: chrome devtools device-toolbar is handy for testing window sizes.

rade:
This does appear to work. I have no idea how. Hopefully @davkal does.

davkal:
This fix prevents the bad label overlaps, great!

Two visual issues:

1. select on node, look at label size, deselect, zoom out, select same node, label size is now bigger

2. when selecting a node, it shortly blows up in size, then the size converges to its end state

davkal:
Code looks good.
At some point we should implement a proper semantic zoom, that spaces out the nodes a bit more and allows labels to expand.

foot:
screen shot 2016-08-09 at 18 25 20

Looks kinda fun

foot:
Ready for re-review

rade:
So does this now also fix #1763?

foot:
Yep, issue updated.

client/app/scripts/charts/node.js
```diff
@@ -126,9 +125,9 @@ class Node extends React.Component {
           svgLabels(label, subLabel, labelClassName, subLabelClassName, labelOffsetY) :
 
           <foreignObject style={{pointerEvents: 'none'}} x={labelOffsetX} y={labelOffsetY}
-            width={labelWidth} height="100em" transform={labelTransform}>
+            width={labelWidth} height="100em">
             <div className="node-label-wrapper"
-              style={{pointerEvents: 'all', fontSize, maxWidth: labelWidth}}
+              style={{pointerEvents: 'all', fontSize, lineHeight: '150%', maxWidth: labelWidth}}
```
davkal: Can we put static numbers into the .less file?

client/app/styles/main.less
```diff
@@ -546,7 +546,7 @@ h2 {
   text-align: center;
 
   &-match {
-    font-size: 0.8rem;
+    font-size: 0.8em;
```
davkal: Why not rem?

client/app/scripts/charts/node.js
foot: Yes, this didn't hop across the foreignObject properly I think I can put in the less file though.

davkal:
LGTM

foot:
This issue fixes label truncation (ellipsis) too, but I don't think thats the same issue as #1760

rade:
> This issue fixes label truncation (ellipsis) too

What exactly is it fixing? Have you got a before/after screenshot?

foot:
screen shot 2016-08-10 at 14 17 37
screen shot 2016-08-10 at 14 17 27

rade:
oh ok. I hadn't actually noticed that. What does one have to do to provoke this? I've played with zoom and +/- to no avail.

foot:
Nothing in particular just have nice long container names, its noticeable for me on proud-wind.

rade:
got it. needed to shrink my window vertically to see the problem.

PR #1774, opened 2016-08-09
Remove kubernetes_id column from the grid view.

### LINKED ISSUE #1759: Table-mode: do we need the k8s id columns?
All the k8s views have an ID column. But AFAICT it's just namespace/name, i.e. the composition of two fields we already show. So can we ditch this?

Somebody with knowledge of k8s and what these fields actually mean needs to chime in here please.

foot:
Redundant info

Fixes #1759

### DIFF
--- a/probe/kubernetes/meta.go
+++ b/probe/kubernetes/meta.go
@@ -10,7 +10,6 @@ import (
 
 // These constants are keys used in node metadata
 const (
-	ID          = "kubernetes_id"
 	Name        = "kubernetes_name"
 	Namespace   = "kubernetes_namespace"
 	Created     = "kubernetes_created"
@@ -20,7 +19,6 @@ const (
 // Meta represents a metadata information about a Kubernetes object
 type Meta interface {
 	UID() string
-	ID() string
 	Name() string
 	Namespace() string
 	Created() string
@@ -34,10 +32,6 @@ type meta struct {
 
 func (m meta) UID() string {
 	return string(m.ObjectMeta.UID)
-}
-
-func (m meta) ID() string {
-	return m.ObjectMeta.Namespace + "/" + m.ObjectMeta.Name
 }
 
 func (m meta) Name() string {
@@ -59,7 +53,6 @@ func (m meta) Labels() map[string]string {
 // MetaNode gets the node metadata
 func (m meta) MetaNode(id string) report.Node {
 	return report.MakeNodeWith(id, map[string]string{
-		ID:        m.ID(),
 		Name:      m.Name(),
 		Namespace: m.Namespace(),
 		Created:   m.Created(),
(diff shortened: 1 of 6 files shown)


davkal:
Wouldnt it make more sense not to include the field in the backend report in the first place?

foot:
Not sure, ids can often be handy though. I'll have a quick look.

davkal:
Code works.

> Not sure, ids can often be handy though. I'll have a quick look.

The original ticket said the field has useless information. So why not drop the field? 

WDYT @rade @2opremio ?

2opremio:
> WDYT @rade @2opremio ?

I think you are right and I started to completely remove the ID (namespace/resource_name) ... but then I realized we are also using them for the ranks. I could keep it just for the ranks but I think it would be fine to simply provide the namespace instead (nodes with the same namespace seem to be getting the same color anyways).

@davkal Can you check if this change would be acceptable?

rade:
Let's please not make #783 worse. See also https://github.com/weaveworks/scope/pull/1567#issuecomment-225641062

davkal:
You can keep the ID, we simply dont need the *field* `kubernetes_id` in the nodes' metadata.

2opremio:
> Let's please not make #783 worse

That's what I am asking, it seems we are only using the three first letters of the rank anyways.

> You can keep the ID, we simply dont need the *field* `kubernetes_id` in the nodes' metadata.

I know I can keep it, but it would be nice to completely remove it if it doesn't make a difference (if it makes a difference then I can simply reconstruct the ID with the namespace and the name in the app without needing to move the ID value around). From a quick look at the JS code from @paulbellamy it seems that it doesn't but I would like to get a confirmation.

rade:
>  it seems we are only using the three first letters of the rank anyways.

Right now we are, but that must change.

2opremio:
> Right now we are, but that must change.

Fair enough, then I will simply generate the exact same ranks in the app (concatenating the namespace and name), without requiring to pass the derivate ID from probe to app.

davkal:
A change here should not affect rank values, whatever it's used for. This PR should only result in the `kubernetes_id` field not showing up for nodes. It's being read from a node's metadata object, so instead of filtering it from there in JS, I suggested of removing it from the metadata.

rade:
agreed. ditch `kubernetes_id` but leave ID well alone - if we want to make changes there let's do that in a separate issue/PR.

2opremio:
> agreed. ditch `kubernetes_id` but leave ID well alone - if we want to make changes there let's do that in a separate issue/PR.

I already did remove it, but the ID was only carried internally between probe and app (so it shouldn't impact the UI)

2opremio:
I reverted the change in the UI and stopped reporting the ID from the probe.

paulbellamy:
LGTM

PR #1775, opened 2016-08-09
Adds support for filtering graph/table by relatives

### LINKED ISSUE #1749: Table-mode: 'container' table does not filter by image name
It should be clear how user can filter things out in table view, e.g. by image name.

foot:
Now that they are available in the summary data.

Fixes #1749


davkal:
LGTM

PR #1780, opened 2016-08-10
Handle loopback addresses correctly when tracking connections

### LINKED ISSUE #1733: connections inside a container shown as going between containers
The container view shows an incorrect link between Rabbit and Riak. 
image

Both services do use Erlang but they are independent containers that are otherwise unaware of each other.
image

2opremio:
Fixes #1733 

Also: revert Endpoint node id cache, since it wasn't doing anything (the code was buggy and it was complicating things)

### DIFF
--- a/probe/endpoint/reporter.go
+++ b/probe/endpoint/reporter.go
@@ -132,7 +132,7 @@ func (r *Reporter) Report() (report.Report, error) {
 			}
 
 			seenTuples[tuple.key()] = tuple
-			r.addConnection(&rpt, tuple, extraNodeInfo, extraNodeInfo)
+			r.addConnection(&rpt, tuple, "", extraNodeInfo, extraNodeInfo)
 		})
 	}
 
@@ -143,7 +143,8 @@ func (r *Reporter) Report() (report.Report, error) {
 		}
 		for conn := conns.Next(); conn != nil; conn = conns.Next() {
 			var (
-				tuple = fourTuple{
+				namespaceID string
+				tuple       = fourTuple{
 					conn.LocalAddress.String(),
 					conn.RemoteAddress.String(),
 					conn.LocalPort,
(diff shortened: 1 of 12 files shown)


probe/endpoint/reporter.go
```diff
@@ -157,6 +158,10 @@ func (r *Reporter) Report() (report.Report, error) {
 				fromNodeInfo[report.HostNodeID] = hostNodeID
 			}
 
+			if conn.Proc.NetNamespaceID > 0 {
```
paulbellamy: I'm assuming that net namespaces are numbered same as pids, starting at 1?

render/container.go
```diff
@@ -264,6 +264,11 @@ func MapContainer2IP(m report.Node) []string {
 			if !ok {
 				continue
 			}
+			// loopback addresses are shared among all namespaces
+			// so we can't use them to attribute connections to a container
+			if report.IsLoopback(addr) {
+				continue
+			}
```
paulbellamy: won't this exclude a container connecting to itself, due to the scoped loopback interface?

Also, `host1;127.0.0.1;8080` maps to a specific pid, right? because it has the host, ip, and port?

probe/endpoint/reporter.go
2opremio: Good point. They are inodes, I am not sure of an inode with value 0 is acceptable, but we can make this conditional to the PID being non-zero.

2opremio: Actually, it seems inodes cannot be zero: http://stackoverflow.com/questions/2099121/why-do-inode-numbers-start-from-1-and-not-0

paulbellamy: ok, just checking, thanks.

render/container.go
2opremio: This is only for short-lived connections (obtained through conntrack and not proc) in which there won't be a PID (and by extension we can't either provide a networking namespace ID either).

In those cases `127.0.0.1:8080` could be the listening socket of **any** container with a loopback connection.

2opremio: To elaborate a bit further, it will still work when PIDs are provided. For instance, when reproducing the bug with the method suggested by @rade on https://github.com/weaveworks/scope/issues/1733#issuecomment-236395353 I get:

screen shot 2016-08-10 at 11 38 42 am

and after the fix:

screen shot 2016-08-10 at 11 31 57 am

Note how loops are still detected when using persistent connections.

paulbellamy:
LGTM

PR #1782, opened 2016-08-11
Elide service-token when logging commandline arguments

### LINKED ISSUE #1779: Do not log service tokens
We log all of Scope's comnand line arguments ... including the service token and users share logs to diagnose problems, which is a horrible combination (security wise).

2opremio:
Fixes #1779

### DIFF
--- a/prog/main.go
+++ b/prog/main.go
@@ -17,9 +17,14 @@ import (
 	"github.com/weaveworks/weave/common"
 )
 
-var version = "dev" // set at build time
-
-const tokenFlag = "service-token"
+var (
+	// set at build time
+	version = "dev"
+	// tokens to be elided when logging
+	serviceTokenFlag = "service-token"
+	probeTokenFlag   = "probe.token"
+	sensitiveFlags   = []string{serviceTokenFlag, probeTokenFlag}
+)
 
 type prefixFormatter struct {
 	prefix []byte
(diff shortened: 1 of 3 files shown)


prog/main.go
```diff
@@ -119,6 +121,25 @@ type appFlags struct {
 	consulInf       string
 }
 
+func logCensoredArgs() {
+	var prettyPrintedArgs string
+	// we show the args followed by the flags which is likely to change the
```
rade: surely it would make more sense to show it the other way round - flags precede args, don't they?

prog/main.go
```diff
@@ -145,7 +166,7 @@ func main() {
 	flag.Bool("no-probe", false, "Don't run the probe.")
 
 	// Probe flags
-	flag.StringVar(&flags.probe.token, "service-token", "", "Token to use to authenticate with cloud.weave.works")
+	flag.StringVar(&flags.probe.token, tokenFlag, "", "Token to use to authenticate with cloud.weave.works")
 	flag.StringVar(&flags.probe.token, "probe.token", "", "Token to use to authenticate with cloud.weave.works")
```
rade: shouldn't "probe.token" be elided too?

prog/main.go
2opremio: I did it this way because we use a launch subcommand. But ... in hindsight it doesn't help because the launch command is translated into a `--mode` flag anyways. I will change it.

prog/main.go
2opremio: very true

rade:
LGTM

PR #1783, opened 2016-08-11
Don't complain when stopping Scope if it wasn't running

### LINKED ISSUE #1777: `scope stop` on Docker-for-Mac complains about "Weave Scope app is not running."
When launching scope in probe-only mode, e.g. by specifying `--service-token=...`, a subsequent `scope stop` results in a 
```
Weave Scope app is not running.
```
complaint on stderr.

@abuehrle tripped over this, and it's likely something that will confuse any new user.

2opremio:
Fixes #1777

### DIFF
--- a/scope
+++ b/scope
@@ -239,12 +239,12 @@ EOF
 
     stop)
         [ $# -eq 0 ] || usage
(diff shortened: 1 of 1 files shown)


rade:
LGTM

PR #1788, opened 2016-08-15
fix MemcacheClient.FetchReports miss & leak on corrupt report

rade:
Problem: Decoding a corrupt report grows the 'missing' list. Since we are waiting for 'len(keys)-len(missing)' decoder go-routines, this results in waiting for fewer go-routines than we should. The surplus go-routines leak and we ignore their reports. And since the keys of the ignored reports are not included in 'missing', we won't attempt to fetch them from S3 either. Oops.

Fix: calculate the number of go-routines once, at the beginning.


jml:
Good catch :(

We need tests. If you'd like to add them, please do, and bounce back to me after. If not, LGTM.

Issue #897, opened 2016-02-01
Nodes appear and disappear intermittently


2opremio:
While testing #889 (b103f93) with the [ECS demo](http://www.weave.works/guides/service-discovery-and-load-balancing-with-weave-on-amazon-ecs-2/), after letting it run for a few hours I see containers appearing a disappearing intermittently.

There are 3 httpserver containers and 3 dataproducer containers which should appear at the same time in the UI but they come and go erratically:

![](http://g.recordit.co/yMt2SDF59I.gif)

Here's the report, although I am not sure it will help here: https://gist.github.com/2opremio/b3b4f435b568fb836306

I don't see a particularly high CPU or memory consumption but the UI is sloppy and loading the topologies takes a considerable time when loading the UI.

The ECS demo uses very small AWS instances (t2.micro) but Scope works just fine after it's freshly spawned.

2opremio:
Related: #869 #827 (and maybe #854)

paulbellamy:
Possibly the probes on different nodes are being slow and missing their deadlines.

tomwilkie:
Paul can't reproduce.

tomwilkie:
@2opremio please try and reproduce.

2opremio:
I can reproduce.

> Possibly the probes on different hosts are being slow and missing their deadlines.

I think @paulbellamy is right.

Looking at the logs I find a lot of:

```
<probe> WARN: 2016/04/25 16:25:22.817941 Topology tagger took longer than 1s
<probe> WARN: 2016/04/25 16:25:37.092299 Endpoint reporter took longer than 1s
<probe> WARN: 2016/04/25 16:25:38.651785 Docker reporter took longer than 1s
<probe> WARN: 2016/04/25 16:25:42.024260 Topology tagger took longer than 1s
<probe> WARN: 2016/04/25 16:25:55.008294 Endpoint reporter took longer than 1s
<probe> WARN: 2016/04/25 16:25:58.244352 Topology tagger took longer than 1s
<probe> WARN: 2016/04/25 16:26:11.398667 Endpoint reporter took longer than 1s
<probe> WARN: 2016/04/25 16:21:15.602084 docker container: dropping stats.
<app> ERRO: 2016/04/25 16:21:16.178981 Error on websocket: websocket: close 1006 (abnormal closure): unexpected EOF
```

Full logs: [logs.txt.gz](https://github.com/weaveworks/scope/files/234948/logs.txt.gz)

[...]

PR #1791, opened 2016-08-15
Release 0.17.0

2opremio:


CHANGELOG.md
```diff
@@ -1,3 +1,87 @@
...
+- Tabular view.
+	[#1673](https://github.com/weaveworks/scope/pull/1673)
+	[#1747](https://github.com/weaveworks/scope/pull/1747)	
+	[#1753](https://github.com/weaveworks/scope/pull/1753)
+	[#1774](https://github.com/weaveworks/scope/pull/1774)
+	[#1775](https://github.com/weaveworks/scope/pull/1775)
+	[#1784](https://github.com/weaveworks/scope/pull/1784)
+- Loading indicator
+	[#1485](https://github.com/weaveworks/scope/pull/1485)
+- Dont show weavescope logo when running in a frame
```
paulbellamy: Don't

CHANGELOG.md
```diff
@@ -1,3 +1,95 @@
+## Release 0.17.0
+
+Highlights:
+- New Grid view as an alternative to Scopes classic graph view.  It provides
```
jml: s/Scopes/Scope's/

CHANGELOG.md
```diff
@@ -1,3 +1,95 @@
...
+  higher information density, proving particularly useful when there's a high
+  node count in the graph view.
```
jml: "there's a high node count" is pretty inside baseball. "there are many nodes" might be more friendly.

CHANGELOG.md
```diff
@@ -1,3 +1,95 @@
...
+- Considerable Performance enhancements: the CPU efficiency of the Scope App has
```
jml: "Performance" should be lower case.

CHANGELOG.md
rade: It's not a Grid, and it's not a View. "Table Mode".

CHANGELOG.md
```diff
@@ -1,3 +1,95 @@
...
+  increased in more than 50% and the Scope probes over 25%.
+
+
+New features and enhancements:
+- Grid view.
```
rade: Table mode.

2opremio:
All good from me. I tested on: 

* The ECS guide: Things are looking good and CPU consumption is ~10% for the app and 6% for the probes (less than 25% of the numbers observed in https://github.com/weaveworks/scope/issues/897). I will let it run for a bit but I think that means we can close https://github.com/weaveworks/scope/issues/897 and https://github.com/weaveworks/guides/issues/181

* Connecting a probe to Weave Cloud: Things are rendered as expected and I don't see any backwards compatibility problems.

idcrosby:
Testing locally, using single docker host with Socks demo app running. Looks like there are some missing connections.
screen shot 2016-08-15 at 17 54 38
[report (2).json.zip](https://github.com/weaveworks/scope/files/418414/report.2.json.zip)

2opremio:
@idcrosby Can you be a bit more specific? What connections were you expecting to see and aren't there?

2opremio:
@idcrosby Also, could you see those connections with prior releases?

idcrosby:
Specifically between catalogue and the front-end.  I will re-test with previous version of scope

2opremio:
Thanks!

idcrosby:
Ok seeing same behaviour with version 0.16.2 so this is potentially not a scope issue, and at least not a regression.

2opremio:
@idcrosby Thanks again. Without any urgency and when you have the time, could you please verify whether it's a bug and create a ticket if it is?

davkal:
LGTM. new table view mode works great!

2opremio:
> I will let it run for a bit but I think that means we can close #897 and weaveworks/guides#181

Nah, unfortunately I still get the same problem after a few hours :S

rade:
> Nah, unfortunately I still get the same problem after a few hours :S

Surely that makes weaveworks/guides#181 the likely culprit. Or were you running on non-micro instances?

2opremio:
> Surely that makes weaveworks/guides#181 the likely culprit. Or were you running on non-micro instances?

I was running micro instances, but the CPU load was pretty low (or so I thought). I'm surprised they surpassed the CPU credit limit.

rade:
According to [The Internet](https://forums.aws.amazon.com/message.jspa?messageID=254836) one call tell whether throttling is taking place by observing the "steal" (st) metric in 'top'.

paulbellamy:
After this release, please merge: https://github.com/weaveworks/scope/pull/1682

2opremio:
You can merge it already if you want, changes to the release should go in this branch

foot:
LGTM

PR #1792, opened 2016-08-15
allow more accurate reporting of memcache hit ratio

rade:
A lot of time could pass between recording the request count and hit count pertaining to a particular report fetching batch, which skewed calculations cache hit ratios.

Fix that by defering the request count recording to the end, which is when we record the hit count.

Hopefully this will cure the >100% memcache hit ratios we sometimes see in our monitoring.


jml:
LGTM. Please merge once green.

PR #1796, opened 2016-08-16
fix report store query chaining

rade:
We fell victim to variable shadowing here. Ouch.

### DIFF
--- a/app/multitenant/aws_collector.go
+++ b/app/multitenant/aws_collector.go
@@ -277,7 +277,8 @@ func (c *awsCollector) getReports(reportKeys []string) ([]report.Report, error) 
 		if store == nil {
 			continue
 		}
-		found, missing, err := store.FetchReports(missing)
+		found, newMissing, err := store.FetchReports(missing)
+		missing = newMissing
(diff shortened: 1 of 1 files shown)


app/multitenant/aws_collector.go
```diff
@@ -277,7 +277,8 @@ func (c *awsCollector) getReports(reportKeys []string) ([]report.Report, error)
 		if store == nil {
 			continue
 		}
-		found, missing, err := store.FetchReports(missing)
+		found, newMissing, err := store.FetchReports(missing)
+		missing = newMissing
```
jml: I wonder if it'd be more idiomatic / clearer to initialize `found` and `err` outside of the loop and set them here with `=`. Genuine question.

jml:
Good catch! I have a question, but I'm happy for this to be merged as-is. Optional second round of review if you change it.

app/multitenant/aws_collector.go
rade: I thought about that. But no,  they ought be scoped to the loop.

Issue #1795, opened 2016-08-15
Scope stops collecting container metrics


foot:
From node-details call, `node.metrics.0.samples` is null.

Which the UI doesn't handle very gracefully right now.

Process/Host is still collected.

env: `docker-machine`

[report.2.json.zip](https://github.com/weaveworks/scope/files/418551/report.2.json.zip)

rade:
what do you mean by "stops"? you see metrics and then you don't?

foot:
Yeah, was testing out the 0.17.0 release, clicking around. At some point I started getting JS errors. Investigating, it was the sparklines failing to render on the details panel of an alpine container. Where there had been sparklines a few minutes ago.

foot:
After the metrics stopped I
1. `./scope stop` succeeded
2.`./scope launch` hung.
3. `docker-machine restart weave-1` succeeded and launched scope again for more testing, metrics are back.

foot:
This did happen last week too. It might be my env.

What command should I try running next time this happens?

1. `docker logs weavescope`
2. ??

2opremio:
@foot Could you try to reproduce with 0.16.2 to confirm whether it's a regression?

> What command should I try running next time this happens?

The logs and the report are the first stage yep.

foot:
I can't solidly repro :(. Here are logs from last time it happened.

[logs.txt](https://github.com/weaveworks/scope/files/420193/logs.txt)

If no-one has seen this on dev/prod I wouldn't block the release, I'll keep an eye out for it and start running scope w/ `--debug`.

2opremio:
Uhm, this is most probably the culprit:

```
<probe> ERRO: 2016/08/15 16:56:00.691200 docker container: error reading event for 5720c372a042f761befd46912841f65b8c621f076d7f7e4c2cbe5a9b56b5d155: only encoded map or array can be decoded into a struct
<probe> ERRO: 2016/08/15 16:56:41.761662 Error gather stats for container: 2b80711ecba7327cec9e2eb3c5195a441f5e16377b5a1940702a3c42e8fc7f68
```

However, the latter type of message could be a leftover (expected error) from https://github.com/weaveworks/scope/pull/1687 . It's difficult to tell because we don't print the error but I also get that for containers which stopped existing.

What docker version are you running? I recently bumped the docker client library ( #1787 ), maybe that's part of the problem?

Can you check whether the log errors are systematic?

[...]

Issue #1683, opened 2016-07-14
probes emit log noise when containers stop


rade:
I have no containers running except scope 0.16.1, and every few seconds the weavescope container log shows a triplet of events like this

```
<probe> INFO: 2016/07/14 18:41:54.975843 docker container: collecting stats for e678d8136f448933f6d04317e6fbab3d5f8fe5743804d833947e28ced162183c
<probe> ERRO: 2016/07/14 18:41:55.059077 docker container: error reading event, did container stop? read unix @->/var/run/docker.sock: use of closed network connection
<probe> INFO: 2016/07/14 18:41:55.059097 docker container: stopped collecting stats for e678d8136f448933f6d04317e6fbab3d5f8fe5743804d833947e28ced162183c

...

<probe> INFO: 2016/07/14 18:42:05.212411 docker container: collecting stats for 0aaf7f771a15411c0d3ca9104624347fc6222a106908de1f1c3a85a25b633c33
<probe> ERRO: 2016/07/14 18:42:05.298484 docker container: error reading event, did container stop? read unix @->/var/run/docker.sock: use of closed network connection
<probe> INFO: 2016/07/14 18:42:05.298508 docker container: stopped collecting stats for 0aaf7f771a15411c0d3ca9104624347fc6222a106908de1f1c3a85a25b633c33
```

The events do also get triggered when I start and then stop a container. So I am guessing that perhaps scope itself is starting and stopping a container every few seconds. In any case though, starting and stopping of containers is not something we should record in the logs at INFO let alone ERRO level.

peterbourgon:
Strong +1 to moving the `collecting stats for` and `stopped collecting stats for` logging to DEBUG, or removing it entirely. Beyond that, you correctly summarize the situation. You could check `!err.(*net.OpError).Temporary()` but it might be too coarse-grained. I don't see an alternative to checking the string; as it is a log optimization rather than a correctness thing, I don't see a huge problem with it.

PR #1798, opened 2016-08-16
Make stats gathering idempotent

2opremio:
Leftover from #1683 , found while diagnosing https://github.com/weaveworks/scope/issues/1795


rade:
LGTM

ekimekim:
There's a race where `StartGatheringStats()` could get called twice, and the second call happens before the line `c.statsConn = conn` is reached, so we'll start a second goroutine and have two things gathering stats.
But that's not a regression in this particular PR, so LGTM.

2opremio:
> There's a race where `StartGatheringStats()` could get called twice, and the second call happens before the line `c.statsConn = conn` is reached, so we'll start a second goroutine and have two things gathering stats.

Good catch, would you mind filing a bug for it?

PR #1769, opened 2016-08-08
Traffic control plugin

krnowak:
Depends on #1682. Last two commits are new.

PR #1818, opened 2016-08-18
Add support for table based controls

davkal:
This is the next version of the UI for #1769.

* if a node table has `controls` they will be rendered in a row
* controls need the to be of the same structure as node-level controls
* control errors are not shown yet

Example:
screen shot 2016-08-18 at 22 26 12

The backend for this is to follow. This PR can be merged without the backend being ready.

### DIFF
--- a/client/app/scripts/components/node-details/node-details-labels.js
+++ b/client/app/scripts/components/node-details/node-details-labels.js
@@ -1,7 +1,9 @@
 import React from 'react';
 import { Map as makeMap } from 'immutable';
+import sortBy from 'lodash/sortBy';
 
 import MatchedText from '../matched-text';
+import NodeDetailsControlButton from './node-details-control-button';
 import ShowMore from '../show-more';
 
 export default class NodeDetailsLabels extends React.Component {
@@ -13,6 +15,7 @@ export default class NodeDetailsLabels extends React.Component {
       limit: this.DEFAULT_LIMIT,
     };
     this.handleLimitClick = this.handleLimitClick.bind(this);
+    this.renderControls = this.renderControls.bind(this);
   }
 
   handleLimitClick() {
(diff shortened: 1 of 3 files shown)


client/app/scripts/components/node-details/node-details-labels.js
```diff
@@ -13,15 +15,25 @@ export default class NodeDetailsLabels extends React.Component {
...
 
   handleLimitClick() {
     const limit = this.state.limit ? 0 : this.DEFAULT_LIMIT;
     this.setState({limit});
   }
 
+  renderControls(controls) {
+    return (
+      <div className="node-details-labels-controls">
+        {sortBy(controls, 'rank').map(control => <NodeDetailsControlButton
```
foot: `rank` everywhere vs. an item's order in a list?

- `rank` feels awkward but useful when transforming lists into maps into lists I suppose.

client/app/styles/main.less
```diff
@@ -958,6 +947,25 @@ h2 {
   }
 }
 
+.node-control-button {
+  .btn-opacity;
+  padding: 6px;
+  margin-left: 2px;
+  font-size: 110%;
```
foot: Do you prefer `em` or `%`? Might be nice to settle on one. `lesshint` (less linting could be interesting, I love our JS linting).

I prefer `em` I think.

foot:
How to test this, checkout the ref'd version of scope, make, start plugin?

Will give that a goo.

foot:
Ah, the controls are still in the normal place in the branch.

foot:
Code LGTM, unsure how to test, maybe it will need a tweak at BE/FE integration anyway.

client/app/scripts/components/node-details/node-details-labels.js
davkal: Yes, this might a go-ism. They like maps, and only output lists when they have to 😉

client/app/styles/main.less
davkal: Yes, let's aim for `em`. (This needs a proper cleanup PR, there is a `body {font-size: 13px}` lurking.)

Issue #1789, opened 2016-08-15
collector shouldn't decode & re-encode reports


rade:
Currently all received reports are decoded. That is fine for the standalone app, but totally unnecessary for the collector, which stuffs reports into memcache and S3, re-encoding them in the process.

We know encoding/decoding is expensive, so this is a significant waste of resources. It also increases latency of reports from probes reaching the app.

There is one gotcha: the collector needs to know whether the report is a shortcut report. We could fix that by sending short-cut reports to a different URL. To maintain backward compatibility with old probes, we'd probably have to add a second new URL, for non-shortcut reports, so we can continue to process reports the old way on the current URL.

jml:
Another gotcha: the Decode / Encode step acts as validation, reducing the amount of garbage we put in our stores. Maybe it's OK to drop that, maybe not.

tomwilkie:
Maybe we shouldn't resersialise them?  That will save some CPU.

rade:
> Maybe we shouldn't resersialise them?

i.e. we always decode, but then pass on the original?

> That will save some CPU.

Yes, though the bulk is used by the decoding. Still, it's a good intermediate steps, especially since it will also work for existing probes.

2opremio:
Fixed via #1819

rade:
No this hasn't been fixed. We are still *de-coding* the report. Which is unnecessary.

2opremio:
> Which is unnecessary.

Uhm, that's questionable I see value in decoding to catch problems early. Otherwise probes (or ill-intentioned clients) could happily push invalid reports without any feedback.

rade:
That is discussed above. The cost of fully decoding a report is *massive*. And it's a cost we pay even when nobody is ever even looking at the data. So, on balance I reckon it is better to save that cost than validate. And it certainly would be far simpler to implement that than the next best thing, which is...

(as suggested by @tomwilkie) a streaming/low-memory validator, perhaps just a json parser.

rade:
> a streaming/low-memory validator, perhaps just a json parser

https://golang.org/pkg/encoding/json/#Decoder.Token might do the trick.

tomwilkie:
Reports aren't JSON.

On Thu, Sep 8, 2016 at 5:46 PM, Matthias Radestock <notifications@github.com

2opremio:
> https://golang.org/pkg/encoding/json/#Decoder.Token might do the trick.

[...]

PR #1819, opened 2016-08-22
Don't reencode reports in the collector

tomwilkie:
Part of #1789

### DIFF
--- a/app/collector.go
+++ b/app/collector.go
@@ -26,7 +26,7 @@ type Reporter interface {
 // Adder is something that can accept reports. It's a convenient interface for
 // parts of the app, and several experimental components.
 type Adder interface {
-	Add(context.Context, report.Report) error
+	Add(context.Context, report.Report, []byte) error
 }
 
 // A Collector is a Reporter and an Adder
@@ -88,7 +88,7 @@ func NewCollector(window time.Duration) Collector {
 }
 
 // Add adds a report to the collector's internal state. It implements Adder.
-func (c *collector) Add(_ context.Context, rpt report.Report) error {
+func (c *collector) Add(_ context.Context, rpt report.Report, _ []byte) error {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 	c.reports = append(c.reports, rpt)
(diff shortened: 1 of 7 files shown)


app/collector.go
```diff
@@ -26,7 +26,7 @@ type Reporter interface {
 // Adder is something that can accept reports. It's a convenient interface for
 // parts of the app, and several experimental components.
 type Adder interface {
-	Add(context.Context, report.Report) error
+	Add(context.Context, report.Report, []byte) error
```
2opremio: Can you please document what's the content and role of the `[]byte` parameter?

app/multitenant/memcache_client.go
```diff
@@ -209,9 +209,14 @@ func (c *MemcacheClient) FetchReports(keys []string) (map[string]report.Report,
 func (c *MemcacheClient) StoreReport(key string, report *report.Report) (int, error) {
```
2opremio: Is there a reason to keep the StoreReport function? Why not inline everything in StoreReportBytes?

rade: Alternatively, make `StoreReport` take an extra (possibly nil) `[]byte` argument. To match `collector.Add`.

app/router.go
```diff
@@ -136,6 +142,12 @@ func RegisterReportPostHandler(a Adder, router *mux.Router) {
 			return
 		}
 
+		// a.Add(..., buf) assumes buf is gzip'd msgpack
+		if !isMsgpack {
```
2opremio: Why not simply reject uncompressed msgpack reports? I think it's safe to do since we haven't done it in the past.

tomwilkie: discussed offline, fixprobe.

jml:
LGTM.

docker/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM alpine:3.3
 MAINTAINER Weaveworks Inc <help@weave.works>
 LABEL works.weave.role=system
 WORKDIR /home/weave
-RUN echo "http://dl-3.alpinelinux.org/alpine/edge/testing" >>/etc/apk/repositories && \
+RUN echo "http://dl-cdn.alpinelinux.org/alpine/edge/community" >>/etc/apk/repositories && \
```
rade: We may want to backport this to `release-0.17`.

PR #1822, opened 2016-08-23
Sort non-number columns ASCending by default.

### LINKED ISSUE #1802: Table-mode: non-metric columns are sorted alphabetically reverse
Seems a bit counter-intuitive.

foot:
- And number columns DESCending by default.
- default: when you sort by that column for the first time.

Fixes #1802

### DIFF
--- a/client/app/scripts/components/node-details/node-details-table.js
+++ b/client/app/scripts/components/node-details/node-details-table.js
@@ -18,6 +18,7 @@ const CW = {
   XXL: '170px',
 };
 
+
 const COLUMN_WIDTHS = {
   count: '70px',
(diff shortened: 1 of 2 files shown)


davkal:
Does not quite fix #1802 

The issue appears when you switch over to tables on the containers topo: containers are reverse sorted by name.

davkal:
Code still looks good, and the alphabetical sorting when there is no metric works. 
But now the Internet is ranked highest, which probably should not be.

screen shot 2016-08-29 at 16 17 23

foot:
T < a! I'm sure someone out there will appreciate that! But yes, good
point, will try w/ toLower().

PR #1823, opened 2016-08-23
Makes services the initial topology if available

### LINKED ISSUE #1809: make 'pods' the initial view, when available
@jml pointed out that when running k8s, the container view is rarely what one wants to look at. So perhaps we should make the 'pods' view the default initial view, if available.

foot:
- Otherwise reverts to containers

Fixes #1809

### DIFF
--- a/client/app/scripts/utils/topology-utils.js
+++ b/client/app/scripts/utils/topology-utils.js
@@ -1,5 +1,22 @@
 import _ from 'lodash';
-import { is as isDeepEqual, Map as makeMap, Set as makeSet } from 'immutable';
+import { is as isDeepEqual, Map as makeMap, Set as makeSet, List as makeList } from 'immutable';
+
+
+const TOPOLOGY_DISPLAY_PRIORITY = {
+  services: 1,
+  containers: 2,
+};
+
+
+export function getDefaultTopology(topologies) {
+  const flatTopologies = topologies
+    .flatMap(t => makeList([t]).concat(t.get('sub_topologies', makeList())));
+
+  return flatTopologies
+    .sortBy(t => TOPOLOGY_DISPLAY_PRIORITY[t.get('id')] || Infinity)
+    .getIn([0, 'id']);
+}
+
 
 /**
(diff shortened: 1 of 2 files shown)


rade:
> Otherwise reverts to containers

The prioritisaion should be services -> deployments -> replica sets  -> pods -> containers. Says @tomwilkie.

foot:
Cool, will update

client/app/scripts/utils/topology-utils.js
```diff
@@ -1,5 +1,31 @@
 import _ from 'lodash';
-import { is as isDeepEqual, Map as makeMap, Set as makeSet } from 'immutable';
+import { is as isDeepEqual, Map as makeMap, Set as makeSet, List as makeList } from 'immutable';
+
+
+//
+// top priority first
+//
+const TOPOLOGY_DISPLAY_PRIORITY = [
```
davkal: I'd prefer it if this was backend-driven via a priority field.

davkal:
Code looks good.

Implementation-nit: priority should be driven by backend, so that when new topos are added, UI does not have to be involved.

UX-nit: If Services is the most important one, it should be at the top. Any introduction of priority ordering trumps representation model ordering (lower topos are groupings of higher topos). This needs to be changed in the backend.

davkal:
> priority should be driven by backend, so that when new topos are added, UI does not have to be involved.
Actually, we should put this up for discussion in the next meeting.

davkal:
Can't we make service the main topo?

screen shot 2016-08-29 at 16 31 26

davkal:
If we cant completely flip the order, I'd say let's stick to the original one.

foot:
Node type, followed by its aggregates in order of most interest breaks
strict "spatial alignment" w/ the priorities we've set up but I think it
still works. If no aggs fall back to basic type.

davkal:
I mostly have a problem with a sub-menu entry being the most important one (decided by priority). I feel that it makes the menu harder to grok. 
An alternative would be dynamic prio ordering in the backend, so that the default one is the top one.

PR #1827, opened 2016-08-23
make smartMerger.Merge merge reports in parallel

rade:
for reduced latency

### DIFF
--- a/app/merger.go
+++ b/app/merger.go
@@ -2,10 +2,7 @@ package app
 
 import (
 	"fmt"
-	"math"
-	"sort"
 
-	"github.com/bluele/gcache"
 	"github.com/spaolacci/murmur3"
 
 	"github.com/weaveworks/scope/report"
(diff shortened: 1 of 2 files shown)


rade:
I put a little bit of load onto our dev cluster by having five browser tabs looking at the "Weave Cloud (Prod)" instance, plus one looking at Dev. I waited for query latencies to stabilise, then deployed this branch, waited to stabilise, then re-deployed master. Here is the result:
screenshot_2016-08-24_00-24-53
The branch was deployed at 23:12, and master re-deployed at 23:19.

app/merger.go
```diff
@@ -34,91 +31,31 @@ func (dumbMerger) Merge(reports []report.Report) report.Report {
...
-}
-
 func (s smartMerger) Merge(reports []report.Report) report.Report {
-	// Start with a sorted list of leaves.
-	// Note we must dedupe reports with the same ID to ensure the
-	// algorithm below doesn't go into an infinite loop.  This is
-	// fine as reports with the same ID are assumed to be the same.
-	nodes := []*node{}
-	seen := map[uint64]struct{}{}
+	l := len(reports)
```
tomwilkie: Since there is nothing in s, you can omit it.

app/merger.go
```diff
@@ -34,91 +31,31 @@ func (dumbMerger) Merge(reports []report.Report) report.Report {
...
-	// two reports is cached.
-	merge := func(left, right *node) *node {
-		return &node{
-			id:  hash(left.rpt.ID, right.rpt.ID),
-			rpt: report.MakeReport().Merge(left.rpt).Merge(right.rpt),
-		}
+	for ; l > 1; l-- {
+		go func(left, right report.Report) {
+			c <- left.Merge(right)
+		}(<-c, <-c)
```
tomwilkie: Am I missing something, or does this not actually guarantee much about the order in which the reports are merged?

It will pair wise merge all the input reports, but then after that the order of merging depends on when the goroutines schedule & finish?

app/merger.go
rade: fixed

app/merger.go
rade: That is correct. Anything else would stall. This can compromise the "merge small reports" somewhat, so it's a trade off. But actually it's less a compromise than it first appears: With fair scheduling (ha ha), merging smaller reports should take less time than merging larger reports - which is why we want the "merge small reports" optimisation in the first place - so the new code should actually work out beneficially for unbalanced inputs / intermediate results; they will "jump the queue" when previously they wouldn't.

tomwilkie: All good points!

tomwilkie:
Lgtm!

2opremio:
Sorry to comment on this late but while the new merger clearly is advantageous for the WeaveCloud use I would be interested to see whether it makes sense when running the app in standalone. @rade Did you get to measure it? I would be particularly interested to see how it behaves in systems with a reduced number of cores.

rade:
I didn't measure it standalone. With few cores there is some cost due to pre-emption, but I doubt it's high. Plus we benefit from the better handling of unbalanced input / intermediate result sizes.

PR #1828, opened 2016-08-25
Improve conntrack kernel-support detection for in-kernel compilations

### LINKED ISSUE #1826: conntrack not used even though it's working
Hello

I am trying a simple install with following docker-compose.yml:

version: '2'

services:
 container1:
   image: nginx:latest

 container2:
   image: nginx:latest

 weavescope:
   image: weaveworks/scope:0.17.1
   privileged: true
   network_mode: "host"
   pid: "host"
   labels:
     - "works.weave.role=system"
   volumes:
     - "/var/run/docker.sock:/var/run/docker.sock"
   command:
     - "--probe.conntrack=true"
     - "--probe.docker=true"

For test, I am using nc -l -p 1300 / nc containerIP 1300
The connection is established since I can see an echo without problems

However, instead of connection between containers, I see http://puu.sh/qLCEQ/b8115764b9.jpg

As we can see from http://puu.sh/qLDcQ/489f20f96b.jpg - Scope can see them, but possibly can no determine that they are between 2 containers, not host local

https://github.com/weaveworks/scope/files/433143/report.gz

Thanks
Nikolay

2opremio:
Fixes #1826

### DIFF
--- a/probe/endpoint/conntrack.go
+++ b/probe/endpoint/conntrack.go
@@ -15,9 +15,8 @@ import (
 )
 
 const (
-	modules = "/proc/modules"
 	// From https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt
-	// Check a tcp-related file since that's what we need
+	// Check a tcp-related file for existence since we need tcp tracking
 	procFileToCheck  = "sys/net/netfilter/nf_conntrack_tcp_timeout_close"
 	xmlHeader        = "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n"
 	conntrackOpenTag = "<conntrack>\n"
--- a/probe/endpoint/conntrack.go
+++ b/probe/endpoint/conntrack.go
@@ -5,7 +5,7 @@ import (
 	"encoding/xml"
 	"io"
 	"os"
-	"strings"
+	"path/filepath"
 	"sync"
 	"time"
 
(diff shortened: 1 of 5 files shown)


probe/endpoint/conntrack.go
```diff
@@ -15,8 +15,9 @@ import (
 )
 
 const (
-	modules          = "/proc/modules"
-	conntrackModule  = "nf_conntrack"
+	// From https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt
+	// Check a tcp-related file for existence since we need tcp tracking
+	procFileToCheck  = "sys/net/netfilter/nf_conntrack_tcp_timeout_close"
```
paulbellamy: Is this supported/available on all version of conntrack? (or all versions we can work with?)

paulbellamy:
Quick q about availability, but other than that LGTM.

probe/endpoint/conntrack.go
2opremio: It's been available at least since 2.6.20 https://github.com/torvalds/linux/blame/a7fd20d1c476af4563e66865213474a2f9f473a4/net/netfilter/nf_conntrack_proto_tcp.c#L1441

2opremio: To be more precise, this confirms it's been part of conntrack since the first commit https://github.com/torvalds/linux/commit/9fb9cbb1082d6b31fb45aa1a14432449a0df6cf1#diff-fbdc7dc5c7c4089799324f79fd85d799R543

PR #1830, opened 2016-08-25
correct cpu metrics

### LINKED ISSUE #1664: CPU Usage is not accurate on hosts
It's being reported out of 100%, on hosts, not num CPUs.

rade:
Fixes #1664.

### DIFF
--- a/probe/host/system_linux.go
+++ b/probe/host/system_linux.go
@@ -71,7 +71,7 @@ var GetUptime = func() (time.Duration, error) {
 
 var previousStat = linuxproc.CPUStat{}
 
-// GetCPUUsagePercent returns the percent cpu usage and max (ie #cpus * 100)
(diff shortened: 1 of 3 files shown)


probe/host/system_linux.go
```diff
@@ -71,7 +71,7 @@ var GetUptime = func() (time.Duration, error) {
 
 var previousStat = linuxproc.CPUStat{}
 
-// GetCPUUsagePercent returns the percent cpu usage and max (ie #cpus * 100)
+// GetCPUUsagePercent returns the percent cpu usage and max (i.e. 100% or 0 if unavailable)
```
jml: Why 0.0 for unavailable? Why not NaN?

rade: it's what was there before. I rather suspect NaN would break a bunch of stuff.

jml:
LGTM.

PR #1831, opened 2016-08-25
Fixes metrics-on-canvas updating.

### LINKED ISSUE #1829: Metrics on canvas get stuck
They don't seem to refresh properly when hovering on them. Clicking on the node to show the details panel seems to _refresh_ them.

I would expect the metrics to refresh while hovering and between hovers.

foot:
- We were being a bit overzealous w/ our layout caching.

Fixes #1829

### DIFF
--- a/client/app/scripts/charts/nodes-chart.js
+++ b/client/app/scripts/charts/nodes-chart.js
@@ -5,14 +5,16 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { Map as makeMap, fromJS, is } from 'immutable';
 import timely from 'timely';
+import { diff } from 'deep-diff';
 
+import { nodeAdjacenciesSelector, adjacentNodesSelector } from '../selectors/chartSelectors';
 import { clickBackground } from '../actions/app-actions';
 import { EDGE_ID_SEPARATOR } from '../constants/naming';
 import { MIN_NODE_SIZE, DETAILS_PANEL_WIDTH, MAX_NODE_SIZE } from '../constants/styles';
 import Logo from '../components/logo';
 import { doLayout } from './nodes-layout';
 import NodesChartElements from './nodes-chart-elements';
-import { getActiveTopologyOptions, getAdjacentNodes } from '../utils/topology-utils';
+import { getActiveTopologyOptions } from '../utils/topology-utils';
 
 const log = debug('scope:nodes-chart');
 
@@ -22,6 +24,24 @@ const ZOOM_CACHE_FIELDS = ['scale', 'panTranslateX', 'panTranslateY'];
 const radiusDensity = d3.scale.threshold()
   .domain([3, 6])
   .range([2.5, 3.5, 3]);
+
+/**
+ * dynamic coords precision based on topology size
+ */
+function getLayoutPrecision(nodesCount) {
+  let precision;
+  if (nodesCount >= 50) {
+    precision = 0;
+  } else if (nodesCount > 20) {
+    precision = 1;
+  } else if (nodesCount > 10) {
+    precision = 2;
+  } else {
+    precision = 3;
+  }
+
+  return precision;
+}
 
 
 function identityPresevingMerge(a, b) {
@@ -152,6 +172,7 @@ class NodesChart extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
+    console.log('componentWillReceiveProps', diff(nextProps, this.props), nextProps);
     // gather state, setState should be called only once here
     const state = _.assign({}, this.state);
 
@@ -224,6 +245,7 @@ class NodesChart extends React.Component {
     const svgClassNames = this.props.isEmpty ? 'hide' : '';
     console.log('nodes-chart.render');
 
+    const layoutPrecision = getLayoutPrecision(nodes.size);
     return (
       <div className="nodes-chart">
         <svg width="100%" height="100%" id="nodes-chart-canvas"
@@ -238,7 +260,7 @@ class NodesChart extends React.Component {
             scale={scale}
             transform={transform}
             selectedNodeScale={this.state.selectedNodeScale}
-            layoutPrecision={this.props.layoutPrecision} />
+            layoutPrecision={layoutPrecision} />
         </svg>
       </div>
     );
@@ -411,7 +433,8 @@ class NodesChart extends React.Component {
 
 function mapStateToProps(state) {
   return {
-    adjacentNodes: getAdjacentNodes(state),
+    nodes: nodeAdjacenciesSelector(state),
+    adjacentNodes: adjacentNodesSelector(state),
     forceRelayout: state.get('forceRelayout'),
     selectedNodeId: state.get('selectedNodeId'),
     topologyId: state.get('currentTopologyId'),
--- a/client/app/scripts/charts/nodes-chart.js
+++ b/client/app/scripts/charts/nodes-chart.js
@@ -3,7 +3,7 @@ import d3 from 'd3';
 import debug from 'debug';
 import React from 'react';
 import { connect } from 'react-redux';
-import { Map as makeMap, fromJS, is as isDeepEqual } from 'immutable';
+import { Map as makeMap, fromJS, is } from 'immutable';
 import timely from 'timely';
 
 import { clickBackground } from '../actions/app-actions';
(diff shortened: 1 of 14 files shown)


foot:
We're back to over-calculating layouts, this doesn't feel too expensive when testing but maybe I'm missing something.

A nicer caching mechanism would be to introduce something like re-select and pull out the graph layout dependencies a bit more.

2opremio:
@foot Would you mind rebasing? The build was broken and we just fixed it in master.

client/app/scripts/charts/nodes-chart.js
```diff
@@ -82,8 +81,7 @@ class NodesChart extends React.Component {
     state.height = nextProps.forceRelayout ? nextProps.height : (state.height || nextProps.height);
     state.width = nextProps.forceRelayout ? nextProps.width : (state.width || nextProps.width);
 
-    // _.assign(state, this.updateGraphState(nextProps, state));
-    if (nextProps.forceRelayout || !isSameTopology(nextProps.nodes, this.props.nodes)) {
+    if (nextProps.forceRelayout || nextProps.nodes !== this.props.nodes) {
```
davkal: Two things are getting mixed up here. Generally, in this class `props.nodes` have to become `state.nodes`, otherwise the changes wont be rendered.

This section makes sure not to run the layout engine code if the topo has not changed, it does create `state.nodes` along the way. (I guess this is bad design.)

What you need is `props.nodes`' properties being merged into `state.nodes` if `props.nodes` and `nextProps.nodes` are not the same. This should be a separate step.

foot:
@davkal PT another quick L, we are 1.5fps slower than master (18.5 vs 20) using reselect and I'm not sure exactly whyyy, similar numbers from react perf re: number of render() calls etc, (ofc master is not rendering MOC correctly).

I'd merge this and then look into using reselect in other parts of the system to hopefully improve perf WDYT?

rade:
@foot what is left to do here? @davkal says you asked him "to hold off" on this PR.

foot:
@davkal fixed hanging edges, PTAL!

client/app/scripts/selectors/chartSelectors.js
```diff
@@ -0,0 +1,124 @@
...
+//   - v1 = getAdjacentNodes(a)
+//   - v2 = getAdjacentNodes(a)
+//   - v1 !== v2
+//   - is(v1, v2) === true
+//
+// - createDeepEqualSelector will wrap those calls with a: is(v1, v2) ? v1 : v2
+//   - Thus you can compare consecutive calls to adjacentNodesSelector(state) with === (which is
+//     what redux is doing with connect()
+//
+// Note: this feels like the wrong way to be using reselect...
```
davkal: Yes, you should be able to rewrite `getAdjacentNodes` so it takes `nodes` (filtered) and `origin` as inputs. Then those two become the selectors, and `getAdjacentNodes` will be fully memoizable.

client/app/scripts/selectors/chartSelectors.js
```diff
@@ -0,0 +1,124 @@
...
+    (nodes) => nodes.map(n => makeMap({
+      id: n.get('id'),
+      adjacency: n.get('adjacency'),
+    }))
+  ),
+  identity
+);
+
+
+export const dataNodesSelector = createSelector(
```
davkal: No potential for deepEqual? This means this selector always returns something new.

client/app/scripts/selectors/chartSelectors.js
```diff
@@ -0,0 +1,124 @@
...
+    //
+    if (layoutNodes.size !== dataNodes.size) {
+      log('Obviously mismatched node data', layoutNodes.size, dataNodes.size);
+    }
+    return mergeDeepKeyIntersection(dataNodes, layoutNodes);
+  }
+);
+
+
+export const completeNodesSelector = createDeepEqualSelector(
```
davkal: Couldnt you simply make `_completeNodesSelector` a `createDeepEqualSelector` ?

client/app/scripts/selectors/chartSelectors.js
```diff
@@ -0,0 +1,124 @@
...
+//
+export const nodeAdjacenciesSelector = createDeepEqualSelector(
+  createSelector(
+    nodesSelector,
+    (nodes) => nodes.map(n => makeMap({
+      id: n.get('id'),
+      adjacency: n.get('adjacency'),
+    }))
+  ),
+  identity
```
davkal: This is not needed, is it?

client/app/scripts/charts/nodes-chart.js
```diff
@@ -3,17 +3,17 @@ import d3 from 'd3';
...
+import { nodeAdjacenciesSelector, adjacentNodesSelector } from '../selectors/chartSelectors';
 import { clickBackground } from '../actions/app-actions';
 import { EDGE_ID_SEPARATOR } from '../constants/naming';
 import { MIN_NODE_SIZE, DETAILS_PANEL_WIDTH, MAX_NODE_SIZE } from '../constants/styles';
 import Logo from '../components/logo';
 import { doLayout } from './nodes-layout';
 import NodesChartElements from './nodes-chart-elements';
-import { getActiveTopologyOptions, getAdjacentNodes,
-  isSameTopology } from '../utils/topology-utils';
+import { getActiveTopologyOptions } from '../utils/topology-utils';
```
davkal: This looks like another selector candidate.

client/app/scripts/charts/nodes-chart.js
```diff
@@ -391,16 +390,19 @@ class NodesChart extends React.Component {
   }
 }
 
+
 function mapStateToProps(state) {
   return {
-    adjacentNodes: getAdjacentNodes(state),
+    nodes: nodeAdjacenciesSelector(state),
```
davkal: This is selector is not working as hoped. My test topology has regular node delta `update` events, but no adding/removing. Yet it's going into `updateGraphState` which is a little more expensive.
The trouble is that `nodes-chart:componentWillReceiveProps` is firing a lot, so there should be extra care in not doing too much work there.

client/app/scripts/charts/nodes-chart.js
foot: Yep! This first pass I didn't selectorify much stuff, lots of things can be selector'd!

client/app/scripts/charts/nodes-chart.js
foot: Strange! Adjacencies are not changing? (`update`). Will try and create a test case kind of thing...

client/app/scripts/selectors/chartSelectors.js
foot: Yes, that would have a similar effect.

client/app/scripts/selectors/chartSelectors.js
foot: Yep, bit of a judgement call here, if nodes has changed something has probably changed in here (metrics). All this talk of expensive deep equality checks has made me a bit cautious about using it everywhere.

client/app/scripts/selectors/chartSelectors.js
foot: Yep, this weird createDeepEqualSelector+identity wrapper ensures that we get back the previous value they were deep equal, I refactored this (sorry) to make that clearer and will push in a sec.

client/app/scripts/selectors/chartSelectors.js
```diff
@@ -0,0 +1,124 @@
...
+);
+
+
+const identity = v => v;
+
+
+const allNodesSelector = state => state.get('nodes');
+
+
+export const nodesSelector = createSelector(
```
foot: `createSelector` checks its arguments (state.get('nodes') in this case) not the result. we could wrap it in the createDeepEqualSelector+identity dance to check its result however! good point.

client/app/scripts/selectors/chartSelectors.js
foot: True!

foot:
I refactored the selectors again (sorry!), to make the intention clearer. That is: (as you say) trying to avoid the calling of `componentWillReceiveProps` and triggering a layout. 

When we hopefully move the layout code into a selector itself we can guard it from being called wastefully by using the traditional immutablejs+reselect pattern suggested in their README.md (a vanilla createDeepEqualSelector).

As our layout code is in a component for now we guard it by being careful that mapStateToProps always returns `===`able values. In the version of the code you reviewed this was awkwardly done w/ an inline `createDeepEqualSelector(anonymousSelector, identity)`, I pulled that out into a documented helper wrapper function called... `returnPreviousRefIfEqual`. It wraps the selectors that feed `nodes-chart` avoiding updating the props if it can.

davkal:
Works well! LGTM

PR #1832, opened 2016-08-25
Only consider pod parents living in the same namespace

### LINKED ISSUE #1824: pod counts and details panel lists do not respect namespace
In Weave Cloud (Dev) we have a memcached service in the default namespace and one in the frankenstein namespace. But when the view has either namespace selected, the "# pods" shown for the service is the aggregate of the two, and when bringing up the details panel, the pods for both services are listed. When selecting 'All Namespaces', two memcached nodes are shown, both containing the aggregrate set of pods.

2opremio:
Fixes #1824


rade:
LGTM

PR #1833, opened 2016-08-26
use go-dockerclient's Client.Stats

### LINKED ISSUE #1799: probe: Stats gathering can be started twice
In `probe/docker/container.go`, `StartGatheringStats()`, we check if `c.statsConn` is set and only start a new stats connection if it is not. However, we do not immediately then set this value (that happens only after some other setup inside a goroutine we then launch), meaning it might not be set by the time we get called again.
In this case, we would have two connections reporting stats for the same probe.

The simplest fix would be to add a `statsStarted bool` to the struct, and do:
```go
c.Lock()
if c.statsStarted {
  c.Unlock()
  return nil
}
c.statsStarted = true
c.Unlock()
```
perhaps modulo some nicer handling of the Unlock

rade:
...instead of rolling our own.

This also fixes #1799 - a race condition that could result in multiple stats gatherers for a container.

### DIFF
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -1,13 +1,9 @@
 package docker
 
 import (
-	"bufio"
 	"fmt"
 	"io"
 	"net"
-	"net/http"
-	"net/http/httputil"
-	"net/url"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


ekimekim:
LGTM.

PR #1839, opened 2016-09-01
Gracefully shutdown app, letting active http requests finish with timeout

ekimekim:
Upon getting SIGINT/TERM, we stop listening and give a configurable timeout (default 5s)
to finish any active requests.

### DIFF
--- a/prog/main.go
+++ b/prog/main.go
@@ -103,11 +103,12 @@ type probeFlags struct {
 }
 
 type appFlags struct {
-	window    time.Duration
-	listen    string
-	logLevel  string
-	logPrefix string
-	logHTTP   bool
+	window      time.Duration
+	listen      string
+	stopTimeout time.Duration
+	logLevel    string
+	logPrefix   string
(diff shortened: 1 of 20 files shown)


prog/main.go
```diff
@@ -229,6 +230,7 @@ func main() {
 	// App flags
 	flag.DurationVar(&flags.app.window, "app.window", 15*time.Second, "window")
 	flag.StringVar(&flags.app.listen, "app.http.address", ":"+strconv.Itoa(xfer.AppPort), "webserver listen address")
+	flag.DurationVar(&flags.app.shutdownTimeout, "app.shutdownTimeout", 5*time.Second, "How long to wait for http requests to finish when shutting down")
```
paulbellamy: Isn't this different from what we've used elsewhere?

ekimekim: ah yes, `shutdownTimeout` vs `stopTimeout`. Slipped my mind.

paulbellamy:
LGTM

PR #1843, opened 2016-09-01
fix internet connection counts

### LINKED ISSUE #1495: Incorrect number of connections shown on internet nodes
The +17 is wrong:

screen shot 2016-05-12 at 16 55 36

rade:
Fixes #1495.

### DIFF
--- a/render/detailed/connections.go
+++ b/render/detailed/connections.go
@@ -62,6 +62,25 @@ type connection struct {
 	port                      string // always the server-side port
 }
 
+func newConnection(n report.Node, node report.Node, port string, endpointID string, localAddr string) connection {
+	c := connection{
+		localNodeID:  n.ID,
+		remoteNodeID: node.ID,
+		port:         port,
+	}
+	// For internet nodes we break out individual addresses, both when
+	// the internet node is remote (an incoming connection from the
+	// internet) and 'local' (ie you are loading details on the
+	// internet node)
+	if isInternetNode(n) {
+		// We use the *endpoint* ID here since that has the reverse
+		// DNS information associated with it.
+		c.localNodeID = endpointID
+		c.localAddr = localAddr
+	}
+	return c
+}
+
 func (row connection) ID() string {
 	return fmt.Sprintf("%s:%s-%s:%s-%s", row.remoteNodeID, row.remoteAddr, row.localNodeID, row.localAddr, row.port)
 }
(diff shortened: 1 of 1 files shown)


rade:
@paulbellamy this does perhaps deserve a unit test, but I am not familiar with all the data structures involved, so this is a bit daunting. If you do think this should have a test, would you mind helping me write one?

render/detailed/connections.go
```diff
@@ -115,7 +110,7 @@ func incomingConnectionsSummary(topologyID string, r report.Report, n report.Nod
 		TopologyID:  topologyID,
 		Label:       "Inbound",
 		Columns:     columnHeaders,
-		Connections: connectionRows(r, counts, isInternetNode(n)),
+		Connections: connectionRows(r, ns, counts, isInternetNode(n)),
```
rade: I am assuming that `ns` contains `n` as well as all the local endpoints. Is that safe?

paulbellamy:
> this does perhaps deserve a unit test

yes, please!

render/detailed/connections.go
paulbellamy: Safe enough, as they have all been pulled out of the rendered nodes at some point. It would be nice if that had better locality.

rade: The former yes. The latter, no. As a result we fail to look up the reverse DNS resolution in `connectionRows()`. Grrr.

rade: Fixed.

davkal:
I'm not getting the following in the details panel for the Internet:
```
Inbound Remote Port Count
X    IP1       PortA    2
Y    IP2       PortB   2
Y    IP2       PortB   2
Y    IP2       PortB   2
X    IP3       PortA    2
```
and with +51 (which works as expected.

Having multiple rows with all values identical looks confusing. I'm guessing those 3 Y rows, are 3 different containers with the same name.

UX remedies:

* aggregate counts to one row if all other fields are the same
* add a column for a field that varies across the rows
* add a hover over the name

rade:
> UX remedies:
>
>    aggregate counts to one row if all other fields are the same

That would break the navigation, i.e. I need to be able to navigate to the 3 individual containers, so need 3 click-able things.

>    add a column for a field that varies across the rows

That would be best, but we are really short on horizontal space here.

>    add a hover over the name

That might be the only choice left :(

Please file an issue.

davkal:
Filed #1844 for UI issues.

paulbellamy:
LGTM

rade:
I've given up on constructing a test for this. It's just too hard, for me at least, not being familiar with the entire report data structure, and the intricacies of `fixture.Report`. I have filed #1845.

PR #1850, opened 2016-09-02
correctly reverse resolve source of inbound connections

### LINKED ISSUE #1847: inbound internet connections reverse-resolved incorrectly
screenshot_2016-09-02_17-22-52-cropped
The connection is from the IP in brackets - 37.157.33.76 - but the reverse resolved name is "ip-172-31-1-14.eu-west-1.compute.internal", which is the machine the connection is going *to*.

rade:
Fixes #1847

### DIFF
--- a/probe/endpoint/reporter.go
+++ b/probe/endpoint/reporter.go
@@ -196,6 +196,9 @@ func (r *Reporter) addConnection(rpt *report.Report, t fourTuple, namespaceID st
 
 	// In case we have a reverse resolution for the IP, we can use it for
 	// the name...
+	if fromNames, err := r.reverseResolver.get(t.fromAddr); err == nil {
+		fromNode = fromNode.WithSet(ReverseDNSNames, report.MakeStringSet(fromNames...))
+	}
 	if toNames, err := r.reverseResolver.get(t.toAddr); err == nil {
 		toNode = toNode.WithSet(ReverseDNSNames, report.MakeStringSet(toNames...))
 	}
--- a/probe/endpoint/reporter.go
+++ b/probe/endpoint/reporter.go
@@ -181,36 +181,27 @@ func (r *Reporter) Report() (report.Report, error) {
 
 func (r *Reporter) addConnection(rpt *report.Report, t fourTuple, namespaceID string, extraFromNode, extraToNode map[string]string) {
 	var (
-		fromEndpointNodeID = report.MakeEndpointNodeID(r.hostID, namespaceID, t.fromAddr, strconv.Itoa(int(t.fromPort)))
-		toEndpointNodeID   = report.MakeEndpointNodeID(r.hostID, namespaceID, t.toAddr, strconv.Itoa(int(t.toPort)))
-
-		fromNode = report.MakeNodeWith(fromEndpointNodeID, map[string]string{
-			Addr: t.fromAddr,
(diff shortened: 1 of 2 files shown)


probe/endpoint/reporter.go
```diff
@@ -181,33 +181,27 @@ func (r *Reporter) Report() (report.Report, error) {
...
-		toNode = report.MakeNodeWith(toEndpointNodeID, map[string]string{
-			Addr: t.toAddr,
-			Port: strconv.Itoa(int(t.toPort)),
-		})
+		fromNode = r.makeEndpointNode(namespaceID, t.fromAddr, t.fromPort, extraFromNode)
+		toNode   = r.makeEndpointNode(namespaceID, t.toAddr, t.toPort, extraToNode)
 	)
+	rpt.Endpoint = rpt.Endpoint.AddNode(fromNode.WithEdge(toNode.ID, report.EdgeMetadata{}))
+	rpt.Endpoint = rpt.Endpoint.AddNode(toNode)
+}
```
paulbellamy: Why not just do it all in inline, and ditch the separate `var`?

rade: We need to add the edge. for which we need a reference to the other node.

render/detailed/connections.go
```diff
@@ -145,7 +145,7 @@ func outgoingConnectionsSummary(topologyID string, r report.Report, n report.Nod
 				if !ok {
 					continue
 				}
-				key := newConnection(n, node, port, remoteEndpointID, localAddr)
```
paulbellamy: Why is this change necessary? `localEndpoint.ID` will only be used in the case of internet nodes, so the id will always be the client endpoint, requesting the outbound connection. Meanwhile we're using the remote port. Shouldn't we be using the `remoteEndpoint.ID`, to match?

probe/endpoint/reporter.go
paulbellamy: Ah yeah, right.

render/detailed/connections.go
rade: That's the whole point of this PR.

localEndpoint is the endpoint on the Internet node, corresponding to where an inbound connection from the Internet originated. It is the reverse-DNS of that endpoint we are interested in.

And we are using the remote port, which is the port *to* which the inbound internet connections are connected because we don't want to distinguish between connections from the same internet IP but different ports, but *do* want to distinguish between connections *to* different container/host IPs/ports.

paulbellamy: > corresponding to where an inbound connection from the Internet originated.

But this is in the `outgoingConnectionsSummary`...

rade: yep. which in case of the internet node is for inbound connections.

paulbellamy: A) Which internet node? B) What?

rade: The "The Internet (Inbound connections)" node in the UI.

Recall that outbound connections of a node are connections from that node to some other node. We list these in the adjacency.

From the perspective of the Internet node, the outbound connections / adjacency contains connections to containers etc. From the perspective of the user, these are inbound connections from the Internet into the system.

paulbellamy: Discussed in person. Of course the inbound internet node *has* no inbound connections. It originates all connections, so it only *has* outbound connections.

paulbellamy:
LGTM

PR #1852, opened 2016-09-05
Allow a nil RouteMatcher in instrumentation

tomwilkie:
So the instrumentation can be used for things other than gorilla.


paulbellamy:
LGTM

PR #1853, opened 2016-09-06
Add localstorage support for saving view state

davkal:
This PR adds support to remember the view state even after the browser
tab has been closed. This helps someone who is always looking at the
table view of a certain topology to which currently they have to perform
a minimum of 2 clicks.

* app view state is backed up in localStorage
* when visiting Scope with no URL view state, the localStorage state is
 used (i.e. the URL state has priority)

### DIFF
--- a/client/app/scripts/utils/router-utils.js
+++ b/client/app/scripts/utils/router-utils.js
@@ -1,6 +1,7 @@
 import page from 'page';
 
 import { route } from '../actions/app-actions';
+import { storageGet, storageSet } from './storage-utils';
 
 //
 // page.js won't match the routes below if ":state" has a slash in it, so replace those before we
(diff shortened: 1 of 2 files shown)


foot:
Works really well!

This is a bit edge casey but do you think we need some kind of localStorage recovery story? If you have a terminal open, notice some bug we've introduced which crashes the app...
- refresh will reload that bad state from the url
- clearing the url will now also load the app into a bad state from localStorage

ekimekim:
The recovery story in that case would be to tell your browser to forget localstorage for that url, no?

foot:
Yep its just not a great user experience might need to document that somewhere in the help section. I think we could wait until this happens to address it.
 - prefs > settings > adv. settings > content settings > all cookies and site data > search for cloud.weave > for each localStorage entry click it and then remove.
 - click the little `i` in the url bar, > cookies > localstorage > remove.
 - Have a `/reset` which has a `localStorage.clear && redirectTo('/')`

davkal:
Created ticket for recover story: #1872

PR #1858, opened 2016-09-12
circleci: push traffic control plugin image to docker hub

alepuccetti:
After this patch, circleci builds the traffic control plugin image and
pushes it on docker hub. Rename traffic control plugin docker image
to "weaveworks/scope-traffic-control-plugin".


davkal:
We've discussed where to put plugins, and decided that the scope repo is not the place for them (apart from the example plugins that are there for the purpose of being examples).

I suggest to host them by the maintainer in their own repo, which also governs their CI (with upload to Docker hub). Once you've set up that repo, could you open a new issue here so we can link to it?

From our side we'll add a bigger plugin section to this repo's readme where 3rd party plugins can be found.

PR #1860, opened 2016-09-13
in details panel, count connections only once

### LINKED ISSUE #1842: details panel connection counts are too high
To reproduce
```
$ scope launch
1c41ae18a677d54463e23e3adc5e63516b7409688634db79a9d8df5dcc9d635e
  [...]
$ google-chrome http://localhost:4040
$ docker run --rm --name=foo -ti alpine /bin/sh
/ # IP=$(getent hosts cloud.weave.works | cut -d ' ' -f 1)
/ # echo $IP
52.3.116.14
/ # nc $IP 80
```
Wait for the 'foo' container and connection to the Internet node to show up in scope and then bring up the details panel on the 'foo' node. You should see something like
screenshot_2016-09-01_22-43-19

Notice that it says there are *two* connections from the node to the Internet, when in reality there is only one.

This doesn't happen all the time, unfortunately, though right now for me it happens more often than not. Sometimes I initially see a count of 1 and it then increases to 2. And then sometimes it drops back to 1.

rade:
Fixes #1842

### DIFF
--- a/render/detailed/connections.go
+++ b/render/detailed/connections.go
@@ -97,12 +97,6 @@ func newConnectionCounters() *connectionCounters {
 func (c *connectionCounters) add(sourceEndpoint report.Node, n report.Node, node report.Node, port string, endpointID string, localAddr string) {
 	// We identify connections by their source endpoint, pre-NAT, to
 	// ensure we only count them once.
-	//
-	// There is some arbitrariness here: We may see the same
-	// connection multiple times, with different destination
-	// endpoints, due to DNATing; the (destination) port under which
-	// we track that connection is determined by the first destination
-	// endpoint we encounter.
 	connectionID := sourceEndpoint.ID
 	if copySourceEndpointID, _, ok := sourceEndpoint.Latest.LookupEntry("copy_of"); ok {
 		connectionID = copySourceEndpointID
@@ -165,7 +159,7 @@ func (c *connectionCounters) rows(r report.Report, ns report.Nodes, includeLocal
 }
 
 func incomingConnectionsSummary(topologyID string, r report.Report, n report.Node, ns report.Nodes) ConnectionsSummary {
-	localEndpointIDs := endpointChildIDsOf(n)
+	localEndpointIDs, localEndpointIDCopies := endpointChildIDsAndCopyMapOf(n)
 	counts := newConnectionCounters()
 
 	// For each node which has an edge TO me
@@ -177,6 +171,7 @@ func incomingConnectionsSummary(topologyID string, r report.Report, n report.Nod
 		// connections to that port.
 		for _, remoteEndpoint := range endpointChildrenOf(node) {
 			for _, localEndpointID := range remoteEndpoint.Adjacency.Intersection(localEndpointIDs) {
+				localEndpointID = canonicalEndpointID(localEndpointIDCopies, localEndpointID)
 				_, localAddr, port, ok := report.ParseEndpointNodeID(localEndpointID)
 				if !ok {
 					continue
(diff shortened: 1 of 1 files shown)


render/detailed/connections.go
```diff
@@ -85,25 +85,111 @@ func (row connection) ID() string {
...
+	// ensure we only count them once.
+	//
+	// There is some arbitrariness here: We may see the same
+	// connection multiple times, with different destination
+	// endpoints, due to DNATing; the (destination) port under which
+	// we track that connection is determined by the first destination
+	// endpoint we encounter.
+	//
+	// The arbitrariness is mitigated by the calling code by
+	// subsituting the destination endpoint with the "copy_of" if
```
2opremio: typo: substituting

render/detailed/connections.go
```diff
@@ -85,25 +85,111 @@ func (row connection) ID() string {
 	return fmt.Sprintf("%s:%s-%s:%s-%s", row.remoteNodeID, row.remoteAddr, row.localNodeID, row.localAddr, row.port)
 }
 
+type connections struct {
```
2opremio: `connection`, `connections` and `[]Connection` make the code a a bit hard to read. I would consider renaming this type to something more explicit like `connectionCounts`

render/detailed/connections.go
```diff
@@ -117,36 +203,37 @@ func incomingConnectionsSummary(topologyID string, r report.Report, n report.Nod
...
 
 		for _, localEndpoint := range localEndpoints {
 			_, localAddr, _, ok := report.ParseEndpointNodeID(localEndpoint.ID)
 			if !ok {
 				continue
 			}
-
 			for _, remoteEndpointID := range localEndpoint.Adjacency.Intersection(remoteEndpointIDs) {
+				if remoteEndpointIDCopy, ok := remoteEndpointIDCopies[remoteEndpointID]; ok {
+					remoteEndpointID = remoteEndpointIDCopy
```
2opremio: I think a comment is in order here to indicate we need this to make the destination unique.

render/detailed/connections.go
rade: That leads to stuttering. `connectionCounts.counts`.

2opremio: I think that the tradeoff is worth it though.

render/detailed/connections.go
rade: fixed

render/detailed/connections.go
rade: ok. fixed.

render/detailed/connections.go
rade: fixed, by creating a helper function with a comment.

rade:
@2opremio PTAL

render/detailed/connections.go
```diff
@@ -174,65 +246,36 @@ func endpointChildrenOf(n report.Node) []report.Node {
...
+// canonicalEndpointID returns the original endpoint ID of which id is
+// a "copy_of" (due to NATing), or, if the id is not a copy, the id
+// itself.
+//
+// This is used for determining a unique destination endpoint ID for a
+// connection, removing any arbitrariness in the destination port we
+// are associating with the connection when it is encountered multiple
+// times in the topology (with different destination endpoints, due to
+// DNATing).
+func canonicalEndpointID(copies map[string]string, id string) string {
```
2opremio: So much nicer, thanks!

2opremio:
LGTM

render/detailed/connections.go
```diff
@@ -85,25 +85,98 @@ func (row connection) ID() string {
...
+	connectionID := sourceEndpoint.ID
+	if copySourceEndpointID, _, ok := sourceEndpoint.Latest.LookupEntry("copy_of"); ok {
+		connectionID = copySourceEndpointID
+	}
+	if _, ok := c.counted[connectionID]; ok {
+		return
+	}
+	c.counted[connectionID] = struct{}{}
+	key := newConnection(n, node, port, endpointID, localAddr)
+	c.counts[key] = c.counts[key] + 1
```
peterbourgon:
```go
c.counts[key]++
```

rade: I didn't write that code.

rade: fixed in 94cd73c72

PR #1863, opened 2016-09-14
Add pseudo-nodes for known services

### LINKED ISSUE #1082: show well known services
Well-known hosts and/or services, such as Amazon RDS, SQS + others TBC should be displayed as a Host node, separate to 'The Internet' node(s)

2opremio:
Fixes #1082

### DIFF
--- a/probe/endpoint/dns_snooper_linux_amd64.go
+++ b/probe/endpoint/dns_snooper_linux_amd64.go
@@ -43,20 +43,11 @@ func NewDNSSnooper() (*DNSSnooper, error) {
 }
 
 func newPcapHandle() (*pcap.Handle, error) {
-	// TODO: use specific interfaces instead?
 	inactive, err := pcap.NewInactiveHandle("any")
 	if err != nil {
 		return nil, err
 	}
 	defer inactive.CleanUp()
-	if err = inactive.SetPromisc(true); err != nil {
-		return nil, err
-	}
-	// TODO: reduce the size of packets being copied? maybe an overoptimization
-	// if err = inactive.SetSnapLen(snaplen); err != nil {
-	//	return
-	// }
-
 	// pcap timeout blackmagic copied from Weave Net to reduce CPU consumption
 	// see https://github.com/weaveworks/weave/commit/025315363d5ea8b8265f1b3ea800f24df2be51a4
 	if err = inactive.SetTimeout(time.Duration(math.MaxInt64)); err != nil {
@@ -132,7 +123,7 @@ func (s *DNSSnooper) run() {
 		sll           layers.LinuxSLL
 	)
 
-	// assumes that the "any" interface in being used (see https://wiki.wireshark.org/SLL)
+	// assumes that the "any" interface is being used (see https://wiki.wireshark.org/SLL)
 	packetParser := gopacket.NewDecodingLayerParser(layers.LayerTypeLinuxSLL, &sll, &eth, &ip4, &ip6, &udp, &tcp, &dns)
 
 	for {
--- a/probe/endpoint/reporter.go
+++ b/probe/endpoint/reporter.go
@@ -21,6 +21,7 @@ const (
 	Conntracked     = "conntracked"
 	Procspied       = "procspied"
 	ReverseDNSNames = "reverse_dns_names"
+	SnoopedDNSNames = "snooped_dns_names"
 )
 
 // Reporter generates Reports containing the Endpoint topology.
@@ -195,11 +196,10 @@ func (r *Reporter) makeEndpointNode(namespaceID string, addr string, port uint16
 	node := report.MakeNodeWith(
 		report.MakeEndpointNodeID(r.hostID, namespaceID, addr, portStr),
 		map[string]string{Addr: addr, Port: portStr})
-	names := r.dnsSnooper.CachedNamesForIP(addr)
-	if resolvedNames, err := r.reverseResolver.get(addr); err == nil {
-		names = append(names, resolvedNames...)
+	if names := r.dnsSnooper.CachedNamesForIP(addr); len(names) > 0 {
+		node = node.WithSet(SnoopedDNSNames, report.MakeStringSet(names...))
 	}
-	if len(names) > 0 {
+	if names, err := r.reverseResolver.get(addr); err == nil && len(names) > 0 {
 		node = node.WithSet(ReverseDNSNames, report.MakeStringSet(names...))
 	}
 	if extra != nil {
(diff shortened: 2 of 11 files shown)


2opremio:
We already show pseudo nodes (with a cloud shape for now), for Dynamo and S3: the two AWS services for which we know reverse DNS lookups lead to an observable pattern.

screen shot 2016-09-14 at 11 09 18 am


Next step is adding a DNS packet inspector in the probes to extend it to IPs without revese lookups.

prog/main.go
```diff
@@ -227,6 +230,10 @@ func main() {
 	flag.StringVar(&flags.probe.weaveAddr, "probe.weave.addr", "127.0.0.1:6784", "IP address & port of the Weave router")
 	flag.StringVar(&flags.probe.weaveHostname, "probe.weave.hostname", app.DefaultHostname, "Hostname to lookup in WeaveDNS")
 
+	// DNS Snooper
+	flag.BoolVar(&flags.probe.dnsSnooperEnabled, "probe.dns-snooper.enabled", true, "Enable DNS Snooper (to identify external services being contacted)")
+	flag.StringVar(&flags.probe.dnsSnooperInterface, "probe.dns-snooper.interface", "any", "Networking interface on which to snoop")
+
```
rade: IMO this is overkill. YAGNI. And we want this 'on' by default.

rade: `eth0` is of no use when connected to a wlan, which many of our developer first-time users will be.

I'd leave it as 'all'.

probe/endpoint/reporter.go
```diff
@@ -193,9 +195,11 @@ func (r *Reporter) makeEndpointNode(namespaceID string, addr string, port uint16
...
 		report.MakeEndpointNodeID(r.hostID, namespaceID, addr, portStr),
 		map[string]string{Addr: addr, Port: portStr})
-	// In case we have a reverse resolution for the IP, we can use it for
-	// the name...
-	if names, err := r.reverseResolver.get(addr); err == nil {
+	names := r.dnsSnooper.CachedNamesForIP(addr)
+	if resolvedNames, err := r.reverseResolver.get(addr); err == nil {
+		names = append(names, resolvedNames...)
+	}
+	if len(names) > 0 {
```
rade: Stuffing the names we obtained from snooping forward resolutions into the `ReverseDNSNames` set is rather misleading.

So I reckon we should either rename it (but the app will also have to pay attention to the old name, for backward compatibility), or introduce a new set, just for the snooped names.

render/process.go
```diff
@@ -96,8 +96,8 @@ func MapEndpoint2Pseudo(n report.Node, local report.Networks) report.Nodes {
 
 	if ip := net.ParseIP(addr); ip != nil && !local.Contains(ip) {
 		// If the dstNodeAddr is not in a network local to this report, we emit an
-		// internet node
-		node = theInternetNode(n)
+		// external pseudoNode
+		node = toInternetNode(n)
```
rade: `externalNode(n)`?

render/theinternet.go
```diff
@@ -2,11 +2,50 @@ package render
...
 import (
 	"net"
+	"regexp"
 
 	"github.com/weaveworks/scope/probe/host"
 	"github.com/weaveworks/scope/report"
 )
 
+var (
+	// ServiceNodeIDPrefix is how the ID all service pseudo nodes begin
```
rade: "... the ID *of* all service pseudo nodes..."

render/theinternet.go
```diff
@@ -2,11 +2,50 @@ package render
...
+		regexp.MustCompile(`^.+\.amazonaws\.com$`),
+		regexp.MustCompile(`^.+\.googleapis\.com$`),
+	}
+
+	knownServicesExcluder = []*regexp.Regexp{
+		// We exclude ec2 machines because they are too generic
+		// and having separate nodes for them makes visualizations worse
+		regexp.MustCompile(`^ec2.*\.amazonaws\.com$`),
+	}
+)
```
rade: drop the plural. `knownServiceBlah`

rade: I wonder whether the excluders should be associated with a matcher, i.e. a matcher becomes a pair of regexps.

rade: The matchers/excludersd should really be obtained by the probes. For the moment I'll settle for a 'TODO' and associated issue.

rade: ah, but that would make them harder to merge (see next point).

prog/main.go
rade: *both* the "enabled" and the interface flags are overkill. Yes, we can think of reasons people may want them. But in practice I just don't think it's going to happen. And if it does then we can always add the configurability then.

probe/endpoint/dns_snooper_linux_amd64.go
```diff
@@ -0,0 +1,217 @@
...
+	return s, nil
+}
+
+func newPcapHandle(networkInterface string) (*pcap.Handle, error) {
+	inactive, err := pcap.NewInactiveHandle(networkInterface)
+	if err != nil {
+		return nil, err
+	}
+	defer inactive.CleanUp()
+	if err = inactive.SetPromisc(true); err != nil {
```
rade: do we need this?

prog/main.go
2opremio: OK, I will (reluctantly :) ) remove them.

probe/endpoint/reporter.go
2opremio: I will add another field.

render/process.go
2opremio: Sure, I wasn't happy with the name.

render/theinternet.go
2opremio: I disagree on this one, I think they should be configurable (or extensible) in the app UI.

probe/endpoint/dns_snooper_linux_amd64.go
2opremio: It works without it (at least in my dev VM), will remove.

2opremio:
@rade PTAL

render/theinternet.go
rade: That just ain't going to happen anytime soon. Way too many complications in doing that. It fundamentally changes the nature of the scope app - to date *all* information comes from the probes.

Meanwhile, releasing this feature w/o giving users any way to adjust the matchers/exclusions to their environment is just mean.

2opremio: I will add a TODO but not mentioning where it should be configurable.

2opremio: Done https://github.com/weaveworks/scope/issues/1876

rade:
needs some squashing. rest LGTM

PR #1868, opened 2016-09-15
fix inbound internet connections counts

### LINKED ISSUE #1867: inbound internet connection counts are too fine-grained
screenshot from 2016-09-15 11-21-43
The two connections are from the same address, so we should only have one row, with a count of 2.

This happens because the counter key for internet connections contains the endpoint id, which includes the port. That is fine for outbound internet connections, since there we do care about the port (though it's redundant, since the port is also part of the key directly). But not for inbound internet connections.

rade:
Fixes #1867.

### DIFF
--- a/render/detailed/node_test.go
+++ b/render/detailed/node_test.go
@@ -25,6 +25,10 @@ func child(t *testing.T, r render.Renderer, id string) detailed.NodeSummary {
 	return s.SummarizeMetrics()
 }
 
+func connectionID(nodeID string) string {
+	return fmt.Sprintf("%s-%s-%d", nodeID, "", 80)
+}
+
 func TestMakeDetailedHostNode(t *testing.T) {
 	renderableNodes := render.HostRenderer.Render(fixture.Report, nil)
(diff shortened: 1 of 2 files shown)


render/detailed/node_test.go
```diff
@@ -149,7 +149,7 @@ func TestMakeDetailedHostNode(t *testing.T) {
 				Columns:    detailed.NormalColumns,
 				Connections: []detailed.Connection{
 					{
-						ID:       fmt.Sprintf("%s:%s-%s:%s-%d", fixture.ServerHostNodeID, "", fixture.ClientHostNodeID, "", 80),
+						ID:       fmt.Sprintf("%s-%s-%d", fixture.ServerHostNodeID, "", 80),
```
rade: I did consider that, but it's not really much of an abstraction - the *contents* of the ID are far more likely to change than the formatting alone, so any change would entail a change in the function and the call sites. I suppose the compiler gives you some help in complaining about mismatches, so perhaps it's worth for that. Will give it a go.

rade: done. PTAL

PR #1875, opened 2016-09-19
show more details of a node's internet connections

### LINKED ISSUE #1713: show me the connections!
Scope collects heaps of information about connections, and uses it to lay out the topology graphs, yet it is remarkably difficult to actually get hold of even the most basic details of these connections. For example, I frequently find myself wondering "Why is my container X talking to the Internet? It shouldn't! *What* is it talking to?"

- clicking on the edge between my container and the "Internet (outbound)" node does precisely nothing. That would be the most obvious place to show connection information. See #1636 
- clicking on the node for my container brings up the details panel, with lovely lists of inbound and outbound connections... but this just gives me node names, ports and connection counts, *not IP addresses* or DNS names. So in case of outbound connections to the Internet, I am just being told that there 37 connections to the Internet on port 443. Marvellous.
- clicking on the "Internet (outbound)" node *does* show me IP addresses / DNS names for all the outbound connection destinations; but this is for *all* containers, so to find the connections for my container I have to search for my container name in a long list that isn't searchable and has the name truncated.

rade:
in the details panel, instead of showing connections to/from the internet as "TheInternet \<port\>", we now show "\<dns-name\>(\<ip_addr\>) \<port\>" (or just "\<ip_addr\> \<port\>" if we don't have a dns name).

Fixes #1713.

### DIFF
--- a/render/detailed/connections.go
+++ b/render/detailed/connections.go
@@ -57,9 +57,9 @@ func (s connectionsByID) Less(i, j int) bool { return s[i].ID < s[j].ID }
 
 // Intermediate type used as a key to dedupe rows
 type connection struct {
-	remoteNodeID string
-	addr         string // for internet nodes only: the address
-	port         string // destination port
+	remoteNodeID          string
+	remoteAddr, localAddr string // for internet nodes only
+	port                  string // destination port
 }
 
 type connectionCounters struct {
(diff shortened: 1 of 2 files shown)


rade:
before:
screenshot_2016-09-19_14-36-36
after:
screenshot_2016-09-19_14-33-25

I chose to denote the source/destination simply by the name/IP, and make them linkable - to TheInternet node. Is this too subtle/confusing? i.e. will users struggle to tell the difference in the list between these links and links to a "real" node? Do we care? NB: this will be made worse with #1863, which presumably denotes pseudo nodes very similar to the nodes here.

So...
1. leave as is
2. choose a different style (italics?)
3. drop the linking, under the assumption that users can always navigate to The Internet node via the topology
4. render as "TheInternet - name (addr)"
5. render the "name (addr)" in a separate column

I have pretty much ruled out 4/5 since they eat too much into the horizontal space we have available for display the info that actually matters to users, i.e. the "name (addr)".

rade:
@foot any thoughts on the "design" concern above?

foot:
Tough call, I think I'd go for dropping the link. 

Unless you feel there's a possible user flow there? But there doesn't seem to be too much utility in clicking through.

I dunno what would be ideal, perhaps all the internet rows grouped under a clickable "the internet" row, sort of tree style.

foot:
I guess you might wanna see what else that aws server is talking too in our scoped world.

Which is a bit rough: 
 - click dns/ip label
 - opens "the internet" card.
 - (probably click show-more first) and try and remember what you just clicked, find other things its talking to too.

I guess in that case you'd want individual pseudo node cards. Which could all be grouped on the topo into a super "The Internet" node, as discussed in some other ticket/feature I can't find right now...

In conclusion leaving/removing the link both sound okay for now.

Issue #1257, opened 2016-04-11
Show aggregated process metrics in containers/pods


2opremio:
As a user I would like to see things such as open file descriptors (or plugin metrics), at the container level (and pod level) instead of clicking on each process inside the container to view them.

<img width="1405" alt="container" src="https://cloud.githubusercontent.com/assets/2362916/14426352/428dc394-ffe6-11e5-8009-3827d77fc930.png">

<img width="1403" alt="screen shot 2016-04-11 at 1 02 54 pm" src="https://cloud.githubusercontent.com/assets/2362916/14426585/e9ee50da-ffe7-11e5-804d-8633c5313c1d.png">

In fact, obtaining the CPU/memory consumption of the container by aggregating the CPU/memory of processes would also fix https://github.com/weaveworks/scope/issues/1133

2opremio:
According to @tomwilkie we probably want to do this sort of aggregations through prometheus.

alepuccetti:
Aggregation Proposal
==================

Proposal on how to implementation data sources aggregation in scope.

Aggregation Architecture for Scope
----------------------------------
To describe an aggregation type we need 2 attributes:
- *AggregationPolicy*: define how to perform the aggregation default, max, min, sum, average, median, list,...
- *AggregationLevels*: where to perform the aggregation container, host, pod, replication-controller,...

All the data source types that will allow aggregation must be extended to include these 2 attributes.

```
type AggregationType {
	AggregationPolicy	string
	AggregationLevels	string
}

type MetricTemplate struct {
	ID				string  `json:"id"`
	Label			string  `json:"label,omitempty"`
	Format			string  `json:"format,omitempty"`
	Priority		float64 `json:"priority,omitempty"`
	Aggregations	[]AggregationType `json:"aggregations,omitempty"`
}
```

Allowing a list of _AggregationType_ makes the aggregation more flexible, it is possible to specify many policies for the same level (or different policies for different levels) for the same data source and as result, the UI could show multiple different aggregate values accordingly with the specified values.
I think that extending the templates is sufficient and the actual data point do not need to be changed because we want to allow aggregate a specific type of measure.

List of data structure to extend:
https://github.com/weaveworks/scope/blob/b27dc99146e8cdfa2ea6b2fca0795ed1a26ad996/report/metadata_template.go#L22
https://github.com/weaveworks/scope/blob/7b0f0cb158d3f8e468913ee5658ddbfae933b8b7/report/metric_template.go#L9
[...]

PR #1879, opened 2016-09-22
plugins/http-requests: documentation and tutorial

alepuccetti:
Write documentation and little tutorial for the HTTP requests plugin.
The goal is to introduce the plugin concepts to users and developers.


examples/plugins/http-requests/README.md
```diff
@@ -0,0 +1,89 @@
+# HTTP requests plugin
+
+The HTTP requests plugin tracks the number of incoming HTTP requests per process over time using eBPF.
+The UI shows the number of requests as a single value and graph which are updated every second.
+The HTTP requests plugin does not use any application-level instrumentation, it has a negligible performance toll (metrics are obtained in-kernel without any packet copying to userspace).
+
+**Note:** This plugin needs a [recent kernel version with eBPF support](https://github.com/iovisor/bcc/blob/master/INSTALL.md#kernel-configuration).
```
2opremio: Can you remove this note from the main README?

examples/plugins/http-requests/README.md
```diff
@@ -0,0 +1,89 @@
...
+It will not compile on current [dlite](https://github.com/nlf/dlite) and boot2docker hosts.
```
2opremio: Why? I would rather explain the requirement than where it works. (Kernel headers I presume)

examples/plugins/http-requests/test-http-requests.sh
```diff
@@ -0,0 +1,12 @@
...
+
+TARGET_IP="127.0.0.1"
+TARGET_PORT="8080"
+
+# long-term connection
+# so there is always at least one connection visible in the httpd view
+/bin/nc $TARGET_IP $TARGET_PORT &
+
+while true; do
+    curl http://$TARGET_IP:$TARGET_PORT &> /dev/null
```
2opremio: I would add a sleep after this

examples/plugins/http-requests/README.md
```diff
@@ -0,0 +1,89 @@
...
+```
+git clone git@github.com:weaveworks/scope.git
+cd scope
+make
+./scope launch
+```
+
+* Run the HTTP requests plugin `cd examples/plugins/http-requests ; make`.
+* In another terminal, run an `nginx` instance `docker run --name http-requests-nginx -p 8080:80 nginx`
+* Run `sh test-http-requests.sh`, this script will open a long-term connection with the web server and then start downloading the main page until Ctrl+c is sent by the user.
```
2opremio: Why? This will probably be confusing for the end user.

examples/plugins/http-requests/README.md
```diff
@@ -0,0 +1,89 @@
...
+* Go to the Weave Scope UI [http://localhost:4040](http://localhost:4040).
+* Open the `http-requests-nginx` container.
+* Click on `nginx: worker process` in the "Processes" table.
+
+## Report
```
2opremio: I would rename this to something like `plugin internals` and explain what a report is if you use the term.

examples/plugins/http-requests/README.md
alepuccetti: I have another PR almost ready about the main plugin README. I will do it there if it is ok for you.

examples/plugins/http-requests/README.md
alepuccetti: If there isn't any long-term connection, then the process detailed view where the graph is rendered doesn't show up.
I cold remove the explanation from the README and write few lines in a comment near the `nc` command in the test script.

2opremio: > I cold remove the explanation from the README and write few lines in a comment near the nc command in the test script.

Yep, I think that's better. I would also mention that it's a workaround until we fix https://github.com/weaveworks/scope/issues/1257

examples/plugins/http-requests/README.md
2opremio: sure

examples/plugins/http-requests/test-http-requests.sh
```diff
@@ -0,0 +1,13 @@
...
+TARGET_IP="127.0.0.1"
+TARGET_PORT="8080"
+
+# This ensures that there is always at least one connection visible in the httpd view.
+# This is a workaround until https://github.com/weaveworks/scope/issues/1257 is fixed.
+/bin/nc $TARGET_IP $TARGET_PORT &
+
+while true; do
+    curl http://$TARGET_IP:$TARGET_PORT &> /dev/null
+    sleep 0.042
```
2opremio: I think that sleep is too small. Why 42 milliseconds? HAL?

alepuccetti: I chose a small sleep just to have a good number of requests show,  I can put 200ms (~=5 request per second).
However, the new plan is to move this plugin to the new [weaveworks-plugins](https://github.com/weaveworks-plugins) organization.

2opremio: I would use 0.2, yep

alepuccetti:
Closing this pull request because this plugin is moved to a different repository  and renamed.

https://github.com/weaveworks-plugins/scope-http-statistics

PR #1892, opened 2016-09-26
Drop our label-col width heuristic for less wasted col. space

### LINKED ISSUE #1882: node names in connection list truncated unnecessarily
Compare
screenshot_2016-09-22_20-45-20
which looks lovely post #1874 - no space wasted, no unnecessary truncation - with
screenshot_2016-09-22_20-47-15
which has a bunch of truncated names in the connection list even though there is acres of space. Furthermore, horizontal space is wasted by the count column taking up way more space than necessary - it should be no wider than in the Internet node.

foot:
- The label column now flexes to the available space. (Old behaviour:
  label width was a function of # of cols).
- Most columns other than the label now have a fixed width specified
  (some large ones are left to share the avaiable space w/ the label)

Fixes #1882

### DIFF
--- a/client/app/scripts/components/node-details/node-details-table.js
+++ b/client/app/scripts/components/node-details/node-details-table.js
@@ -13,7 +13,7 @@ function isNumber(data) {
 const CW = {
   XS: '32px',
   S: '50px',
-  M: '64px',
+  M: '70px',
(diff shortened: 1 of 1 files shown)


davkal:
LGTM

PR #1894, opened 2016-09-26
Move truncation of docker ids to FE to allow full value in tooltips

### LINKED ISSUE #1139: Container view, detail panel, info table, ID & IMAGE ID fields contain truncated text as tooltip
Per https://github.com/weaveworks/scope/issues/1059.
The truncated text is not indicated with '...' and the tooltip is truncated text rather than untruncated text.
Same applies to the image id field.

truncated-tooltip

foot:
- Reveals full id if you search for it.
- Difficult to copy and paste the full id if you want it for some reason

Fixes #1139

### DIFF
--- a/client/app/scripts/components/matched-results.js
+++ b/client/app/scripts/components/matched-results.js
@@ -18,7 +18,8 @@ class MatchedResults extends React.Component {
           <span className="matched-results-match-label">
             {match.label}:
           </span>
-          <MatchedText text={text} match={match} maxLength={MAX_MATCH_LENGTH} />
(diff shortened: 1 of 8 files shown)


davkal:
Works well for container topology. On Containers By DNS, I'm not getting the full IDs on tooltips neither in the details panel title nor in the connections table.

screen shot 2016-09-27 at 13 41 18

Report: https://gist.github.com/davkal/610baf649b64ded6b06dc6a26a3eff99

foot:
I don't think we're responsible for those ones, I think they're just DNS names chosen w/ a similar truncation heuristic.

PR #1895, opened 2016-09-26
Add custom label-based filters in container view

CarltonSemple:
Added the ability to add container filters by editing an environment variable, and kept the ability to filter out weavescope system containers

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -15,48 +15,61 @@ import (
 	"github.com/weaveworks/scope/report"
 )
 
-const apiTopologyURL = "/api/topology/"
+const (
+	apiTopologyURL                     = "/api/topology/"
+	processesTopologyDescID            = "processes"
+	processesByNameTopologyDescID      = "processes-by-name"
+	containerLabelFiltersGroupID       = "container_label_filters_group"
+	containersTopologyDescID           = "containers"
+	containersByHostnameTopologyDescID = "containers-by-hostname"
+	containersByImageTopologyDescID    = "containers-by-image"
+	podsTopologyDescID                 = "pods"
+	replicaSetsTopologyDescID          = "replica-sets"
+	deploymentsTopologyDescID          = "deployments"
+	servicesTopologyDescID             = "services"
+	hostsTopologyDescID                = "hosts"
+)
 
 var (
-	topologyRegistry = &registry{
-		items: map[string]APITopologyDesc{},
-	}
-	k8sPseudoFilter = APITopologyOptionGroup{
+	topologyRegistry = MakeRegistry()
+	k8sPseudoFilter  = APITopologyOptionGroup{
 		ID:      "pseudo",
 		Default: "hide",
 		Options: []APITopologyOption{
-			{"show", "Show Unmanaged", nil, false},
-			{"hide", "Hide Unmanaged", render.IsNotPseudo, true},
+			{Value: "show", Label: "Show Unmanaged", filter: nil, filterPseudo: false},
+			{Value: "hide", Label: "Hide Unmanaged", filter: render.IsNotPseudo, filterPseudo: true},
 		},
 	}
 )
 
 func init() {
+	AddInitialTopologiesToRegistry(topologyRegistry)
+}
+
+// AddInitialTopologiesToRegistry does the initial setup for a Registry.
+// This is needed for testing.
+func AddInitialTopologiesToRegistry(registry *Registry) {
 	containerFilters := []APITopologyOptionGroup{
 		{
-			ID:      "system",
+			ID:      containerLabelFiltersGroupID,
 			Default: "application",
-			Options: []APITopologyOption{
-				{"system", "System containers", render.IsSystem, false},
-				{"application", "Application containers", render.IsApplication, false},
-				{"both", "Both", nil, false},
-			},
+			Options: []APITopologyOption{{Value: "all", Label: "All", filter: nil, filterPseudo: false}, {Value: "system", Label: "System Containers", filter: render.IsSystem, filterPseudo: false}, {Value: "notsystem", Label: "Application Containers", filter: render.IsApplication, filterPseudo: false}},
 		},
 		{
 			ID:      "stopped",
 			Default: "running",
 			Options: []APITopologyOption{
-				{"stopped", "Stopped containers", render.IsStopped, false},
-				{"running", "Running containers", render.IsRunning, false},
-				{"both", "Both", nil, false},
+				{Value: "stopped", Label: "Stopped containers", filter: render.IsStopped, filterPseudo: false},
+				{Value: "running", Label: "Running containers", filter: render.IsRunning, filterPseudo: false},
+				{Value: "both", Label: "Both", filter: nil, filterPseudo: false},
 			},
 		},
 		{
 			ID:      "pseudo",
 			Default: "hide",
 			Options: []APITopologyOption{
-				{"show", "Show Uncontained", nil, false},
-				{"hide", "Hide Uncontained", render.IsNotPseudo, true},
+				{Value: "show", Label: "Show Uncontained", filter: nil, filterPseudo: false},
+				{Value: "hide", Label: "Hide Uncontained", filter: render.IsNotPseudo, filterPseudo: true},
 			},
 		},
 	}
@@ -68,16 +81,16 @@ func init() {
 			Options: []APITopologyOption{
 				// Show the user why there are filtered nodes in this view.
 				// Don't give them the option to show those nodes.
-				{"hide", "Unconnected nodes hidden", nil, false},
+				{Value: "hide", Label: "Unconnected nodes hidden", filter: nil, filterPseudo: false},
 			},
 		},
 	}
 
 	// Topology option labels should tell the current state. The first item must
 	// be the verb to get to that state
-	topologyRegistry.add(
+	registry.Add(
 		APITopologyDesc{
-			id:          "processes",
+			id:          processesTopologyDescID,
 			renderer:    render.FilterUnconnected(render.ProcessWithContainerNameRenderer),
 			Name:        "Processes",
 			Rank:        1,
@@ -85,7 +98,7 @@ func init() {
 			HideIfEmpty: true,
 		},
 		APITopologyDesc{
-			id:          "processes-by-name",
+			id:          processesByNameTopologyDescID,
 			parent:      "processes",
 			renderer:    render.FilterUnconnected(render.ProcessNameRenderer),
 			Name:        "by name",
@@ -93,56 +106,56 @@ func init() {
 			HideIfEmpty: true,
 		},
 		APITopologyDesc{
-			id:       "containers",
+			id:       containersTopologyDescID,
 			renderer: render.ContainerWithImageNameRenderer,
 			Name:     "Containers",
 			Rank:     2,
 			Options:  containerFilters,
 		},
 		APITopologyDesc{
-			id:       "containers-by-hostname",
+			id:       containersByHostnameTopologyDescID,
 			parent:   "containers",
 			renderer: render.ContainerHostnameRenderer,
 			Name:     "by DNS name",
 			Options:  containerFilters,
 		},
 		APITopologyDesc{
-			id:       "containers-by-image",
+			id:       containersByImageTopologyDescID,
 			parent:   "containers",
 			renderer: render.ContainerImageRenderer,
 			Name:     "by image",
 			Options:  containerFilters,
 		},
 		APITopologyDesc{
-			id:          "pods",
+			id:          podsTopologyDescID,
 			renderer:    render.PodRenderer,
 			Name:        "Pods",
 			Rank:        3,
 			HideIfEmpty: true,
 		},
 		APITopologyDesc{
-			id:          "replica-sets",
+			id:          replicaSetsTopologyDescID,
 			parent:      "pods",
 			renderer:    render.ReplicaSetRenderer,
 			Name:        "replica sets",
 			HideIfEmpty: true,
 		},
 		APITopologyDesc{
-			id:          "deployments",
+			id:          deploymentsTopologyDescID,
 			parent:      "pods",
 			renderer:    render.DeploymentRenderer,
 			Name:        "deployments",
 			HideIfEmpty: true,
 		},
 		APITopologyDesc{
-			id:          "services",
+			id:          servicesTopologyDescID,
 			parent:      "pods",
 			renderer:    render.PodServiceRenderer,
 			Name:        "services",
 			HideIfEmpty: true,
 		},
 		APITopologyDesc{
-			id:       "hosts",
+			id:       hostsTopologyDescID,
 			renderer: render.HostRenderer,
 			Name:     "Hosts",
 			Rank:     4,
@@ -165,10 +178,10 @@ func kubernetesFilters(namespaces ...string) APITopologyOptionGroup {
 			options.Default = namespace
 		}
 		options.Options = append(options.Options, APITopologyOption{
-			namespace, namespace, render.IsNamespace(namespace), false,
+			Value: namespace, Label: namespace, filter: render.IsNamespace(namespace), filterPseudo: false,
 		})
 	}
-	options.Options = append(options.Options, APITopologyOption{"all", "All Namespaces", nil, false})
+	options.Options = append(options.Options, APITopologyOption{Value: "all", Label: "All Namespaces", filter: nil, filterPseudo: false})
 	return options
 }
 
@@ -192,7 +205,7 @@ func updateFilters(rpt report.Report, topologies []APITopologyDesc) []APITopolog
 	}
 	sort.Strings(ns)
 	for i, t := range topologies {
-		if t.id == "pods" || t.id == "services" || t.id == "deployments" || t.id == "replica-sets" {
+		if t.id == podsTopologyDescID || t.id == servicesTopologyDescID || t.id == deploymentsTopologyDescID || t.id == replicaSetsTopologyDescID {
 			topologies[i] = updateTopologyFilters(t, []APITopologyOptionGroup{
 				kubernetesFilters(ns...), k8sPseudoFilter,
 			})
@@ -210,10 +223,23 @@ func updateTopologyFilters(t APITopologyDesc, options []APITopologyOptionGroup) 
 	return t
 }
 
-// registry is a threadsafe store of the available topologies
-type registry struct {
+// MakeAPITopologyOption provides an external interface to the package for creating an APITopologyOption.
+func MakeAPITopologyOption(value string, label string, filterFunc render.FilterFunc, pseudo bool) APITopologyOption {
+	return APITopologyOption{Value: value, Label: label, filter: filterFunc, filterPseudo: pseudo}
+}
+
+// Registry is a threadsafe store of the available topologies
+type Registry struct {
 	sync.RWMutex
 	items map[string]APITopologyDesc
+}
+
+// MakeRegistry returns a new Registry
+func MakeRegistry() *Registry {
+	newRegistry := &Registry{
+		items: map[string]APITopologyDesc{},
+	}
+	return newRegistry
 }
 
 // APITopologyDesc is returned in a list by the /api/topology handler.
@@ -261,7 +287,27 @@ type topologyStats struct {
 	FilteredNodes      int `json:"filtered_nodes"`
 }
 
-func (r *registry) add(ts ...APITopologyDesc) {
+// AddContainerFilters adds to the default Registry (topologyRegistry)'s containerFilters
+func AddContainerFilters(newFilters ...APITopologyOption) {
+	topologyRegistry.AddContainerFilters(newFilters...)
+}
+
+// AddContainerFilters adds container filters to this Registry
+func (r *Registry) AddContainerFilters(newFilters ...APITopologyOption) {
+	r.Lock()
+	defer r.Unlock()
+	for _, key := range []string{containersTopologyDescID, containersByHostnameTopologyDescID, containersByImageTopologyDescID} {
+		for i := range r.items[key].Options {
+			if r.items[key].Options[i].ID == containerLabelFiltersGroupID {
+				r.items[key].Options[i].Options = append(r.items[key].Options[i].Options, newFilters...)
+				break
+			}
+		}
+	}
+}
+
+// Add inserts a topologyDesc to the Registry's items map
+func (r *Registry) Add(ts ...APITopologyDesc) {
 	r.Lock()
 	defer r.Unlock()
 	for _, t := range ts {
@@ -277,14 +323,14 @@ func (r *registry) add(ts ...APITopologyDesc) {
 	}
 }
 
-func (r *registry) get(name string) (APITopologyDesc, bool) {
+func (r *Registry) get(name string) (APITopologyDesc, bool) {
 	r.RLock()
 	defer r.RUnlock()
 	t, ok := r.items[name]
 	return t, ok
 }
 
-func (r *registry) walk(f func(APITopologyDesc)) {
+func (r *Registry) walk(f func(APITopologyDesc)) {
 	r.RLock()
 	defer r.RUnlock()
 	descs := []APITopologyDesc{}
@@ -301,7 +347,7 @@ func (r *registry) walk(f func(APITopologyDesc)) {
 }
 
 // makeTopologyList returns a handler that yields an APITopologyList.
-func (r *registry) makeTopologyList(rep Reporter) CtxHandlerFunc {
+func (r *Registry) makeTopologyList(rep Reporter) CtxHandlerFunc {
 	return func(ctx context.Context, w http.ResponseWriter, req *http.Request) {
 		report, err := rep.Report(ctx)
 		if err != nil {
@@ -312,14 +358,14 @@ func (r *registry) makeTopologyList(rep Reporter) CtxHandlerFunc {
 	}
 }
 
-func (r *registry) renderTopologies(rpt report.Report, req *http.Request) []APITopologyDesc {
+func (r *Registry) renderTopologies(rpt report.Report, req *http.Request) []APITopologyDesc {
 	topologies := []APITopologyDesc{}
 	req.ParseForm()
 	r.walk(func(desc APITopologyDesc) {
-		renderer, decorator, _ := r.rendererForTopology(desc.id, req.Form, rpt)
+		renderer, decorator, _ := r.RendererForTopology(desc.id, req.Form, rpt)
 		desc.Stats = decorateWithStats(rpt, renderer, decorator)
 		for i, sub := range desc.SubTopologies {
-			renderer, decorator, _ := r.rendererForTopology(sub.id, req.Form, rpt)
+			renderer, decorator, _ := r.RendererForTopology(sub.id, req.Form, rpt)
 			desc.SubTopologies[i].Stats = decorateWithStats(rpt, renderer, decorator)
 		}
 		topologies = append(topologies, desc)
@@ -349,7 +395,8 @@ func decorateWithStats(rpt report.Report, renderer render.Renderer, decorator re
 	}
 }
 
-func (r *registry) rendererForTopology(topologyID string, values url.Values, rpt report.Report) (render.Renderer, render.Decorator, error) {
+// RendererForTopology ..
+func (r *Registry) RendererForTopology(topologyID string, values url.Values, rpt report.Report) (render.Renderer, render.Decorator, error) {
 	topology, ok := r.get(topologyID)
 	if !ok {
 		return nil, nil, fmt.Errorf("topology not found: %s", topologyID)
@@ -393,7 +440,7 @@ func captureReporter(rep Reporter, f reporterHandler) CtxHandlerFunc {
 
 type rendererHandler func(context.Context, render.Renderer, render.Decorator, report.Report, http.ResponseWriter, *http.Request)
 
-func (r *registry) captureRenderer(rep Reporter, f rendererHandler) CtxHandlerFunc {
+func (r *Registry) captureRenderer(rep Reporter, f rendererHandler) CtxHandlerFunc {
 	return func(ctx context.Context, w http.ResponseWriter, req *http.Request) {
 		topologyID := mux.Vars(req)["topology"]
 		if _, ok := r.get(topologyID); !ok {
@@ -406,7 +453,7 @@ func (r *registry) captureRenderer(rep Reporter, f rendererHandler) CtxHandlerFu
 			return
 		}
 		req.ParseForm()
-		renderer, decorator, err := r.rendererForTopology(topologyID, req.Form, rpt)
+		renderer, decorator, err := r.RendererForTopology(topologyID, req.Form, rpt)
 		if err != nil {
 			respondWith(w, http.StatusInternalServerError, err)
 			return
--- a/app/api_topologies_test.go
+++ b/app/api_topologies_test.go
@@ -3,6 +3,7 @@ package app_test
 import (
 	"bytes"
 	"net/http/httptest"
+	"net/url"
 	"testing"
 	"time"
 
@@ -12,8 +13,15 @@ import (
 
 	"github.com/weaveworks/scope/app"
 	"github.com/weaveworks/scope/probe/kubernetes"
+	"github.com/weaveworks/scope/render"
+	"github.com/weaveworks/scope/render/detailed"
 	"github.com/weaveworks/scope/report"
 	"github.com/weaveworks/scope/test/fixture"
+)
+
+const (
+	containerLabelFiltersGroupID    = "container_label_filters_group"
+	customAPITopologyOptionFilterID = "containerLabelFilter0"
 )
 
 func TestAPITopology(t *testing.T) {
@@ -48,6 +56,56 @@ func TestAPITopology(t *testing.T) {
 			t.Errorf("NonpseudoNodeCount isn't positive for %s: %d", topology.Name, have)
 		}
 	}
+}
+
+func TestContainerLabelFilter(t *testing.T) {
+	topologySummaries, err := getTestContainerLabelFilterTopologySummary(t, false)
+	if err != nil {
+		t.Fatalf("Topology Registry Report error: %s", err)
+	}
+
+	// only the filtered container with fixture.TestLabelKey1 should be present
+	equals(t, 1, len(topologySummaries))
+	for key := range topologySummaries {
+		equals(t, fixture.ClientContainerID+";<container>", key)
+	}
+}
+
+func TestContainerLabelFilterExclude(t *testing.T) {
+	topologySummaries, err := getTestContainerLabelFilterTopologySummary(t, true)
+	if err != nil {
+		t.Fatalf("Topology Registry Report error: %s", err)
+	}
+
+	// all containers but the excluded container should be present
+	for key := range topologySummaries {
+		if fixture.ServerContainerNodeID+";<container>" == key {
+			t.Errorf("TestAPITopologyNegativeContainerLabelFilter Failed. Expected to not find " + fixture.ServerContainerNodeID + ";<container> in report")
+		}
+	}
+}
+
+func getTestContainerLabelFilterTopologySummary(t *testing.T, exclude bool) (detailed.NodeSummaries, error) {
+	ts := topologyServer()
+	defer ts.Close()
+
+	topologyRegistry := app.MakeRegistry()
+	app.AddInitialTopologiesToRegistry(topologyRegistry)
+
+	if exclude == true {
+		topologyRegistry.AddContainerFilters(app.MakeAPITopologyOption(customAPITopologyOptionFilterID, "title", render.DoesNotHaveLabel(fixture.TestLabelKey2, fixture.ApplicationLabelValue2), false))
+	} else {
+		topologyRegistry.AddContainerFilters(app.MakeAPITopologyOption(customAPITopologyOptionFilterID, "title", render.HasLabel(fixture.TestLabelKey1, fixture.ApplicationLabelValue1), false))
+	}
+
+	urlvalues := url.Values{}
+	urlvalues.Set(containerLabelFiltersGroupID, customAPITopologyOptionFilterID)
+	renderer, decorator, err := topologyRegistry.RendererForTopology("containers", urlvalues, fixture.Report)
+	if err != nil {
+		return nil, err
+	}
+
+	return detailed.Summaries(fixture.Report, renderer.Render(fixture.Report, decorator)), nil
 }
 
 func TestAPITopologyAddsKubernetes(t *testing.T) {
--- a/app/api_topologies_test.go
+++ b/app/api_topologies_test.go
@@ -67,7 +67,7 @@ func TestContainerLabelFilter(t *testing.T) {
 	// only the filtered container with fixture.TestLabelKey1 should be present
 	equals(t, 1, len(topologySummaries))
 	for key := range topologySummaries {
-		equals(t, fixture.ClientContainerID+";<container>", key)
+		equals(t, report.MakeContainerNodeID(fixture.ClientContainerID), key)
 	}
 }
 
@@ -79,8 +79,8 @@ func TestContainerLabelFilterExclude(t *testing.T) {
 
 	// all containers but the excluded container should be present
 	for key := range topologySummaries {
-		if fixture.ServerContainerNodeID+";<container>" == key {
-			t.Errorf("TestAPITopologyNegativeContainerLabelFilter Failed. Expected to not find " + fixture.ServerContainerNodeID + ";<container> in report")
+		if report.MakeContainerNodeID(fixture.ServerContainerNodeID) == key {
+			t.Errorf("TestAPITopologyNegativeContainerLabelFilter Failed. Expected to not find " + report.MakeContainerNodeID(fixture.ServerContainerNodeID) + " in report")
 		}
 	}
 }
--- a/app/api_topology.go
+++ b/app/api_topology.go
@@ -107,7 +107,7 @@ func handleWebsocket(
 			log.Errorf("Error generating report: %v", err)
 			return
 		}
-		renderer, decorator, err := topologyRegistry.rendererForTopology(topologyID, r.Form, report)
+		renderer, decorator, err := topologyRegistry.RendererForTopology(topologyID, r.Form, report)
 		if err != nil {
 			log.Errorf("Error generating report: %v", err)
 			return
--- a/prog/main.go
+++ b/prog/main.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -16,6 +17,7 @@ import (
 	"github.com/weaveworks/scope/common/xfer"
 	"github.com/weaveworks/scope/probe/appclient"
 	"github.com/weaveworks/scope/probe/kubernetes"
+	"github.com/weaveworks/scope/render"
 	"github.com/weaveworks/weave/common"
 )
 
@@ -33,6 +35,8 @@ var (
 		kubernetesPasswordFlag,
 		kubernetesTokenFlag,
 	}
+	colonFinder         = regexp.MustCompile(`[^\\](:)`)
+	unescapeBackslashes = regexp.MustCompile(`\\(.)`)
 )
 
 type prefixFormatter struct {
@@ -138,6 +142,54 @@ type appFlags struct {
 	consulInf       string
 }
 
+type containerLabelFiltersFlag struct {
+	apiTopologyOptions []app.APITopologyOption
+	filterNumber       int
+	filterIDPrefix     string
+	exclude            bool
+}
+
+func (c *containerLabelFiltersFlag) String() string {
+	return fmt.Sprint(c.apiTopologyOptions)
+}
+
+func (c *containerLabelFiltersFlag) Set(flagValue string) error {
+	filterID := fmt.Sprintf(c.filterIDPrefix+"%d", c.filterNumber)
+	newAPITopologyOption, err := c.toAPITopologyOption(flagValue, filterID)
+	if err != nil {
+		return err
+	}
+	c.filterNumber++
+
+	c.apiTopologyOptions = append(c.apiTopologyOptions, newAPITopologyOption)
+	return nil
+}
+
+func (c *containerLabelFiltersFlag) toAPITopologyOption(flagValue string, filterID string) (app.APITopologyOption, error) {
+	indexRanges := colonFinder.FindAllStringIndex(flagValue, -1)
+	if len(indexRanges) != 1 {
+		if len(indexRanges) == 0 {
+			return app.APITopologyOption{}, fmt.Errorf("No unescaped colon found. This is needed to separate the title from the label")
+		}
+		return app.APITopologyOption{}, fmt.Errorf("Multiple unescaped colons. Escape colons that are part of the title and label")
+	}
+	splitIndices := indexRanges[0]
+	titleStringEscaped := flagValue[:splitIndices[0]+1]
+	labelStringEscaped := flagValue[splitIndices[1]:]
+	containerFilterTitle := unescapeBackslashes.ReplaceAllString(titleStringEscaped, `$1`)
+	containerFilterLabel := unescapeBackslashes.ReplaceAllString(labelStringEscaped, `$1`)
+	labelKeyValuePair := strings.Split(containerFilterLabel, "=")
+	if len(labelKeyValuePair) != 2 {
+		return app.APITopologyOption{}, fmt.Errorf("Docker label in the app.container-label-filter(-exclude) flag isn't in the correct key=value format")
+	}
+
+	filterFunction := render.HasLabel
+	if c.exclude {
+		filterFunction = render.DoesNotHaveLabel
+	}
+	return app.MakeAPITopologyOption(filterID, containerFilterTitle, filterFunction(labelKeyValuePair[0], labelKeyValuePair[1]), false), nil
+}
+
 func logCensoredArgs() {
 	var prettyPrintedArgs string
 	// We show the flags followed by the args. This may change the original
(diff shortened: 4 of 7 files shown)


2opremio:
This PR incorporates unrelated changes already part of master, can you please clean up the PR before we review it?

CarltonSemple:
@2opremio alright, I got rid of the extra commits

2opremio:
Thanks @CarltonSemple I've had a quick look and I don't like using env variables and passing json around.

How about passing one (or many) command line arguments like:

```
scope launch --app.label-filter=id1:title1:label1 --app.label-filer=id2:title2:label2
```

To avoid dealing with escaping of the colons we could start by only accepting alphanumeric characters in all `id`, `title` and `label`.

Thoughts?

CarltonSemple:
@2opremio Good idea. I didn't like the escaping either. I'll try that

2opremio:
Great. Also, I think we can ditch the id field and generate it automatically. User's shouldn't worry about that.

2opremio:
That is:

```
scope launch --app.label-filter=title1:label1 --app.label-filter=title2:label2
```

2opremio:
Two more things:

* I would call the arguments `--app.container-label-filter` (since they only apply to the container topology).
* As per our label in the default filter (`works.weave.role=system`) we will need to accept non-alphanumeric characters and escape the `:` :S . Any other suggestions are welcome.

2opremio:
Still getting lint errors:

```
app/api_topologies.go:34:1: comment on exported function InitializeTopologies should be of the form "InitializeTopologies ..."

app/load_container_filters.go:20:1: comment on exported var ContainerLabelFlags should be of the form "ContainerLabelFlags ..."
```

You should be able to run the lint checks locally with `make lint`

prog/main.go
```diff
@@ -259,6 +261,8 @@ func main() {
 
 	flag.Parse()
 
+	app.InitializeTopologies()
```
2opremio: Instead of using `InitializeTopologies()` I would leave the `init()` from `api_topologies.go` as it was and instead incorporate a function to add a container filter (e.g. `AddTopologyOptions(id string, options ...APITopologyOption)` ) to the topologies registry. Then you can simply invoke `AddTopologyOptions()` from here.

render/filters.go
```diff
@@ -244,6 +245,24 @@ func IsApplication(n report.Node) bool {
 // IsSystem checks if the node is a "system" node
 var IsSystem = Complement(IsApplication)
 
+// IsDesired checks if the node has the desired label
+func IsDesired(label string) FilterFunc {
```
2opremio: Please use a more descriptive name: `HasLabel` for instance

render/filters.go
```diff
@@ -244,6 +245,24 @@ func IsApplication(n report.Node) bool {
...
+	return func(n report.Node) bool {
+		desiredKeyValue := strings.Split(label, "=")
```
2opremio: All this parsing should be done earlier, when parsing the flags. Scope should fail to start if the syntax is incorrect.

2opremio: You can include it in the label field.

app/load_container_filters.go
```diff
@@ -0,0 +1,66 @@
+package app
```
2opremio: I'm not sure it's worth having a separate module just for this, particularly once you simplify the id generation.

app/load_container_filters.go
```diff
@@ -0,0 +1,66 @@
...
+	// Add option to view weave system containers
+	system := APITopologyOption{Value: "system", Label: "System Containers", filter: render.IsSystem, filterPseudo: false}
+	toptions = append(toptions, system)
+
+	// Add option to not view weave system containers
+	notSystem := APITopologyOption{Value: "notsystem", Label: "Application Containers", filter: render.IsApplication, filterPseudo: false}
+	toptions = append(toptions, notSystem)
+
+	for _, f := range ContainerLabelFlags {
+		titleLabel := strings.Split(f, ":")
```
2opremio: This is not good enough since labels may contain colons.

app/load_container_filters.go
```diff
@@ -0,0 +1,66 @@
...
+	"time"
+)
+
+type arrayFlags []string
+
+func (i *arrayFlags) String() string {
+	return "my string representation"
+}
+func (i *arrayFlags) Set(value string) error {
+	*i = append(*i, value)
```
2opremio: You should do the parsing here.

2opremio:
Thanks a lot for your contribution. Apart from the review comments I am missing unit tests.

app/api_topologies.go
```diff
@@ -31,10 +31,27 @@ var (
 	}
 )
 
+func init() {
+	// Initialize topologies here since the tests require this
+	InitializeTopologies()
```
2opremio: Please, just inline the code (i.e. leaving it as it was before).

app/load_container_filters.go
CarltonSemple: @2opremio 
So you mean to have the user insert an escape?
Like this:
--app.container-label-filter=Ti:tle\\:weave.role=sys:tem

2opremio:
No, the other way around.
--app.container-label-filter=Ti\:tle:weave.role=sys\:tem (assuming that the
title you want is Ti:tle and the label weave.role=sys:tem)

CarltonSemple:
I was trying the regex route, but couldn't figure it out

prog/main.go
```diff
@@ -144,12 +144,37 @@ func (i *filterFlags) String() string {
...
-	// TODO - allow escaping
-	v := app.CreateFilterOption(fmt.Sprintf("cmdlinefilter%d", len(*i)), titleLabel[0], render.HasLabel(titleLabel[1]))
+	//fmt.Printf("%q\n",value)
+	u := fmt.Sprintf("%q", value)
+	u = u[1 : len(u)-1]
+
+	soloColonFound := false
+	var title string
+	var label string
+	for k := 1; k < len(u); k++ {
```
2opremio: Why not simply use regexp `[^\\](:)` ? i.e. find a colon (first subgroup) not preceded by a backslash

prog/main.go
```diff
@@ -144,12 +144,37 @@ func (i *filterFlags) String() string {
...
+					return nil
+				}
+				soloColonFound = true
+				title = u[0:k]
+				label = u[k+1 : len(u)]
+			}
+		}
+	}
+
+	title = strings.Replace(title, "\\:", ":", -1)
```
2opremio: Please use backticks to avoid escaping backslashes (i.e. `` `\:` ``)

2opremio: Also, please factor out the escaping in a separate function (instead of invoking replace 4 times for title a label)

2opremio:
> I was trying the regex route, but couldn't figure it out

It can be done, see my comment above.

prog/main.go
```diff
@@ -134,6 +135,53 @@ type appFlags struct {
...
+	title = strings.Replace(title, "\\", "", -1)
+	label = strings.Replace(label, "\\:", ":", -1)
+	label = strings.Replace(label, "\\", "", -1)
+
+	v := app.CreateFilterOption(fmt.Sprintf("cmdlinefilter%d", len(*i)), title, render.HasLabel(label))
+	app.ContainerOpts = append(app.ContainerOpts, v)
```
2opremio: This is unsafe since it can happen in a different thread. Please add a method to the topology registry instead, using its lock.

Also, this should happen out of this function since it's not part of setting the flag per se.

prog/main.go
2opremio: In addition, you can do it in single regexp replacement https://play.golang.org/p/Bs-N5O-Z-Q

(no need to create an extra function if you do it like this)

prog/main.go
CarltonSemple: @2opremio i'll move the code in lines 177 & 178, but ultimately I do use topologyRegistry's add function inside of the RefreshTopologyOptions() function in api_topologies.go

2opremio: > I guess I can move the code in that function into a new topologyRegistry function, using the lock

Yep

2opremio: > ultimately I do use topologyRegistry's add function inside of the RefreshTopologyOptions() function in api_topologies.go

I still don't understand why you need to build the full options from main. Why not build the options like it was done before (just in init) and simply modify the container options through a topology registry function built specifically for that (which you can also call from main)?

CarltonSemple: I'm making the topology registry function now, and I'll call it from main. Also, if I move the code from lines 177 & 178, I might as well take everything out of the Set function.

CarltonSemple: @2opremio Now that I'm trying it again, I remember why I recreated parts of the registry. In golang there's an issue in assigning to fields of a map element. See here https://github.com/golang/go/issues/3117
So for example, I can't do `r.items["containers-by-image"].Options = ContainerOpts`

CarltonSemple: @2opremio you actually can't. I got this error: `app/api_topologies.go:296: cannot assign to struct field r.items["containers"].Options in map`
So in the latest commit I did `var c = r.items["containers"]` and went from there

2opremio: You are right, a bit more involved, but doable in any case.

app/api_topologies.go
```diff
@@ -272,39 +268,34 @@ type topologyStats struct {
 	FilteredNodes      int `json:"filtered_nodes"`
 }
 
-// RefreshTopologyOptions loads the container filter options into the topology
-func RefreshTopologyOptions() { //AddTopologyOptions(options ...APITopologyOption) {
-	/*for _, opt := range options {
-		ContainerOpts = append(ContainerOpts, opt)
-	}*/
+// AddContainerFilters adds to the topology registry's containerFilters
+func AddContainerFilters(newOptions []APITopologyOption) {
```
2opremio: I would make this `func AddContainerFilters(newOptions ...APITopologyOption)`

app/api_topologies.go
```diff
@@ -255,6 +268,30 @@ type topologyStats struct {
...
+	topologyRegistry.addContainerFilters(newOptions)
+}
+
+func (r *registry) addContainerFilters(newOptions []APITopologyOption) {
+	r.Lock()
+	defer r.Unlock()
+
+	for _, o := range newOptions {
+		ContainerOpts = append(ContainerOpts, o)
```
2opremio: you can simply do `ContainerOpts = append(ContainerOpts, newOptions...)` without the loop

2opremio: https://gobyexample.com/variadic-functions

app/api_topologies.go
```diff
@@ -25,50 +25,58 @@ var (
...
-			{"show", "Show Unmanaged", nil, false},
-			{"hide", "Hide Unmanaged", render.IsNotPseudo, true},
+			{Value: "show", Label: "Show Unmanaged", filter: nil, filterPseudo: false},
+			{Value: "hide", Label: "Hide Unmanaged", filter: render.IsNotPseudo, filterPseudo: true},
 		},
 	}
 )
 
+// ContainerOpts holds the container topology view options such as "all" and "system"
+var ContainerOpts []APITopologyOption
```
2opremio: I don't see a  reason to use global variables, you can simply append to the map.

app/api_topologies.go
2opremio: (i.e. items in the registry)

app/api_topologies.go
2opremio: (i.e. items in the registry)

app/api_topologies.go
```diff
@@ -255,6 +268,30 @@ type topologyStats struct {
...
+	defer r.Unlock()
+
+	for _, o := range newOptions {
+		ContainerOpts = append(ContainerOpts, o)
+	}
+
+	var c = r.items["containers"]
+	var h = r.items["containers-by-hostname"]
+	var i = r.items["containers-by-image"]
+	c.Options[0].Options = ContainerOpts
```
CarltonSemple: I'm not sure why it was made like this, but containerFilters is of type []APITopologyOptionGroup. However, each option group has an ID, as if it was intended to be a map instead of an array

prog/main.go
```diff
@@ -134,6 +136,52 @@ type appFlags struct {
 	consulInf       string
 }
 
+type containerLabelFilterFlagArray []string
+
+func (i *containerLabelFilterFlagArray) String() string {
+	var s string
+	for _, x := range *i {
+		s = x
```
CarltonSemple: yeah i fixed it locally

prog/main.go
```diff
@@ -134,6 +136,52 @@ type appFlags struct {
...
+	var newTopologyOptions []app.APITopologyOption
+	i := 0
+	for _, value := range containerLabels {
+		u := fmt.Sprintf("%q", value)
+		u = u[1 : len(u)-1]
+
+		colonFinder := regexp.MustCompile(`[^\\](:)`)
+		indices := colonFinder.FindStringIndex(u)
+		if len(indices) == 0 {
+			fmt.Println("No unescaped colon found. This is needed to separate the title from the label")
```
2opremio: Please do not use `fmt`. Use `log`

prog/main.go
```diff
@@ -134,6 +136,52 @@ type appFlags struct {
 	consulInf       string
 }
 
+type containerLabelFilterFlagArray []string
```
2opremio: `s/containerLabelFilterFlagArray/containerLabelFilters/` and instead of using an array of strings use an array of the final parsed values (which you should parse in `Set()`. Something like

```go
type containerLabelFilter struct {
    title string
    label struct {
      key string
      value string
    }
}

type containerLabelFilters []containerLabelFilter
```

Then you can parse it in the `Set()` method and add a `toAPITopologyOption()` method (replacing `makeContainerFiltersFromFlags`)

prog/main.go
```diff
@@ -134,6 +136,52 @@ type appFlags struct {
...
+		label := unescapeColons.ReplaceAllString(u[indices[1]:], `:`)
+
+		v := app.MakeFilterOption(fmt.Sprintf("cmdlinefilter%d", i), title, render.HasLabel(label), false)
+		newTopologyOptions = append(newTopologyOptions, v)
+	}
+	return newTopologyOptions
+}
+
+// containerLabelFlags is set from the command line argument app.container-label-filter in /scope/prog/main.go
+var containerLabelFlags containerLabelFilterFlagArray
```
2opremio: Please don't use global variables.

prog/main_test.go
```diff
@@ -0,0 +1,22 @@
...
+	containerLabelFlags = append(containerLabelFlags, `ti\:tle2:lab\:el=2`)
+	apiOptions := makeContainerFiltersFromFlags(containerLabelFlags)
+
+	fmt.Println(apiOptions)
+
+	equals(t, 2, len(apiOptions))
+	equals(t, `title1`, apiOptions[0].Value)
+	equals(t, `label=1`, apiOptions[0].Label)
+	equals(t, `ti:tle2`, apiOptions[1].Value)
+	equals(t, `lab:el=2`, apiOptions[1].Value)
```
2opremio: Please add a test with spaces

render/filters.go
```diff
@@ -244,6 +245,24 @@ func IsApplication(n report.Node) bool {
 // IsSystem checks if the node is a "system" node
 var IsSystem = Complement(IsApplication)
 
+// HasLabel checks if the node has the desired docker label
+func HasLabel(label string) FilterFunc {
```
2opremio: You should be passing a parsed value (with separated key and value parts) here.

render/filters.go
```diff
@@ -244,6 +245,24 @@ func IsApplication(n report.Node) bool {
...
+		desiredKeyValue := strings.Split(label, "=")
+		value, _ := n.Latest.Lookup(docker.LabelPrefix + desiredKeyValue[0])
+
+		if len(desiredKeyValue) == 2 {
+			if value == desiredKeyValue[1] {
+				log.Println(value, "=", desiredKeyValue[1])
+				return true
+			}
+		} else {
+			log.Printf("label isn't in the correct key=value format")
```
2opremio: You shouldn't be parsing here

app/api_topologies.go
```diff
@@ -204,10 +224,30 @@ func updateTopologyFilters(t APITopologyDesc, options []APITopologyOptionGroup)
...
+// MakeFilterOption provides an external interface to the package for creating an APITopologyOption
+func MakeFilterOption(value string, label string, filterFunc render.FilterFunc, pseudo bool) APITopologyOption {
+	return APITopologyOption{Value: value, Label: label, filter: filterFunc, filterPseudo: pseudo}
+}
+
+// Registry is a threadsafe store of the available topologies
+type Registry struct {
 	sync.RWMutex
-	items map[string]APITopologyDesc
+	Items map[string]APITopologyDesc
```
2opremio: Oh, for testing. Let me think if there's a way around it.

app/api_topologies_test.go
```diff
@@ -50,6 +52,95 @@ func TestAPITopology(t *testing.T) {
...
+			false,
+		),
+	)
+
+	equals(t, 3, topologyRegistry.CountContainerFilters())
+
+	v := app.MakeFilterOption(fmt.Sprintf("cmdlinefilter%d", 0), "title", render.HasLabel("role=system"), false)
+	topologyRegistry.AddContainerFilters(v)
+
+	equals(t, 4, topologyRegistry.CountContainerFilters())
```
2opremio: This test is not testing the extra functionality you added. You should be testing that your parsing works as expected and that filtering works (i.e. rendering only includes containers with the criteria of the label filter when applied).

prog/main.go
CarltonSemple: Rather than make a struct just to create an APITopologyOption, maybe I should just go with
`type containerLabelFilters []APITopologyOption`
and use the `toAPITopologyOption()` method in `Set()` like you said

2opremio: That's another option

2opremio: > Why worry about splitting up the label? That's already handled elsewhere

Parsing the label when parsing the arguments lets us print an error when launching Scope. As a rule of thumb, the earlier the errors, the better.

prog/main.go
```diff
@@ -134,6 +136,44 @@ type appFlags struct {
...
+	if len(indices) == 0 {
+		log.Infof("No unescaped colon found. This is needed to separate the title from the label")
+		return nil
+	}
+	if len(colonFinder.FindAllStringIndex(u, -1)) > 1 {
+		log.Infof("Escape colons that are part of the title and label")
+		return nil
+	}
+
+	unescapeColons := regexp.MustCompile(`\\(\\:)`)
```
2opremio: Same here.

app/api_topologies_test.go
CarltonSemple: To test the filtering, would I call topologyRegistry.renderTopologies() ? I'm not too certain about that part

2opremio: Use`topologyRegistry.rendererForTopology()` to obtain the right renderer and decorator (you will need to pass the right filtering parameter to use your label filter) and then `renderer.Render(report, decorator)` to obtain the final nodes. 

Here's an example https://github.com/weaveworks/scope/blob/master/app/api_topology.go#L110 . Feel free to use/extend the report in `report_fixture.go` to play with label filters in containers.

app/api_topologies.go
CarltonSemple: @fons i made a function for creating a Registry, and returned items to being private

CarltonSemple:
I didn't yet get a chance to look at why the other tests were breaking when I added a container to the fixed report

2opremio:
@CarltonSemple Can you please squash all the commits?

app/api_topologies.go
```diff
@@ -15,134 +15,154 @@ import (
...
 	}
 )
 
+var containerFilters []APITopologyOptionGroup
+var unconnectedFilter []APITopologyOptionGroup
+
+// InitializeTopologies the topologies
 func init() {
-	containerFilters := []APITopologyOptionGroup{
+	// containerOpts holds the container topology view options such as "all" and "system"
```
2opremio: No need to comment this

app/api_topologies.go
```diff
@@ -255,7 +303,30 @@ type topologyStats struct {
...
+}
+
+// AddContainerFilters adds container filters to this registry
+func (r *Registry) AddContainerFilters(newOptions ...APITopologyOption) {
+	r.Lock()
+	defer r.Unlock()
+
+	for _, key := range []string{containersTopologyDescID, containersByHostnameTopologyDescID, containersByImageTopologyDescID} {
+		var tmp = r.items[key]
+		//for _, optionGroup := range tmp.Options {
```
2opremio: Remove comment

app/api_topologies.go
```diff
@@ -271,14 +342,30 @@ func (r *registry) add(ts ...APITopologyDesc) {
 	}
 }
 
-func (r *registry) get(name string) (APITopologyDesc, bool) {
+// CountContainerFilters returns the number of container filters in the registry
+func (r *Registry) CountContainerFilters() int {
```
2opremio: This shouldn't be here if it's only used in tests.

app/api_topologies.go
```diff
@@ -255,7 +303,30 @@ type topologyStats struct {
...
+	topologyRegistry.AddContainerFilters(newOptions...)
+}
+
+// AddContainerFilters adds container filters to this registry
+func (r *Registry) AddContainerFilters(newOptions ...APITopologyOption) {
+	r.Lock()
+	defer r.Unlock()
+
+	for _, key := range []string{containersTopologyDescID, containersByHostnameTopologyDescID, containersByImageTopologyDescID} {
+		var tmp = r.items[key]
```
2opremio: Please get rid of `tmp` simply use `t.items[key]`

app/api_topologies.go
```diff
@@ -255,7 +303,30 @@ type topologyStats struct {
...
+		//for _, optionGroup := range tmp.Options {
+		for i := 0; i < len(tmp.Options); i++ {
+			if tmp.Options[i].ID == "system" {
```
2opremio: Why do the filters need to go specifically after the "system" filter? Why not append them at the end?

app/api_topologies.go
2opremio: Or at least give it a meaningful name.

app/api_topologies.go
```diff
@@ -255,7 +303,30 @@ type topologyStats struct {
...
+func AddContainerFilters(newOptions ...APITopologyOption) {
+	topologyRegistry.AddContainerFilters(newOptions...)
+}
+
+// AddContainerFilters adds container filters to this registry
+func (r *Registry) AddContainerFilters(newOptions ...APITopologyOption) {
+	r.Lock()
+	defer r.Unlock()
+
+	for _, key := range []string{containersTopologyDescID, containersByHostnameTopologyDescID, containersByImageTopologyDescID} {
```
2opremio: You probably want to create a global variable with the container topology  IDs:

```
containerTopologyIDs = []string{containersTopologyDescID, containersByHostnameTopologyDescID, containersByImageTopologyDescID}
```

app/api_topologies_test.go
```diff
@@ -11,7 +13,10 @@ import (
 	"k8s.io/kubernetes/pkg/api"
 
 	"github.com/weaveworks/scope/app"
+	//"github.com/weaveworks/scope/probe/docker"
```
2opremio: Please remove comment

app/api_topology.go
```diff
@@ -106,7 +106,7 @@ func handleWebsocket(
 			log.Errorf("Error generating report: %v", err)
 			return
 		}
-		renderer, decorator, err := topologyRegistry.rendererForTopology(topologyID, r.Form, report)
+		renderer, decorator, err := topologyRegistry.RendererForTopology(topologyID, r.Form, report)
```
2opremio: OK, this justifies the `Registry` being public, although you could create a wrapper like `AddContainerFilters`

prog/main.go
```diff
@@ -158,12 +218,14 @@ func logCensoredArgs() {
...
-		weaveHostname string
-		dryRun        bool
+		flags                       = flags{}
+		mode                        string
+		debug                       bool
+		weaveEnabled                bool
+		weaveHostname               string
+		dryRun                      bool
+		containerLabelFlagsPositive containerLabelFiltersPositive
+		containerLabelFlagsNegative containerLabelFiltersNegative
```
2opremio: Instead of positive and negative please call them `containerLabelFlags` and `containerLabelFlagsExclude`

prog/main.go
```diff
@@ -134,6 +138,62 @@ type appFlags struct {
...
+func (filters *containerLabelFiltersPositive) String() string {
+	return fmt.Sprint(*filters)
+}
+
+func (filters *containerLabelFiltersPositive) Set(value string) error {
+	*filters = append(*filters, makeAPITopologyOption(true, value, len(*filters)))
+	return nil
+}
+
+type containerLabelFiltersNegative []app.APITopologyOption
```
2opremio: There is no need to have two different types of flags. Please use an intermediate representation usable for both and transform it to an `APITopologyOption` (inverting the filter where needed) later on.

app/api_topologies.go
CarltonSemple: If you look at containerFilters, "system" is just the name of the API Topology Option group that contains the label filters. The naming confusion probably is due to containerFilters being named that way before I started adding code

prog/main.go
```diff
@@ -134,6 +138,62 @@ type appFlags struct {
...
+func (filters *containerLabelFiltersNegative) String() string {
+	return fmt.Sprint(*filters)
+}
+
+func (filters *containerLabelFiltersNegative) Set(value string) error {
+	*filters = append(*filters, makeAPITopologyOption(false, value, len(*filters)))
+	return nil
+}
+
+func makeAPITopologyOption(positive bool, value string, filterNumber int) app.APITopologyOption {
```
2opremio: Please use a more suitable name, something like filterToApiTopologyOption

2opremio: Also, use more suitable parameter names, `value` doesn't say anything

prog/main.go
```diff
@@ -134,6 +138,62 @@ type appFlags struct {
...
+	u := fmt.Sprintf("%q", value)
+	u = u[1 : len(u)-1]
+
+	indices := colonFinder.FindStringIndex(u)
+	if len(indices) == 0 {
+		log.Infof("No unescaped colon found. This is needed to separate the title from the label")
```
2opremio: Instead of logging, please propagate errors to the flag parser (until `Set()`).

prog/main.go
```diff
@@ -134,6 +138,62 @@ type appFlags struct {
...
+		return app.APITopologyOption{}
+	}
+	if len(colonFinder.FindAllStringIndex(u, -1)) > 1 {
+		log.Infof("Escape colons that are part of the title and label")
```
2opremio: Same here

prog/main.go
```diff
@@ -134,6 +138,62 @@ type appFlags struct {
...
+}
+
+func (filters *containerLabelFiltersNegative) Set(value string) error {
+	*filters = append(*filters, makeAPITopologyOption(false, value, len(*filters)))
+	return nil
+}
+
+func makeAPITopologyOption(positive bool, value string, filterNumber int) app.APITopologyOption {
+	u := fmt.Sprintf("%q", value)
+	u = u[1 : len(u)-1]
```
2opremio: I don't understand what you are doing here. Why are you adding quotes?

prog/main.go
```diff
@@ -134,6 +138,62 @@ type appFlags struct {
...
+
+	indices := colonFinder.FindStringIndex(u)
+	if len(indices) == 0 {
+		log.Infof("No unescaped colon found. This is needed to separate the title from the label")
+		return app.APITopologyOption{}
+	}
+	if len(colonFinder.FindAllStringIndex(u, -1)) > 1 {
```
2opremio: Why are you invoking `FindStringIndex` *and* `FindAllStringIndex`

prog/main.go
```diff
@@ -134,6 +138,62 @@ type appFlags struct {
...
+	label := unescapeColons.ReplaceAllString(u[indices[1]:], `:`)
+
+	var labelKey string
+	var labelValue string
+	p := strings.Split(label, "=")
+	if len(p) == 2 {
+		labelKey = p[0]
+		labelValue = p[1]
+	} else {
+		log.Fatalf("docker label isn't in the correct format")
```
2opremio: Don't use fatal, propagate the error

prog/main.go
```diff
@@ -134,6 +138,62 @@ type appFlags struct {
...
+		log.Infof("No unescaped colon found. This is needed to separate the title from the label")
+		return app.APITopologyOption{}
+	}
+	if len(colonFinder.FindAllStringIndex(u, -1)) > 1 {
+		log.Infof("Escape colons that are part of the title and label")
+		return app.APITopologyOption{}
+	}
+
+	title := unescapeColons.ReplaceAllString(u[0:indices[0]+1], `:`)
+	label := unescapeColons.ReplaceAllString(u[indices[1]:], `:`)
```
2opremio: You also need to unescape the backslash, `\` we've discussed this, and I even provided you with a regular expression to do both, sigh ...

prog/main.go
```diff
@@ -134,6 +138,62 @@ type appFlags struct {
...
+	if len(p) == 2 {
+		labelKey = p[0]
+		labelValue = p[1]
+	} else {
+		log.Fatalf("docker label isn't in the correct format")
+	}
+
+	//*filters = append(*filters, app.MakeAPITopologyOption(fmt.Sprintf("cmdlinefilter%d", len(*filters)), title, render.HasLabel(labelKey, labelValue), false))
+	if positive == true {
+		return app.MakeAPITopologyOption(fmt.Sprintf("cmdlinefilter%d", filterNumber), title, render.HasLabel(labelKey, labelValue), false)
```
2opremio: It's probably better to keep a global counter instead of passing a filter number (which is a bit strange). No strong opinions here.

render/filters.go
```diff
@@ -244,6 +244,30 @@ func IsApplication(n report.Node) bool {
...
+
+		if value == labelValue {
+			return true
+		}
+		return false
+	}
+}
+
+// DoesNotHaveLabel checks if the node does NOT have the specified docker label, allowing for a more complete user-customizable filtering feature
+func DoesNotHaveLabel(labelKey string, labelValue string) FilterFunc {
```
2opremio: Please use the `Complement` combinator instead of defining another filter.

prog/main.go
CarltonSemple: i was trying to stick to what you said about functions beginning with "make"

prog/main.go
CarltonSemple: this was here originally to get rid of quotes. the quotes would show up in the UI, which we don't want

render/filters.go
CarltonSemple: I would have done that, but I wasn't sure if that works the same way as when it was used with `IsSystem`, since this function doesn't simply return a bool

prog/main.go
2opremio: Sure, but you are not making an option from scratch you are translating a filter flag into it.

render/filters.go
2opremio: It does return a bool: `return func(n report.Node) bool`

app/api_topologies.go
2opremio: Oh, I see, please rename it to something more meaningful like `"container_label_filters"` and use a constant for it.

prog/main.go
CarltonSemple: I was using `regexp.FindStringIndex` to find the index of the unescaped colon so that I could split the title from the label.

prog/main.go
CarltonSemple: fixed

prog/main.go
CarltonSemple: got it

app/api_topologies.go
CarltonSemple: I'll just remove it then

app/api_topologies_test.go
```diff
@@ -50,6 +58,66 @@ func TestAPITopology(t *testing.T) {
...
+}
+
+func TestContainerLabelFilterExclude(t *testing.T) {
+	topologySummaries, err := getTestContainerLabelFilterTopologySummary(t, true)
+	if err != nil {
+		t.Fatalf("Topology Registry Report error: %s", err)
+	}
+
+	// all containers but the excluded container should be present
+	for keyy := range topologySummaries {
```
ijsnellf: See above.

prog/main_test.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func TestMakeContainerFiltersFromFlags(t *testing.T) {
+	containerLabelFlags := containerLabelFiltersFlag{exclude: false}
+	containerLabelFlags.Set(`title1:label=1`)
+	containerLabelFlags.Set(`ti\:tle2:lab\:el=2`)
+	containerLabelFlags.Set(`ti tile3:label=3`)
+	apiTopologyOptions := containerLabelFlags.apiTopologyOptions
+
+	fmt.Println(apiTopologyOptions)
```
ijsnellf: Remove debugging print statement to keep the tests clean.

prog/main.go
```diff
@@ -134,6 +138,41 @@ type appFlags struct {
...
+	} else if len(colonFinder.FindAllStringIndex(flagValue, -1)) > 1 {
+		return app.APITopologyOption{}, fmt.Errorf("Multiple unescaped colons. Escape colons that are part of the title and label")
+	} else {
+		var labelKeyValuePair []string
+		containerFilterTitle, containerFilterLabel := unescapeBackslashes.ReplaceAllString(flagValue[0:indices[0]+1], `$1`), unescapeBackslashes.ReplaceAllString(flagValue[indices[1]:], `$1`)
+		if labelKeyValuePair = strings.Split(containerFilterLabel, "="); len(labelKeyValuePair) != 2 {
+			return app.APITopologyOption{}, fmt.Errorf("Docker label in the app.container-label-filter(-exclude) flag isn't in the correct key=value format")
+		}
+		switch {
+		case c.exclude == true:
```
ijsnellf: ```case c.exclude:```

app/api_topologies.go
```diff
@@ -255,7 +299,27 @@ type topologyStats struct {
...
+func AddContainerFilters(newFilters ...APITopologyOption) {
+	topologyRegistry.AddContainerFilters(newFilters...)
+}
+
+// AddContainerFilters adds container filters to this registry
+func (r *registry) AddContainerFilters(newFilters ...APITopologyOption) {
+	r.Lock()
+	defer r.Unlock()
+	for _, key := range []string{containersTopologyDescID, containersByHostnameTopologyDescID, containersByImageTopologyDescID} {
+		for i := 0; i < len(r.items[key].Options); i++ {
```
ijsnellf: `for i := range r.items[key].Options {`

prog/main_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+	equals(t, 3, len(apiTopologyOptions))
+	equals(t, `title1`, apiTopologyOptions[0].Value)
+	equals(t, `label=1`, apiTopologyOptions[0].Label)
+	equals(t, `ti:tle2`, apiTopologyOptions[1].Value)
+	equals(t, `lab:el=2`, apiTopologyOptions[1].Label)
+	equals(t, `ti tle3`, apiTopologyOptions[2].Value)
+	equals(t, `label=3`, apiTopologyOptions[2].Label)
+}
```
ijsnellf: We should also probably that invalid `containerLabelFlags` are caught.

app/api_topologies.go
```diff
@@ -15,48 +15,62 @@ import (
...
 		},
 	}
 )
 
 func init() {
+	AddInitialTopologiesToRegistry(topologyRegistry)
+}
+
+// AddInitialTopologiesToRegistry does the initial setup for a Registry
+func AddInitialTopologiesToRegistry(tRegistry *Registry) {
```
2opremio: Minor: s/tRegistry/registry/

app/api_topologies.go
```diff
@@ -210,12 +224,40 @@ func updateTopologyFilters(t APITopologyDesc, options []APITopologyOptionGroup)
...
 
-// registry is a threadsafe store of the available topologies
-type registry struct {
+// MakeAPITopologyDesc makes an APITopologyDesc
+func MakeAPITopologyDesc(id string, renderer render.Renderer, name string, rank int, options []APITopologyOptionGroup, hideIfEmpty bool) APITopologyDesc {
+	return APITopologyDesc{id: id, renderer: renderer, Name: name, Rank: rank, HideIfEmpty: hideIfEmpty, Options: options}
+}
+
+//MakeAPITopologyDescWithParent makes an APITopologyDesc with a parent string specified
+func MakeAPITopologyDescWithParent(id string, parent string, renderer render.Renderer, name string, rank int, options []APITopologyOptionGroup, hideIfEmpty bool) APITopologyDesc {
```
2opremio: This doesn't seem to be used anywhere

app/api_topologies.go
```diff
@@ -210,12 +224,40 @@ func updateTopologyFilters(t APITopologyDesc, options []APITopologyOptionGroup)
...
+	return APITopologyDesc{id: id, parent: parent, renderer: renderer, Name: name, Rank: rank, HideIfEmpty: hideIfEmpty, Options: options}
+}
+
+//MakeAPITopologyOptionGroup makes an APITopologyOptionGroup
+func MakeAPITopologyOptionGroup(id string, Default string, options []APITopologyOption) APITopologyOptionGroup {
```
2opremio: This doesn't seem to be used anywhere

app/api_topologies_test.go
```diff
@@ -83,9 +141,6 @@ func TestAPITopologyAddsKubernetes(t *testing.T) {
 				{ContainerID: "container2"},
 			},
 		},
-		Spec: api.PodSpec{
```
2opremio: Why did you remove this?

prog/main.go
```diff
@@ -138,6 +142,41 @@ type appFlags struct {
...
+	apiTopologyOptions []app.APITopologyOption
+	exclude            bool
+}
+
+func (c *containerLabelFiltersFlag) String() string {
+	return fmt.Sprint(c.apiTopologyOptions)
+}
+
+func (c *containerLabelFiltersFlag) Set(flagValue string) error {
+	newAPITopologyOption, err := c.toAPITopologyOption(flagValue)
```
2opremio: You should return the error if not `nil` empty after this point. Not at the end.

prog/main.go
```diff
@@ -138,6 +142,41 @@ type appFlags struct {
...
+	c.apiTopologyOptions = append(c.apiTopologyOptions, newAPITopologyOption)
+	return err
+}
+
+func (c *containerLabelFiltersFlag) toAPITopologyOption(flagValue string) (app.APITopologyOption, error) {
+	var indices []int
+	if indices = colonFinder.FindStringIndex(flagValue); len(indices) == 0 {
+		return app.APITopologyOption{}, fmt.Errorf("No unescaped colon found. This is needed to separate the title from the label")
+	} else if len(colonFinder.FindAllStringIndex(flagValue, -1)) > 1 {
```
2opremio: Hold on a second. I already reviewed this. Why not calling `colonFinder.FindAllIndex` and checking that he number of indices is 1? Why do you need to call  `FindStringIndex` and `FindAllStringIndex` separately?

prog/main.go
```diff
@@ -138,6 +142,41 @@ type appFlags struct {
...
+		return app.APITopologyOption{}, fmt.Errorf("Multiple unescaped colons. Escape colons that are part of the title and label")
+	}
+	var labelKeyValuePair []string
+	containerFilterTitle, containerFilterLabel := unescapeBackslashes.ReplaceAllString(flagValue[0:indices[0]+1], `$1`), unescapeBackslashes.ReplaceAllString(flagValue[indices[1]:], `$1`)
+	if labelKeyValuePair = strings.Split(containerFilterLabel, "="); len(labelKeyValuePair) != 2 {
```
2opremio: Same comment about the variable declaration.

app/api_topologies.go
```diff
@@ -210,12 +224,40 @@ func updateTopologyFilters(t APITopologyDesc, options []APITopologyOptionGroup)
...
+	return APITopologyOptionGroup{ID: id, Default: Default, Options: options}
+}
+
+// MakeAPITopologyOption provides an external interface to the package for creating an APITopologyOption
```
2opremio: Actually, I stand corrected, I see you are using it in `main.go`. My bad.

prog/main.go
```diff
@@ -162,12 +201,14 @@ func logCensoredArgs() {
...
-		weaveEnabled  bool
-		weaveHostname string
-		dryRun        bool
+		flags                            = flags{}
+		mode                             string
+		debug                            bool
+		weaveEnabled                     bool
+		weaveHostname                    string
+		dryRun                           bool
+		containerLabelFilterFlags        = containerLabelFiltersFlag{exclude: false}
```
2opremio: Minor: This is equivalent to `containerLabelFiltersFlag{}`. Although it may be clearer to make it explicit. Your call.

prog/main.go
```diff
@@ -138,6 +142,41 @@ type appFlags struct {
...
+		return app.APITopologyOption{}, fmt.Errorf("Multiple unescaped colons. Escape colons that are part of the title and label")
+	}
+	var labelKeyValuePair []string
+	containerFilterTitle, containerFilterLabel := unescapeBackslashes.ReplaceAllString(flagValue[0:indices[0]+1], `$1`), unescapeBackslashes.ReplaceAllString(flagValue[indices[1]:], `$1`)
+	if labelKeyValuePair = strings.Split(containerFilterLabel, "="); len(labelKeyValuePair) != 2 {
+		return app.APITopologyOption{}, fmt.Errorf("Docker label in the app.container-label-filter(-exclude) flag isn't in the correct key=value format")
+	}
+	switch {
+	case c.exclude:
+		return app.MakeAPITopologyOption(fmt.Sprintf("containerLabelFilterExclude%d", len(c.apiTopologyOptions)), containerFilterTitle, render.DoesNotHaveLabel(labelKeyValuePair[0], labelKeyValuePair[1]), false), nil
```
2opremio: This line is too long to read comfortably. Please break it down and/or use multiple lines, e.g.:

```
app.MakeAPITopologyOption(
      foo,
      bar,
....
```

Minor: Interesting use of `len(c.apiTopologyOptions)` to avoid global state. It's a bit sneaky to use it in the middle of the function though, so I am having second thoughts.  Maybe it would be clearer if you passed the final option `id` as a parameter of `toAPITopologyOption` (as you did originally if I recall properly)? Your call.

prog/main.go
```diff
@@ -138,6 +142,41 @@ type appFlags struct {
...
+	default:
+		return app.MakeAPITopologyOption(fmt.Sprintf("containerLabelFilter%d", len(c.apiTopologyOptions)), containerFilterTitle, render.HasLabel(labelKeyValuePair[0], labelKeyValuePair[1]), false), nil
```
2opremio: Line too long.

prog/main_test.go
2opremio: I agree

app/api_topologies.go
```diff
@@ -210,12 +224,40 @@ func updateTopologyFilters(t APITopologyDesc, options []APITopologyOptionGroup)
...
+	return APITopologyOption{Value: value, Label: label, filter: filterFunc, filterPseudo: pseudo}
+}
+
+// Registry is a threadsafe store of the available topologies
+type Registry struct {
 	sync.RWMutex
 	items map[string]APITopologyDesc
 }
 
+// MakeRegistry returns a new Registry
```
2opremio: Please mention this is only used for testing

2opremio: Actually, you could initialize the global topology registry calling this as well.

prog/main.go
```diff
@@ -138,6 +142,41 @@ type appFlags struct {
...
+
+func (c *containerLabelFiltersFlag) toAPITopologyOption(flagValue string) (app.APITopologyOption, error) {
+	var indices []int
+	if indices = colonFinder.FindStringIndex(flagValue); len(indices) == 0 {
+		return app.APITopologyOption{}, fmt.Errorf("No unescaped colon found. This is needed to separate the title from the label")
+	} else if len(colonFinder.FindAllStringIndex(flagValue, -1)) > 1 {
+		return app.APITopologyOption{}, fmt.Errorf("Multiple unescaped colons. Escape colons that are part of the title and label")
+	}
+	var labelKeyValuePair []string
+	containerFilterTitle, containerFilterLabel := unescapeBackslashes.ReplaceAllString(flagValue[0:indices[0]+1], `$1`), unescapeBackslashes.ReplaceAllString(flagValue[indices[1]:], `$1`)
```
2opremio: Escaping before will require readjusting the colon separator index which may make things a bit complicated. Your call: give it a try a see how it looks.

Regardless, please try to create smaller lines, breaking the process in steps/intermediate variables if needed.

prog/main_test.go
CarltonSemple: Alright, I added it.

prog/main.go
CarltonSemple: Well now that's part of a struct, I could add an ID variable.

2opremio: I am not sure what you mean (since the ID is dynamic). But give it a go and see if it's easier to read.

app/api_topologies.go
CarltonSemple: alright, i initialized the global registry with it in the latest commit

prog/main.go
```diff
@@ -152,28 +154,50 @@ func (c *containerLabelFiltersFlag) String() string {
 }
 
 func (c *containerLabelFiltersFlag) Set(flagValue string) error {
-	newAPITopologyOption, err := c.toAPITopologyOption(flagValue)
+	c.filterIDBase = fmt.Sprintf(c.filterIDBase+"%d", c.filterNumber)
```
2opremio: This doesn't do what you want. I think you want a local variable instead of overwriting the base.

Optional: keep using the length instead of a counter like you did before.

prog/main.go
```diff
@@ -152,28 +154,50 @@ func (c *containerLabelFiltersFlag) String() string {
...
 }
 
-func (c *containerLabelFiltersFlag) toAPITopologyOption(flagValue string) (app.APITopologyOption, error) {
-	var indices []int
-	if indices = colonFinder.FindStringIndex(flagValue); len(indices) == 0 {
-		return app.APITopologyOption{}, fmt.Errorf("No unescaped colon found. This is needed to separate the title from the label")
-	} else if len(colonFinder.FindAllStringIndex(flagValue, -1)) > 1 {
+func (c *containerLabelFiltersFlag) toAPITopologyOption(flagValue string, filterID string) (app.APITopologyOption, error) {
+	indexRanges := colonFinder.FindAllStringIndex(flagValue, -1)
+	if len(indexRanges) != 1 {
```
2opremio: So much nicer!

prog/main.go
```diff
@@ -152,28 +154,50 @@ func (c *containerLabelFiltersFlag) String() string {
...
+	titleStringEscaped := flagValue[:splitIndices[0]+1]
+	labelStringEscaped := flagValue[splitIndices[1]:]
+	containerFilterTitle := unescapeBackslashes.ReplaceAllString(titleStringEscaped, `$1`)
+	containerFilterLabel := unescapeBackslashes.ReplaceAllString(labelStringEscaped, `$1`)
+	labelKeyValuePair := strings.Split(containerFilterLabel, "=")
+	if len(labelKeyValuePair) != 2 {
 		return app.APITopologyOption{}, fmt.Errorf("Docker label in the app.container-label-filter(-exclude) flag isn't in the correct key=value format")
 	}
+
 	switch {
```
2opremio: I would use an `if` `else` instead of a switch (sorry I missed this before)

2opremio: Also, instead of returning on each branch, simply create a result variable and do `return result, nil` at the end of the function.

2opremio: Even better, it would be to create a variable for the filtering function (`render.DoesNotHaveLabel` or `render.HasLabel`) which is the only difference between branches.

prog/main.go
```diff
@@ -138,6 +142,55 @@ type appFlags struct {
...
+	labelKeyValuePair := strings.Split(containerFilterLabel, "=")
+	if len(labelKeyValuePair) != 2 {
+		return app.APITopologyOption{}, fmt.Errorf("Docker label in the app.container-label-filter(-exclude) flag isn't in the correct key=value format")
+	}
+
+	filterFunction := render.HasLabel(labelKeyValuePair[0], labelKeyValuePair[1])
+	if c.exclude {
+		filterFunction = render.DoesNotHaveLabel(labelKeyValuePair[0], labelKeyValuePair[1])
+	}
+	result := app.MakeAPITopologyOption(filterID, containerFilterTitle, filterFunction, false)
```
2opremio: No need for a variable since you are returning it right away. Sorry if my previous comment was confusing.

app/api_topologies_test.go
```diff
@@ -50,6 +58,56 @@ func TestAPITopology(t *testing.T) {
...
+
+func TestContainerLabelFilterExclude(t *testing.T) {
+	topologySummaries, err := getTestContainerLabelFilterTopologySummary(t, true)
+	if err != nil {
+		t.Fatalf("Topology Registry Report error: %s", err)
+	}
+
+	// all containers but the excluded container should be present
+	for key := range topologySummaries {
+		if fixture.ServerContainerNodeID+";<container>" == key {
```
2opremio: Please use `MakeContainerNodeID`

prog/main.go
```diff
@@ -138,6 +142,54 @@ type appFlags struct {
...
+		return app.APITopologyOption{}, fmt.Errorf("Multiple unescaped colons. Escape colons that are part of the title and label")
+	}
+	splitIndices := indexRanges[0]
+	titleStringEscaped := flagValue[:splitIndices[0]+1]
+	labelStringEscaped := flagValue[splitIndices[1]:]
+	containerFilterTitle := unescapeBackslashes.ReplaceAllString(titleStringEscaped, `$1`)
+	containerFilterLabel := unescapeBackslashes.ReplaceAllString(labelStringEscaped, `$1`)
+	labelKeyValuePair := strings.Split(containerFilterLabel, "=")
+	if len(labelKeyValuePair) != 2 {
+		return app.APITopologyOption{}, fmt.Errorf("Docker label in the app.container-label-filter(-exclude) flag isn't in the correct key=value format")
```
2opremio: Please remove `in the app.container-label-filter(-exclude) flag` from the error, it's redundant.

Here's an example:

```
invalid value "foo:bar" for flag -app.container-label-filter: Docker label in the app.container-label-filter(-exclude) flag isn't in the correct key=value format
```

2opremio:
Good job!

I'll merge as soon as CircleCI gives me a green.

PR #1897, opened 2016-09-27
Be more strict when snooping DNS over TCP traffic

2opremio:
Getting non-DNS traffic on TCP port 53 was causing the parser to flood the logs with errors (it happened in Weave Cloud, Prometheus is incorrectly sending http requests).

This makes DNS traffic identification a bit more robust and strict, enforcing our relaxation of only considering self-contained TCP segments.


rade:
LGTM

PR #1901, opened 2016-09-30
Allow user to specify URLs on the command line, and use that to allow per-target tokens.

### LINKED ISSUE #1696: Allow setting a service token per target app
Right now, when using a service token with several target apps (i.e. `scope launch --service-token=foo app1.com:80 app2.com:80`), the same token must be shared among all the targets.

For testing purposes, when we mirror reports to multiple environments it would be desirable to be able to set the service token individually for each environment.

@tomwilkie suggests fixing this by using URLs for the target apps. 

Maybe something like: `scope launch app://app1.com/foo app://app2.com/bar`

tomwilkie:
Fixes #1696

### DIFF
--- a/app/controls_test.go
+++ b/app/controls_test.go
@@ -4,6 +4,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"strings"
 	"testing"
 	"time"
@@ -43,7 +44,8 @@ func TestControl(t *testing.T) {
 			Value: "foo",
 		}
 	})
-	client, err := appclient.NewAppClient(probeConfig, ip+":"+port, ip+":"+port, controlHandler)
+	url := url.URL{Scheme: "http", Host: ip + ":" + port}
+	client, err := appclient.NewAppClient(probeConfig, ip+":"+port, url, controlHandler)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/app/pipes_internal_test.go
+++ b/app/pipes_internal_test.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"strings"
 	"testing"
 	"time"
@@ -83,7 +84,8 @@ func TestPipeClose(t *testing.T) {
 	probeConfig := appclient.ProbeConfig{
 		ProbeID: "foo",
 	}
-	client, err := appclient.NewAppClient(probeConfig, ip+":"+port, ip+":"+port, nil)
+	url := url.URL{Scheme: "http", Host: ip + ":" + port}
+	client, err := appclient.NewAppClient(probeConfig, ip+":"+port, url, nil)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/extras/fixprobe/main.go
+++ b/extras/fixprobe/main.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/url"
 	"time"
 
 	"github.com/ugorji/go/codec"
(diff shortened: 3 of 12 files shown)


tomwilkie:
So currently you need to use this like this:

```
./scope launch http://user:<token>@frontend.dev.weave.works:80/ https://user:<token>@cloud.weave.works:443/
```

I feel like we should:
- [ ] default to port 80/443 is http(s) is specified, and default to 4040 is just a hostname is used (to preserve existing behaviour)
- [ ] potentially treat the username as the token, so one can do https://<token>@cloud.weave.works/
- [ ] make it so badly formed URLs cause `scope launch` to fail.

@2opremio WDYT?

2opremio:
> WDYT?

When were users introduced as part of the authentication?  I thought that the probe only required a token.

Also, we should review, update and redeploy the Scope documentation/guides (and maybe blogposts) before merging this since it changes how we launch Scope.

> default to port 80/443 is http(s) is specified, and default to 4040 is just a hostname is used (to preserve existing behaviour)?

Uhm, that's a tricky one. I would only default 80/443 if omitted when providing the scheme. Defaulting 4040 when you are providing the scheme feels too magical to me. 

If you want to preserve the current behaviour then I would propose to omit the scheme, default to 4040 if the port is omitted and force using 80/443(https) explicitly. This is what I believe we do right now.

> potentially treat the username as the token, so one can do https://<token>@cloud.weave.works?

I still don't fully understand the role of users when pushing reports from probes. Users have their own token now?

tomwilkie:
> When were users introduced as part of the authentication?

They are not, the user part of the URL is ignored.

> changes how we launch Scope.

This is backwards compatible - if you don't specify a password in the URL, the one from `--service-token` is used.

> I would only default 80/443 if omitted when providing the scheme. 

Sorry if it wasn't clear - that was the suggestion.  Looks like we agree!

> I still don't fully understand the role of users when pushing reports from probes. Users have their own token now?

As an aside, there are user tokens now, but this is not what I'm using here (and I don't propose we do, as it makes it harder)

2opremio:
> This is backwards compatible - if you don't specify a password in the URL, the one from `--service-token` is used.

Not when using explicit targets, at least if we don't default to 4040 when not providing the scheme (which would be a bit odd).

> They are not, the user part of the URL is ignored.

Then let's use the user part of the URL as a token instead of the password.

> Sorry if it wasn't clear - that was the suggestion.  Looks like we agree!

Great :)

tomwilkie:
> Not when using explicit targets, at least if we don't default to 4040 when not providing the scheme (which would be a bit odd).

I'm confused - how about if we:
- default to port 4040 when no scheme AND no port is specified
- if a port is specified always use that
- if a scheme is specified AND no port is specified, use 80/443 as appropriate

That should be backwards compatible, right?  Can you give an example where it isn't?

2opremio:
It would, but defaulting to different ports based on whether the scheme is
provided or not seems a bit unintuitive to me.

However, I don't see a better solution, so go ahead.

tomwilkie:
Okay with latest changes user can now do:

```
./scope launch http://<token>@frontend.dev.weave.works https://<token>@cloud.weave.works
```

Which is much cleaner.  Should be good to merge now?

2opremio:
LGTM, please squash.

app/controls_test.go
```diff
@@ -43,7 +44,7 @@ func TestControl(t *testing.T) {
 			Value: "foo",
 		}
 	})
-	client, err := appclient.NewAppClient(probeConfig, ip+":"+port, ip+":"+port, controlHandler)
+	client, err := appclient.NewAppClient(probeConfig, ip+":"+port, url.URL{Scheme: "http", Host: ip + ":" + port}, controlHandler)
```
2opremio: Actually, could you initialize the `url.URL{}` separately (the line is took long to read.

app/pipes_internal_test.go
```diff
@@ -83,7 +84,7 @@ func TestPipeClose(t *testing.T) {
 	probeConfig := appclient.ProbeConfig{
 		ProbeID: "foo",
 	}
-	client, err := appclient.NewAppClient(probeConfig, ip+":"+port, ip+":"+port, nil)
+	client, err := appclient.NewAppClient(probeConfig, ip+":"+port, url.URL{Scheme: "http", Host: ip + ":" + port}, nil)
```
2opremio: Same here.

app/controls_test.go
tomwilkie: Done

app/pipes_internal_test.go
tomwilkie: Done

PR #1903, opened 2016-10-05
Clarify terminal is child window of details panel.

### LINKED ISSUE #1152: it's not obvious that terminals are extensions of the details panel
I only discovered by accident that closing the details panel also closes any associated terminal panel - up until that point I had been dutifully closing the latter and then the former.

Can we make the relationship between the two more obvious?

Perhaps the terminal could look like (and reveal like) a side pull-out from the details panel?

@peterbourgon says such ideas were discussed before. Might be worth revisiting.

foot:
So when you close the details panel and the terminal closes too its not a surprise!

Fixes #1152

### DIFF
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -88,9 +88,11 @@ class Terminal extends React.Component {
       pixelPerCol: 0,
       pixelPerRow: 0
     };
+
     this.handleCloseClick = this.handleCloseClick.bind(this);
     this.handlePopoutTerminal = this.handlePopoutTerminal.bind(this);
     this.handleResize = this.handleResize.bind(this);
+    this.focusTerminal = this.focusTerminal.bind(this);
   }
 
   createWebsocket(term) {
@@ -106,7 +108,13 @@ class Terminal extends React.Component {
     };
 
     socket.onclose = () => {
-      log('socket closed');
+      //
+      // componentWillUnmount has called close and tidied up! don't try and do it again here
+      // (setState etc), its too late.
+      //
+      if (!this.socket) {
+        return;
+      }
       this.socket = null;
       const wereConnected = this.state.connected;
       this.setState({connected: false});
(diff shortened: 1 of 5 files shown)


rade:
screen cap?

rade:
any chance we could change the 'close' to a 'slide in'?

rade:
and perhaps change the icon too, e.g. ">>"

foot:
`>>` / `slide in` suggests some kind of permanence/continuity to me, as in, _I'm done with this terminal session for now, lets tuck it away for later._

I like: boom, its gone, gotta make a new one now.

Though you could interpret the idea of the having access to the container as continuous thing here and it would make more sense summoning it back, thoughts? (+ @davkal)

rade:
I see your point.

My concern is that I see the "x" button and "boom, its gone" action as indicators that the terminal has an independent life from the detail panel.

foot:
They feel quite connected now.

There are other onClose animations we could look into but I don't wanna make it too animation heavy, and I think a snappy close is usually nice: Done here, I'm off to do something else. Vs establishing context etc as we show something new.

Alt close animation: 
 - Keep `x`
 - fade out + sliding back, signifying destruction + going back home to its parent the details panel.

davkal:
I'm happy with the immediate close on X. 

screen shot 2016-10-18 at 14 46 30

Not so happy with the aesthetic. Both panels look a bit bolted together, colors are clashing a bit. There is no clear hierarchy (other than suggested by the animation). Most oddly, the details panel has a shadow, but the terminal has not.

Counter-proposal: remove the dark color border, add padding (to make it smaller to suggest hierarchy), and add shadow.

screen shot 2016-10-18 at 14 46 12

client/app/scripts/components/terminal.js
```diff
@@ -328,4 +353,10 @@ class Terminal extends React.Component {
   }
 }
 
+
+Terminal.defaultProps = {
+  connect: true
```
davkal: What's this for?

client/app/styles/main.less
```diff
@@ -979,35 +980,34 @@ h2 {
...
-    right: (@details-window-width + @details-window-padding-left + 10px);
+    position: relative;
+    flex: 1;
+    overflow-x: hidden;
+  }
+
+  &-animation-wrapper {
+    width: 100%;
+    height: 100%;
+    transition: transform 0.5s cubic-bezier(0.230, 1.000, 0.320, 1.000);
```
davkal: Maybe extract transform timing function?

client/app/scripts/components/terminal.js
foot: Terminal doesn't try and connect until it receives the connect attr.
`<Terminal connect />`

We don't want to connect until after the drawer animation has completed. Term.js hits the dom a lot and makes the animation super janky.

davkal:
YES! I like the hover bit. It feels like the details panel owns the terminal. Well done. LGTM

PR #1904, opened 2016-10-05
Apply filters from current view to details panel

### LINKED ISSUE #1864: Topology filters are not applied to detail panel connections or children
Because we don't send topology filters to the details panel.

I argue that this is correct. The topology filters are about *which nodes* should appear, but the data on the details panel is an integral part of the node, and should not be modified, based on topology filters.

@rade argues the opposite.

2opremio:
Fixes #1864

### DIFF
--- a/render/filters.go
+++ b/render/filters.go
@@ -10,6 +10,28 @@ import (
 	"github.com/weaveworks/scope/report"
 )
 
+// PreciousNodeRenderer ensures a node is never filtered out by decorators
+type PreciousNodeRenderer struct {
+	PreciousNodeID string
+	Renderer
+}
+
+func (p PreciousNodeRenderer) Render(rpt report.Report, dct Decorator) report.Nodes {
+	undecoratedNodes := p.Renderer.Render(rpt, nil)
+	preciousNode, foundBeforeDecoration := undecoratedNodes[p.PreciousNodeID]
+	finalNodes := applyDecorator{ConstantRenderer(undecoratedNodes)}.Render(rpt, dct)
+	if _, ok := finalNodes[p.PreciousNodeID]; !ok && foundBeforeDecoration {
+		finalNodes[p.PreciousNodeID] = preciousNode
+	}
+	return finalNodes
+}
+
+func (p PreciousNodeRenderer) Stats(rpt report.Report, dct Decorator) Stats {
+	// default to the underlying renderer
+	// TODO: we don't take into account the precious node, so we may be off by one
+	return p.Renderer.Stats(rpt, dct)
+}
+
 // CustomRenderer allow for mapping functions that received the entire topology
 // in one call - useful for functions that need to consider the entire graph.
 // We should minimise the use of this renderer type, as it is very inflexible.
--- a/render/filters.go
+++ b/render/filters.go
@@ -16,6 +16,7 @@ type PreciousNodeRenderer struct {
 	Renderer
 }
 
+// Render implements Renderer
 func (p PreciousNodeRenderer) Render(rpt report.Report, dct Decorator) report.Nodes {
 	undecoratedNodes := p.Renderer.Render(rpt, nil)
 	preciousNode, foundBeforeDecoration := undecoratedNodes[p.PreciousNodeID]
@@ -26,6 +27,7 @@ func (p PreciousNodeRenderer) Render(rpt report.Report, dct Decorator) report.No
 	return finalNodes
 }
 
+// Stats implements Renderer
 func (p PreciousNodeRenderer) Stats(rpt report.Report, dct Decorator) Stats {
 	// default to the underlying renderer
 	// TODO: we don't take into account the precious node, so we may be off by one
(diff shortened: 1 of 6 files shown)


2opremio:
Before:

screen shot 2016-10-05 at 10 46 17

After:
screen shot 2016-10-05 at 10 51 01

rade:
wow, that turned out to be much easier than I expected.

foot:
Nice! Works really well on container topo.

There's a small issue w/ cross-topo stuff as the options get mixed up: (e.g. when on the processes topo clicking the relatives to load up the weavescope container in the details panel).

`/api/topology/containers/<weavescope-id>` 404
`/api/topology/containers/<weavescope-id>?&unconnected=hide` 404 (processes topo options)
`/api/topology/containers/<weavescope-id>?&system=system&stopped=running&pseudo=hide` 200

We could change the UI a bit more to send the correct topo options for the node type but maybe we wanna use the topo options to filter the contents of the requested node rather than the node itself?

foot:
Sent topo options now correspond to node type being requested.

BE filtering is still a bit tight: if containers.topoOptions = 'application', you are unable to view 'weavescope' in the details panel.

2opremio:
`/api/topology/containers/<weavescope-id>`  returns a 404 because when no explicit filtering options are given it applies the default ones for the topology (filter out system containers).

This is a change of behavior since no filtering was applied before, but I see it as correct and reasonable.

2opremio:
> BE filtering is still a bit tight: if containers.topoOptions = 'application', you are unable to view 'weavescope' in the details panel.

By clicking through details panels you can get to the host (e.g. app-curler > host > weavescope ), which will return 404 since system containers are not viewable through the filters.

This one's tricky. I see two options:

* Make sure the host details panel doesn't show those containers (since they are filtered by the container topology). But this would make the topology filters global, which is probably unintuitive.

* Make filtering affect the content of the panel but not the existence of the panel itself (suggested by @rade). This is non-trivial with the current code. We could try gathering the node without filtering if it's not found but it's an ugly hack ....

rade:
> Make filtering affect the content of the panel but not the existence of the panel itself (suggested by @rade). This is non-trivial with the current code.

Ah, so *that* is why this PR has so little code!

I do rather think we need that.

rade:
> I do rather think we need that.

btw, that does not (necessarily) mean that this PR cannot be merged w/o that. I haven't got the time to play with this to form an opinion. Please consult others in the team.

foot:
I wouldn't merge this PR as is.

To get this in now we could 
 1. change the behaviour of `/api/topology/containers/<weavescope-id>` (no querystring) so that it doesn't do any filtering. Then use this route when opening up relatives via the details panel. Only the first level (clicking nodes  on the topo) would use the filtered topology.
 2. Only apply the querystring filters to pseudo nodes like the internet... ?

2opremio:
I've fixed this by never filtering out the target node if it exists.

> change the behaviour of /api/topology/containers/<weavescope-id> (no querystring) so that it doesn't do any filtering. Then use this route when opening up relatives via the details panel. Only the first level (clicking nodes on the topo) would use the filtered topology.

On top of my change, we could do this when getting information about topologies other than the one in view. It requires a minimal modification in the backend. Sounds good?

render/filters.go
```diff
@@ -10,6 +10,30 @@ import (
 	"github.com/weaveworks/scope/report"
 )
 
+// PreciousNodeRenderer ensures a node is never filtered out by decorators
```
paulbellamy: Is filtering the only thing decorators do?

2opremio: AFAIK

2opremio:
This should be ready to go. @paulbellamy / @foot please take a final look.

rade:
@fons please summarise what the behaviour is.

2opremio:
The filters in the topology in view are applied to the details shown for any node in that topology. However, the filter is not applied to the existente of the node itself to ensure no broken links are found between topologies or if the state of the node changes (I.e. a container from running to stopped).

No filters are applied to the details of nodes out of the topology in view (that would make filtersrs from other topologies global, which would be confusing for the end user). Note that pseudo nodes are attributed to the topology in view.

paulbellamy:
> The filters in the topology in view are applied to the details shown for any node in that topology. 

From testing, this statement doesn't seem true.

2opremio:
> From testing, this statement doesn't seem true.

Can you provide specific evidence?

2opremio:
> From testing, this statement doesn't seem true.

Can you provide specific evidence?

foot:
Code LGTM and seems to work okay for internet node + other topos.

paulbellamy:
> Can you provide specific evidence?

1) Fire up scope with the demo report
2) Go to Containers-By-Image view
3) Open details panel for "Swarm" image. (all containers are running)
4) Select "Stopped containers"

Expected behaviour: No containers are shown in the details panel children table.
Actual behaviour: All containers are shown in the details panel children table.

2opremio:
> Expected behaviour: No containers are shown in the details panel children table.
> Actual behaviour: All containers are shown in the details panel children table.

That's an edge case, in which the filter change makes the node disappear from view. `PreciousNodeRenderer` ensures that the node won't get filtered out in that case but doesn't apply filtering to the details of the node itself.

Admittedly it should behave like you state but I don't see a simple way to solve this.

2opremio:
Discussed offline with @paulbellamy and he didn't find any other problems except the corner case commented above, merging.

PR #1909, opened 2016-10-07
Extend logging middleware to optionally only log failed HTTP requests

tomwilkie:


common/middleware/logging.go
```diff
@@ -10,16 +10,35 @@ import (
...
+
+// Wrap implements Middleware
+func (l Log) Wrap(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		begin := time.Now()
 		uri := r.RequestURI // capture the URI before running next, as it may get rewritten
 		i := &interceptor{ResponseWriter: w, statusCode: http.StatusOK}
 		next.ServeHTTP(i, r)
-		log.Infof("%s %s (%d) %s", r.Method, uri, i.statusCode, time.Since(begin))
+		if l.LogSuccess || !(200 <= i.statusCode && i.statusCode < 300) {
```
ekimekim: 100s and 300s should be considered successes in this context. This check should be `|| (400 <= i.statusCode && i.statusCode < 600)`

I'd maybe even argue that from a server perspective, only 500s are true "errors", the rest is a succesful invalidation of a bad user request. But logging 400s is sometimes useful.

PR #1912, opened 2016-10-10
Add label_minor to tooltips in connections table

### LINKED ISSUE #1844: connection count in details panel has visually duplicate rows
After #1843 is merged:

When you have multiple containers of the same name connect to the Internet, you easily get to this:

```
Inbound Remote Port Count
X    IP1       PortA    2
Y    IP2       PortB   2
Y    IP2       PortB   2
Y    IP2       PortB   2
X    IP3       PortA    2
```

Having multiple rows with all values identical looks confusing. 

UX remedies:

* aggregate counts to one row if all other fields are the same (would lose clickability for drilldown)
* add a column for a field that varies across the rows (IDs are usually wide and pretty useless to see)
* add a hover over the name with a field that distinguishes the rows
* add a column for IDs or a distinguishing field, but keep it really narrow (maybe dont even show the value, only on hover)

foot:
Sometimes label is not enough to distinguish between rows.

Fixes #1844


foot:
screen shot 2016-10-10 at 15 14 14

render/detailed/connections.go
```diff
@@ -43,11 +43,12 @@ type ConnectionsSummary struct {
...
 type Connection struct {
-	ID       string               `json:"id"`     // ID of this element in the UI.  Must be unique for a given ConnectionsSummary.
-	NodeID   string               `json:"nodeId"` // ID of a node in the topology. Optional, must be set if linkable is true.
-	Label    string               `json:"label"`
-	Linkable bool                 `json:"linkable"`
-	Metadata []report.MetadataRow `json:"metadata,omitempty"`
+	ID         string               `json:"id"`     // ID of this element in the UI.  Must be unique for a given ConnectionsSummary.
+	NodeID     string               `json:"nodeId"` // ID of a node in the topology. Optional, must be set if linkable is true.
+	Label      string               `json:"label"`
+	LabelMinor string               `json:"label_minor,omitempty"`
```
davkal: I thought we used camelCase in the json?

davkal:
Apart from the snake case, LGTM

foot:
We use camelCase everywhere except label_minor!

I can change that in this PR too.

Should be safe to have FE/BE out of sync during update. I think we interpret label_minor as optional everywhere.

davkal:
LGTM

PR #1917, opened 2016-10-10
Timeout reads and writes in the http server.

### LINKED ISSUE #1916: Goroutine leak in scope app?
screen shot 2016-10-10 at 14 08 51

tomwilkie:
Fixes #1916

### DIFF
--- a/prog/app.go
+++ b/prog/app.go
@@ -31,6 +31,7 @@ import (
 
 const (
 	memcacheUpdateInterval = 1 * time.Minute
+	httpTimeout            = 90 * time.Second
 )
 
 var (
@@ -278,8 +279,8 @@ func appMain(flags appFlags) {
 		Server: &http.Server{
 			Addr:           flags.listen,
 			Handler:        handler,
-			ReadTimeout:    10 * time.Second,
-			WriteTimeout:   10 * time.Second,
+			ReadTimeout:    httpTimeout,
+			WriteTimeout:   httpTimeout,
 			MaxHeaderBytes: 1 << 20,
 		},
 	}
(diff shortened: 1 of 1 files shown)


tomwilkie:
Ah yes websockets!

rade:
Won't this interact badly with persistent http connections?

2opremio:
@tomwilkie  Out of curiosity, how have you confirmed that this fixes #1916 ?

prog/app.go
```diff
@@ -276,8 +277,11 @@ func appMain(flags appFlags) {
 		// we want to manage the stop condition ourselves below
 		NoSignalHandling: true,
 		Server: &http.Server{
-			Addr:    flags.listen,
-			Handler: handler,
+			Addr:           flags.listen,
+			Handler:        handler,
+			ReadTimeout:    httpTimeout,
+			WriteTimeout:   httpTimeout,
+			MaxHeaderBytes: 1 << 20,
```
2opremio: I think this can be omitted since it matches `DefaultMaxHeaderBytes`

rade:
And I am still waiting for an answer to my question re persistent connections.

rade:
@ekimekim could you take over on this, e.g. address review comments, answer questions above, perhaps do a deploy to dev to test whether it solves #1916 (unless you can think of a better way to test this)?

ekimekim:
I'm attempting to reproduce locally. So far what I've tried: making a connection which hangs (stops sending data) during the sending of a POST body. This causes goroutine count to rise until those connections are closed on the client side, at which point they drop again.

ekimekim:
Disregard the bad news above, it's an unrelated problem with my local setup since it's happening on master too (when accessing via /admin/scope). Connecting to scope.kube-system directly works fine with this PR.

ekimekim:
ok, i'm now confident in my testing:
* works fine in normal browser usage
* websockets aren't affected, they remain open for > 90s
* manually testing with telnet - persistent connections remain open for > 90s
* manually testing with telnet - connections where I stop sending halfway through request headers are closed after 90s
I think this is good to merge.

PR #1919, opened 2016-10-12
Add a new search section to the help popover

### LINKED ISSUE #1762: explain search syntax (better)
From a user...

> It would also be good to have a list of all the search params you can use in the search field, for example I had to figure out I can use namespace:prod

foot:
Fixes #1762

### DIFF
--- a/client/app/scripts/components/help-panel.js
+++ b/client/app/scripts/components/help-panel.js
@@ -1,4 +1,10 @@
 import React from 'react';
+import { connect } from 'react-redux';
+
+import { searchableFieldsSelector } from '../selectors/chartSelectors';
+import { CANVAS_MARGINS } from '../constants/styles';
+import { hideHelp } from '../actions/app-actions';
+
 
 const GENERAL_SHORTCUTS = [
   {key: 'esc', label: 'Close active panel'},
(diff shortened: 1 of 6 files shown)


foot:
screen shot 2016-10-12 at 17 37 05

davkal:
UI:

> Any field matching pid for the value 12345

I'd highlight `pid`

> memory less than 4mb

Should be 10mb

Generally I'd visually distinguish the input string a bit more from the explanation.

davkal:
The adaptive field help works nicely!

client/app/scripts/components/help-panel.js
```diff
@@ -27,20 +35,135 @@ function renderShortcuts(cuts) {
...
+
+const REGEX_SEARCHES = [
+  {term: 'foo|bar', label: 'All fields for foo or bar'},
+  {term: 'command: foo(bar|baz)', label: 'Command field for foobar or foobaz'},
+];
+
+
+const METRIC_SEARCHES = [
+  {term: 'cpu > 4%', label: 'CPU greater than 4%'},
+  {term: 'memory < 10mb', label: 'memory less than 4mb'},
```
davkal: this should be `...than 10MB`, capital M is for Mega, and capital B is for byte, as opposed to lowercase b (bit).

client/app/scripts/components/search.js
```diff
@@ -130,14 +133,17 @@ class Search extends React.Component {
             </label>
           </div>
           {!showPinnedSearches && <div className="search-hint">
-            {getHint(nodes)}
+            {getHint(nodes)} <span className="search-help-link" onClick={onClickHelp}>
```
davkal: HELP! is hard to read. Why not use a question mark in a circle?

Also, the search bar collapses again when help is clicked, which was a bit confusing.

client/app/scripts/selectors/chartSelectors.js
```diff
@@ -95,6 +96,12 @@ export const dataNodesSelector = createSelector(
 );
 
 
+export const searchableFieldsSelector = createSelector(
```
davkal: Shouldnt this be a selector that checks for deep equality (fields are always new sets)?

client/app/scripts/utils/string-utils.js
```diff
@@ -65,7 +65,7 @@ export const formatMetric = makeFormatMetric(renderHtml);
 export const formatMetricSvg = makeFormatMetric(renderSvg);
 export const formatDate = d3.time.format.iso;
 
-const CLEAN_LABEL_REGEX = /\W/g;
+const CLEAN_LABEL_REGEX = /[^A-Za-z]/g;
```
davkal: I'm curious, why was that needed?

PR #1920, opened 2016-10-12
Add timeouts to the probe http client

paulbellamy:
Probe side of: #1917

## TODO

- [ ] Try this out


tomwilkie:
Checkout https://github.com/hashicorp/go-cleanhttp

tomwilkie:
LGTM!  I'll hold off merging #1917 until this is deployed and we can see that it works.

PR #1923, opened 2016-10-14
Add new middleware for adding headers

ekimekim:
This extends the middleware library to be able to easily add headers to responses.


tomwilkie:
I've tended to prefer structs that implement the middleware interface to closures, as they've proven to be more future proof wrt extensibility and default/zero values.

ekimekim:
I'm actually just going to drop this since it turned out to not be needed for what I was originally going to use it for.

PR #1924, opened 2016-10-14
Check existence of SHA1 before running `git diff` 

asymmetric:
This fixes a bug whereby a cached Docker image would always end up being used, if the git commit it was built from was not reachable anymore.

### DIFF
--- a/tools/rebuild-image
+++ b/tools/rebuild-image
@@ -40,11 +40,25 @@ commit_timestamp() {
 	git show -s --format=%ct "$rev"
 }
 
+# is the SHA1 actually present in the repo?
+# it could be it isn't, e.g. after a force push
+commit_invalid() {
+	local rev=$1
+	git rev-parse --quiet --verify "$rev^{commit}"
+	[ $? -ne 0 ]
+}
+
 cached_revision=$(cached_image_rev)
 if [ -z "$cached_revision" ]; then
 	echo ">>> No cached image found; rebuilding"
 	rebuild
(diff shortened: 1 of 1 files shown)


tools/rebuild-image
```diff
@@ -40,13 +40,27 @@ commit_timestamp() {
 	git show -s --format=%ct "$rev"
 }
 
+# is the SHA1 actually present in the repo?
+# it could be it isn't, e.g. after a force push
+commit_valid() {
+	local rev=$1
+	git rev-parse --quiet --verify "$rev^{commit}"
```
krnowak: Maybe add `>/dev/null` at the end? The `--quiet` option prevents the printing of the error, but in case of the correct rev, it will print it in its full form.

tools/rebuild-image
```diff
@@ -40,13 +40,27 @@ commit_timestamp() {
...
+
 cached_revision=$(cached_image_rev)
 if [ -z "$cached_revision" ]; then
 	echo ">>> No cached image found; rebuilding"
 	rebuild
 	exit 0
 fi
 
+if ! commit_valid "$cached_revision"; then
+  echo ">>> cached SHA1 not found in repo, rebuilding"
```
krnowak: Fix indentation to one tab instead of two spaces. Also, `s/cached/Cached` and `s/,/;`.

tomwilkie:
Can you file this against build-tools.git pls?  Then update the subtree in here.

PR #1926, opened 2016-10-17
Improve view-node-in-topo button usability

### LINKED ISSUE #1151: navigation action from details panel to topology is non-obvious
It took me ages to discover that there was a way to navigate back from the details panel to the topology, via the '<-->' button next to the close button.

Can we think of a more obvious way of perform this action?

One, perhaps too radical (and quite possibly already dismissed) approach would be to automatically update the topology view, so what you are seeing in the topology always corresponds to the top details panel.

foot:
Fixes #1151

### DIFF
--- a/client/app/scripts/actions/app-actions.js
+++ b/client/app/scripts/actions/app-actions.js
@@ -286,18 +286,12 @@ export function clickRelative(nodeId, topologyId, label, origin) {
       topologyId
     });
     updateRoute(getState);
-    resetUpdateBuffer();
     const state = getState();
     getNodeDetails(
       state.get('topologyUrlsById'),
       state.get('currentTopologyId'),
       getActiveTopologyOptions(state),
       state.get('nodeDetails'),
-      dispatch
-    );
-    getNodesDelta(
-      getCurrentTopologyUrl(state),
-      getActiveTopologyOptions(state),
       dispatch
     );
   };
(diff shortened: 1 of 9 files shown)


davkal:
> IMO it's not obvious that the shape is a control. Let alone what that control will do.

I agree, it's not clear what the control will do. The same is true for the old button, so I see no improvement here. In these situations I suggest to use a plain text label to describe what the button  will do. 

> navigation in a sub-view updating the whole UI, a lot of things changing in reaction to following a link.

Yes, I would find that confusing, because at that point you're in the context of a node, the change of a topology would be too big. 

There is another option, let's simply remove the button, and see if anyone complains.

rade:
> In these situations I suggest to use a plain text label to describe what the button will do. 

"Find in Topology"? "Find in Graph"? "Find"?

> There is another option, let's simply remove the button, 

The fact that the single button changes a lot is exactly what makes it useful.

foot:
screen shot 2016-11-03 at 12 39 49

testing out labels. feels pretty good.

rade:
"View"? "Find"? "Locate"?

foot:
I like View! Find/Locate sound expensive like we're gonna have to do some work to perform that action... but maybe that works well in envs where it takes a couple of seconds to load up the next topo. j/k!

I like: View / Show
Also: Focus / Examine / Inspect / Select
Thesaurus provides: Behold / Witness / Scrutinise. 


Wonder if there is webpack module to thesaursize all your copy.

foot:
screen shot 2016-11-03 at 15 50 43

Another variant w/ the more familiar ALLCAPS topology label.

rade:
I quite like that.

davkal:
Circle fails on BE test.

davkal:
Code and copy LGTM

Issue #1866, opened 2016-09-15
Time format in table and details panel is weird


foot:
`02 Aug 16 10:47 UTC` is OK in isolation, but harder to quickly parse in a list. Let's try a readable ISO format.

![screen shot 2016-08-16 at 20 54 13](https://cloud.githubusercontent.com/assets/859729/17713601/a90460da-63f3-11e6-8e58-27403eac7f04.png)

jpellizzari:
Fixed by PR #1927

PR #1927, opened 2016-10-17
1866: Improve metadata table 'date' format

jpellizzari:
From this issue: https://github.com/weaveworks/scope/issues/1866

This PR is to improve the date formatting of the metadata table values. They will change from this: 
1866_before
To this:
1866_after

The full mouse-over title would be something like Mon, 17 Oct 2016 17:23:00 GMT

### DIFF
--- a/client/app/scripts/components/node-details/node-details-table-row.js
+++ b/client/app/scripts/components/node-details/node-details-table-row.js
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 
 import NodeDetailsTableNodeLink from './node-details-table-node-link';
 import NodeDetailsTableNodeMetric from './node-details-table-node-metric';
-
+import { formatDataType } from '../../utils/string-utils';
 
 function getValuesForNode(node) {
   const values = {};
@@ -31,7 +31,6 @@ function getValuesForNode(node) {
   return values;
 }
 
-
 function renderValues(node, columns = [], columnStyles = []) {
   const fields = getValuesForNode(node);
   return columns.map(({id}, i) => {
@@ -39,11 +38,12 @@ function renderValues(node, columns = [], columnStyles = []) {
     const style = columnStyles[i];
     if (field) {
       if (field.valueType === 'metadata') {
+        const {value, title} = formatDataType(field);
         return (
-          <td className="node-details-table-node-value truncate" title={field.value}
+          <td className="node-details-table-node-value truncate" title={title}
             style={style}
             key={field.id}>
-            {field.value}
+            {value}
           </td>
         );
       }
--- a/client/app/scripts/utils/string-utils.js
+++ b/client/app/scripts/utils/string-utils.js
@@ -2,6 +2,7 @@ import React from 'react';
 import filesize from 'filesize';
 import d3 from 'd3';
 import LCP from 'lcp';
+import moment from 'moment';
 
 const formatLargeValue = d3.format('s');
 
@@ -72,4 +73,23 @@ export function slugify(label) {
 
 export function longestCommonPrefix(strArr) {
   return (new LCP(strArr)).lcp();
+}
+
+// Formats metadata values. Add a key to the `formatters` obj
+// that matches the `dataType` of the field. You must return an Object
+// with the keys `value` and `title` defined.
+export function formatDataType(field) {
+  const formatters = {
+    datetime(dateString) {
+      const date = moment(new Date(dateString));
+      return {
+        value: date.fromNow(),
+        title: date.format('YYYY-MM-DD HH:mm:ss.SSS')
+      };
+    }
+  };
+  const format = formatters[field.dataType];
+  return format
+    ? format(field.value)
+    : {value: field.value, title: field.value};
 }
--- a/client/app/scripts/components/node-details/node-details-info.js
+++ b/client/app/scripts/components/node-details/node-details-info.js
@@ -3,6 +3,7 @@ import { Map as makeMap } from 'immutable';
 
 import MatchedText from '../matched-text';
 import ShowMore from '../show-more';
+import { formatDataType } from '../../utils/string-utils';
 
 export default class NodeDetailsInfo extends React.Component {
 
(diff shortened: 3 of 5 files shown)


rade:
We don't really want users to have to hover to get the time. So either the column width needs increasing or we pick a more compact date & time format.

davkal:
@jpellizzari getting started early 😄 

The issue #1866 suggested to use an ISO date, ISO 8601 that is. So something like 2016-10-12 14:12:56.234

davkal:
Maybe for the rows a relative time would be helpful too, e.g. `4 mins ago`, and then show the real datetime in ISO format on hover.

rade:
"23ms ago" :)

jpellizzari:
I was able to use Moment to get a nice human readable string:
screenshot from 2016-10-17 12-32-17

@davkal The input date string doesn't give enough granularity to go down to the second or millisecond, so you would end up with a false timestamp that is rounded to the second:
`17 Oct 16 17:23 UTC => 2016-10-17 17:23:00.000`

rade:
Nice.

The granularity thing is odd. @2opremio any idea?

We want this formatting to apply to all timestamp data, not just `docker_container_created`.

I am worried about the on-hover "YYYY-MM-DD HH:MM" confusing our American users.

jpellizzari:
@rade I don't see any other timestamp-like data in the node:
```json
[
  "docker_container_id",
  "docker_container_state_human",
  "docker_container_command",
  "docker_image_id",
  "docker_container_uptime",
  "docker_container_restart_count",
  "docker_container_ips",
  "docker_container_created",
  "weave_dns_hostname"
]
```
`docker_container_uptime` maybe, but that sounds like more of a counter than a duration. All of these values come through as strings; it might be nice to have a `type` in the metadata payload field that the UI could use to select the right formatting.

```json
{
  "id": "docker_container_created",
  "label": "Created",
  "value": "17 Oct 16 17:23 UTC",
  "type" : "datetime",
  "priority": 17
}
```

rade:
> I don't see any other timestamp-like data in the node

Look at some other views, e.g. processes, hosts. k8s views should have a few too, but I am guessing you aren't running that.

> All of these values come through as strings

Eek, that is gross.

> it might be nice to have a type in the metadata payload field that the UI could use to select the right formatting.

Not just nice, essential. We don't want to have to list all the date fields in the f/e code!

2opremio:
> The granularity thing is odd. @2opremio any idea?

I vaguely recall being surprised to notice the lack of sub-second timestamps in some information provided by Docker. But I must be wrong:

```
$ docker inspect weavescope | grep Created                                                                           
        "Created": "2016-10-17T14:07:06.404190024Z",
```

@jpellizzari What metadata lacks sub-second precision?

jpellizzari:
I traced the metadata object all the way back to the Websockets traffic to make sure the client wasn't formatting the data.

@2opremio This was the field in question
```json
{
  "id": "docker_container_created",
  "label": "Created",
  "value": "17 Oct 16 21:10 UTC",
  "priority": 17
}
```

rade:
re lack of precision... #1928.

re tagging the data... that's harder than it may first appear and is covered in #1735.

re other timestamp data... looks like currently the only place besides ContainerCreated is in the k8s data.

client/app/scripts/components/node-details/node-details-table-row.js
```diff
@@ -30,7 +31,24 @@ function getValuesForNode(node) {
 
   return values;
 }
-
+// Formats metadata table values. Add a key to the `formatters` obj
+// that matches the `id` arg to format columns. You must return an Object
+// with the keys `value` and `title` defined.
+function formatValue(field, id) {
+  const formatters = {
```
davkal: I suggest formatter be backend driven:

```diff
diff --git a/probe/docker/reporter.go b/probe/docker/reporter.go
index f9bf682..f591c2e 100644
--- a/probe/docker/reporter.go
+++ b/probe/docker/reporter.go
@@ -32,7 +32,7 @@ var (
                ContainerNetworks:     {ID: ContainerNetworks, Label: "Networks", From: report.FromSets, Priority: 14},
                ContainerIPs:          {ID: ContainerIPs, Label: "IPs", From: report.FromSets, Priority: 15},
                ContainerPorts:        {ID: ContainerPorts, Label: "Ports", From: report.FromSets, Priority: 16},
-               ContainerCreated:      {ID: ContainerCreated, Label: "Created", From: report.FromLatest, Priority: 17},
+               ContainerCreated:      {ID: ContainerCreated, Label: "Created", From: report.FromLatest, Datatype: "datetime", Priority: 17},
        }
 
        ContainerMetricTemplates = report.MetricTemplates{
```

client/app/scripts/components/node-details/node-details-table-row.js
```diff
@@ -30,7 +31,24 @@ function getValuesForNode(node) {
...
+    docker_container_created(dateString) {
```
davkal: then this becomes `datetime`.

client/app/scripts/components/node-details/node-details-table-row.js
```diff
@@ -30,7 +31,24 @@ function getValuesForNode(node) {
...
+      const date = moment(dateString);
+      return {
+        value: date.fromNow(),
+        title: date.format('YYYY-M-D hh:mm')
+      };
+    }
+  };
+  const format = formatters[id];
```
davkal: and this `formatters[field.dataType]`.

davkal:
For the purpose of this PR, the datetime precision is not that important. I do however think we should move the format type to the backend, and have the frontend check all fields. That would be consistent with previous design decisions.

jpellizzari:
@davkal Updated with the changes you suggested! I considered filling out the rest of the Datatype fields in the reporter, but I thought I might be getting ahead of myself ;).

davkal:
Waiting on #1933 to be merged, which adds time precision and changes the date format of the backend data, so we need to see if the this PR still works with that. Additionally we have to verify if the datetime format in the table and the details panel are the same.

Currently they are not, see screenshot (where the table tooltip says 2016-10-20 12:55, local time)

screen shot 2016-10-20 at 12 58 58

rade:
> re other timestamp data... looks like currently the only place besides ContainerCreated is in the k8s data.

That still needs doing, doesn't it?

jpellizzari:
I've added the `Datatype` field to the kubernetes reporter.go Metadata templates variables, but I am having trouble building the binary to test. I have a k8s installation running on a fresh VM, and when I try to run `make` the build fails with this error:

`> Unable to find image 'weaveworks/scope-backend-build:latest' locally`
`> Pulling repository docker.io/weaveworks/scope-backend-build`
`> docker: Error: image weaveworks/scope-backend-build:latest not found.`

I got the same error when I removed the image from my macbook and ran `make` again. Did that docker image get removed? I don't see it on hub.docker.com.

Steps to reproduce:
1. `docker rmi weaveworks/scope-backend-build` (careful! you may not be able to build after this)
2. `docker pull weaveworks/scope-backend-build` or run `make` to try and build

jpellizzari:
I also moved the formatter func to the `string-utils.js` file so it can be reused by the table and the details view (and whatever else) and applied the formatting to the details view.

rade:
weaveworks/scope-backend-build:latest never gets pushed; 'make' should build it. try 'make clean'.

jpellizzari:
@davkal I was able to pull in the changes from #1933 and test them locally. The mouseover values have the extra precision on both the table row and the detail view: 


The only thing left on this PR should be the open question about the ideal date format that @rade mentioned.

client/app/scripts/utils/string-utils.js
```diff
@@ -73,3 +74,22 @@ export function slugify(label) {
...
+// Formats metadata values. Add a key to the `formatters` obj
+// that matches the `dataType` of the field. You must return an Object
+// with the keys `value` and `title` defined.
+export function formatDataType(field) {
+  const formatters = {
+    datetime(dateString) {
+      const date = moment(new Date(dateString));
+      return {
+        value: date.fromNow(),
+        title: date.format('YYYY-M-D hh:mm:ss.SSS')
```
davkal: This should be `'YYYY-MM-DD HH:mm:ss.SSS'`.

jpellizzari:
@davkal Time format updated.

davkal:
All looking good! 🍡

PR #1931, opened 2016-10-18
probe: process: optionally use the proc connector

alban:
Before this patch:
- The process walker (probe/process/walker_linux.go) iterates over /proc
  at every Tick (every second) to get all the processes with their
  details.
- There are two kind of process details: the constant ones (pid, name,
  cmdline) and the dynamic ones (# threads, memory, cpu, # open files).
  The constant details were read only one time (cached via
  cachedReadFile) and the dynamic ones were read every time.
- Short-lived processes were rarely caught.

This patch adds a new package "procconnector" (for proc connector) that
receives notifications for fork, exec and exit events. It maintains the
list of processes in the map "activePids" and update it whenever a proc
connector event is received. Since it is event based, it is unlikely to
miss any process.

ProcConnector has a Walk() function that iterates over both activePids
(running processes) and bufferedPids (processes terminated since the
last iteration). This is exactly the same logic than in
probe/endpoint/conntrack.go with activeFlows and bufferedFlows: this
ensures that all processes are displayed in at least one iteration.

Since the Scope process view only displays processes that have
connections, and short-lived processes have short-lived connections,
this patch is not immediately useful by itself. But with the coming work
to get connections from eBPF, this will allow us to display short-lived
processes and short-lived connections.

When running this patch on top of the eBPF work, I can successfully see
the connection between the two "nc" processes generated by the following
script:

> # !/bin/bash
> 
> PORT=$RANDOM
> let "PORT %= 1000"
> let "PORT += 50000"
> echo "PORT=$PORT"
> nc -l $PORT &
> sleep 0.2
> echo Hello | nc 127.0.0.1 $PORT
> wait

The proc connector requires a kernel compiled with CONFIG_PROC_EVENTS=y.
The feature has been added in v2.6.15 released on 3 January, 2006, see
https://github.com/torvalds/linux/commit/9f46080c41d5f3f7c00b4e169ba4b0b2865258bf

CONFIG_PROC_EVENTS=y is usually enabled in Linux distributions. However,
Scope should gracefully fall back to the previous /proc polling logic.

---

/cc @asymmetric @iaguis

### DIFF
--- a/integration/313_short_lived_process.sh
+++ b/integration/313_short_lived_process.sh
@@ -1,0 +1,21 @@
+#! /bin/bash
+
+# shellcheck disable=SC1091
+. ./config.sh
+
+start_suite "Test short lived connections in short lived processes between containers, with ebpf connection tracking enabled, require proc connector"
+
+weave_on "$HOST1" launch
+scope_on "$HOST1" launch --probe.ebpf.connections=true
+weave_on "$HOST1" run -d --name server alpine /bin/sh -c "sleep 2 ; nc -l -p 8080 ; sleep 2"
+weave_on "$HOST1" run -d --name client alpine /bin/sh -c "sleep 5 ; echo Hello | nc server.weave.local 8080 ; sleep 2"
+
+wait_for_containers "$HOST1" 60 server client
+
+has_container "$HOST1" server
+has_container "$HOST1" client
+has_connection containers "$HOST1" client server
+
+endpoints_have_ebpf "$HOST1"
+
+scope_end_suite
--- a/probe/process/proc_connector/proc_connector.go
+++ b/probe/process/proc_connector/proc_connector.go
@@ -1,0 +1,339 @@
+package procconnector
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"os"
+	"path"
+	"strconv"
+	"strings"
+	"sync"
+	"syscall"
+	"unsafe"
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/weaveworks/common/fs"
+)
+
+const (
+	// <linux/connector.h>
+	cnIdxProc = 0x1
+	cnValProc = 0x1
+
+	// <linux/cn_proc.h>
+	procCnMcastListen = 1
+
+	procEventFork = 0x00000001 // fork() events
+	procEventExec = 0x00000002 // exec() events
+	procEventExit = 0x80000000 // exit() events
+)
+
+var (
+	byteOrder binary.ByteOrder
+)
+
+func init() {
+	var i int32 = 0x01020304
+	u := unsafe.Pointer(&i)
+	pb := (*byte)(u)
+	b := *pb
+	if b == 0x04 {
+		byteOrder = binary.LittleEndian
+	} else {
+		byteOrder = binary.BigEndian
+	}
+}
+
+// ProcConnector receives events from the proc connector and maintain the set
+// of processes.
+type ProcConnector struct {
+	sockfd       int
+	seq          uint32
+	lock         sync.RWMutex
+	activePids   map[int]Process
+	bufferedPids []Process
+	running      bool
+}
+
+// Process represents a single process. Only include the constant details here.
+type Process struct {
+	Pid     int
+	Name    string
+	Cmdline string
+}
+
+// linux/connector.h: struct cb_id
+type cbID struct {
+	Idx uint32
+	Val uint32
+}
+
+// linux/connector.h: struct cb_msg
+type cnMsg struct {
+	ID    cbID
+	Seq   uint32
+	Ack   uint32
+	Len   uint16
+	Flags uint16
+}
+
+// linux/cn_proc.h: struct proc_event.{what,cpu,timestamp_ns}
+type procEventHeader struct {
+	What      uint32
+	CPU       uint32
+	Timestamp uint64
+}
+
+// linux/cn_proc.h: struct proc_event.fork
+type forkProcEvent struct {
+	ParentPid  uint32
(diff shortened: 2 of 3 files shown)


asymmetric:
What's changed since `alban/short-lived-processes-3`?

alban:
@asymmetric I added some comments on exported functions that were missing.

It seems I forgot to check lint. I will fix that.

2opremio:
Thanks a lot @alban ! Why not reuse https://github.com/mindreframer/golang-stuff/tree/master/github.com/jondot/gosigar/psnotify instead of implementing it yourself? Does it lack any functionality you need?

2opremio:
Actually, https://github.com/cloudfoundry/gosigar/tree/master/psnotify seems to be the original implementation.

alban:
I did a fix after the original code: the [byte ordering](https://github.com/cloudfoundry/gosigar/blob/master/psnotify/psnotify_linux.go#L33) (my code no longer assumes that the machine is little endian).

I also added state in my ProcConnector, to keep the set of pids with their constant details (pid, name, cmdline). I don't use channels to pass the information but a Walk() function. I guess I could vendor the original implementation and move the logic about maintaining the state in another file. I don't know if it worth it...

When I receive the "new process" message, I would prefer to parse /proc/pid/cmdline as soon as possible without extra IPC (Golang channels) because there is always the race where the process terminates before I could read /proc/pid/cmdline. I didn't have that race in practice so far but if I used Golang channels, it could make it a bit more likely.

2opremio:
> I don't know if it worth it...

Uhm, unless we plan to make larger modifications, I would rather try to upstream the changes and reuse the upstream repository (assuming is actively maintained). 

I am not sure we want to maintain this kind of library code. As a prior example, we took the procspy code and turned it into a bit of a monster.

2opremio:
Did you manage to get that CPU comparison between master and this PR?

As discussed, we will need to see a noticeable improvement  to justify this. (A code simplification would also justify it but IMO that doesn't seem to be the case).

alban:
No, I didn't get it. I am leaving this aside for now, and focus on ebpf.

2opremio:
@alban Since ebpf is merged now, are you willing to retake this work?

alban:
Yes, I'll go back to this.

alban:
I rebased this branch, fixed conflicts and fixed some test issues. I added a new test but it is not ready. The performances have not been tested yet.

I'll go back to this when I get some time after the ebpf remaining tasks.

schu:
Results from testing the performance of this branch vs. master on a GCE n1-standard-4 (4 vCPUs, 15 GB memory) instance.

With 50 running containers and 10 connections per container + stopping an old / starting a new container every 2 seconds:

proc-connector-perf

With 50 running containers and 10 connections per container:

proc-connector-perf-static

We attribute the performance difference between this branch and master to better tracking of processes (e.g. short-lived processes) and thus the processing of more events. The second graph seems to hint that (less overhead with a fixed number of containers and connections).

Also, further optimizations might be possible. For instance, other events than exec/fork/exit are received over proc connector. Perhaps those could be filtered in the kernel with BPF.

I used the dialer scripts from #2350.for the test. For the second run, `dialer` was modified to not stop/start containers after reaching the desired number (here 50).

cc @alban

2opremio:
> With 50 running containers and 10 connections per container + stopping an old / starting a new container every 2 seconds

This doesn't seem to create a lot of processes. Conceptually the proc connector should be cheaper than walking the whole proc (even if rate-limited) ...

> For instance, other events than exec/fork/exit are received over proc connector

Maybe this is it. It may be experimenting with this.

The proc connector will give us more accuracy but, considering reports are sent every 3 seconds, it's not really worth it unless it also gives us better performance.

schu:
We also tested with 100 containers with 100 processes each and no network activity, with similar results (master performing a bit better):

proc-connector-perf-100-100

This is due to the fact that the proc connector approach has to do extra work (processing events) while the same amount of data has to be collected/processed from on places (e.g. `/proc/pid/stat`) with and without proc connector.

>> For instance, other events than exec/fork/exit are received over proc connector

> Maybe this is it. It may be worth experimenting with this.

> The proc connector will give us more accuracy but, considering reports are sent every 3 seconds, it's not really worth it unless it also gives us better performance.

We see other events during container/process starts (`sid` + `comm`) and a few `uid`/`gid` in between, but not so many that we think there would be a benefit (i.e. performance gain) from a socket filter (BPF).

2opremio:
> This is due to the fact that the proc connector approach has to do extra work (processing events) 

Yes, but there is no polling, which should be expensive in a busy system. I still don't get this, it's like saying that HW interruptions are less performant than HW polling  because you need to run interruption handlers. 

There may be corner cases in which that's the case, but I still don't see how it can be slower.

Are you certain that the proc walker is off?

alban:
Even with the proc connector enabled, we still read regularly the files `/proc/<pid>/{stat,fd,limits}` in the `walkOne` function because the values in those files can change and we don't get the information about them from the proc connector.

With the proc connector, we don't iterate on the /proc/* entries but this saves us only one [Readdirnames](https://golang.org/pkg/os/#File.Readdirnames) call.

Both with and without the proc connector, the files `/proc/<pid>/{cmdline,comm}` are read only one time, so the proc connector does not bring much here.
- with the proc connector, they are read when the process is created (fork/exec events)
- without the proc connector, there is a file system cache to avoid reading it several times.

We could try to read the statistics about cpu/memory/fd_counter/limits less often because it's not that critical to get the uptodate data quickly. The basic data (pid, comm, cmdline) would still be retrieved quickly when we get the proc connector events.

alban:
As discussed, closing because it does not show good performance.

PR #1932, opened 2016-10-18
Enable scroll locking on the terminal

### LINKED ISSUE #1481: Allow scroll-locking the terminal views
When using the attach control on a container or running commands with continuos output I cannot go back in the terminal history, since the terminal automatically scrolls down to the newest output.

The ideal behavior would be to only get automatic scrolling when the terminal is showing the very end of the output.

foot:
Fixes #1481

### DIFF
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -184,7 +184,8 @@ class Terminal extends React.Component {
       cols: this.state.cols,
       rows: this.state.rows,
       convertEol: !this.props.raw,
-      cursorBlink: true
+      cursorBlink: true,
+      scrollback: 10000,
     });
 
     const innerNode = ReactDOM.findDOMNode(component.innerFlex);
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -168,6 +168,7 @@ class Terminal extends React.Component {
       return;
     }
     this.term.ydisp = y;
+    this.term.emit('scroll', y);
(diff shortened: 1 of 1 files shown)


foot:
@2opremio could you give this a quick test please!

2opremio:
Sure, I'll do that before EOD.

davkal:
If I run the following and try to scroll up, the screen goes all black (before the fix it would jump around but show output.
```
while true; do find /etc; done                                
```

foot:
Good one, that blows through the scroll back nice and quickly..

2opremio:
I am missing a side bar to indicate where am I in the history.

2opremio:
Also, I would expect scrolling to happen when selecting text and moving the mouse up. That is:

foot:
Scrollbars are a nice to have, mentioned in #1940.

Scrolling up as we drag up is an interesting one. I'll mention it in that issue too.

foot:
But you're looking for a way to export the contents of the terminal somehow. Other (less ideal but still) options:
- Copy terminal contents to clipboard
- Show terminal contents in some other mode that allows for better selection. (textarea/plain `<p>`s)

foot:
Issue w/ this: cannot make a selection after having scrolled up.

davkal:
Scrolling up works well, and the position stays there. Selections dont work when scrolled up, but that's mentioned. Definitely an improvement. LGTM

2opremio:
>  Selections dont work when scrolled up, but that's mentioned

Is there an issue for this?

PR #1933, opened 2016-10-18
Increase timestamp precision

### LINKED ISSUE #1928: probe sends timestamps with precision of one minute
The probe sends timestamps as a string formatted with `t.Format(time.RFC822)`, which only has minute precision.

AFAIK the probe data currently contains timestamps in just four places:
- docker container creation timestamp
- k8s 'created' metadata
- report timestamp
- metrics first/last timestamp

The former two user `time.RFC822`, the latter use `time.RFC3339Nano`.

Poking around our vendored code, `RFC3339Nano` is quite popular, so is `RFC3339` and `RFC1123Z`.

I suggest we switch to `RFC3339Nano` throughout. It will cost us a few more bytes; if we are concerned about that then RFC3339.

And we should leave the user-friendly formatting to the f/e; see #1866.

2opremio:
Fixes #1928 

screen shot 2016-10-18 at 16 55 18

### DIFF
--- a/client/app/scripts/components/node-details/node-details-table.js
+++ b/client/app/scripts/components/node-details/node-details-table.js
@@ -30,7 +30,7 @@ const XS_LABEL = {
 const COLUMN_WIDTHS = {
   count: CW.XS,
   container: CW.XS,
-  docker_container_created: CW.XL,
+  docker_container_created: CW.XXL,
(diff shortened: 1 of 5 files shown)


davkal:
Gave the datetime more space (was showing only minutes before the `...`).

Otherwise LGTM

Issue #925, opened 2016-02-05
Show multiple images ids in details panel


tomwilkie:
When image is on multiple hosts.

Also test with multiple versions.

tomwilkie:
With https://github.com/weaveworks/scope/pull/1531 we're merging images properly, so we should display this in the details panel.  We should add a table with image id, tag and number of containers using it.

2opremio:
> We should add a table with image id, tag and number of containers using it.

The only way I see to make this this to work _naturally_ with merges is to make tables based on StringSets instead of LatestMap.

@tomwilkie any alternative suggestions?

PR #1942, opened 2016-10-20
Add image information table to container details panel

### LINKED ISSUE #1800: Show which image a container is running
In order of importance:
- image name
- image label
- link to where image is from (e.g. dockerhub, quay.io URL)

Secondarily:
- size
- age
- other stuff

Ideally also directly on pod, replication controller, and deployment views too, since it should be the same on all of them.

Kubernetes views should highlight if pods in the same rc, deployment, etc. differ (they really never should, but if they _do_, then that's very interesting).

2opremio:
Fixes #1800 

screen shot 2016-10-20 at 13 24 07

### DIFF
--- a/probe/docker/reporter.go
+++ b/probe/docker/reporter.go
@@ -26,16 +26,16 @@ const (
 // Exposed for testing
 var (
 	ContainerMetadataTemplates = report.MetadataTemplates{
-		ContainerID:           {ID: ContainerID, Label: "ID", From: report.FromLatest, Truncate: 12, Priority: 1},
-		ContainerStateHuman:   {ID: ContainerStateHuman, Label: "State", From: report.FromLatest, Priority: 2},
-		ContainerCommand:      {ID: ContainerCommand, Label: "Command", From: report.FromLatest, Priority: 3},
-		ImageID:               {ID: ImageID, Label: "Image ID", From: report.FromLatest, Truncate: 12, Priority: 11},
-		ContainerUptime:       {ID: ContainerUptime, Label: "Uptime", From: report.FromLatest, Priority: 12},
-		ContainerRestartCount: {ID: ContainerRestartCount, Label: "Restart #", From: report.FromLatest, Priority: 13},
-		ContainerNetworks:     {ID: ContainerNetworks, Label: "Networks", From: report.FromSets, Priority: 14},
-		ContainerIPs:          {ID: ContainerIPs, Label: "IPs", From: report.FromSets, Priority: 15},
-		ContainerPorts:        {ID: ContainerPorts, Label: "Ports", From: report.FromSets, Priority: 16},
-		ContainerCreated:      {ID: ContainerCreated, Label: "Created", From: report.FromLatest, Datatype: "datetime", Priority: 17},
+		ImageName:             {ID: ImageName, Label: "Image", From: report.FromLatest, Priority: 1},
+		ContainerCommand:      {ID: ContainerCommand, Label: "Command", From: report.FromLatest, Priority: 2},
+		ContainerStateHuman:   {ID: ContainerStateHuman, Label: "State", From: report.FromLatest, Priority: 3},
+		ContainerUptime:       {ID: ContainerUptime, Label: "Uptime", From: report.FromLatest, Priority: 4},
+		ContainerRestartCount: {ID: ContainerRestartCount, Label: "Restart #", From: report.FromLatest, Priority: 5},
+		ContainerNetworks:     {ID: ContainerNetworks, Label: "Networks", From: report.FromSets, Priority: 6},
+		ContainerIPs:          {ID: ContainerIPs, Label: "IPs", From: report.FromSets, Priority: 7},
+		ContainerPorts:        {ID: ContainerPorts, Label: "Ports", From: report.FromSets, Priority: 8},
+		ContainerCreated:      {ID: ContainerCreated, Label: "Created", From: report.FromLatest, Datatype: "datetime", Priority: 9},
+		ContainerID:           {ID: ContainerID, Label: "ID", From: report.FromLatest, Truncate: 12, Priority: 10},
 	}
 
 	ContainerMetricTemplates = report.MetricTemplates{
--- a/probe/docker/reporter.go
+++ b/probe/docker/reporter.go
@@ -4,6 +4,7 @@ import (
 	"net"
 	"strings"
 
+	humanize "github.com/dustin/go-humanize"
 	docker_client "github.com/fsouza/go-dockerclient"
 
 	"github.com/weaveworks/scope/probe"
@@ -15,8 +16,11 @@ import (
 const (
 	ImageID          = "docker_image_id"
 	ImageName        = "docker_image_name"
+	ImageSize        = "docker_image_size"
+	ImageVirtualSize = "docker_image_virtual_size"
 	ImageLabelPrefix = "docker_image_label_"
 	IsInHostNetwork  = "docker_is_in_host_network"
+	ImageTableID     = "image_table"
 )
 
 // Exposed for testing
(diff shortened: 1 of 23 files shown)


jml:
Thanks!

I think we should re-order the headings: STATUS, IMAGE, DOCKER LABELS, ENVIRONMENT VARIABLES, PROCESSES.

rade:
Can we shift the whole thing to the left, so less space is used by keys and more by values?

probe/docker/reporter.go
```diff
@@ -237,16 +249,17 @@ func (r *Reporter) containerImageTopology() report.Topology {
...
-		node = node.AddTable(ImageLabelPrefix, image.Labels)
-
+		latests := map[string]string{
+			ImageID:          imageID,
+			ImageSize:        humanize.Bytes(uint64(image.Size)),
+			ImageVirtualSize: humanize.Bytes(uint64(image.VirtualSize)),
+		}
 		if len(image.RepoTags) > 0 {
-			node = node.WithLatests(map[string]string{ImageName: image.RepoTags[0]})
+			latests[ImageName] = image.RepoTags[0]
```
jml: How does this work? Is the first thing in `RepoTags` the latest?

2opremio:
> Can we shift the whole thing to the left, so less space is used by keys and more by values?

@davkal / @foot Wanna do the honors? :)

report/table.go
```diff
@@ -17,8 +17,8 @@ const (
 	TruncationCountPrefix = "table_truncation_count_"
 )
 
-// AddTable appends arbirary key-value pairs to the Node, returning a new node.
-func (node Node) AddTable(prefix string, labels map[string]string) Node {
+// AddPrefixTable appends arbirary key-value pairs to the Node, returning a new node.
```
jml: Nit: "arbitrary"

report/table_test.go
```diff
@@ -28,6 +28,29 @@ func TestTables(t *testing.T) {
...
+	nmd := report.MakeNodeWith("foo1", map[string]string{
+		"foo1key": "bar1",
+		"foo2key": "bar2",
+	})
+
+	template := report.TableTemplate{FixedRows: map[string]string{
+		"foo1key": "foo1",
+		"foo2key": "foo2",
+	},
+	}
```
jml: Kind of surprised this what gofmt likes. No action required.

rade:
> I think we should re-order the headings: STATUS, IMAGE, DOCKER LABELS, ENVIRONMENT VARIABLES, PROCESSES. 

I am not convinced. 'Processes' is navigable (i.e. you can click on things and get new detail panels). It's weird to have to scroll in order to navigate.

jml:
> I am not convinced. 'Processes' is navigable (i.e. you can click on things and get new detail panels). It's weird to have to scroll in order to navigate.

I just listed things in order of interest to me.

2opremio:
> I think we should re-order the headings: STATUS, IMAGE, DOCKER LABELS, ENVIRONMENT VARIABLES, PROCESSES.

I can definitely show the image info first, but  showing the processes at the end (which I am also doubtful about) requires a UI change (which will likely affect all details panels unless we want to start making exceptions).

IMAGE, DOCKER LABELS and ENVIRONMENT VARIABLES are normal tables and PROCESSES is a child table. The relative rendering order between tables and child tables is determined by the UI at https://github.com/weaveworks/scope/blob/master/client/app/scripts/components/node-details.js#L192 (kudos to @paulbellamy for finding out)

davkal:
It's a bit weird showing all those details in the container details panel, but not in the container-image details panel. The current metadata section looks a bit barren:

screen shot 2016-10-20 at 22 48 57

2opremio:
> It's a bit weird showing all those details in the container details panel, but not in the container-image details panel. The current metadata section looks a bit barren:

I agree but that is part of #925. The nodes in the image topoly represent an image name (which potentially corresponds to multiple images). Doing this right involves rendering tables to include the IDs/Sizes/Tags of multiple images is a bit harder to achieve and is out of the scope of this PR.

> This also raises the question, shouldnt it be enough to be there?

I asked this to @jml and he claimed it's not the case (please see above)

> If it's so important as @jml claims, I suggest to put the first 3 items in the meta data section of the container

Uhm, that section is already crowded and requires expanding, that's why I went for a separate table (and added the extra table type required to make that possible)

davkal:
> Uhm, that section is already crowded and requires expanding, that's why I went for a separate table (and added the extra table type required to make that possible)

jml's point was that currently shown priority (i.e. non-expanded) fields are useless to him. My suggestion was to add the newly important fields as priority fields and demote some current priority fields. That way it will look less cluttered/crowded and will require no scrolling.

2opremio:
I agree but that is part of #925. The nodes in the image to poly represent
an image name (which potentially corresponds to multiple images). That is a
bit harder to achieve and is out of the scope of this PR.

2opremio:
> My suggestion was to add the newly important fields as priority fields and demote some current priority fields

Yep, got it :) However I don't think the image details (e.g. image size) aren't the most important information about the container and we only display 3 fields in the image.

> That way it will look less cluttered/crowded and will require no scrolling.

It doesn't require scrolling on a Desktop screen (as per the screenshot attached).

jml:
@2opremio asked me:

> [October 27th, 2016 6:23 AM] fons: @jml Can you unblock https://github.com/weaveworks/scope/pull/1942 by choosing the option which is better IYO?

After reading this carefully, I am not clear on what the alternatives are.

2opremio:
> After reading this carefully, I am not clear on what the alternatives are.

The alternatives are:
1. Leave the image information in its own separate table, as it is in the current PR. (my inclination)
2. Move the image information to the main metadata section (_Info_ section). (@davkal's inclination)

jml:
Thanks for clarifying, @2opremio. 

tl;dr: I think we should:

* put the image name & label as one field "Image" in the top Info table
* either 
  * change Info to allow more fields as visible by default 
  * push container ID "below the fold" (i.e. only visible when Info expanded)
* put the rest of the image details in an image section at the bottom, as it is now, possibly repeating the image name & label stuff for context

Thoughts and a huge pile of questions below.

> we only display 3 fields in the metadata without expanding

This seems pretty arbitrary (we display more environment variables, for example). It also seems that the things we've chosen to be at the top are also fairly arbitrary. Who do you think needs the Container ID first thing? What would they use it for?

What does the command mean separately from the context of the image? How would it be useful to someone who doesn't know what the image is? How useful is it _really_ to show only the first 20 characters of the command line?

> IMAGE, DOCKER LABELS and ENVIRONMENT VARIABLES are normal tables and PROCESSES is a child table.

This seems like an implementation detail? I don't know the difference between these, and I suspect most users will be similar. If a layout is more helpful to users, wouldn't we change our internal structure to accommodate them, rather than expecting them to adapt to what's easy for us?

> However I don't think the image details (e.g. image size) are the most important information about the container

Image size is definitely not most important. It's basically an irrelevance almost all of the time.

However, image name & label are almost certainly the most important thing about the container. At least, they are to me.

rade:
Let's not turn this PR into a complete redesign and re-architecting of the details panel. Modulo easy tweaks, let's get this merged provided we believe it improves on the status quo, which it does, doesn't it? And file a follow-up issue for improvements.

2opremio:
>  Let's not turn this PR into a complete redesign and re-architecting of the details panel. Modulo easy tweaks, let's get this merged provided we believe it improves on the status quo, which it does, doesn't it? And file a follow-up issue for improvements.

Agreed.

From the changes proposed by @jml I can easily do the following:

> * put the image name & label as one field "Image" in the top Info table
> * push container ID "below the fold" (i.e. only visible when Info expanded)
> * put the rest of the image details in an image section at the bottom, as it is now, possibly repeating the image name & label stuff for context

Issue #1940, opened 2016-10-20
Switch terminal library to xterm.js


foot:
We're using term.js which is no longer being maintained.

xterm.js:
- Provides scrollbars
- Has IME support
- Is maintained.

foot:
Investigate whether dragging selection up to scroll up is possible.

jpellizzari:
Fixed in PR #1948

PR #1948, opened 2016-10-25
Change term.js library to xterm.js

jpellizzari:
From this issue: https://github.com/weaveworks/scope/issues/1940

Term.js is no longer supported, so this PR is to use xterm.js instead: https://github.com/sourcelair/xterm.js. There should be no change to what the users sees.

@davkal and @foot , I added a directory called `styles/vendor` for the xterm styles and did `@import vendor/xterm.less` right in the `main.less` file. Let me know if that is the preferred way to include vendor stylesheets.


foot:
Nice one!

Small issues:
- term.js was swallowing keypresses when focused. xterm.js appears not to be doing that, our `App.onKeyPress|onKeyUp` is triggered on shortcut keys: (e.g. term loses focus on `/` as the search bar is focused). Some options:
 - add more code to the App key handlers to return immediately if showTerminal. eh
 - listen for and cancel key events on terminal wrapper element. ugh
 - something more elegant?
- ctrl-left-arrow / ctrl-right-arrow (which we hacked into vendor/term.js ourselves) don't work for me, confirm if they work for you? 

> the preferred way to include vendor stylesheets.

I don't think we have one atm but it would be nice to keep scope/vendor/* as small as possible. Perhaps either
 - In main.less:
`@import (inline) '../../node_modules/xterm/dist/xterm.css';`
 - Or (untested) something like `require('xterm/xterm.css')` in app/scripts/main.js

jpellizzari:
- [ ] Intercept hotkeys while terminal is up
- [ ] Ctrl-Left/Right while terminal is up
- [ ] Ctrl-W to close terminal?
- [ ] Reconfigure stylesheets
- [ ] Test on a windows keyboard

jpellizzari:
@foot I added a `showingTerminal` condition to the hotkeys handler. This would prevent any hotkeys from working while the terminal is open. Good idea or bad idea?
```javascript
  onKeyPress(ev) {
    const { dispatch, searchFocused, showingTerminal } = this.props;
    //...
    if (!searchFocused && !showingTerminal) {
      keyPressLog('onKeyPress', 'keyCode', ev.keyCode, ev);
      const char = String.fromCharCode(ev.charCode);
      if (char === '<') {
        dispatch(pinNextMetric(-1));
      } else if (char === '>') {
```
Re: the Ctrl-arrow keys not working, I am not able to reproduce. On my macbook, I am able to use option-arrows to go to the beginning of each word. Under the hood, xterm has a hidden textarea that does all the input, so it should do all the stuff you can do in a textarea. 

I also tried to do a option-arrow while using the `top` program within the shell to see if it does stuff, and it seems to move things around (not sure what it does, but things move around when I press the key).

Are you using a normal (windows) or a mac keyboard?

foot:
Yes! option-arrows works \o/. My bad.

> This would prevent any hotkeys from working while the terminal is open. Good idea or bad idea?

This actually works as its only been applied to onKeyPress and not onKeyUp (which handles the `<esc>: closeDisconnectedTerminal` mapping).

We should clean it up though, will make a techdebt ticket...

foot:
Perhaps `git rebase -i origin/master` and squash away the introduction of the xterm.css to keep history a little tidier and then LGTM!

Issue #1943, opened 2016-10-20
High contrast mode needs more contrast


bowenli:
High contrast mode is mostly used for projectors/presentations where contrast is already low. The contrast still needs to be improved as it sometimes isn't enough.

This could go beyond colors, but also touch on font weight, size, etc.

See previously https://github.com/weaveworks/scope/commit/20b3c42

bowenli:
Originally via Slack discussion.

I looked at high contrast mode on the SF office projector and it it was readable only when the lights were turned off. The text is near black on white currently, but not exactly. The node colors are too subtle (light blue vs light purple vs teal).

davkal:
How about applying the (2 tone) node coloring as proposed in the last couple of comments of #1952, but only in contrast mode?

rade:
Why is this issue still open? Didn't #2006 resolve it?

davkal:
> Why is this issue still open? Didn't #2006 resolve it?

That only addressed the text. According to Bowen's last comment, contrast mode could benefit from more changes. I think applying the color change in contrast mode could work well, while keeping the look of the normal mode unchanged.

PR #1952, opened 2016-10-25
update text and background color for contrast. add fill and border to…

bowenli:
Fixes https://github.com/weaveworks/scope/issues/1943
- On contrast mode, text color set to black, background set to white (no gradient)
- Label background set to white (no transparency)
- Visual update of nodes

untitled-1
- Note: there is an existing bug where sometimes the node rendering has a gap between the fill and border. This PR does not fix this.

### DIFF
--- a/client/app/scripts/charts/node-shape-circle.js
+++ b/client/app/scripts/charts/node-shape-circle.js
@@ -4,7 +4,9 @@ import {getMetricValue, getMetricColor, getClipPathDefinition} from '../utils/me
 import {CANVAS_METRIC_FONT_SIZE} from '../constants/styles.js';
 
 
-export default function NodeShapeCircle({id, highlighted, size, color, metric}) {
+export default function NodeShapeCircle({
+  id, highlighted, size, metric, color, lightColor, darkColor
+}) {
   const clipId = `mask-${id}`;
   const {height, hasMetric, formattedValue} = getMetricValue(metric, size);
   const metricStyle = { fill: getMetricColor(metric) };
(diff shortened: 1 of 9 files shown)


davkal:
Looking at the big screen, my first reaction is that it looks less slick, but that was expected. To qualify this a bit more, I feel like the number of things on the screen vying for my attention increased.

On closer inspection I feel like the 3-color version of a node looks a bit cluttered with the two outlines.

screen shot 2016-10-28 at 13 35 08

I've created a version with two colors that makes use of a third transparent outline.

screen shot 2016-10-28 at 13 34 39
 
Here is the diff https://gist.github.com/davkal/c290c6fea25c6930eb15b43effac94d7

@foot WDYT?

rade:
How is this supposed to work with metrics-on-canvas?

davkal:
> How is this supposed to work with metrics-on-canvas?

The background turns transparent again:

screen shot 2016-10-28 at 15 02 47

foot:
Nice variations guys!

- Solid rank color on nodes does look a bit jellybeany and "less slick"
- But! makes rank much much clearer which is cool.
- But! Rank color is not actually that useful right now. #1567

- I like the two tones
- But! I also think having a node as an "unbroken" solid, unclutters things and simplifies the diagram a bit.

davkal:
Unbroken variant with 2 tones:

screen shot 2016-11-02 at 11 19 18

To add detail we could introduce the empty space at close zoom levels :eyes:

foot:
Yep! Either the 3-tone, or the transparent (or bg colored) ring could make an appearance at closer zoom levels to make it look more... slick?

rade:
>  - On contrast mode, text color set to black, background set to white (no gradient)
>  - Label background set to white (no transparency)
>  - Visual update of nodes

I suggest we do the first two only, which seem uncontroversial.

bowenli:
The outer border ended up way bigger than I envisioned. The outline controls were a bit difficult to accurately position. Here is how I designed it:
screen shot 2016-11-02 at 14 23 46

bowenli:
I think the 2 tone version is still nicer than it currently is.

pidster:
If the original issue is about readability, the PR fixing it should address just that and not a change of colour - which seems unrelated to me.

davkal:
I'm fine with the first two items as well.

As for the node changes, I would propose to go with 2 tones, but have the light one be lighter than what `getNodeColorLight` currently returns for the normal mode, and maybe the current `getNodeColorLight` for the HC mode. Also, `getNodeColorLight` is probably used in other places too, so you may need two different one.

rade:
This has been dragging on for far too long. Move the node change into a separate PR, leaving just the first two items in this PR. Review. Merge.

bowenli:
Closing this PR.

Text only change here: https://github.com/weaveworks/scope/pull/2006

PR #1953, opened 2016-10-25
Fix a JS bug where certain query strings would contain a && instead of &

ekimekim:
This was causing 400s when talking to overly-strict services

### DIFF
--- a/client/app/scripts/utils/web-api-utils.js
+++ b/client/app/scripts/utils/web-api-utils.js
@@ -27,7 +27,7 @@ let firstMessageOnWebsocketAt = 0;
 
 function buildOptionsQuery(options) {
   if (options) {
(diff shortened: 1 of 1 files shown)


davkal:
I cannot reproduce the issue, so I cannot verify if this in fact fixes the 400.

I've tried `/api/topology/services?t=5s&&&&namespace=default&pseudo=hide` on our prod cloud instance, and got a normal 200.

davkal:
Ah, I see what happened, this line messed it up:
`${wsUrl}${topologyUrl}/ws?t=${updateFrequency}&${optionsQuery}`
because it prepends a `&` to the query string that has `&` as prefix.

Closing this, in favor of #1956, which uses the immutable object function and adds a test.

PR #1956, opened 2016-10-27
Prevent querystrings from starting with &

davkal:
Fixes #1953


davkal:
> This was causing 400s when talking to overly-strict services

(from #1953)

PR #1958, opened 2016-10-28
middleware/errorhandler: Fix a bug which meant it never works

ekimekim:
Because methods operated on the value, not the reference, their changes
were discarded.


ekimekim:
oob LGTM from tom

Issue #1947, opened 2016-10-24
Upgrade client dev runtime to Node 6 LTS


jpellizzari:
Currently, running the webpack dev server and express proxy is eating up a ton of CPU cycles, even when the build isn't running. Upgrading to Node 6 appears to have fixed the problem. This change does break `jest` tests, however, because the `jest-cli` package uses an unsupported path argument:

https://github.com/facebook/jest/issues/1088

Upgrading`jest-cli` should fix this, but there were breaking changes which will need to be resolved. 

Also Node 6.9.1 just went into LTS, so it should be supported until Spring 2018.
https://nodejs.org/en/blog/release/v6.9.0/

jpellizzari:
Test builds:
`time docker build -t node4-build .` : `real	16m12.900s
user	0m0.103s
sys	0m0.103s`

PR #1959, opened 2016-10-28
Update client deps to use Node v6.9.0

jpellizzari:
From issue #1947. 
@davkal, @foot, @bowenli This PR upgrades some dependencies to work with Node 6 LTS. The major change is that `jest` auto-mocking is turned off by default now. Webpack dev server should be more efficient, and builds should be a little faster. I gathered some data on how long docker image builds are taking before/after:
`time docker build -t node4-build .` : `real	16m12.900s`& `real 8m39.124s`
`time docker build -t node6-build .`: `real	4m13.992s` <-- This PR

### DIFF
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:4.2.2
+FROM node:6.9.0
 WORKDIR /home/weave
 COPY package.json /home/weave/
 ENV NPM_CONFIG_LOGLEVEL=warn NPM_CONFIG_PROGRESS=false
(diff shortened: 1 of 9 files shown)


davkal:
Yup, great time improvements. LGTM

Also, I tried the node6 presets for babel, as opposed to the currently used ES2015 presets, but I'm getting syntax errors in webpack's uglify step.

jpellizzari:
I would imagine that's because Uglify wants ES5 and babel is only transforming stuff that Node 6 doesn't have. The only thing that will get transformed is the ES6 module syntax with the `node6` preset.

Issue #1325, opened 2016-04-18
Clicking outside keyboard shortcut 'window' should dismiss it.


tomwilkie:
I had expected clicking outside the keyboard shortcut help screen would dismiss it, but it doesn’t.

paulbellamy:
Should probably have an X in the corner as well.

jpellizzari:
Fixed by #1960

PR #1960, opened 2016-10-28
Close help dialog when the canvas is clicked

jpellizzari:
@davkal , @bowenli From #1325. Closes the 'Help' dialog when it is open and the background is clicked. I looked for other related dialog windows to see if there needed to be a dedicated action for this, but this was the only one I could find that didn't have this behavior already.


davkal:
LGTM

and tests! well done.

Issue #1964, opened 2016-10-28
Cannot copy/paste out of a scope terminal


paulbellamy:
Highlight goes away when releasing mouse-button.

jpellizzari:
@paulbellamy Was this working before? We recently changed the terminal library.

rade:
> @paulbellamy Was this working before?

Yes, as you can see for yourself by running the latest released version of Scope.

foot:
Ooops, seems to the focusing stuff, I'm remove it in #1966 and that should fix this.

foot:
Fixed in #1966

PR #1966, opened 2016-10-29
Resize TTYs

### LINKED ISSUE #746: Pipes should handle window resizing (and pop-out)
When a window resizes we should send docker `POST docker:2375/v1.21/containers/alpine/resize?h=35&w=80`, to trigger ncurses apps to redraw

Apps only redraw when the window size _actually_ changes (so the POST may have no effect). So, when we pop-out (or first open) we should send `POST docker:2375/v1.21/containers/alpine/resize?h=0&w=0`, then resize to the proper size.

There is an equivalent api endpoint for execs.

2opremio:
This PR adds the possibility to resize TTYs. It adds (optional) arguments to controls and modifies the UI<->App contract in the following way:
- Control requests now take optional arguments (a JSON string map)
- Pipe request add an additional key in the response: `resize_tty_control` which indicates what control to use in order to resize the TTY
- That control takes three (mandatory) parameters: `width`, `height` and `pipeID`

Here's an example:
1. The UI requests `POST http://172.16.0.3:4040/api/control/3b89ec4023a94f3e/516cc48f99788d7d6b558d4ebf3d0e3bd1d3f2eae65ceb4e405a53498ed99b45%3B%3Ccontainer%3E/docker_exec_container` to create a containter TTY
2. The App responds with `{"pipe":"pipe-4012265930510560261","raw_tty":true,"resize_tty_control":"docker_resize_exec_tty"}`. Note the new `resize_tty_control` field.
3. We can set the terminal size by doing `curl -X POST -d '{ "pipeID": "pipe-4012265930510560261", "height": "100", "width": "100"}' http://172.16.0.3:4040/api/control/3b89ec4023a94f3e/516cc48f99788d7d6b558d4ebf3d0e3bd1d3f2eae65ceb4e405a53498ed99b45%3B%3Ccontainer%3E/docker_resize_exec_tty`
4. The tty gets adjusted accordingly by the App:

screen shot 2016-10-29 at 13 25 20

TODO:
- [x] Implement (3) in the UI (CC @foot )
- [x] Implement resizing for host TTYs in the backend

I would appreciate a design review (maybe by @tomwilkie )

Fixes #746.
Fixs #1964

### DIFF
--- a/client/app/scripts/actions/app-actions.js
+++ b/client/app/scripts/actions/app-actions.js
@@ -580,16 +580,17 @@ export function receiveControlNodeRemoved(nodeId) {
   };
 }
 
-export function receiveControlPipeFromParams(pipeId, rawTty) {
+export function receiveControlPipeFromParams(pipeId, rawTty, resizeTtyControl) {
   // TODO add nodeId
   return {
     type: ActionTypes.RECEIVE_CONTROL_PIPE,
     pipeId,
-    rawTty
+    rawTty,
+    resizeTtyControl
   };
 }
 
-export function receiveControlPipe(pipeId, nodeId, rawTty) {
+export function receiveControlPipe(pipeId, nodeId, rawTty, resizeTtyControl) {
   return (dispatch, getState) => {
     const state = getState();
     if (state.get('nodeDetails').last()
@@ -608,7 +609,8 @@ export function receiveControlPipe(pipeId, nodeId, rawTty) {
       type: ActionTypes.RECEIVE_CONTROL_PIPE,
       nodeId,
       pipeId,
-      rawTty
+      rawTty,
+      resizeTtyControl
     });
 
     updateRoute(getState);
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -261,7 +261,7 @@ class Terminal extends React.Component {
     const cols = Math.floor(width / this.state.characterWidth);
     const rows = Math.floor(height / this.state.characterHeight);
 
-    doResizePipe(this.getPipeId(), this.props.control, cols, rows)
+    doResizePipe(this.getPipeId(), this.props.pipe.get('resizeTtyControl'), cols, rows)
       .then(() => this.setState({cols, rows}));
   }
 
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -8,7 +8,7 @@ import classNames from 'classnames';
 import { clickCloseTerminal } from '../actions/app-actions';
 import { getNeutralColor } from '../utils/color-utils';
 import { setDocumentTitle } from '../utils/title-utils';
-import { getPipeStatus, basePath } from '../utils/web-api-utils';
+import { getPipeStatus, basePath, doResizePipe } from '../utils/web-api-utils';
 import Term from 'xterm';
 
 const wsProto = location.protocol === 'https:' ? 'wss' : 'ws';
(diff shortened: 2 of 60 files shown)


foot:
Works great for containers!

Using the controls endpoint is a bit awkward right now: to construct the url `/api/control/:control.probeId}/:control.nodeId/docker_resize_exec_tty` we have to look up `probeId` and `nodeId` from some other control.

A pipe based endpoint feels like it would be cleaner: `/api/pipes/:pipeId/resize`.

If thats not feasible I guess either: 
 - `nodeDetails.resizeControl` property
 - as suggested somewhere else include a `docker_resize_exec_tty` control in nodeDetails.control w/ a `hidden: true`, a bit awkward.

2opremio:
> Works great for containers!

Shouldn't it be generic? (i.e. if the pipe response sets `raw_tty` and has a `resize_tty_control` field, always set the initial size and every time we resize the window)

> Using the controls endpoint is a bit awkward right now: to construct the url `/api/control/:control.probeId}/:control.nodeId/docker_resize_exec_tty` we have to look up `probeId` and `nodeId` from some other control.

Can't you save them with the pipeID and associate them to the terminal instance?

> A pipe based endpoint feels like it would be cleaner: `/api/pipes/:pipeId/resize`

That makes it awkward for the backend. I gave it a lot of thought and this was the best option I could come up with, respecting what we already have for controls (we cannot use the pipes directly because we don't have a way to send out of band control information cleanly without breaking backwards compatibility in the pipes protocol).

foot:
> Can't you save them with the pipeID and associate them to the terminal instance?

Yep, true, we can do that.

> if the pipe response sets raw_tty and has a resize_tty_control field

Ah hah! This is probably why host is not working right now.

Will clean up and fix this.

foot:
Working on hosts and popped out windows too now.

Can the resize_tty_control be missing sometimes?

2opremio:
> Can the resize_tty_control be missing sometimes?

It won't be missing in the response docker_exec/host_exec controls but it's optional, so I would handle it as such.

common/xfer/controls.go
```diff
@@ -52,6 +55,41 @@ func (c ControlHandlerFunc) Handle(req Request, res *Response) error {
...
+		height, err = strconv.ParseUint(heightS, 10, 32)
+		if err != nil {
+			return ResponseErrorf("Bad parameter: height (%q): %v", heightS, err)
+		}
+		width, err = strconv.ParseUint(widthS, 10, 32)
+		if err != nil {
+			return ResponseErrorf("Bad parameter: width (%q): %v", widthS, err)
+		}
+
+		return next(pipeID, uint(height), uint(width))
```
jml: Optional.

This is probably more of a subjective taste thing than necessary feedback, but I'd probably have made a data type that represents the resize args, and had a function that parses the map into said args (i.e. all of this without the call to `next`), and then used that parser here.

Would make the intent & behaviour more clear and easier to test, I think.

client/app/scripts/components/terminal.js
```diff
@@ -32,27 +32,29 @@ function ab2str(buf) {
...
+    characterHeight = parseInt(subjectRow.offsetHeight, 10);
+    subjectRow.innerHTML = contentBuffer;
   }
 
-  const pixelPerCol = width / cols;
-  const pixelPerRow = height / rows;
-
-  log('Caculated (col, row) sizes in px: ', pixelPerCol, pixelPerRow);
-  return {pixelPerCol, pixelPerRow};
+  log('Caculated (charWidth, charHeight) sizes in px: ', characterWidth, characterHeight);
```
davkal: Do we need this?

client/app/scripts/utils/web-api-utils.js
```diff
@@ -248,6 +252,22 @@ export function doControlRequest(nodeId, control, dispatch) {
   });
 }
 
+
+export function doResizePipe(pipeId, control, cols, rows) {
```
davkal: This should probably be called `doResizeTty` as we're not resizing the pipe.

client/app/scripts/utils/web-api-utils.js
```diff
@@ -232,7 +232,11 @@ export function doControlRequest(nodeId, control, dispatch) {
       if (res) {
         if (res.pipe) {
           dispatch(blurSearch());
-          dispatch(receiveControlPipe(res.pipe, nodeId, res.raw_tty, true));
+          dispatch(receiveControlPipe(
+            res.pipe,
+            nodeId,
+            res.raw_tty,
+            {id: res.resize_tty_control, probeId: control.probeId, nodeId: control.nodeId}));
```
davkal: Extracting that into a variable above would make the intention clearer.

client/app/scripts/actions/app-actions.js
```diff
@@ -580,16 +580,17 @@ export function receiveControlNodeRemoved(nodeId) {
   };
 }
 
-export function receiveControlPipeFromParams(pipeId, rawTty) {
+export function receiveControlPipeFromParams(pipeId, rawTty, resizeTtyControl) {
```
davkal: I'd try to avoid to have the signature become too TTY heavy. Pipes should be general. Maybe condense to a parameter field?

common/xfer/controls.go
2opremio: I initially coded it like that bit then I thought it was a bit too verbose (leading to this implementation).

jml: Fair enough!

client/app/scripts/components/terminal.js
foot: I like logs.

client/app/scripts/utils/web-api-utils.js
foot: 👍

foot:
So we don't receive a resize control from `attach`. Does it makes sense to have one there? @2opremio

2opremio:
> So we don't receive a resize control from attach. Does it makes sense to have one there? 

I don't think it does. Logs are not in raw mode, so it shouldn't matter (in general at least)

2opremio:
Docker provides https://godoc.org/github.com/fsouza/go-dockerclient#Client.ResizeContainerTTY so it should be easy to add, but I don't think it's even worth it. That could change if I am presented with a reasonable usecase.

Issue #1260, opened 2016-04-11
Short-lived connections not tracked for containers in shared networking namespaces


2opremio:
We currently can only track short-lived connections (using conntrack) for containers which have a unique IP.

This is because  the `/proc/` walker is not fast enough to track short-lived connections, for which we use conntrack. However, conntrack doesn't provide PIDs (just IPs and ports), which is not good enough to narrow down individual processes. Instead, we obtain the IPs from conntrack and match them against the networking interfaces of the containers, which must have a unique IP to be properly attributed. 

In other words, when there's a short-lived connection from/to an IP shared by multiple containers (which happens when they share the networking namespace) we don't know the specific container the connection is from/to. Instead of showing edges to/from all those containers (which would be inaccurate), we decide not to show any edges at all.

Although this is by design, I am tagging it as a bug because that's how an end-user would see it (the design explanation is too low-level).

A path forward would be to use ebpf to track connections, as proposed in #1168 .

2opremio:
See https://github.com/weaveworks/scope/issues/1258 to understand how this is confusing for the end user.

Issue #1168, opened 2016-03-15
walking /proc to obtain connections is very expensive


rade:
@tomwilkie suggests we should replace the existing logic with something eBPF based like https://github.com/iovisor/bcc/blob/master/tools/tcpaccept.py and https://github.com/iovisor/bcc/blob/master/tools/tcpconnect.py

Issue #1962, opened 2016-10-28
Port ebpf tracker to Go


2opremio:
We don't want to depend on bcc and Python to run the ebpf connection tracker.

We will need a wrapper library on the low level ebpf system calls and probably replicate some of the bcc features in Go (maybe even create a bcc Go backend)

PR #1967, opened 2016-10-31
Add eBPF connection tracking

alban:
Introduces connection tracking via eBPF. This allows scope to get
notified of every connection event, without relying on the parsing of
/proc/$pid/net/tcp{,6} and /proc/$pid/fd/*, and therefore improve
performance.

The eBPF program is in a python script using bcc: docker/tcpv4tracer.py.
It is contributed upstream via https://github.com/iovisor/bcc/pull/762
It is using kprobes on the following kernel functions:
- tcp_v4_connect
- inet_csk_accept
- tcp_close

connection tuple but also the pid and the netns.

The python script is piped into the Scope Probe and
conntrack, we keep the dead connections for one iteration in order to
report the short-lived connections.

state is now done in foreground instead of background, via
newForegroundReader().

Scope Probe also falls back on the the old /proc parsing if eBPF is not
working (e.g. too old kernel, or missing kernel headers). There is also
a flag "probe.ebpf.connections" that could disable eBPF if set to false.

NAT resolutions on connections from eBPF works in the same way as it did
on connections from /proc: by using conntrack.

The Scope Docker image is bigger because we need a few more packages
for bcc:
- weaveworks/scope in current master:  22 MB
- weaveworks/scope with this patch:   147 MB

Limitations:
- [ ] Does not support IPv6
- [ ] Sets `procspied: true` on connections coming from eBPF
- [ ] Size of the Docker images: 6 times bigger
- [ ] Requirement on kernel headers
- [ ] Location of kernel headers: https://github.com/iovisor/bcc/issues/743

Fixes https://github.com/weaveworks/scope/issues/1168 (walking /proc to
obtain connections is very expensive)

Fixes https://github.com/weaveworks/scope/issues/1260 (Short-lived
connections not tracked for containers in shared networking namespaces)

-----

This is the code from @asymmetric's ebpf branch after rebasing on master and squashing it.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,14 +1,12 @@
-FROM alpine:3.3
+FROM zlim/bcc
 MAINTAINER Weaveworks Inc <help@weave.works>
 LABEL works.weave.role=system
 WORKDIR /home/weave
-RUN echo "http://dl-cdn.alpinelinux.org/alpine/edge/community" >>/etc/apk/repositories && \
-	apk add --update bash runit conntrack-tools iproute2 util-linux curl && \
-	rm -rf /var/cache/apk/*
+RUN apt-get update -y && apt-get install -y bash runit conntrack iproute2 util-linux curl python bcc-tools python-bcc libbcc
 ADD ./docker.tgz /
 ADD ./demo.json /
 ADD ./weave /usr/bin/
-COPY ./scope ./runsvinit ./entrypoint.sh /home/weave/
+COPY ./scope ./runsvinit ./entrypoint.sh ./tcpv4tracer.py /home/weave/
 COPY ./run-app /etc/service/app/run
 COPY ./run-probe /etc/service/probe/run
 EXPOSE 4040
--- a/docker/tcpv4tracer.py
+++ b/docker/tcpv4tracer.py
@@ -1,0 +1,291 @@
+#!/usr/bin/python
+#
+# tcpv4tracer   Trace TCP IPv4 connections.
+#               For Linux, uses BCC, eBPF. Embedded C.
+#
+# USAGE: tcpv4tracer [-h] [-p PID]
+#
+from __future__ import print_function
+from bcc import BPF
+
+import argparse
+import ctypes
+
+parser = argparse.ArgumentParser(
+    description="Trace TCP IPv4 connections",
+    formatter_class=argparse.RawDescriptionHelpFormatter)
+parser.add_argument("-p", "--pid",
+    help="trace this PID only")
+args = parser.parse_args()
+
+# define BPF program
+bpf_text = """
+#include <uapi/linux/ptrace.h>
+#include <net/sock.h>
+#include <net/inet_sock.h>
+#include <net/net_namespace.h>
+#include <bcc/proto.h>
+
+#define TCP_EVENT_TYPE_CONNECT 1
+#define TCP_EVENT_TYPE_ACCEPT  2
+#define TCP_EVENT_TYPE_CLOSE   3
+
+struct tcp_event_t {
+        u32 type;
+        u32 netns;
+        u32 pid;
+        u32 saddr;
+        u32 daddr;
+        u16 sport;
+        u16 dport;
+};
+
+BPF_PERF_OUTPUT(tcp_event);
+BPF_HASH(connectsock, u64, struct sock *);
+BPF_HASH(closesock, u64, struct sock *);
+
+int kprobe__tcp_v4_connect(struct pt_regs *ctx, struct sock *sk)
+{
+        u64 pid = bpf_get_current_pid_tgid();
+
+        ##FILTER_PID##
+
+        // stash the sock ptr for lookup on return
+        connectsock.update(&pid, &sk);
+
+        return 0;
+};
+
+int kretprobe__tcp_v4_connect(struct pt_regs *ctx)
+{
+        int ret = PT_REGS_RC(ctx);
+        u64 pid = bpf_get_current_pid_tgid();
+
+        struct sock **skpp;
+        skpp = connectsock.lookup(&pid);
+        if (skpp == 0) {
+                return 0;       // missed entry
+        }
+
+        if (ret != 0) {
+                // failed to send SYNC packet, may not have populated
+                // socket __sk_common.{skc_rcv_saddr, ...}
+                connectsock.delete(&pid);
+                return 0;
+        }
+
+
+        // pull in details
+        struct sock *skp = *skpp;
+        struct ns_common *ns;
+        u32 saddr = 0, daddr = 0;
+        u16 sport = 0, dport = 0;
+        u32 net_ns_inum = 0;
+        bpf_probe_read(&sport, sizeof(sport), &((struct inet_sock *)skp)->inet_sport);
+        bpf_probe_read(&saddr, sizeof(saddr), &skp->__sk_common.skc_rcv_saddr);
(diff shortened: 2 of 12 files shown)


2opremio:
Fixes #1168 but unfortunately cannot be merged into master (at least as is) until https://github.com/weaveworks/scope/issues/1962 is addressed due to the big image increase (from 22MB to 147MB).

2opremio:
@alban Is the loopback connection tracking problem seen in the last demo fixed in this PR?

alban:
> @alban Is the loopback connection tracking problem seen in the last demo fixed in this PR?

No, it is not fixed. You directed me to #1780 but I didn't have time to look at it yet.

iaguis:
Limitation:

- [ ] For refused connections, we get a `connect` event without a matching `close` event causing the node to seem permanently connected

WIP fix in https://github.com/kinvolk/bcc/commits/iaguis/tcp_tracer_established

2opremio:
Are there any other limitations apart from refused connections and the loopback connections?

Can you document what testing has been done?

alban:
I could reproduce the issue with the loopback connection with the Scope master branch. In fact, I reproduce the issue both when connecting to the 127.0.0.1 address and connecting to the eth0 address.

I can see the two "nc" processes. When I look at /proc, I see they are connected, but Scope does not show it.

Reproducer:
```
apiVersion: v1
kind: Pod
metadata:
  name: connectlo
  labels:
    name: connectlo
spec:
  containers:
  - name: connectlo
    image: albanc/toolbox
    args:
      - '/bin/sh'
      - '-c'
      - "MYIP=$(ip addr show eth0 | grep \"inet\\b\" | awk '{print $2}' | cut -d/ -f1) ; echo MYIP=$MYIP ; (/bin/nc -l --keep-open -p 8081 & ) ; sleep 1 ; echo CLIENT ; (echo Hello ; sleep 1000 ; echo 'Bye' ) | /bin/nc $MYIP 8081"
```

So it does not seem to be a regression from this branch.

alban:
> Are there any other limitations apart from refused connections and the loopback connections?

The other limitations listed in the description of the PR, but after that I don't know other limitations.

> Can you document what testing has been done?

Works for me:

- :white_check_mark: get initial connections before Scope was started
- :white_check_mark: works well with NAT (tested with Kubernetes Services)
- :white_check_mark: displaying connection in the container view when both the source and the destination are pods with multiple containers

Not tested:

- :question: falling back to the old parsing method when ebpf does not work at all (e.g. old kernel)
- :question: falling back to the old parsing method when ebpf stops working (e.g. python script crashed - I have not seen that happening, but fallback would be nice)
- :question: connection between 2 containers using the same network namespace (It should work but I don't remember the result of that test)

Does not work:

- :x: IPv6

2opremio:
> I could reproduce the issue with the loopback connection with the Scope master branch

Ah, I see. Could you please create a separate issue for it?

alban:
> Ah, I see. Could you please create a separate issue for it?

Done in #1972

alban:
To compare the performance between the master branch and the ebpf branch, I used the following script [dialer.sh](https://gist.github.com/alban/186c9e0dec4977fa732f440a1bcacb4b). It reads the cpu time spent by the scope-probe process in kernel space and userspace from /proc. It does 2 measures: the first 10 seconds, and the next 10 seconds. I only consider the second part so that we don't count the initialization part. The script makes sure there is 1000 tcp connections established to a nginx container.

- master branch: 3.5 seconds
- golang-ebpf branch: 4.5 seconds

It is not as expected, unfortunately. To investigate what's going on, we ran the "go tool pprof". pprof does not work with Golang 1.7, so we need to revert to using Golang 1.6:
- git revert 002770d3949b52bcc0c0dba444e74090b97b53e8
- git revert 28213a00a463ca68bd32cf972ebe74280f427885
- git revert d1cf9f60fced05ccb9283d281c9405209bb4b0b0

Then I can edit dialer.sh to run Scope for longer and enable probe.http.listen with:
```
sudo VERSION=${VERSION} ./scope launch -probe.http.listen 0.0.0.0:4041
```

Then, we can run commands like:
```
go tool pprof -raw -seconds 30 http://localhost:4041/debug/pprof/heap
go tool pprof -raw -seconds 30 http://localhost:4041/debug/pprof/profile
go tool pprof ./docker/scope $HOME/pprof/pprof.localhost\:4041.samples...
```

See images
- pprof-master.svg
image

- pprof-ebpf.svg
image

As expected, we see that spyLoop takes the most of time in master but not anymore in golang-ebpf. But the regression is in the GC-related functions in golang-ebpf: runtime.gcDrain, runtime.bgsweep, runtime.scanobject take most of the time.

By looking at the code, I don't see much difference in the way objects are allocated and released between master and golang-ebpf. So I'm still investigating.

2opremio:
@alban It would help to get:

* Flame graphs
* An object allocation profile (with object allocation counts)
* An explanation of what `runtime.sysmon` is and in case of it being a symptom of contention, a blocking profile.

2opremio:
Side note: We should try to reduce the allocations (caching? pools?) in:

* `makeEndpointNode` 
* `ReadDirNames`  (@alban , is this still used in the new `/proc` Connector code?)

alban:
> ReadDirNames (@alban , is this still used in the new /proc Connector code?)

No, with the proc connector branch, ReadDirNames() is called only once at startup to get the initial state. Note that I am testing without the proc connector code, since it is independent from the ebpf branch.

alban:
## Flamegraphs

It does not seem possible to upload flamegraphs SVG on github, imgur or similar. And converting the flamegraphs to png lose information. I have the flamegraphs
- torch-master-perf.svg
- torch-golang-ebpf.svg

## object allocation profile

- pprof-master-alloc_objects.png
pprof-master-alloc_objects


- pprof-ebpf-alloc_objects.png
pprof-ebpf-alloc_objects


## Blocking profiles

- pprof-master-block.png
pprof-master-block

- pprof-ebpf-block.png
pprof-ebpf-block

alban:
On master, with my test of 1000 connections, the reporter gets 1000 connections from conntrack but only 66 connections from /proc (checked with @iaguis' patch https://github.com/kinvolk/scope/commit/ec98c486e5b152851377fcd3671a46a17cf69e99).

So the test with 1000 connections is not a good comparison between master and the ebpf branch because they don't handle the same amount of connections.

I tried to change the rate limiting parameters with [this patch](https://gist.github.com/alban/455aefa0c2bc989ac703273229685b51) on master. It becomes slower, but the reporters adds all the connections from /proc (instead of only 66), so it seems more correct.

- master with this change on the rate limits: 7490ms
- golang-ebpf: 6400ms
(both with Golang version reverted to 1.6)

Changing the timings in this way might not be what we want to test either since it makes master run the /proc parsing more often... @2opremio any idea of what would be a better test?

2opremio:
Why is it a bad comparison? IMO it's a good overall black-box comparison.

> they don't handle the same amount of connections

They do, it's 1000 in both. The internals shouldn't matter for absolute CPU metrics.

If I am following your reasoning correctly, it seems the problem is most likely that we are duplicating work with conntrack and the ebpf. Did you disable conntrack for everything except nat?

2opremio:
> Did you disable conntrack for everything except nat?

It doesn't seem you did:
https://github.com/kinvolk/scope/blob/alban/conn-perf-ebpf-pr-1/probe/endpoint/reporter.go#L68

alban:
This PR #1967 (ebpf using bcc/python) is superseded by #2024 (ebpf using bcc/gobpf) and #2070 (ebpf in Go without bcc).

PR #1970, opened 2016-11-01
[WIP] controls: plugins can add generic controls to start a container

alban:
This patch adds a way for plugins to add controls to start a new
container such as GDB for debugging. This does not require the plugin to
know the list of Docker containers, nor does it require to communicate
with the Docker daemon at all. The plumbing is done in the probe itself.
The plugin merely advertises the name of the GDB-like container to use.

This is achieved through two new fields in the report that the plugin
sends to the probe.

```
    "controls": {
      "ctrl-one": {
        "id": "ctrl-one",
        "human": "Ctrl One",
        "icon": "fa-futbol-o",
        "rank": 1,
        "always_propagated": true,
        "start_image": "albanc/toolbox"
      },
    }
```

Notice the two new optional fields:

- AlwaysPropagated (default: false): this asks the new control tagger
(probe/controls/tagger.go) to add the control on all nodes of the topology. In
this way, the plugin does not need to know the list of nodes and it does not
need to duplicate the work from the probe.

- StartImage (default: empty): if not empty, the probe will not forward the
command to the plugin but do the job itself instead. It creates a new Docker
container from the specified image and starts it. A window in the Scope UI is
immediately created and attached to the container.

The plugin developer can simply create an image that will start a
debugger such as GDB or strace. The container will be created with the
following environment variables: $SCOPE_PLUGIN_ID, $SCOPE_CONTROL_ID,
$SCOPE_IMAGE, $SCOPE_APP_ID, $SCOPE_NODE_ID. From $SCOPE_NODE_ID, the
image can determine which process to gdb or strace.

Previously, the controlID given by the plugin was converted before being
added in the report in the following form: `pluginID~controlID`. Now, if
the StartImage field is not empty, the controlID added in the report
will now have this form: `pluginID~controlID~encodedImage`. encodedImage
uses url.QueryEscape() to make sure that slashes in image names don't
bother Scope App.

-----

TODO & questions:

- [ ] This patch was partially tested with https://github.com/kinvolk/scope-debugger but at the moment, "start_image" does not point to an image that uses $SCOPE_NODE_ID, so the full chain has not been tested.
- [ ] I am not so happy with the way I patched the struct controlsByPlugin (probe/plugins/registry.go). I think it will need to be clarified.
- [ ] This does not compile anymore after a rebase on master
- [ ] The report data structure (report/controls.go) is changed by this patch and this is visible by Scope App. Even though the go json parser ignore unknown fields, I am not sure about the compatibility with an older Scope App. I am pondering about using a different data structure for the communication between the plugin and the Probe.

### DIFF
--- a/client/app/scripts/utils/web-api-utils.js
+++ b/client/app/scripts/utils/web-api-utils.js
@@ -223,7 +223,7 @@ export function getApiDetails(dispatch) {
 export function doControlRequest(nodeId, control, dispatch) {
   clearTimeout(controlErrorTimer);
   const url = `api/control/${encodeURIComponent(control.probeId)}/`
-    + `${encodeURIComponent(control.nodeId)}/${control.id}`;
+    + `${encodeURIComponent(control.nodeId)}/${encodeURIComponent(control.id)}`;
   reqwest({
     method: 'POST',
     url,
--- a/probe/controls/tagger.go
+++ b/probe/controls/tagger.go
@@ -1,0 +1,42 @@
+package controls
+
+import (
+	"time"
+
+	"github.com/weaveworks/scope/report"
+)
+
+// Tagger adds controls to nodes
+type Tagger struct {
+}
+
+// NewTagger tags each node with generic controls
+func NewTagger() Tagger {
+	return Tagger{}
+}
+
+// Name of this tagger, for metrics gathering
+func (Tagger) Name() string { return "Controls" }
+
+// Tag implements Tagger.
+func (t Tagger) Tag(r report.Report) (report.Report, error) {
+	for _, topology := range []report.Topology{r.Process, r.Container, r.ContainerImage, r.Host, r.Pod} {
+		// Each topology has most likely not many controls that need
+		// to be propagated. Gather the list in propagatingControls.
+		var propagatingControls []report.Control
+		for _, control := range topology.Controls {
+			if control.AlwaysPropagated {
+				propagatingControls = append(propagatingControls, control)
+			}
+		}
+
+		for _, node := range topology.Nodes {
+			for _, control := range propagatingControls {
+				metadata := map[string]string{"thisisatest": control.ID}
+				topology.AddNode(node.WithLatests(metadata))
+				topology.AddNode(node.WithLatestControl(control.ID, time.Now(), report.NodeControlData{Dead: false}))
+			}
+		}
+	}
+	return r, nil
+}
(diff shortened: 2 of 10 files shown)


2opremio:
Nice work. A few comments after reading the description (I haven't checked the code).

> `"always_propagated": true,`

I am not sure the name is descriptive enough, also, I am not sure if propagating it on the probes is the best way to go (we could simply do it in the app when rendering and save all that duplicated information in all the nodes).

Also, systematically adding the control without any filtering seems simplistic but it's a good start.

> "start_image": "albanc/toolbox"

Cool! Again, it's probably a bit simplistic but a good start. Restart policies? When should it be garbage collected? Will the used always be responsible of ending it (e.g. with `^D`).

alban:
> I am not sure the name is descriptive enough, also, I am not sure if propagating it on the probes is the best way to go (we could simply do it in the app when rendering and save all that duplicated information in all the nodes).

Right, that would save bandwidth. If Scope App is an old version without this and Scope Probe + plugins are using a new version, propagating in the app would not work. But maybe we don't care about not-updated app.

> Cool! Again, it's probably a bit simplistic but a good start. Restart policies? When should it be garbage collected? Will the used always be responsible of ending it (e.g. with ^D).

I wanted to let it to the user terminate the container (either with ^D or by closing the pseudo-terminal inside Scope App). So the container would only exist as long as it is visible in the app. My patch does not currently terminate correctly the container when closing the terminal, but that was the intention.

Do you have a use case with restart policy?

I have not thought about garbage collection. Scope could do the `docker rm` as soon as the container stops.

2opremio:
>  If Scope App is an old version without this and Scope Probe + plugins are using a new version, propagating in the app would not work. But maybe we don't care about not-updated app.

We normally suffer the opposite problem (probes being older than the app, we update Weave Cloud very regularly. So, yeah, we don't care too much about the opposite problem.

> Do you have a use case with restart policy?

Not at this point, and maybe that's a sign that we shouldn't make it too complicated at this point.

> I have not thought about garbage collection. Scope could do the docker rm as soon as the container stops.

I am fine with this, we should make sure that's the behavior before merging.

Also, would you mind rebasing?

schu:
> Also, would you mind rebasing?

We have done that now, thanks.

2opremio:
@alban @schu Do you plan to work on this anytime soon? Otherwise, can we close this PR?

PR #1971, opened 2016-11-02
middleware/errorhandler: Implement Hijacker so it works with ws proxy

ekimekim:
The lack of this was causing 500s when errorhandler-wrapped requests were hijacked, such as for websocket proxying

### DIFF
--- a/common/middleware/errorhandler.go
+++ b/common/middleware/errorhandler.go
@@ -83,7 +83,7 @@ func (i *errorInterceptor) Write(data []byte) (int, error) {
 }
 
 // errorInterceptor also implements net.Hijacker, to let the downstream Handler
-// hijack the connection. This is needed by the app-mapper's proxy.
+// hijack the connection. This is needed, for example, for working with websockets.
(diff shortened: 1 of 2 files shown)


rade:
Some details please. What is the user-visible impact of the problem this is fixing, and since when has it been present?

ekimekim:
No scope user impact - this middleware is only used by authfe in the service. In the service, this bug means that we can't use the admin scope instance, and has been in dev and prod since late last week. This only affects the admin endpoints, and scope is the only admin endpoint to use websockets.

common/middleware/errorhandler.go
```diff
@@ -78,3 +81,14 @@ func (i *errorInterceptor) Write(data []byte) (int, error) {
 	}
 	return len(data), nil
 }
+
+// errorInterceptor also implements net.Hijacker, to let the downstream Handler
+// hijack the connection. This is needed by the app-mapper's proxy.
```
2opremio: I would remove the `app-mapper` reference since it doesn't formally exist anymore (and it's unkown to OSS devs)

2opremio:
LGTM

PR #1973, opened 2016-11-03
Extend metadata in details panel for Weave Net nodes

### LINKED ISSUE #1938: Weave Net peers view
Users of Weave Net expect to see properties of that product in Scope, which has no special handling of it.

As a user I want...
- … to see Weave Net peers in Scope so I can verify they have started and are connected so I can debug network issues
- … to see Weave Net peers in Scope so I can verify whether peer connections are using Fast Datapath or encryption so I can debug network issues

Information similar to the `status` and `report` outputs from the `weave` script can presented in Scope.

2opremio:
Fixes #1938 

Here's how it looks:

screen shot 2016-11-03 at 16 04 29

There's still no connections table, that will come later. I don't see it critical for the kubecon deadline, but let me know if I am wrong.

### DIFF
--- a/common/weave/client.go
+++ b/common/weave/client.go
@@ -27,29 +27,47 @@ type Client interface {
 type Status struct {
 	Version string
 	Router  Router
-	DNS     DNS
-	IPAM    IPAM
+	DNS     *DNS
+	IPAM    *IPAM
 }
 
 // Router describes the status of the Weave Router
 type Router struct {
-	Name  string
-	Peers []struct {
+	Name               string
+	Encryption         bool
+	ProtocolMinVersion int
+	ProtocolMaxVersion int
+	PeerDiscovery      bool
+	Peers              []Peer
+	Connections        []struct {
+		Address  string
+		Outbound bool
+		State    string
+		Info     string
+	}
+	Targets        []string
+	TrustedSubnets []string
+}
+
+// Peer describes a peer in the weave network
+type Peer struct {
+	Name        string
+	NickName    string
+	Connections []struct {
 		Name        string
 		NickName    string
-		Connections []struct {
-			Name        string
-			NickName    string
-			Address     string
-			Outbound    bool
-			Established bool
-		}
+		Address     string
+		Outbound    bool
+		Established bool
 	}
 }
 
 // DNS describes the status of Weave DNS
 type DNS struct {
-	Entries []struct {
+	Domain   string
+	Upstream []string
+	TTL      uint32
+	Entries  []struct {
 		Hostname    string
 		ContainerID string
 		Tombstone   int64
@@ -58,7 +76,18 @@ type DNS struct {
 
 // IPAM describes the status of Weave IPAM
 type IPAM struct {
+	Paxos *struct {
+		Elector    bool
+		KnownNodes int
+		Quorum     uint
+	}
+	Range         string
 	DefaultSubnet string
+	Entries       []struct {
+		Size        uint32
+		IsKnownPeer bool
+	}
+	PendingAllocates []string
 }
 
 var weavePsMatch = regexp.MustCompile(`^([0-9a-f]{12}) ((?:[0-9a-f][0-9a-f]\:){5}(?:[0-9a-f][0-9a-f]))(.*)$`)
--- a/probe/overlay/weave.go
+++ b/probe/overlay/weave.go
@@ -1,36 +1,116 @@
 package overlay
 
 import (
+	"bytes"
+	"fmt"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/weaveworks/scope/common/backoff"
-	"github.com/weaveworks/scope/common/mtime"
 	"github.com/weaveworks/scope/common/weave"
 	"github.com/weaveworks/scope/probe/docker"
 	"github.com/weaveworks/scope/probe/host"
 	"github.com/weaveworks/scope/report"
+
+	docker_client "github.com/fsouza/go-dockerclient"
 )
 
+// Keys for use in Node
 const (
-	// WeavePeerName is the key for the peer name, typically a MAC address.
-	WeavePeerName = "weave_peer_name"
-
-	// WeavePeerNickName is the key for the peer nickname, typically a
-	// hostname.
-	WeavePeerNickName = "weave_peer_nick_name"
-
-	// WeaveDNSHostname is the ket for the WeaveDNS hostname
-	WeaveDNSHostname = "weave_dns_hostname"
-
-	// WeaveMACAddress is the key for the mac address of the container on the
-	// weave network, to be found in container node metadata
-	WeaveMACAddress = "weave_mac_address"
-
-	// WeaveVersion is the key for the weave version running on the peer
-	WeaveVersion = "weave_version"
+	WeavePeerName          = "weave_peer_name"
+	WeavePeerNickName      = "weave_peer_nick_name"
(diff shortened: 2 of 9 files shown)


rade:
> There's still no connection table...  let me know if I am wrong.

You are wrong.

2opremio:
> You are wrong.

OK, I will work on that next, but let's merge this first

rade:
I find the "#<num>" notation a bit odd. I don't think we are using that elsewhere.

probe/overlay/weave.go
```diff
@@ -1,61 +1,152 @@
...
+	WeaveDNSDomain         = "weave_dns_domain"
+	WeaveDNSUpstream       = "weave_dns_upstream"
+	WeaveDNSTTL            = "weave_dns_ttl"
+	WeaveDNSEntryCount     = "weave_dns_entry_count"
+	WeaveProxyTableID      = "weave_proxy_table"
+	WeaveProxyStatus       = "weave_proxy_status"
+	WeaveProxyAddress      = "weave_proxy_address"
+	WeavePluginTableID     = "weave_plugin_table"
+	WeavePluginStatus      = "weave_plugin_status"
+	WeavePluginDriver      = "weave_plugin_driver"
```
jml: lolgo

probe/overlay/weave.go
```diff
@@ -1,61 +1,152 @@
...
+		WeaveMACAddress:  {ID: WeaveMACAddress, Label: "Weave MAC", From: report.FromLatest, Priority: 17},
+		WeaveDNSHostname: {ID: WeaveDNSHostname, Label: "Weave DNS Name", From: report.FromLatest, Priority: 18},
+	}
 
-	// WeaveVersion is the key for the weave version running on the peer
-	WeaveVersion = "weave_version"
+	weaveMetadata = report.MetadataTemplates{
+		WeaveVersion:        {ID: WeaveVersion, Label: "Version", From: report.FromLatest, Priority: 1},
+		WeaveProtocol:       {ID: WeaveProtocol, Label: "Protocol", From: report.FromLatest, Priority: 2},
+		WeavePeerName:       {ID: WeavePeerName, Label: "Peer Name", From: report.FromLatest, Truncate: 17, Priority: 3},
```
jml: Why is this truncated? Why 17? (Please add a comment)

probe/overlay/weave.go
```diff
@@ -106,25 +207,98 @@ func (w *Weave) status() (bool, error) {
...
 }
 
+func filterContainerNotFound(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	switch err.(type) {
+	case *docker_client.Error:
+		// This is really ugly, but there's no way around it :(
```
jml: Maybe say why there's no way around it?

probe/overlay/weave.go
```diff
@@ -162,43 +336,120 @@ func (w *Weave) Report() (report.Report, error) {
...
-				report.MakeSets().Add(host.LocalNetworks, report.MakeStringSet(w.statusCache.IPAM.DefaultSubnet)),
-			),
+			report.MakeNode(report.MakeOverlayNodeID(report.WeaveOverlayPeerPrefix, w.statusCache.Router.Name)).
+				WithSet(host.LocalNetworks, report.MakeStringSet(w.statusCache.IPAM.DefaultSubnet)),
 		)
 	}
 	return r, nil
 }
+
+func (w *Weave) getPeerNode(peer weave.Peer) report.Node {
```
jml: Probably worth adding a doc comment for this.

prog/probe.go
```diff
@@ -40,7 +40,10 @@ const (
 	defaultServiceHost = "https://cloud.weave.works:443"
 )
 
-var pluginAPIVersion = "1"
+var (
+	pluginAPIVersion = "1"
+	dockerEndpoint   = "unix:///var/run/docker.sock"
```
jml: Shouldn't this get `DOCKER_HOST` or something, to handle docker-machine / minikube cases?

probe/overlay/weave.go
```diff
@@ -162,43 +336,120 @@ func (w *Weave) Report() (report.Report, error) {
...
+	node := report.MakeNode(report.MakeOverlayNodeID(report.WeaveOverlayPeerPrefix, peer.Name))
+	latests := map[string]string{
+		WeavePeerName:     peer.Name,
+		WeavePeerNickName: peer.NickName,
+	}
+	// Peer corresponding to current host
+	if peer.Name == w.statusCache.Router.Name {
+		latests[report.HostNodeID] = w.hostID
```
jml: I wonder if almost everything inside this conditional should be moved to yet another function? It's purpose isn't super clear to me, so giving it a name might help. Also, it would make the flow of `getPeerNode` more obvious to those who don't know what it's for.

probe/overlay/weave.go
2opremio: No real reason, I will remove the truncation.

probe/overlay/weave.go
2opremio: Sure

probe/overlay/weave.go
2opremio: Sure

2opremio:
> I find the "#" notation a bit odd. I don't think we are using that elsewhere.

We are (in the container count of the container image topology). I tried to be consistent here but I can happily change it.

prog/probe.go
2opremio: Maybe, not part in this PR though. I was nice enough to refactor it a bit :)

I have created https://github.com/weaveworks/scope/issues/1975

rade:
> We are (using the "#" notation) in the container count of the container image topology

The notation there is "# CONTAINERS: 2", not "CONTAINERS: #2".

2opremio:
> The notation there is "# CONTAINERS: 2", not "CONTAINERS: #2".

Ups, my bad, will fix

probe/overlay/weave.go
2opremio: I considered it during the process, but you need to pass all the locals (latests and node). Then decided it was not worth it. Now that it's grown quite a bit seems to make sense. Thanks

prog/probe.go
2opremio: > to handle docker-machine / minikube cases?

However, this won't be the usecase since scope also needs to live in the machine being monitored (it needs `/proc` and conntrack for instance)

2opremio:
Everything should be addressed now addressed (and fixed a bug on the way). @jml Thanks for the review, plese take another look.

Here's a screenshot after the fixes:

screen shot 2016-11-03 at 23 02 02

rade:
"DEF. SUBNET"? Why not spell it out? I recommend using *exactly* the same keys as in the `weave status` output, except perhaps for replacing "CompoundNoun" with "Compound Noun".

And why do the keys take up so much horizontal space, instead of leaving that to the values? Hence the unfortunate truncation of the proxy address.

common/weave/client.go
```diff
@@ -27,29 +27,47 @@ type Client interface {
...
+	Peers              []Peer
+	Connections        []struct {
+		Address  string
+		Outbound bool
+		State    string
+		Info     string
+	}
+	Targets        []string
+	TrustedSubnets []string
+}
```
2opremio: and since we couldn't vendor the whole data structure (that would imply vendoring `weave/prog/weaver/http.go` which is part of a program not a library) I decided to continue replicate the parts I needed.

2opremio: I will be more than happy to refactor this once we can vendor the full data structure.

PR #1976, opened 2016-11-03
Fixes the terminal spamming POST requests on resize

### LINKED ISSUE #1974: TTY resizing goes nuts when detaching terminal window
Tested on http://frontend.dev.weave.works/app/nameless-night-79

You can see a video showing the problem at http://recordit.co/BGPayC8clP

foot:
Forgot to debounce the initial implementation

Fixes #1974

### DIFF
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -4,6 +4,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { connect } from 'react-redux';
 import classNames from 'classnames';
+import _ from 'lodash';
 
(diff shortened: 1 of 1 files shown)


client/app/scripts/components/terminal.js
```diff
@@ -4,6 +4,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { connect } from 'react-redux';
 import classNames from 'classnames';
+import _ from 'lodash';
```
davkal: we should start using partial imports.

davkal:
LGTM

PR #1979, opened 2016-11-04
Allow resizing on attached pipes too.

### LINKED ISSUE #1977: kubernetes logs don't use full width

foot:
Fixes #1977

### DIFF
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -185,7 +185,7 @@ class Terminal extends React.Component {
     this.term = new Term({
       cols: this.state.cols,
       rows: this.state.rows,
-      convertEol: !this.props.raw,
+      convertEol: !this.props.pipe.get('raw'),
       cursorBlink: true,
       scrollback: 10000,
     });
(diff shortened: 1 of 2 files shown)


errordeveloper:
It works in general, i.e. the terminal uses full-width:



And resizing affects environment variables:



However, having resized to a very large size, and making it smaller again leads to shell prompt way below the last visible line. I'm not sure if this is something we may be able to fix, or it's a more general issue, I still don't fully grasp how exactly this stuff works.

errordeveloper:
> However, having resized to a very large size, and making it smaller again leads to shell prompt way below the last visible line. I'm not sure if this is something we may be able to fix, or it's a more general issue, I still don't fully grasp how exactly this stuff works.

Having said observed it my initial test, I am not able to reproduce it now... I think this may be a separate issue to do some kind of edge-case. I'll report if I see it again, this PR LGTM 👍

2opremio:
@errordeveloper The terminals (`>_`) are not relevant for this PR. Did you test the logs (attach)?

2opremio:
@foot It would be cool if the control icon could be placed on a top corner of the terminal window (to know what control triggered it)

davkal:
LGTM

errordeveloper:
I've retested with logs.

Here is first one, looks good.



However, making the window wider doesn't reflow the log lines.



But closing and re-opening works ok.



Making the window smaller and bigger again leave empty space...

2opremio:
Thanks @errordeveloper,  better late than never. Can you please create an issue with this?

Issue #1980, opened 2016-11-04
Add support for multicolumn tables from the probes


2opremio:
Right now we only support [two-column tables](https://github.com/weaveworks/scope/blob/375ce6b33c606dd967f005c81089085d85476c2d/report/table.go#L21).

As a workaround you can merge columns but it results in an ugly layout.

We should support multiple columns for both prefix-tables and fixed-tables.

We should also support setting headings, alignment and datatype metadata (for ordering and alignment).

2opremio:
Motivated by https://github.com/weaveworks/scope/pull/1981

fbarl:
The current table definition we use looks like this (taken from `report/table.go`):

```go
// Table is the type for a table in the UI.
type Table struct {
	ID              string        `json:"id"`
	Label           string        `json:"label"`
	Rows            []MetadataRow `json:"rows"`
	TruncationCount int           `json:"truncationCount,omitempty"`
}
```

where `MetadataRow` is essentially a key-value entry.

I see two ways we could add support for proper multi-column tables:

### 1. Extending current `Table` structure

```go
// Table is the type for a table in the UI.
type Table struct {
	ID              string   `json:"id"`
	Label           string   `json:"label"`
	Columns         []Column `json:"columns"`
	Rows            []Row    `json:"rows"`
	TruncationCount int      `json:"truncationCount,omitempty"`
}
```

To use this structure the *old way*, we would convert all the `MetadataRow` entries to 2-column `Row`s and simply omit the `Columns` field.

**good**: We could control the order of tables rendered without adding extra logic for that on frontend.

**bad**: `Table` and `Labels` (a list of key-value pairs) are not semantically that similar so it might be bad to use the same structure to describe them.

### 2. Adding a separate `Table` structure

In that case, it would make sense to rename current `Table` to something more accurate (maybe `Labels`):

```go
type Labels struct {
	ID              string        `json:"id"`
	Label           string        `json:"label"`
	Rows            []MetadataRow `json:"rows"`
	TruncationCount int           `json:"truncationCount,omitempty"`
}

type Table struct {
	ID              string   `json:"id"`
	Columns         []Column `json:"columns"`
	Rows            []Row    `json:"rows"`
	TruncationCount int      `json:"truncationCount,omitempty"`
}
```

**good**: Structures' names better describe what they are and there are no optional/unused attributes.

[...]

PR #1981, opened 2016-11-04
Add connections table to Weave Net details panel

### LINKED ISSUE #1938: Weave Net peers view

2opremio:
Fixes #1938 

Here's how it looks:

screen shot 2016-11-04 at 10 40 01

The two-column layout and alignment is ... ugly to say the least. I have created #1980 to fix that later on.

### DIFF
--- a/probe/overlay/weave.go
+++ b/probe/overlay/weave.go
@@ -19,33 +19,34 @@ import (
 
 // Keys for use in Node
 const (
-	WeavePeerName          = "weave_peer_name"
-	WeavePeerNickName      = "weave_peer_nick_name"
-	WeaveDNSHostname       = "weave_dns_hostname"
-	WeaveMACAddress        = "weave_mac_address"
-	WeaveVersion           = "weave_version"
-	WeaveEncryption        = "weave_encryption"
-	WeaveProtocol          = "weave_protocol"
-	WeavePeerDiscovery     = "weave_peer_discovery"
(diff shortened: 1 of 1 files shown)


probe/overlay/weave.go
```diff
@@ -428,11 +434,32 @@ func (w *Weave) addCurrentPeerInfo(latests map[string]string, node report.Node)
...
+		inboundArrow  = "<-"
+	)
+	table := make(map[string]string, len(router.Connections))
+	for _, conn := range router.Connections {
+		arrow := inboundArrow
+		if conn.Outbound {
+			arrow = outboundArrow
+		}
+		// TODO: we should probably use a multicolumn table for this
+		//       but there is no mechanism to support it yet.
```
tomwilkie: To do this you'll need to model the connections as children

tomwilkie:
LGTM

probe/overlay/weave.go
2opremio: How do you mean exactly? I considered using EdgeMetadata and render them in the app but then discarded it.

Issue #1986, opened 2016-11-04
Incorrect height of terminal window in Safari


pidster:
When launching the terminal control on a SockShop container in Safari (Version 10.0.1 (12602.2.14.0.7)), the height of the internal window is very small - barely 1 character high.

Chrome works OK. See screenshot below.

<img width="1237" alt="safari-scope" src="https://cloud.githubusercontent.com/assets/303151/20015695/557dd270-a2b4-11e6-98a0-ea2780101d63.png">

2opremio:
As a reminder please create a PR on top of branch `release-1.0` if you manage to fix this today.

jpellizzari:
Root cause of this appears to be Safari measuring `innerNode.clientHeight` differently from Chrome:
https://github.com/weaveworks/scope/blob/master/client/app/scripts/components/terminal.js#L274

foot:
Fixed in #1988

PR #1988, opened 2016-11-04
Fix Safari terminal window height bug

jpellizzari:
Fix for #1986 
Fixes the issue where the terminal in Safari isn't the right size.
Before:
screen shot 2016-11-04 at 1 09 57 pm
After:
screen shot 2016-11-04 at 1 10 05 pm


jpellizzari:
@bowenli could you review?

jpellizzari:
@foot Could you take a look at this?

foot:
Its kind of a fun effect where it slides out compact and then grows vertically.

Nice catch on making the terminal padding "border" black to match the term.

foot:
I changed the solution a bit and instead fixed up the CSS layout so that height was propagated down the DOM tree in Safari.

All the nesting for the Terminal here is a bit icky, mix of absolute+top0+bottom0 and height 100%s but works for now.

Issue #1721, opened 2016-07-26
automatically switch to table on high node/connection count


rade:
Until we have a dedicated high-node-count mode (#1614), the table mode is a pretty good substitute...

scope with a thousand connected processes created with

```
nc -lk 1122 &
for i in $(seq 1 1000) ; do nc localhost 1122 & done
```

topology visualisation - this takes a while to render and is rather unusable:
![screenshot_2016-07-26_08-20-25](https://cloud.githubusercontent.com/assets/109109/17129307/2a75ac0a-530a-11e6-899a-854f3dd3cb27.png)
topology table (from #1673) - this renders quite quickly and is perfectly usable:
![screenshot_2016-07-26_08-21-04](https://cloud.githubusercontent.com/assets/109109/17129335/5464564c-530a-11e6-94db-58a07e774dc6.png)

I therefore propose that we automatically switch to the table for high node/connection count. As noted in  https://github.com/weaveworks/scope/issues/1614#issuecomment-230343329, this should be done on a hysteresis.

rade:
So I think the behaviour we are looking for here is:

Whenever a UI action takes you to a visualisation that exceeds the complexity threshold, we switch to table mode instead. Except when that UI action was "switch to visualisation mode" - this is mainly because our heuristics for what is "too complex" aren't perfect; we don't want to prevent users from seeing the visualisation when it does in fact work.

I don't think we need to automatically switch from the visualisation mode to table mode when complexity increases due to extra nodes and edges being added. Hence we don't need a hysteresis either.

So the automatic switch to table mode is *always* triggered by a UI action. Such as selecting a different view or changing filters.

That leaves the question what the threshold should be. Something like node_count + 2*edge_count < 500? Or should there be some ^2 in there somewhere? We probably don't want to get too clever here, at least to begin with.

jpellizzari:
@davkal and @rade, I would like to give this issue a try. The way I am understanding this, we would switch the to the table view if the graph view exceeds a complexity threshold when the user does an action.
 
My first inclination is to figure out which of the [action-types](https://github.com/weaveworks/scope/blob/master/client/app/scripts/constants/action-types.js) would need this behavior, then apply the "threshold" logic to each (using some sort of white-list so its easy enable/disable this behavior for individual actions).

@davkal can you think of a more clever way of doing this than listing the individual actions manually? Should I look to do something further downstream in a component instead?

jpellizzari:
[...]

PR #1994, opened 2016-11-07
Add graph complexity check on page load

jpellizzari:
Fix for #1721 .

When the page is loaded, the first received topology will trigger a check to see if the graph will be too complex to render. If so, the view will switch to the table view.  The complexity check formula:
`(currentTopology.stats.node_count * (2 * currentTopology.stats.edge_count)) > 500`

The user can switch to the graph view at any time. This logic only applies at page load. To test with a high node count (on linux):
`nc -lk 1122 & for i in $(seq 1 300) ; do nc localhost 1122 & done`

@davkal Let me know what you think.

### DIFF
--- a/client/app/scripts/reducers/__tests__/root-test.js
+++ b/client/app/scripts/reducers/__tests__/root-test.js
@@ -1,5 +1,4 @@
-const is = require('immutable').is;
-
+import {is, fromJS} from 'immutable';
 // Root reducer test suite using Jasmine matchers
 
 describe('RootReducer', () => {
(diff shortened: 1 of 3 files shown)


client/app/scripts/components/__tests__/node-details-test.js
```diff
@@ -34,7 +34,8 @@ describe('NodeDetails', () => {
     details = {label: 'Node 1'};
     const c = TestUtils.renderIntoDocument(
       <Provider store={configureStore()}>
-        <NodeDetails nodes={nodes}
+        <NodeDetails topologyId="containers" nodes={nodes}
```
davkal: Could you revert that hunk? It's been fixed elsewhere.

jpellizzari: Fixed

client/app/scripts/constants/naming.js
```diff
@@ -1,2 +1,3 @@
 
 export const EDGE_ID_SEPARATOR = '-';
+export const GRAPH_COMPLEXITY_THRESH = 500;
```
davkal: Hmm, is this really a naming constant?

davkal:
Code looks good.

One more thing: if Scope has been used before, the view mode may be stored in localStorage (see router-utils.js and storage-utils.js). Should we then take those values, or still check for the graph complexity?

jpellizzari:
@davkal Since the graph could have gotten more complex from one browser session to the next, I think we should switch to the table mode no matter what the URL or localStorage says.

client/app/scripts/constants/naming.js
jpellizzari: Changed to move this threshold into the helper, since it will probably only every be used there.

davkal:
LGTM

Issue #1997, opened 2016-11-08
Clicking the canvas does not always close the details panel


jpellizzari:
If you do not click perfectly (ie your mouse is not moving at all) on the canvas, the details panel does not go away. Some times it takes a couple clicks in order for the details panel to go away.

jpellizzari:
The root cause was the `isZooming` flag. D3 registers drag events as zoom events and was setting the flag if the mouse click wasn't perfect.

PR #1998, opened 2016-11-08
Fix details panel not closing on canvas click

jpellizzari:
Fix for #1997.

@foot Could you review please?


davkal:
Hmm, I dont really have that problem on my track pad, but if you experienced this frequently, this fix does the trick. LGTM

PR #2003, opened 2016-11-10
plug goroutine leak in control

### LINKED ISSUE #2002: 'control' is leaking goroutines
_From @rade on November 9, 2016 18:59_

observed in 'prod':
screenshot from 2016-11-09 18-45-11

The 'control' containers are leaking go-routines at a steady rate; resulting in ~17k go-routines per container at the daily deploy.

Memory and CPU usage of these containers is low, FD usage is in the low hundreds and stable. AFAICT there is no user visible impact.

In dev the goroutine count is ~150 x 2 instances, after 18 hours of running. It tends to hover around one level for long periods and then increase before hovering again.

_Copied from original issue: weaveworks/service#1001_

rade:
Fixes #2002.

### DIFF
--- a/app/multitenant/sqs_control_router.go
+++ b/app/multitenant/sqs_control_router.go
@@ -315,6 +315,7 @@ func (pw *probeWorker) stop() {
 }
 
 func (pw *probeWorker) loop() {
+	defer pw.done.Done()
 	for {
 		// have we been stopped?
 		select {


rade:
I re-ran the test from https://github.com/weaveworks/scope/issues/2002#issuecomment-259690147, then deployed this change and re-ran:

screenshot_2016-11-10_19-37-09

The new instances show no increase in their go-routine count.

PR #2006, opened 2016-11-10
increase black/white text contrast in contrast mode

bowenli:
Splitting out the text portion, re conversation here https://github.com/weaveworks/scope/pull/1952


davkal:
LGTM

PR #2007, opened 2016-11-11
Table-mode: sort ips numerically

fbarl:


client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -50,6 +51,16 @@ const COLUMN_WIDTHS = {
 };
 
 
+// TODO: Consider introducing the 'ip' dataType for
+// the fields instead of maintaining a list here.
+const IP_FIELDS = ['kubernetes_ip'];
```
fbarl: I found the place(s) where I need to modify it on the backend, but I can't really test it manually right now, as I'd need to run the updated backend code and generate a new report that I'd need to plug into the frontend client that I'm running locally. I'm sure there is a better way to dynamically observe changes on the backend, but I don't know yet how to run the backend service in the development mode - maybe you guys could show me that on Monday?

So, I'll just stash those changes for now and play with some other stuff the rest of the day :)

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -74,13 +80,22 @@ function maybeToLower(value) {
 }
 
 
+function toArrayOfThreeDigitStrings(value) {
+  const padToThreeDigits = (n) => `000${n}`.slice(-3);
```
davkal: Better to have that outside the function?

client/app/scripts/components/node-details/__tests__/node-details-table-test.js
```diff
@@ -0,0 +1,106 @@
...
+          { id: 'kubernetes_namespace', label: 'Namespace', value: '10.244.253.100' },
+        ]
+      },
+    ];
+  });
+
+  function matchColumnValues(columnLabel, expectedValues) {
+    const columnIndex = columns.findIndex(column => column.id === columnLabel);
+    const values = TestUtils.scryRenderedDOMComponentsWithClass(component, 'node-details-table-node-value').map(d => d.title);
+    const filteredValues = values.filter((element, index) => index % columns.length === columnIndex);
```
davkal: What's happening here?

client/app/scripts/components/node-details/__tests__/node-details-table-test.js
```diff
@@ -0,0 +1,106 @@
...
+    expect(filteredValues).toEqual(expectedValues);
+  }
+
+  describe('Sorts by column', () => {
+    describe('kubernetes_ip', () => {
+      it('sorts ascendingly', () => {
+        component = TestUtils.renderIntoDocument(
+          <Provider store={configureStore()}>
```
davkal: You should be able to export NodeDetailsTable in its unconnected form, so that you dont need the redux provider.

client/app/scripts/components/node-details/__tests__/node-details-table-test.js
```diff
@@ -0,0 +1,106 @@
...
+  beforeEach(() => {
+    columns = [
+      { id: 'kubernetes_ip', label: 'IP' },
+      { id: 'kubernetes_namespace', label: 'Namespace' },
+    ];
+    nodes = [
+      {
+        id: 'node-1',
+        metadata: [
+          { id: 'kubernetes_ip', label: 'IP', value: '10.244.253.24' },
```
davkal: How can those work if the `datatype` is never set?

client/app/scripts/components/node-details/__tests__/node-details-table-test.js
fbarl: Do you know how exactly I could do it? If I just remove the `Provider` tag, the test fails with the message:

```
Invariant Violation: Could not find "store" in either the context or props of "Connect(NodeDetailsTableNodeLink)". Either wrap the root component in a <Provider>, or explicitly pass "store" as a prop to "Connect(NodeDetailsTableNodeLink)".
```

client/app/scripts/components/node-details/__tests__/node-details-table-test.js
fbarl: Good point, forgot to run the tests once more in the end! :) I modified `line 14`, so the tests should pass now (they weren't passing before).

client/app/scripts/components/node-details/__tests__/node-details-table-test.js
fbarl: I added the comments. It is a bit hacky, but I didn't think it's a big problem in tests? It would be simpler if columns had different class tags I guess..

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -74,13 +80,22 @@ function maybeToLower(value) {
 }
 
 
+function toArrayOfThreeDigitStrings(value) {
```
fbarl: Hm.. I really couldn't think of a more fitting name here, do you have any suggestion? :D Test and location are good points, I'm gonna address them now!

client/app/scripts/components/node-details/node-details-table.js
fbarl: Sure

client/app/scripts/components/node-details/__tests__/node-details-table-test.js
foot: in node-details-panel there's an `export default connect(...`

If you add an export to the main class too:

`export class NodeDetailsPanel...`

Then you can import that w/out the connect() stuff:

`import { NodeDetailsPanel }  from ....`

client/app/scripts/utils/__tests__/string-utils-test.js
```diff
@@ -20,4 +20,13 @@ describe('StringUtils', () => {
       expect(fun(['prefix', 'suffix'])).toBe('');
     });
   });
+
+  describe('iPtoPaddedString', () => {
+    const fun = StringUtils.iPtoPaddedString;
```
foot: `fun` sounds a bit... `fun`ny. I would prefer `fn` or `f` as a shortenting

foot:
Nicely done! Keeping the `ip` as a padded string rather than converting it into a number works really well and still allows for comparisons w/ domains etc too. 👍 

I'm not sure if we really need to render the components to test the sorting though.. you could `export function getSortedNodes` in node-details-table and test that instead (i think?).

foot:
Ooops! The rebase looks like its gone a bit funny (commits from master now in this branch),

You wanna do something like:
```
git fetch
git rebase origin/master
git push -f
```

client/app/scripts/utils/__tests__/string-utils-test.js
fbarl: Noted! I also modified the other tests in this file for consistency.

client/app/scripts/utils/string-utils.js
```diff
@@ -75,6 +80,11 @@ export function longestCommonPrefix(strArr) {
   return (new LCP(strArr)).lcp();
 }
 
+// Converts IPs from '10.244.253.4' to '010.244.253.004' format.
+export function iPtoPaddedString(value) {
```
fbarl: Yup, makes sense! :)

fbarl:
@davkal PTAL before I merge (I will squash all the commits into one as Simon suggested).

davkal:
LGTM

PR #2010, opened 2016-11-11
Fix kubectl port-forward command to access the scope app locally

sdurrheimer:
HI there,

I was testing weave scope on kubernetes and discovered that the `kubectl port-forward` command in the doc was not corresponding to current labels.

weave-scope-labels

Signed-off-by Steve Durrheimer <s.durrheimer@gmail.com>


2opremio:
Thanks!

@errordeveloper Please take this problems into consideration when updating the launch generator.

Issue #1999, opened 2016-11-09
Fix source-maps/debugging in browser


jpellizzari:
Currently, setting a breakpoint/debugger statement jumps to some weird index.js page. Could be a webpack issue, could be a chrome issue. 
https://github.com/webpack/webpack/issues/2145

PR #2011, opened 2016-11-13
Change webpack local config to use source maps

jpellizzari:
Fix for #1999.

Updated local webpack config to use source maps, according to this doc:
http://webpack.github.io/docs/configuration.html#devtool

Re-builds might be slightly slower, but breakpoints should work properly now. Also fixed webpack config file linting errors, which caused in-editor linters to freak out.

@foot Could you take a look at this?


foot:
Nice one, lgtm.

PR #2014, opened 2016-11-15
Combined external and prod webpack config files

### LINKED ISSUE #1941: Consolidate webpack configs after external config was introduced
95% of webpack production and external config is shared. So one could include the other.

jpellizzari:
Fixes #1941.

Combines `webpack.external.config` and `webpack.production.conifg` and uses an env variable to trigger external builds.


jpellizzari:
@davkal Please review. Also, do you think there is any value in combining the files further to include "local"?

PR #2017, opened 2016-11-18
Update package.json

fbarl:
The only two things I didn't update are:

1. **d3** to version 4, because they introduced [a lot of changes](https://github.com/d3/d3/blob/master/CHANGES.md) - I think this update should be done in a separate PR where we also make sure all the elements are displayed correctly
2. **eslint** together with **eslint-config-airbnb**, which are also pretty much behind - I think we should do this in a separate PR as well; I tried doing it now but I got a lot of lint warnings, so I assume either our lint configuration would need to be revised or we would need to do a bunch of small fixes across a lot of files

I used http://npm.click/ for libraries' version comparison.


fbarl:
@foot: Maybe you can take a look at this since @davkal is not here today, so that I can merge it soon :)

foot:
- Hot reloading is not working (would be realllly nice to keep this working).

But most of the UI seems to be working fine.

davkal:
> Hot reloading is not working (would be realllly nice to keep this working).

I'd say that's a hard requirement.
Maybe it would work more smoothly if we used the webpack-middleware in the express server as opposed to starting our own webpack-dev-server.

fbarl:
@davkal PTAL: I think I managed to make it work without `webpack-middleware`, but I had to weaken the cross-origin resource policy explicitly in the dev-mode, which I'm not sure is ok:

```javascript
  new WebpackDevServer(webpack(config), {
    hot: true,
    noInfo: true,
    historyApiFallback: true,
    stats: 'errors-only',
    // We need the access from the express server for the hot-loader.
    // See: https://github.com/gaearon/react-hot-loader/issues/56
    headers: { "Access-Control-Allow-Origin": '*' },
  }).listen(4041, '0.0.0.0', function (err, result) {
```

Without this `headers` line, the problem would occur when hot loader would try to compile javascript with syntax error - it would first say it failed to access `localhost:4041` from `localhost:4042` and from there hot loading would basically get broken until you'd refresh the page.

fbarl:
@foot FYI If you think changing the cross-origin policy as I did is ok, I'd like to merge this with master and then try switching to `webpack-middleware` in a separate story as @davkal suggested.

foot:
Yeah the cross origin stuff looks fine.

1. Breakpoints don't work after a hot-reload (not sure if fixable, and kind of work-around-able: ctrl-r).
2. its very slow to hot-reload too (10s), though I think this was #1999 
   - moving to cheap source maps brings it down to 3-4s for me, but you lose breakpoints completely.
   - Is it slow for you too?

It would be good to fix both the above but we could move them into new issues at this stage.

PR #2019, opened 2016-11-18
Use intuitive standalone service name in compose

2opremio:
Context: https://weaveworks.slack.com/archives/weave-users/p1479477446003653


pidster:
LGTM

PR #2022, opened 2016-11-21
Improve plugin errors tooltip

### LINKED ISSUE #1905: Plugin errors are difficult to view
I keep expecting to click the `(!)` and see the error, but you have to hover for the tooltip.

fbarl:
Fixes #1905 - Plugin errors are difficult to view

### DIFF
--- a/client/app/styles/main.less
+++ b/client/app/styles/main.less
@@ -187,11 +187,12 @@ h2 {
 }
 
 .footer {
+  padding: 5px;
   position: absolute;
-  bottom: 16px;
-  right: 48px;
-  z-index: 20;
+  bottom: 11px;
+  right: 43px;
   color: @text-tertiary-color;
+  background-color: fade(@background-average-color, 90%);
   font-size: 0.7rem;
   display: flex;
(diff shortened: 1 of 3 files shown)


rade:
We tend to have more descriptive commit messages, rather than just referencing an issue number.

fbarl:
Triggered both on _click_ and on _hover_:

image1
image2

I also added some opacity to the footer background, so that the it's readable when it floats above the nodes texts.

foot:
Nice one.

- I'd use `effect="solid"` so it doesn't chase the cursor around.

Please change the commit message too!

client/app/styles/main.less
```diff
@@ -187,11 +188,14 @@ h2 {
 }
 
 .footer {
+  padding: @footer-padding;
+  margin: -@footer-padding;
```
foot: Maybe just adjust the `bottom` and `right` properties below instead of using negative margins. A little easier to read/comprehend.

foot:
Other than that LGTM

PR #2024, opened 2016-11-22
Add eBPF connection tracking with gobpf

### LINKED ISSUE #1168: walking /proc to obtain connections is very expensive

### LINKED ISSUE #1260: Short-lived connections not tracked for containers in shared networking namespaces

alban:
This is the second attempt to add eBPF connection tracking. The first one was via https://github.com/weaveworks/scope/pull/1967 by forking a python script using bcc. This one is done in Golang directly thanks to [gobpf](https://github.com/iovisor/gobpf).

This is not enabled by default. For now, it should be enabled manually with:
```
```
Scope Probe also falls back on the the old /proc parsing if eBPF is not working (e.g. too old kernel, or missing kernel headers).

This allows scope to get notified of every connection event, without relying on the parsing of `/proc/$pid/net/tcp{,6}` and `/proc/$pid/fd/*`, and therefore improve performance.

The eBPF program is in `probe/endpoint/ebpf.go`. It was discussed in bcc via iovisor/bcc#762.
It is using kprobes on the following kernel functions:
- tcp_v4_connect
- inet_csk_accept
- tcp_close

It generates "connect", "accept" and "close" events containing the connection tuple but also the pid and the netns.

`probe/endpoint/ebpf.go` maintains the list of connections. Similarly to conntrack, we keep the dead connections for one iteration in order to report the short-lived connections.

The code for parsing `/proc/$pid/net/tcp{,6}` and `/proc/$pid/fd/*` is still there and still used at start-up because eBPF only brings us the events and not the initial state. However, the /proc parsing for the initial state is now done in foreground instead of background, via `newForegroundReader()`.

NAT resolutions on connections from eBPF works in the same way as it did on connections from /proc: by using conntrack. One of the two conntrack instances was removed since eBPF is able to get short-lived connections.

The Scope Docker image is bigger because we need a few more packages
for bcc:
- weaveworks/scope in current master:  22 MB (compressed),  71 MB (uncompressed)
- weaveworks/scope with this patchset: 83 MB (compressed), 223 MB (uncompressed)

But @iaguis still has ongoing work to reduce the size of the image.

Limitations:
- [ ] Sets `procspied: true` on connections coming from eBPF
- [ ] Size of the Docker images
- [ ] Requirement on kernel headers for now
- [ ] Location of kernel headers: iovisor/bcc#743
- [ ] Does not support IPv6
- [ ] Uncompleted connections that time out don't disappear (WIP branch to fix this and support IPv6: [iaguis/tcp_tracer_established](https://github.com/kinvolk/bcc/commits/iaguis/tcp_tracer_established))

Fixes #1260 (Short-lived connections not tracked for containers in shared networking namespaces)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ GO_HOST=$(NO_CROSS_COMP); $(GO)
 WITH_GO_HOST_ENV=$(NO_CROSS_COMP); $(GO_ENV)
 GO_BUILD_INSTALL_DEPS=-i
 GO_BUILD_TAGS='netgo unsafe'
-GO_BUILD_FLAGS=$(GO_BUILD_INSTALL_DEPS) -ldflags "-extldflags \"-static\" -X main.version=$(SCOPE_VERSION) -s -w" -tags $(GO_BUILD_TAGS)
+GO_BUILD_FLAGS=$(GO_BUILD_INSTALL_DEPS) -ldflags "-X main.version=$(SCOPE_VERSION) -s -w" -tags $(GO_BUILD_TAGS)
 IMAGE_TAG=$(shell ./tools/image-tag)
 
 all: $(SCOPE_EXPORT)
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,6 +1,9 @@
-FROM golang:1.7.1
+FROM ubuntu:yakkety
+ENV GOPATH /go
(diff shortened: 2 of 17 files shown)


alban:
This PR (ebpf using bcc/gobpf) is replaced by #2070 (ebpf in Go without bcc, to avoid run-time dependency on kernel headers).

PR #1391, opened 2016-04-27
Add kubernetes checkpoint flag

2opremio:
Fixes #1380

PR #2026, opened 2016-11-23
Add ECS views

ekimekim:
This PR adds two new ECS-specific topologes, ECSTasks and ECSServices,
and adds a probe which collects info to populate it.

The probe uses the instance's IAM role to look up task and service info. It requires the following permissions:
ListServices
DescribeTasks
DescribeServices

@2opremio: For rendering: I've associated a container with 0 or 1 ECSTasks and ECSServices, under the Node.Parents Sets.
ie. if a container X is associated with task Y and service Z, then rpt.Container.Nodes[X].Parents contains:
```json
{
	"ecstask": [
		"Y;ecstask"
	],
	"ecsservice": [
		"Z;ecsservice"
	]
}
```
and so any endpoints associated with the container should be assoicated with the given task and service,
so that communication between containers is visible on the tasks view, etc.

TODO:
- [x] ~Unit tests~ (moved to #2026 )
- [x] Test on Socks shop: https://microservices-demo.github.io/microservices-demo/deployment/ecs.html
- [x] Look for performance regressions (in the socks shop for instance)
- [x] ~Document required permissions~ (No need, created #2039 , see  https://github.com/weaveworks/scope/pull/2026#issuecomment-263597159 for details) 
- [x] Add checkpoint flag (like we do for k8s, see https://github.com/weaveworks/scope/pull/1391)
- [x] Extend metadata shown in the details panel
- [x] Rendering
- [x] Make the topology appear along the top (`app/api_topology.go`)
- [x] Actually get the parent stuff described above working - there's a bug I can't work out right now and I need to go sleep. Something to do with the immutable/copying stuff, because I appear to be successfully making the change but it doesn't actually appear in the report.

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -183,11 +183,10 @@ func AddInitialTopologiesToRegistry(registry *Registry) {
 			Rank:     4,
 		},
 		APITopologyDesc{
-			id:          weaveID,
-			parent:      hostsID,
-			renderer:    render.WeaveRenderer,
-			Name:        "weave net",
-			HideIfEmpty: true,
+			id:       weaveID,
+			parent:   hostsID,
+			renderer: render.WeaveRenderer,
+			Name:     "Weave Net",
 		},
 	)
 }
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -35,7 +35,7 @@ const (
 
 var (
 	topologyRegistry = MakeRegistry()
-	k8sPseudoFilter  = APITopologyOptionGroup{
+	unmanagedFilter  = APITopologyOptionGroup{
 		ID:      "pseudo",
 		Default: "hide",
 		Options: []APITopologyOption{
@@ -165,6 +165,7 @@ func AddInitialTopologiesToRegistry(registry *Registry) {
 			renderer:    render.ECSTaskRenderer,
 			Name:        "Tasks",
 			Rank:        3,
+			Options:     []APITopologyOptionGroup{unmanagedFilter},
 			HideIfEmpty: true,
 		},
 		APITopologyDesc{
@@ -172,6 +173,7 @@ func AddInitialTopologiesToRegistry(registry *Registry) {
 			parent:      ecsTasksID,
 			renderer:    render.ECSServiceRenderer,
 			Name:        "services",
+			Options:     []APITopologyOptionGroup{unmanagedFilter},
 			HideIfEmpty: true,
 		},
 		APITopologyDesc{
@@ -228,7 +230,7 @@ func updateFilters(rpt report.Report, topologies []APITopologyDesc) []APITopolog
 	for i, t := range topologies {
 		if t.id == podsID || t.id == servicesID || t.id == deploymentsID || t.id == replicaSetsID {
 			topologies[i] = updateTopologyFilters(t, []APITopologyOptionGroup{
-				kubernetesFilters(ns...), k8sPseudoFilter,
+				kubernetesFilters(ns...), unmanagedFilter,
 			})
 		}
 	}
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -16,18 +16,21 @@ import (
 )
 
 const (
-	apiTopologyURL                     = "/api/topology/"
-	processesTopologyDescID            = "processes"
-	processesByNameTopologyDescID      = "processes-by-name"
-	containerLabelFiltersGroupID       = "container_label_filters_group"
-	containersTopologyDescID           = "containers"
-	containersByHostnameTopologyDescID = "containers-by-hostname"
-	containersByImageTopologyDescID    = "containers-by-image"
-	podsTopologyDescID                 = "pods"
-	replicaSetsTopologyDescID          = "replica-sets"
-	deploymentsTopologyDescID          = "deployments"
-	servicesTopologyDescID             = "services"
-	hostsTopologyDescID                = "hosts"
+	apiTopologyURL               = "/api/topology/"
+	processesID                  = "processes"
+	processesByNameID            = "processes-by-name"
+	containerLabelFiltersGroupID = "container-label-filters-group"
+	containersID                 = "containers"
+	containersByHostnameID       = "containers-by-hostname"
+	containersByImageID          = "containers-by-image"
+	podsID                       = "pods"
+	replicaSetsID                = "replica-sets"
+	deploymentsID                = "deployments"
+	servicesID                   = "services"
+	hostsID                      = "hosts"
+	weaveID                      = "weave"
+	ecsTasksID                   = "ecs-tasks"
+	ecsServicesID                = "ecs-services"
 )
 
 var (
@@ -53,7 +56,10 @@ func AddInitialTopologiesToRegistry(registry *Registry) {
 		{
 			ID:      containerLabelFiltersGroupID,
 			Default: "application",
-			Options: []APITopologyOption{{Value: "all", Label: "All", filter: nil, filterPseudo: false}, {Value: "system", Label: "System Containers", filter: render.IsSystem, filterPseudo: false}, {Value: "notsystem", Label: "Application Containers", filter: render.IsApplication, filterPseudo: false}},
+			Options: []APITopologyOption{
+				{Value: "all", Label: "All", filter: nil, filterPseudo: false},
+				{Value: "system", Label: "System Containers", filter: render.IsSystem, filterPseudo: false},
+				{Value: "notsystem", Label: "Application Containers", filter: render.IsApplication, filterPseudo: false}},
 		},
 		{
 			ID:      "stopped",
@@ -90,7 +96,7 @@ func AddInitialTopologiesToRegistry(registry *Registry) {
 	// be the verb to get to that state
 	registry.Add(
 		APITopologyDesc{
-			id:          processesTopologyDescID,
+			id:          processesID,
 			renderer:    render.FilterUnconnected(render.ProcessWithContainerNameRenderer),
 			Name:        "Processes",
 			Rank:        1,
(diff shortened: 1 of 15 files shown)


ekimekim:
more testing is needed since my test environment turned out to be flawed (not enough cpu to run app and probe), but apart from the rendering stuff that @2opremio is working on, this should be ready to go

2opremio:
@abuehrle We are going to need documentation for this. Showing ECS views on Scope requires adding permissions to the AWS machines in which it runs.

@ekimekim knows the details and can draft/write the requirements but it would be great if you can review it (for Re:Invent which is when we plan to release this)

app/merger.go
```diff
@@ -48,14 +48,15 @@ func (smartMerger) Merge(reports []report.Report) report.Report {
...
 	for ; l > 1; l-- {
-		go func(left, right report.Report) {
-			c <- left.Merge(right)
+		go func(left, right *report.Report) {
+			r := left.Merge(*right)
+			c <- &r
 		}(<-c, <-c)
 	}
-	return <-c
+	return *<-c
```
rade: In slack I had proposed the following as a simpler change, which is blissfully pointer-free:
```diff
diff --git a/app/merger.go b/app/merger.go
index be4d149..94231ba 100644
--- a/app/merger.go
+++ b/app/merger.go
@@ -53,9 +53,10 @@ func (smartMerger) Merge(reports []report.Report) report.Report {
               c <- r
       }
       for ; l > 1; l-- {
-               go func(left, right report.Report) {
+               left, right := <-c, <-c
+               go func() {
                       c <- left.Merge(right)
-               }(<-c, <-c)
+               }()
       }
       return <-c
}
```

ekimekim: i've modified the function once again to use your approach, it's definitely the cleanest solution.

ekimekim:
Tested with sock shop. No obvious performance regressions after looking at CPU load on master (0.70) vs branch with ECS stuff disabled (0.81) vs branch with ECS stuff enabled (0.79). It's a really, really poor test, but we can discount crazy order-of-magnitude changes.

2opremio:
We should probably add a filter for Clusters, pretty much like we do for K8s namespaces.

It is probably not urgent (since a Container Instance can only belong to a cluster) and you can create different organizations for different clusters in Weave Cloud.

@ekimekim Please add the cluster to the metadata (probably to the host instead of the task/service) in order to make this possible.

app/api_topologies.go
```diff
@@ -90,81 +96,98 @@ func AddInitialTopologiesToRegistry(registry *Registry) {
 	// be the verb to get to that state
 	registry.Add(
 		APITopologyDesc{
-			id:          processesTopologyDescID,
+			id:          processesID,
```
tomwilkie: I think its best not to mix lots of renamings into a PR that adds features.  Makes it hard to see the real changes.

app/api_topologies.go
```diff
@@ -90,81 +96,98 @@ func AddInitialTopologiesToRegistry(registry *Registry) {
...
 		APITopologyDesc{
-			id:       "weave",
-			parent:   "hosts",
-			renderer: render.WeaveRenderer,
-			Name:     "Weave Net",
+			id:          weaveID,
+			parent:      hostsID,
+			renderer:    render.WeaveRenderer,
+			Name:        "weave net",
+			HideIfEmpty: true,
```
tomwilkie: Change here seems unrelated?

app/merger.go
```diff
@@ -53,9 +53,10 @@ func (smartMerger) Merge(reports []report.Report) report.Report {
 		c <- r
 	}
 	for ; l > 1; l-- {
-		go func(left, right report.Report) {
+		left, right := <-c, <-c
+		go func() {
 			c <- left.Merge(right)
-		}(<-c, <-c)
+		}()
```
tomwilkie: Again, unrelated change!

probe/awsecs/client.go
```diff
@@ -0,0 +1,138 @@
...
+
+				for _, failure := range resp.Failures {
+					// log the failures but still continue with what succeeded
+					log.Warnf("Failed to describe ECS service %s, ECS service report may be incomplete: %s", failure.Arn, failure.Reason)
+				}
+
+				lock.Lock()
+				for _, service := range resp.Services {
+					for _, deployment := range service.Deployments {
+						results[*deployment.Id] = *service.ServiceName
```
tomwilkie: Can a deployment "belong" to multiple services?

probe/awsecs/client.go
```diff
@@ -0,0 +1,138 @@
...
+		&ecs.ListServicesInput{Cluster: &c.cluster},
+		func(page *ecs.ListServicesOutput, lastPage bool) bool {
+			// describe each page of 10 (the max for one describe command) concurrently
+			group.Add(1)
+			go func() {
+				defer group.Done()
+
+				resp, err := c.client.DescribeServices(&ecs.DescribeServicesInput{
+					Cluster:  &c.cluster,
+					Services: page.ServiceArns,
```
tomwilkie: Its strikes me that it may not be safe to use page here, as it maybe reused within ListServicesPages.  At least, its hard to reason either way.

probe/awsecs/client.go
```diff
@@ -0,0 +1,138 @@
...
+	return &ecsClient{
+		client:  ecs.New(sess, &aws.Config{Region: aws.String(region)}),
+		cluster: cluster,
+	}, nil
+}
+
+// returns a map from deployment ids to service names
+// cannot fail as it will attempt to deliver partial results, though that may end up being no results
+func (c ecsClient) getDeploymentMap() map[string]string {
+	results := make(map[string]string)
```
tomwilkie: We prefer `results := map[string]string{}`

probe/awsecs/client.go
```diff
@@ -0,0 +1,138 @@
...
+	resp, err := c.client.DescribeTasks(&ecs.DescribeTasksInput{
+		Cluster: &c.cluster,
+		Tasks:   taskPtrs,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, failure := range resp.Failures {
+		// log the failures but still continue with what succeeded
```
tomwilkie: comments shouldn't just restate what the code does in english.

2opremio:
Fixes #722

probe/awsecs/client.go
```diff
@@ -0,0 +1,138 @@
...
+		log.Warnf("Failed to describe ECS task %s, ECS service report may be incomplete: %s", failure.Arn, failure.Reason)
+	}
+
+	results := make(map[string]string)
```
tomwilkie: prefer `results := map[string]string{}`

probe/awsecs/client.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// do these two fetches in parallel
+	taskDeployments, err := c.getTaskDeployments(taskArns)
+	deploymentMap := <-deploymentMapChan
+
+	if err != nil {
+		return nil, err
+	}
+
+	results := make(map[string]string)
```
tomwilkie: prefer `results := map[string]string{}`

probe/awsecs/reporter.go
```diff
@@ -0,0 +1,121 @@
...
+	for nodeID, node := range rpt.Container.Nodes {
+
+		taskArn, taskArnOk := node.Latest.Lookup(docker.LabelPrefix + "com.amazonaws.ecs.task-arn")
+		cluster, clusterOk := node.Latest.Lookup(docker.LabelPrefix + "com.amazonaws.ecs.cluster")
+		family, familyOk := node.Latest.Lookup(docker.LabelPrefix + "com.amazonaws.ecs.task-definition-family")
+
+		if taskArnOk && clusterOk && familyOk {
+			taskMap, ok := results[cluster]
+			if !ok {
+				taskMap = make(map[string]*taskInfo)
```
tomwilkie: prefer `{}` over `make`.

probe/awsecs/reporter.go
```diff
@@ -0,0 +1,121 @@
...
+func getLabelInfo(rpt report.Report) map[string]map[string]*taskInfo {
+	results := make(map[string]map[string]*taskInfo)
+	log.Debug("scanning for ECS containers")
+	for nodeID, node := range rpt.Container.Nodes {
+
+		taskArn, taskArnOk := node.Latest.Lookup(docker.LabelPrefix + "com.amazonaws.ecs.task-arn")
+		cluster, clusterOk := node.Latest.Lookup(docker.LabelPrefix + "com.amazonaws.ecs.cluster")
+		family, familyOk := node.Latest.Lookup(docker.LabelPrefix + "com.amazonaws.ecs.task-definition-family")
+
+		if taskArnOk && clusterOk && familyOk {
```
tomwilkie: to reduce the nesting in this function, I'd probably push `if !ok { continue }` up after each lookup.

probe/awsecs/reporter.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+
+			for _, containerID := range info.containerIDs {
+				if containerNode, ok := rpt.Container.Nodes[containerID]; ok {
+					rpt.Container.Nodes[containerID] = containerNode.WithParents(parentsSets)
+				} else {
+					log.Warnf("Got task info for non-existent container %v, this shouldn't be able to happen", containerID)
+				}
+			}
+
```
tomwilkie: Remove newline.

probe/awsecs/reporter.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+
```
tomwilkie: Remove empty line.

prog/main.go
```diff
@@ -282,6 +284,9 @@ func main() {
 	flag.StringVar(&flags.probe.kubernetesConfig.User, "probe.kubernetes.user", "", "The name of the kubeconfig user to use")
 	flag.StringVar(&flags.probe.kubernetesConfig.Username, "probe.kubernetes.username", "", "Username for basic authentication to the API server")
 
+	// AWS ECS
+	flag.BoolVar(&flags.probe.ecsEnabled, "probe.ecs", false, "collect ecs-related attributes for containers on this node")
```
tomwilkie: Capitalise description.

tomwilkie:
Lots of nits, but nothing major.

probe/awsecs/client.go
```diff
@@ -0,0 +1,138 @@
...
+
+// returns a map from deployment ids to service names
+// cannot fail as it will attempt to deliver partial results, though that may end up being no results
+func (c ecsClient) getDeploymentMap() map[string]string {
+	results := make(map[string]string)
+	lock := sync.Mutex{} // lock mediates access to results
+
+	group := sync.WaitGroup{}
+
+	err := c.client.ListServicesPages(
```
tomwilkie: Ah no I see.

app/merger.go
ekimekim: This one's releated. See https://github.com/weaveworks/scope/pull/2026/commits/4007a902a264e5ff2c3be6b269ade515c9c1c145

Adding the new topologies put the `Report` struct over a critical size, such that passing two of them by value in a single function call causes a panic (something about running out of stack during goroutine startup).

ekimekim: It's a fair point that it could've been a seperate PR as long as it was merged first - this was just less accounting.

probe/awsecs/client.go
ekimekim: Yeah, what I wouldn't give for some lifetime annotations. This works, but I'll pull out serviceArns early for clarity.

probe/awsecs/reporter.go
```diff
@@ -0,0 +1,121 @@
...
+	log.Debug("Got ECS container info: %v", results)
+	return results
+}
+
+// Reporter implements Tagger
+type Reporter struct {
+}
+
+// Tag needed for Tagger
+func (r Reporter) Tag(rpt report.Report) (report.Report, error) {
```
ekimekim: what, like this? `func (Reporter) Tag(...`?
Had no idea that was valid syntax...

probe/awsecs/reporter.go
```diff
@@ -0,0 +1,121 @@
...
+
+		client, err := newClient(cluster)
+		if err != nil {
+			return rpt, err
+		}
+
+		taskArns := make([]string, 0, len(taskMap))
+		for taskArn := range taskMap {
+			taskArns = append(taskArns, taskArn)
+		}
```
ekimekim: I'd rather keep the weirdness of AWS's api types contained within that function.
And it's not really a matter of "back to" a `[]*string`, this is just a `keys()` operation on the map. Is there an easier way to get `[]*string` out of a map?

app/api_topologies.go
2opremio: I will create a different PR with this

app/api_topologies.go
2opremio: Something I learned after #1182 

I will create a separate PR for this.

probe/awsecs/client.go
2opremio: I would actually use make, but I would supply a length (since we know if in advance)

app/api_topologies.go
2opremio: Done https://github.com/weaveworks/scope/pull/2035

probe/awsecs/reporter.go
ekimekim: Also, it would be non-trivial to get pointers here - simply taking the pointer of the loop variable doesn't work, as I've discovered (twice) already.

2opremio:
I fixed the tests, which in turn required rebasing.

2opremio:
Given the time pressure, I have created a separate ticket to create unit tests for the ECS reporter https://github.com/weaveworks/scope/issues/2036

2opremio:
I've documented the permissions at https://github.com/weaveworks/integrations/pull/117 but we should also include it on the website (under `/site`) which is the only thing left. Over to you @ekimekim

errordeveloper:
Sorry for jumping in here, but if I may suggest a small feature, it'd be very very handy – I'd love to see link back to AWS resources, like task definitions and EC2 instances. Seems like a small addition, but it'd extremely useful.

rade:
@errordeveloper please file a separate issue.

errordeveloper:
> It requires the following permissions:
> - ListServices
> - DescribeTasks
> - DescribeServices

As I don't see `ListTasks` here, I am implying that the code retrieves list of services and picks tasks from there... That's okay, but there ways to get a task deployed without a service, a which can be done via [`aws ecs run-task`](http://docs.aws.amazon.com/cli/latest/reference/ecs/run-task.html) command or the equivalent API call (which is the API you'd use to build a custom schedulers, as the default service scheduler is pretty basic).

errordeveloper:
@rade https://github.com/weaveworks/scope/issues/2038

2opremio:
Regarding docs, after some thought, I don't think we should do anything here.

The website doesn't document any of the other actions required by the AMIs. They are included in  https://github.com/weaveworks/integrations/blob/master/aws/ecs/README.md and https://github.com/weaveworks/integrations/pull/117 already takes care include the new permissions.

If anything, we should move https://github.com/weaveworks/integrations/blob/master/aws/ecs/README.md to the website. ( created https://github.com/weaveworks/scope/issues/2039 )

ekimekim:
> @errordeveloper: I am implying that the code retrieves list of services and picks tasks from there

It scans the list of containers on the machine, extracts task names from the containers, and then goes from there. So you won't see tasks that don't have at least one container running on a probe-monitored instance.

errordeveloper:
> So you won't see tasks that don't have at least one container running on a probe-monitored instance.

Ah, sure, makes sense with respect to probe's view... Personally, I'd love to see tasks with no scheduled containers, but that'd be a big feature request, and I think I've already filed that as an issue sometime ago in the Kubernetes context or mentioned it elsewhere. Thanks for clarifying, glad we don't have a last minute bug.

2opremio:
LGTM

2opremio:
@pidster Also signed-off offline

Issue #1187, opened 2016-03-21
JS error on edges with lots of waypoints


davkal:
This becomes important again after #1186 lifted the 100 node limit on topologies.

For edges to be animated they have to keep the number of waypoints during their lifetime. Currently edges being extended to a fixed length. With high-node-count topologies, this becomes more frequent and should be addressed.

Here is what @foot found during the PR review:

> If you select a node, then click bg to deselect it often fails to return to the canvas due to the violation of the assumption we have max of 30 points on a line (I've seen 45 points and 61 points on a 500node layout). It would be neat to make this dynamic but I dunno how...

PR #2045, opened 2016-11-29
Remove upper bound for waypoints count on edges

fbarl:
Fix for #1187.

As long as we are using `dagre` in combination with `react-motion`, we have to have a way of transforming the array of edge waypoints with variable length that `dagre` is producing to a fixed length array that works with `Motion` animations. So far we were assuming the number of waypoints per edge will never go over `30`, but that doesn't hold anymore for big graphs.

In this PR, I introduced `WAYPOINTS_CAP` configurable constant which now basically decides to which precision we want to render the edges.

1. If the `dagre` waypoints array size is **lower** than `WAYPOINTS_CAP`, we add some *fake* copies of the first waypoint (I just left this case intact)
2. If the `dagre` waypoints array size is **higher** than `WAYPOINTS_CAP`, we select exactly `WAYPOINTS_CAP` waypoints uniformly across the `dagre` waypoints array (e.g. `[1, 2, 3, 4, 5] -> [1, 3, 5]` for `WAYPOINTS_CAP = 3`), which is arguably a a good simple heuristics that gives us a very similar resulting edge curve

Obviously, the lower we set the `WAYPOINTS_CAP`, the rougher our edges will look like, but the higher we set it, the worse performance we might get. IMO, value `8` looks like a good choice, even on graphs with a large number of nodes.

### DIFF
--- a/client/app/scripts/charts/edge-container.js
+++ b/client/app/scripts/charts/edge-container.js
@@ -5,11 +5,17 @@ import { Motion, spring } from 'react-motion';
 import { Map as makeMap } from 'immutable';
 import { line, curveBasis } from 'd3-shape';
 
+import { uniformSelect } from '../utils/array-utils';
 import { round } from '../utils/math-utils';
 import Edge from './edge';
 
-const animConfig = [80, 20]; // stiffness, damping
-const pointCount = 30;
+// Spring stiffness & damping respectively
+const ANIMATION_CONFIG = [80, 20];
+// Tweak this value for the number of control
+// points along the edge curve, e.g. values:
+//   * 2 -> edges are simply straight lines
+//   * 4 -> minimal value for loops to look ok
+const WAYPOINTS_CAP = 8;
 
 const spline = line()
   .curve(curveBasis)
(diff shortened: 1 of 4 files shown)


foot:
Have you had any luck getting a before and after shot of a report from say, prod?

I'm not getting a very stable layout for either branch right now! Hmm!

fbarl:
@foot Comparing the two graphs (i.e. their edges before & after this PR) generated by prod report, I couldn't really tell the difference - I think it might only become visible on graphs with 300+ nodes, when the number of waypoints given by `dagre` would be `>> 8`.

To be fair, I couldn't really make a very precise comparison because my layout was also not deterministic, but I think that's another issue to be solved..

client/app/scripts/charts/edge-container.js
```diff
@@ -50,14 +56,9 @@ class EdgeContainer extends React.Component {
   }
 
   render() {
-    const { layoutPrecision, points } = this.props;
+    const { layoutPrecision } = this.props;
     const other = _.omit(this.props, 'points');
 
-    if (layoutPrecision === 0) {
```
davkal: This is needed for really busy topologies. It returns a non-animated edge, which otherwise would look yanky.

client/app/scripts/charts/edge-container.js
```diff
@@ -71,24 +72,29 @@ class EdgeContainer extends React.Component {
...
+    // Motion requires a constant number of waypoints along the path of each edge
+    // for the animation to work correctly, but dagre might be changing their number
+    // depending on the dynamic topology reconfiguration. Here we are transforming
+    // the waypoints array given by dagre to the fixed size of `WAYPOINTS_CAP` that
+    // Motion could take over.
+    const pointsMissing = WAYPOINTS_CAP - nextPoints.length;
+    if (pointsMissing > 0) {
+      // Whenever there are some waypoints missing, we simply populate the beginning of the
+      // array with the first element, as this leaves the curve interpolation unchanged.
+      nextPoints = Array(pointsMissing).fill(_.first(nextPoints)).concat(nextPoints);
```
davkal: Is `Array.prototype.fill` polyfilled by babel?

client/app/scripts/charts/edge-container.js
fbarl: But if we don't want to animate the edges on graphs with `>= 50` nodes, then we are very unlikely to ever cross the old limit of `30` waypoints used with `Motion`. I thought the whole point of this story was to lift this limit, so that the bigger graphs could also animate properly. :)

 `pointsMap` will only be used on small graphs, so it would be very unlikely that we'd cross the limit of `30` waypoints there.

client/app/scripts/charts/edge-container.js
fbarl: Not sure, but I think it'd be better to replace it with `_.times(pointsMissing, nextPoints[0])` anyway.

client/app/scripts/charts/edge-container.js
davkal: That is true. But we had no way of knowing how many points there could be even with small topologies. The main goal was to not live in fear of this code blowing up. With your fix, it's robust.

davkal: To clarify, we need the optimization with non-animated edges now, but may take it out in the future.

fbarl:
@davkal I addressed your comments, PTAL.

davkal:
LGTM

Issue #2047, opened 2016-11-29
Scope metrics fill displayed to the side


pidster:
<img width="195" alt="fill" src="https://cloud.githubusercontent.com/assets/303151/20728935/d977a306-b634-11e6-886b-d30d8818381c.png">

The metrics fill is appearing at the right hand side, instead of the bottom.

2opremio:
@pidster Please add more information on how to reproduce

rade:
both dev and prod show this. Works fine for processes and hosts, but not containers and pods.

pidster:
I am running a Sock Shop which you are now invited to @2opremio.

errordeveloper:
I have also seen this, I thought it was intentional...

foot:
:), no, back to vertical for now.

PR #2048, opened 2016-11-30
Fix two bugs caused by transition to D3 v4

fbarl:
* D3 `rangeBands` doesn't exist anymore
* Scope metrics fill displayed to the side (#2047)


foot:
@fbarl the networks overlay fix looks good. 

Wanna test the hex/hepta shape fills and if okay please merge!

fbarl:
@foot The shape fills look good!

Eventually, I'd like to refactor all those expressions in `getClipPathDefinition`, but since it works as it is, I'll just merge this as a hotfix.

PR #2057, opened 2016-12-05
Fix all the npm library versions

fbarl:
As @davkal proposed, we decided to not use `^` or `~` version selectors for any of our `package.json` libraries, but keep all the versions fixed instead:

* `"react-tooltip": "~3.2.2"` -> **bad**
* `"react-tooltip": "^3.2.2"` -> **bad**
* `"react-tooltip": "3.2.2"` -> **good**


davkal:
LGTM

PR #2058, opened 2016-12-05
Upgraded eslint & eslint-config-airbnb

fbarl:
There is a bunch of new lint rules that I temporarily switched off, because they either require changes across multiple files or I wasn't sure whether we wanted to override them in some way.

@davkal, @foot: I will be working on resolving some of the items from the list on my own now, but I'm sure there will be at least a couple of them which we'll need to discuss. As I expect we'll soon be upgrading `eslint` on `service-ui` as well, maybe @jpellizzari & @bowenli would also like to contribute if they feel we should disable/override some of these rules:

* ~~[import/first](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/first.md)~~ **used**
* ~~[import/prefer-default-export](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/prefer-default-export.md)~~ - **disabled**
* ~~[import/no-extraneous-dependencies](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md)~~ - **used**, but only for `dependencies`, which is forcing the used *D3* modules to be listed explicitly in `package.json`; the default rule would force us to move some `react` dev stuff from `devDependencies` to `dependencies`
* ~~[import/no-named-as-default](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-named-as-default.md)~~ **used**
* ~~[jsx-a11y/no-static-element-interactions](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md)~~ - **disabled**, at least for now as it applies to a lot of elements and in some cases it seems too strict
* ~~[jsx-a11y/label-has-for](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md)~~ - **used**
* ~~[react/jsx-indent](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-indent.md)~~ **used**
* ~~[react/jsx-filename-extension](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-filename-extension.md)~~ - **tweaked** like @jpellizzari posted below
* ~~[react/jsx-first-prop-new-line](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-first-prop-new-line.md)~~ **used**
* ~~[react/jsx-no-target-blank](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md)~~ - **used**
* ~~[react/no-find-dom-node](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-find-dom-node.md)~~ - **used**
* ~~[react/no-string-refs](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-string-refs.md)~~ **used**
* ~~[react/self-closing-comp](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/self-closing-comp.md)~~ **used**
* ~~[react/forbid-prop-types](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-prop-types.md)~~ - **used**
* ~~[no-plusplus](https://github.com/eslint/eslint/blob/master/docs/rules/no-plusplus.md)~~ **used**
* ~~[no-mixed-operators](http://eslint.org/docs/rules/no-mixed-operators)~~ - **used**
* ~~[no-restricted-properties](http://eslint.org/docs/rules/no-restricted-properties)~~ - **disabled**, as we get warnings for using `**` instead of `Math.pow` which is only a part of *ES2017* syntax that *Airbnb* uses, but we don't
* ~~[no-underscore-dangle](http://eslint.org/docs/rules/no-underscore-dangle)~~ - **used**
* ~~[arrow-parens](http://eslint.org/docs/rules/arrow-parens)~~ - **used**
* ~~[class-methods-use-this](http://eslint.org/docs/rules/class-methods-use-this)~~ - **used**, but pulled such functions out of component logic and made them helpers instead of class static methods
* ~~[global-require](http://eslint.org/docs/rules/global-require)~~ - **disabled** because of how our hot loading is set up

**Note:** I will be removing the items from this list as I resolve them


jpellizzari:
Here is the `.eslintrc` file I have been using with version 13.0.0 of `eslint-config-airbnb`.
 
```json
{
  "extends": "airbnb",
  "parser": "babel-eslint",
  "env": {
    "browser": true,
    "jest": true,
    "node": true
  },
  "rules": {
    "one-var" : 0,
    "one-var-declaration-per-line": 0,
    "func-names": 0,
    "class-methods-use-this": 0,
    "import/prefer-default-export": 0,
    "import/no-named-as-default" : 0,
    "import/no-webpack-loader-syntax": 0,
    "import/no-extraneous-dependencies": [
      "error",
      {
        "devDependencies": true,
        "optionalDependencies": true,
        "peerDependencies": true
      }
    ],
    "react/jsx-filename-extension": [
      2,
      {
        "extensions": [
          ".js",
          ".jsx"
        ]
      }
    ],
    "comma-dangle": 0,
    "no-param-reassign": 0,
    "object-curly-spacing": 0,
    "react/jsx-closing-bracket-location": 0,
    "react/prefer-stateless-function": 0,
    "react/sort-comp": 0,
    "react/prop-types": 0,
    "jsx-a11y/no-static-element-interactions" : 0
  }
}
```

client/app/scripts/components/search.js
```diff
@@ -121,13 +125,14 @@ class Search extends React.Component {
           <div className="search-input">
             {showPinnedSearches && pinnedSearches.toIndexedSeq()
               .map(query => <SearchItem query={query} key={query} />)}
-            <input className="search-input-field" type="text" id={inputId}
+            <input
+              className="search-input-field" type="text" id={inputId}
               value={value} onChange={this.handleChange}
               onFocus={this.handleFocus} onBlur={this.handleBlur}
-              disabled={disabled} ref="queryInput" />
+              disabled={disabled} ref={this.saveQueryInputRef} />
```
fbarl: I think both patterns existed in parallel for a while now (the [official documentation](https://facebook.github.io/react/docs/refs-and-the-dom.html) uses the functional one).

The new `no-string-refs` rule goes against our *old* pattern, here's a discussion why:
http://stackoverflow.com/questions/37468913/why-ref-string-is-legacy

client/app/scripts/main.dev.js
```diff
@@ -1,18 +1,18 @@
...
 import Immutable from 'immutable';
 import installDevTools from 'immutable-devtools';
-installDevTools(Immutable);
 
+import configureStore from './stores/configureStore.dev';
+import DevTools from './components/dev-tools';
+
+require('font-awesome-webpack');
+require('../styles/main.less');
+require('../images/favicon.ico');
```
fbarl: Right, I replaced every `require(X);` with `import X;` in `main.js`, `main.dev.js`, `contrast-main.js` and `terminal-main.js`.

client/package.json
```diff
@@ -38,18 +45,20 @@
...
-    "eslint": "2.4.0",
-    "eslint-config-airbnb": "6.1.0",
-    "eslint-loader": "1.3.0",
-    "eslint-plugin-jasmine": "1.6.0",
-    "eslint-plugin-react": "4.2.2",
+    "eslint": "3.11.1",
+    "eslint-config-airbnb": "13.0.0",
+    "eslint-loader": "1.6.1",
+    "eslint-plugin-import": "2.2.0",
+    "eslint-plugin-jasmine": "2.2.0",
```
fbarl: Good question - @davkal do you think it can be removed?

davkal: I thought jest tests use jasmine syntax, so you get new keywords like `describe` and friends that the linter needs to know about.

client/app/scripts/components/grid-mode-selector.js
```diff
@@ -4,6 +4,21 @@ import classNames from 'classnames';
 
 import { toggleGridMode } from '../actions/app-actions';
 
+
+function renderItem(icons, label, isSelected, onClick) {
```
davkal: I think we should distinguish functions that return JSX a bit more, to make it visible that they return UI components. I have seen this format in other projects:
```
const Item = (icons, label, isSelected, onClick) => (
  <div>...</div>
);
```
Notice the first letter capitalized, like it is for component classes. That makes it easy to read when used in other places, e.g., `{items.map(Item)}`

client/package.json
jpellizzari: I believe `env: { node, browser, jest }` in the .eslintrc should take care of BDD globals.

jpellizzari:
LGTM

client/package.json
davkal: @fbarl could you please test if the linter still works on tests without that plugin being installed?

client/app/scripts/components/grid-mode-selector.js
fbarl: I quite like your convention, will use it from now on!

Issue #2050, opened 2016-12-01
ECS reporter throttled by AWS API


2opremio:
![](http://g.recordit.co/iZcc8l3fjF.gif)

2opremio:
This is the Sock Shop, run with the CloudFormation template (3x m4.xlarge instances) in Weave Cloud

2opremio:
It seems ECS is throttling us:

```
<probe> WARN: 2016/11/30 19:23:17.144169 Error listing ECS services, ECS service report may b
e incomplete: ThrottlingException: Rate exceeded
        status code: 400, request id: 7210cbfd-b732-11e6-a879-0b8af0abb45a
<probe> ERRO: 2016/11/30 19:23:17.154211 error applying tagger: ThrottlingException: Rate exc
eeded
        status code: 400, request id: 721252a0-b732-11e6-a879-0b8af0abb45a
<probe> WARN: 2016/11/30 19:23:18.549526 Error describing some ECS services, ECS service repo
rt may be incomplete: ThrottlingException: Rate exceeded
```

2opremio:
Also we should correct the printf string format of some warnings:

```
<probe> WARN: 2016/11/30 20:54:25.202746 Failed to describe ECS task %!s(*string=0xc42121bf10
), ECS service report may be incomplete: %!s(*string=0xc42121bf30)
```

2opremio:
First 1000 lines of the logs: http://sprunge.us/SNIb

ekimekim:
my short-term thoughts on a long-term solution: we may need to get clever here with caching and careful use of immutable fields. For example, StartedBy for a task isn't going to change, which means we don't need to DescribeTasks every time - except this means any other metadata we may want to collect will also get stale. :S

2opremio:
I've worked around it for now by creating the cluster in a separate region (AWS rate limits per region)

2opremio:
A robust way to fix this should be using the ECS event stream https://aws.amazon.com/blogs/compute/monitor-cluster-state-with-amazon-ecs-event-stream/

However, I am not sure whether or how easily we can plug Scope.

ekimekim:
I think it's too much to ask to need the user to create CloudWatch rules and an SQS topic as part of setup - or even if we go that route later, it'd be nice if it was optional.
So for now I'm moving forward with caching. My thoughts so far:
[...]

PR #2060, opened 2016-12-05
ecs reporter: Fix some log lines that were passing *string instead of string

ekimekim:
See https://github.com/weaveworks/scope/issues/2050#issuecomment-264231558

### DIFF
--- a/probe/awsecs/client.go
+++ b/probe/awsecs/client.go
@@ -73,7 +73,7 @@ func (c ecsClient) getServices() map[string]*ecs.Service {
 				}
 
 				for _, failure := range resp.Failures {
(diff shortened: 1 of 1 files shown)


probe/awsecs/client.go
```diff
@@ -73,7 +73,7 @@ func (c ecsClient) getServices() map[string]*ecs.Service {
 				}
 
 				for _, failure := range resp.Failures {
-					log.Warnf("Failed to describe ECS service %s, ECS service report may be incomplete: %s", failure.Arn, failure.Reason)
+					log.Warnf("Failed to describe ECS service %s, ECS service report may be incomplete: %s", *failure.Arn, *failure.Reason)
```
2opremio: Can `failure.Arn` or `failure.Reason` be `nil`?

ekimekim:
PTAL?

PR #2061, opened 2016-12-06
Use newly-created 'common' repo

### LINKED ISSUE #2509: Truncate prefixes in container image names
Container image names are often quite long, so are truncated in the node view.
Truncation often leaves only the prefix visible, which often offers very little information about the container e.g: `google_containers/...`.

It'd be nice if common prefixes were truncated so that the information-high parts of the name could be seen at a glance, e.g: `google.../kube-apiser...`

jml:
Fixes #2509

This means we can re-use this code in other weave repos without having to depend on scope.


jml:
I don't know.

jml:
Running `gvt restore` on scope produces huge numbers of changes. Maybe I'm doing it wrong?

jml:
I've made a totally new PR (#2062) that does `gvt restore` on master. `gvt fetch github.com/weaveworks/common` still brings in new vendoring. I don't know why, and I don't know why it should prevent this PR from being merged. 

PTAL.

tomwilkie:
Extra vendoring can be explained by weaveworks/common including tools, which has a bunch of go stuff, which you've pulled in the libraries for.  Its not a big deal.

jml:
#2066 talks a bit about that.

PR #2062, opened 2016-12-06
No op gvt restore

jml:
When I tried to add a dependency with my gvt, I got a huge delta.

So I decided to run `gvt restore` on master to see what I got. This is it.


jml:
We talked about this in the London office. Here are the highlights:

* not necessary for getting #2061 in
* in the past, people (esp. @tomwilkie) have manually:
   * deleted unnecessary files after vendoring (e.g. the fuzzing corpus of go-fuzz)
   * explicitly vendored in our transitive dependencies and deleted the `vendor` / `Godeps` directories from our vendored dependencies
* `gvt` itself has probably changed what it decides to pull in, and how

PR #2065, opened 2016-12-07
ECS reporter: Minimize API calls by caching task and service data

ekimekim:
Fixes https://github.com/weaveworks/scope/issues/2050

Due to AWS API rate limits, we need to minimize API calls as much as
possible.

Our stated objectives:
* for all displayed tasks and services to have up-to-date metadata
* for all tasks to map to services if able

My approach here:
* Tasks only contain immutable fields (that we care about). We cache tasks forever.
 We only DescribeTasks the first time we see a new task.
* We attempt to match tasks to services with what info we have. Any "referenced" services,
 ie. a service with at least one matching task, needs to be updated to refresh changing data.
* In the event that a task doesn't match any of the (updated) services, ie. a new service entirely
 needs to be found, we do a full list and detail of all services (we don't re-detail ones we just refreshed).
* To avoid unbounded memory usage, we evict tasks and services from the cache after 1 minute without use.
 This should be long enough for things like temporary failures to be glossed over.

This gives us exactly one call per task, and one call per referenced service per report, which is unavoidable to maintain fresh data. Expensive "describe all" service queries are kept to only when newly-referenced services appear, which should be rare.

We could make a few very minor improvements here, such as trying to refresh unreferenced but known services before doing a list query, or getting details one by one when "describing all" and stopping when all matches have been found, but I believe these would produce very minor, if any, gains in number of calls while having an unjustifiable effect on latency since we wouldn't be able to do requests as concurrently.

Speaking of which, this change has a minor performance impact. Even though we're now doing less calls, we can't do them as concurrently.

Old code:
```
concurrently:
	describe tasks (1 call)
	sequentially:
		list services (1 call)
		describe services (N calls concurrently)
```
Assuming full concurrency, total latency: 2 end-to-end calls

New code (worst case):
```
sequentially:
	describe tasks (1 call)
	describe services (N calls concurrently)
	list services (1 call)
	describe services (N calls concurrently)
```
Assuming full concurrency, total latency: 4 end-to-end calls

In practical terms, I don't expect this to matter.

### DIFF
--- a/probe/awsecs/client.go
+++ b/probe/awsecs/client.go
@@ -1,6 +1,7 @@
 package awsecs
 
 import (
+	"strings"
 	"sync"
 	"time"
 
@@ -13,17 +14,17 @@ import (
 
 // a wrapper around an AWS client that makes all the needed calls and just exposes the final results
 type ecsClient struct {
-	client  *ecs.ECS
-	cluster string
-	taskCache map[string]ecsTask
+	client       *ecs.ECS
+	cluster      string
+	taskCache    map[string]ecsTask
 	serviceCache map[string]ecsService
 }
 
 // Since we're caching tasks heavily, we ensure no mistakes by casting into a structure
 // that only contains immutable attributes of the resource.
 type ecsTask struct {
-	taskARN string
-	createdAt time.Time
+	taskARN           string
+	createdAt         time.Time
 	taskDefinitionARN string
 
 	// These started fields are immutable once set, and guarenteed to be set once the task is running,
@@ -32,7 +33,7 @@ type ecsTask struct {
 	startedBy string // tag or deployment id
 
 	// Metadata about this cache copy
-	fetchedAt time.Time
+	fetchedAt  time.Time
 	lastUsedAt time.Time
 }
 
@@ -41,17 +42,16 @@ type ecsTask struct {
 // but we avoid re-listing services unless we can't find a service for a task.
 type ecsService struct {
 	serviceName string
-	createdAt time.Time
 
 	// The following values may be stale in a cached copy
-	deploymentIDs []string
-	desiredCount int64
-	pendingCount int64
-	runningCount int64
+	deploymentIDs     []string
+	desiredCount      int64
+	pendingCount      int64
+	runningCount      int64
 	taskDefinitionARN string
 
 	// Metadata about this cache copy
-	fetchedAt time.Time
+	fetchedAt  time.Time
 	lastUsedAt time.Time
 }
 
@@ -70,42 +70,41 @@ func newClient(cluster string) (*ecsClient, error) {
 	}
 
 	return &ecsClient{
-		client:  ecs.New(sess, &aws.Config{Region: aws.String(region)}),
-		cluster: cluster,
-		taskCache: map[string]ecsTask{},
+		client:       ecs.New(sess, &aws.Config{Region: aws.String(region)}),
+		cluster:      cluster,
+		taskCache:    map[string]ecsTask{},
 		serviceCache: map[string]ecsService{},
 	}, nil
 }
 
 func newECSTask(task *ecs.Task) ecsTask {
-	now = time.Now()
+	now := time.Now()
 	return ecsTask{
-		taskARN: *task.TaskARN,
-		createdAt: *task.CreatedAt,
+		taskARN:           *task.TaskArn,
+		createdAt:         *task.CreatedAt,
 		taskDefinitionARN: *task.TaskDefinitionArn,
-		startedAt: *task.StartedAt,
-		startedBy: *task.StartedBy,
-		fetchedAt: now,
-		lastUsedAt: now,
+		startedAt:         *task.StartedAt,
+		startedBy:         *task.StartedBy,
+		fetchedAt:         now,
+		lastUsedAt:        now,
 	}
 }
 
 func newECSService(service *ecs.Service) ecsService {
-	now = time.Now()
-	deploymentIDs = make([]string, 0, len(service.Deployments))
+	now := time.Now()
+	deploymentIDs := make([]string, 0, len(service.Deployments))
 	for _, deployment := range service.Deployments {
-		deploymentIDs = append(deploymentIDs, *deployment.ID)
+		deploymentIDs = append(deploymentIDs, *deployment.Id)
 	}
 	return ecsService{
-		serviceName: *service.ServiceName,
-		createdAt: *service.CreatedAt,
-		deploymentIDs: deploymentIDs,
-		desiredCount: *service.DesiredCount,
-		pendingCount: *service.PendingCount,
-		runningCount: *service.RunningCount,
-		taskDefinitionARN: *service.TaskDefinitionARN,
-		fetchedAt: now,
-		lastUsedAt: now,
+		serviceName:       *service.ServiceName,
+		deploymentIDs:     deploymentIDs,
+		desiredCount:      *service.DesiredCount,
+		pendingCount:      *service.PendingCount,
+		runningCount:      *service.RunningCount,
+		taskDefinitionARN: *service.TaskDefinition,
+		fetchedAt:         now,
+		lastUsedAt:        now,
 	}
 }
 
@@ -118,9 +117,10 @@ func (c ecsClient) listServices() <-chan string {
 			&ecs.ListServicesInput{Cluster: &c.cluster},
 			func(page *ecs.ListServicesOutput, lastPage bool) bool {
 				for _, arn := range page.ServiceArns {
-					results <- arn
+					results <- *arn
 				}
-			}
+				return true
+			},
 		)
 		if err != nil {
 			log.Warnf("Error listing ECS services, ECS service report may be incomplete: %v", err)
@@ -134,6 +134,7 @@ func (c ecsClient) listServices() <-chan string {
 // with full ecsService objects being put into the cache. Closes done when finished.
 func (c ecsClient) describeServices() (chan<- string, <-chan bool) {
 	input := make(chan string)
+	done := make(chan bool)
 
 	go func() {
 		const MAX_SERVICES = 10 // How many services we can put in one Describe command
@@ -149,9 +150,14 @@ func (c ecsClient) describeServices() (chan<- string, <-chan bool) {
 				go func(arns []string) {
 					defer group.Done()
 
+					arnPtrs := make([]*string, 0, len(arns))
+					for i := range arns {
+						arnPtrs = append(arnPtrs, &arns[i])
+					}
+
 					resp, err := c.client.DescribeServices(&ecs.DescribeServicesInput{
 						Cluster:  &c.cluster,
-						Services: arns,
+						Services: arnPtrs,
 					})
 					if err != nil {
 						log.Warnf("Error describing some ECS services, ECS service report may be incomplete: %v", err)
@@ -162,11 +168,11 @@ func (c ecsClient) describeServices() (chan<- string, <-chan bool) {
 						log.Warnf("Failed to describe ECS service %s, ECS service report may be incomplete: %s", *failure.Arn, failure.Reason)
 					}
 
-					mutex.Lock()
+					lock.Lock()
 					for _, service := range resp.Services {
 						c.serviceCache[*service.ServiceName] = newECSService(service)
 					}
-					mutex.Unlock()
+					lock.Unlock()
 				}(page)
 
 				page = make([]string, 0, MAX_SERVICES)
@@ -181,10 +187,10 @@ func (c ecsClient) describeServices() (chan<- string, <-chan bool) {
 }
 
 // get details on given tasks, updating cache with the results
-func (c ecsClient) getTasks(taskArns []string) {
-	taskPtrs := make([]*string, len(taskArns))
-	for i := range taskArns {
-		taskPtrs[i] = &taskArns[i]
+func (c ecsClient) getTasks(taskARNs []string) {
+	taskPtrs := make([]*string, len(taskARNs))
+	for i := range taskARNs {
+		taskPtrs[i] = &taskARNs[i]
 	}
 
 	// You'd think there's a limit on how many tasks can be described here,
@@ -210,16 +216,16 @@ func (c ecsClient) getTasks(taskArns []string) {
 // Evict entries from the caches which have not been used within the eviction interval.
 func (c ecsClient) evictOldCacheItems() {
 	const EVICT_TIME = time.Minute
-	now = time.Now()
+	now := time.Now()
 
 	for arn, task := range c.taskCache {
-		if now - task.lastUsedAt > EVICT_TIME {
+		if now.Sub(task.lastUsedAt) > EVICT_TIME {
 			delete(c.taskCache, arn)
 		}
 	}
 
 	for name, service := range c.serviceCache {
-		if now - service.lastUsedAt > EVICT_TIME {
+		if now.Sub(service.lastUsedAt) > EVICT_TIME {
 			delete(c.serviceCache, name)
 		}
 	}
@@ -233,7 +239,7 @@ func (c ecsClient) matchTasksServices(taskARNs []string) (map[string]string, []s
 
 	deploymentMap := map[string]string{}
 	for serviceName, service := range c.serviceCache {
-		for _, deployment := range service.DeploymentIDs {
+		for _, deployment := range service.deploymentIDs {
 			deploymentMap[deployment] = serviceName
 		}
 	}
@@ -242,16 +248,16 @@ func (c ecsClient) matchTasksServices(taskARNs []string) (map[string]string, []s
 	unmatched := []string{}
 	for _, taskARN := range taskARNs {
 		task, ok := c.taskCache[taskARN]
-		if ! ok {
+		if !ok {
 			// this can happen if we have a failure while describing tasks, just pretend the task doesn't exist
 			continue
 		}
-		if ! strings.HasPrefix(task.startedBy, SERVICE_PREFIX) {
+		if !strings.HasPrefix(task.startedBy, SERVICE_PREFIX) {
 			// task was not started by a service
 			continue
 		}
-		if service, ok := deploymentMap[task.startedBy]; ok {
-			results[taskARN] = service.serviceName
+		if serviceName, ok := deploymentMap[task.startedBy]; ok {
+			results[taskARN] = serviceName
 		} else {
 			unmatched = append(unmatched, taskARN)
 		}
@@ -267,7 +273,7 @@ func (c ecsClient) getInfo(taskARNs []string) ecsInfo {
 	// First, we ensure we have all the tasks we need, and fetch the ones we don't.
 	// We also mark the tasks as being used here to prevent eviction.
 	tasksToFetch := []string{}
-	now = Time.Now()
+	now := time.Now()
 	for _, taskARN := range taskARNs {
 		if task, ok := c.taskCache[taskARN]; ok {
 			task.lastUsedAt = now
@@ -284,9 +290,9 @@ func (c ecsClient) getInfo(taskARNs []string) ecsInfo {
 	taskServiceMap, unmatched := c.matchTasksServices(taskARNs)
 
 	// In order to ensure service details are fresh, we need to refresh any referenced services
-	toDescribe, done := describeServices()
+	toDescribe, done := c.describeServices()
 	servicesRefreshed := map[string]bool{}
-	for taskARN, serviceName := range taskServiceMap {
+	for _, serviceName := range taskServiceMap {
 		if servicesRefreshed[serviceName] {
 			continue
 		}
@@ -305,11 +311,11 @@ func (c ecsClient) getInfo(taskARNs []string) ecsInfo {
 	// If we still have tasks unmatched, we'll have to try harder. Get a list of all services and,
 	// if not already refreshed, fetch them.
 	if len(unmatched) > 0 {
-		serviceNamesChan := listServices()
-		toDescribe, done := describeServices()
+		serviceNamesChan := c.listServices()
+		toDescribe, done := c.describeServices()
 		go func() {
 			for serviceName := range serviceNamesChan {
-				if ! servicesRefreshed[serviceName] {
+				if !servicesRefreshed[serviceName] {
 					toDescribe <- serviceName
 					servicesRefreshed[serviceName] = true
 				}
--- a/probe/awsecs/client.go
+++ b/probe/awsecs/client.go
@@ -27,7 +27,7 @@ type ecsTask struct {
 	createdAt         time.Time
 	taskDefinitionARN string
 
-	// These started fields are immutable once set, and guarenteed to be set once the task is running,
+	// These started fields are immutable once set, and guaranteed to be set once the task is running,
 	// which we know it is because otherwise we wouldn't be looking at it.
 	startedAt time.Time
 	startedBy string // tag or deployment id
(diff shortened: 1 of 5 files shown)


ekimekim:
Tested, added a ton of debug logging and fixed a bunch of bugs.
Can confirm we do much less calls now.

In this example system, we have one service with 3 tasks across 3 machines, so one task on our example machine.
This is what the first report on startup looks like: (many extra logs elided)
```
<probe> DEBU: 2016/12/10 02:09:06.235489 Creating new ECS client
<probe> DEBU: 2016/12/10 02:09:06.324142 Describing 1 ECS tasks
<probe> DEBU: 2016/12/10 02:09:06.367489 Refreshing ECS services
<probe> DEBU: 2016/12/10 02:09:06.367542 Described 0 services in 0 calls
<probe> DEBU: 2016/12/10 02:09:06.367569 After refreshing services, 1 tasks unmatched
<probe> DEBU: 2016/12/10 02:09:06.367574 Listing ECS services
<probe> DEBU: 2016/12/10 02:09:06.376347 Listed 1 services
<probe> DEBU: 2016/12/10 02:09:06.376414 Described 1 services in 1 calls
<probe> DEBU: 2016/12/10 02:09:06.387699 Got info from ECS: 1 tasks, 1 services
```
for a total of 3 calls (1 DescribeTasks, 1 ListServices, 1 DescribeServices)

In subsequent reports:
```
<probe> DEBU: 2016/12/10 02:09:07.162438 Refreshing ECS services
<probe> DEBU: 2016/12/10 02:09:07.162537 Described 1 services in 1 calls
<probe> DEBU: 2016/12/10 02:09:07.178871 After refreshing services, 0 tasks unmatched
<probe> DEBU: 2016/12/10 02:09:07.179110 Got info from ECS: 1 tasks, 1 services
```
for a total of 1 call (DescribeServices)

2opremio:
> To avoid unbounded memory usage, we evict tasks and services from the cache after 1 minute without use.

Shouldn't this be size-bound instead of time-bound? Also, 1-minute seems very conservative (1 minute in scheduler time is very small).

2opremio:
> such as trying to refresh unreferenced but known services before doing a list query

Would this fix #2085 ? (i.e. what does _known_ mean here? previously referenced?)

2opremio:
> In practical terms, I don't expect this to matter.

I won't either

probe/awsecs/reporter.go
```diff
@@ -78,39 +78,51 @@ func getLabelInfo(rpt report.Report) map[string]map[string]*taskLabelInfo {
 
 // Reporter implements Tagger, Reporter
 type Reporter struct {
+	clients map[string]*ecsClient
```
2opremio: Mind documenting the keys?

2opremio: Maybe renaming to `clusterClients` or `clientsPerCluster` would be enough

probe/awsecs/reporter.go
```diff
@@ -78,39 +78,51 @@ func getLabelInfo(rpt report.Report) map[string]map[string]*taskLabelInfo {
...
 	for cluster, taskMap := range clusterMap {
 		log.Debugf("Fetching ECS info for cluster %v with %v tasks", cluster, len(taskMap))
 
-		client, err := newClient(cluster)
-		if err != nil {
-			return rpt, err
+		client, ok := r.clients[cluster]
+		if !ok {
+			log.Debugf("Creating new ECS client")
+			var err error // can't use := on the next line without shadowing outer client var
```
2opremio: I would remove the comment

probe/awsecs/client.go
```diff
@@ -106,45 +219,173 @@ func (c ecsClient) getTasks(taskArns []string) (map[string]*ecs.Task, error) {
...
+		}
 	}
 
-	deploymentMap := c.getDeploymentMap(services)
+	log.Debugf("Matched %d from %d tasks, %d unmatched", len(results), len(taskARNs), len(unmatched))
+	return results, unmatched
+}
+
+// Returns a ecsInfo struct containing data needed for a report.
+func (c ecsClient) getInfo(taskARNs []string) ecsInfo {
```
2opremio: Mind breaking down this function into simpler units? It got really long

2opremio:
Can you make a rough estimation on how many clusters/container instances/services could be run with these improvements?

Did you test this in AWS with a few clusters? 

Also, could you please add some unit tests?

probe/awsecs/client.go
```diff
@@ -106,45 +219,173 @@ func (c ecsClient) getTasks(taskArns []string) (map[string]*ecs.Task, error) {
...
 }
 
-// returns a map from task ARNs to service names
-func (c ecsClient) getInfo(taskArns []string) (ecsInfo, error) {
-	servicesChan := make(chan map[string]*ecs.Service)
-	go func() {
-		servicesChan <- c.getServices()
-	}()
+// Evict entries from the caches which have not been used within the eviction interval.
+func (c ecsClient) evictOldCacheItems() {
```
2opremio: I would suggest using an existing cache library. A quick search reveals we use github.com/bluele/gcache and github.com/coocood/freecache in some parts of the code

probe/awsecs/client.go
```diff
@@ -12,13 +14,50 @@ import (
 
 // a wrapper around an AWS client that makes all the needed calls and just exposes the final results
 type ecsClient struct {
-	client  *ecs.ECS
-	cluster string
+	client       *ecs.ECS
+	cluster      string
+	taskCache    map[string]ecsTask
+	serviceCache map[string]ecsService
```
ekimekim: done

probe/awsecs/client.go
ekimekim: I don't think a standard cache would be suitable. Our goal here isn't to keep a certain size budget of things cached for performance, it's to absolutely minimize the number of calls we're making by explicitly managing the cache. Things shouldn't be evicted if they're still in use - in the case of services in particular this will trigger an expensive list-and-describe-all.
The timed eviction logic is only here so we can withstand a given task flickering out of existence due to a restarting container or other temporary weirdness. Otherwise we'd be explicitly only carrying forward the items we used last round.

With this in mind, maybe the name "cache" is misleading - perhaps "known" or even just simply "tasks" and "services"?

probe/awsecs/reporter.go
```diff
@@ -78,39 +78,51 @@ func getLabelInfo(rpt report.Report) map[string]map[string]*taskLabelInfo {
 
 // Reporter implements Tagger, Reporter
 type Reporter struct {
+	clients map[string]*ecsClient
+}
+
+// New creates a new Reporter
+func New() Reporter {
```
ekimekim: done

probe/awsecs/reporter.go
ekimekim: done

ekimekim:
PTAL

probe/awsecs/client.go
```diff
@@ -31,72 +70,146 @@ func newClient(cluster string) (*ecsClient, error) {
...
 	return results
 }
 
-// cannot fail as it will attempt to deliver partial results, though that may end up being no results
-func (c ecsClient) getServices() map[string]*ecs.Service {
-	results := map[string]*ecs.Service{}
-	lock := sync.Mutex{} // lock mediates access to results
+// Returns (input, done) channels. Service ARNs given to input are batched and details are fetched,
+// with full ecsService objects being put into the cache. Closes done when finished.
+func (c ecsClient) describeServices() (chan<- string, <-chan bool) {
```
2opremio: It doesn't look like you are sending anything over the done channel. I would use a `chan struct{}` instead of a `chan bool`

probe/awsecs/client.go
```diff
@@ -31,72 +70,146 @@ func newClient(cluster string) (*ecsClient, error) {
...
-				}
+	go func() {
+		const maxServices = 10 // How many services we can put in one Describe command
+		group := sync.WaitGroup{}
+		lock := sync.Mutex{} // mediates access to the service cache when writing results
 
-				for _, failure := range resp.Failures {
-					log.Warnf("Failed to describe ECS service %s, ECS service report may be incomplete: %s", failure.Arn, failure.Reason)
-				}
+		describePage := func(arns []string) {
```
2opremio: To make this even easier to read I would move `describePage` out of `describeServices`. Also, instead of initializing the lock every time, wouldn't it be cleaner if it lived on the client struct?

2opremio:
I gave it another read. I asked some questions in my first comments which remain to be answered.

probe/awsecs/client.go
ekimekim: Huh, cool.

probe/awsecs/client.go
```diff
@@ -31,72 +70,146 @@ func newClient(cluster string) (*ecsClient, error) {
...
+		startedAt:         *task.StartedAt,
+		startedBy:         *task.StartedBy,
+		fetchedAt:         now,
+		lastUsedAt:        now,
+	}
+}
+
+func newECSService(service *ecs.Service) ecsService {
+	now := time.Now()
+	deploymentIDs := make([]string, 0, len(service.Deployments))
```
ekimekim: Yeah, I'm still in the python habit of "avoid using indices". done.

probe/awsecs/client.go
ekimekim: done

ekimekim:
> Can you make a rough estimation on how many clusters/container instances/services could be run with these improvements?

Taking socks shop as an (admittedly poor) example:
Socks shop has 14 services and 1 task per service, across 3 machines.
Let's take the average of 14/3 = 4.6 tasks per machine and round up to 5.

Note it's unclear if rate limits are per actual http request, or per thing being described (ie. if a describe services call with 10 services is 10x the 'cost' of a describe services with 1 service, or the same 'cost')

Old:
for each machine, list and describe all and services, and describe all tasks on that machine
= 3 machines * (1 list services + 14 describe services + 5 describe tasks)
or 3 machines * (1 list services + 2 describe service calls + 1 describe tasks call)
= 60 list or describe operations
or 12 list or describe calls

New:
first report is as per old. for subsequent reports:
for each machine, describe all services on that machine
= 3 machines * 5 describe services
or 3 machines * 1 describe service call
= 15 describe operations
or 3 describe calls

In summary: We make 1/4 the calls. But note this is somewhat of an ideal scenario, since most services aren't present on most machines.
But I'd expect we'll always save at least half of all calls, just by virtue of caching all tasks.

ekimekim:
rebased, resolved trivial conflict

ekimekim:
I'm running a test with 3 socks shop clusters in one region, to see if I hit rate limits. Leaving it 12h, which should be enough to hit any longer-term limits.

ekimekim:
Ran for about 15h with no rate limiting.

Can I get a PTAL % no unit tests?

2opremio:
Please rebase, #2094 has messed up your commits and it's hard to review.

2opremio:
Also, there's a pending question about size-based eviction vs time-based eviction (I would go for the first one or a combination of the two if you want to refine it).

ekimekim:
rebased.

The reason I went for a time-based cache is because:
* The sizes involved are small. Even if the strings end up being large, a ecsTask or ecsService struct could only be at most 100s of bytes. Supposing a worst-case scenario of 512B, we could store 100s of such tasks and services for a total cost of less than 100KB.
* In contrast, the cost of an additional AWS call is very high, since we don't know what the limits are and when you hit that limit, there's no recourse except "don't run as much stuff in one region".

Therefore, we _always_ want to prefer more caching over calls.
I added the time-based eviction as an afterthought just to ensure the cache wouldn't grow without bound in scenarios with high amounts of task churn.
Increasing the "1 minute" to "1 hour" would probably make sense, now I think of it, just to ensure we never evict until it's unreasonable to think it might ever be useful again.

The other reason I'm hesitant to use a size-based cache is because it would complicate the code which refreshes items and puts them in the cache, since we can no longer be 100% sure that items we get in step A would still be there by the time we reach step B. We would need to either introduce extra "working" maps to track things that we are currently using, or find/implement a size-based cache with an explicit "evict now" step instead of upon every addition.

2opremio:
I think that not setting a maximum size bound is unsafe particularly if we increase the time limit (which we should).

Otherwise the code looks good to me (module the missing tests)

2opremio:
> We would need to either introduce extra "working" maps to track things that we are currently using, or find/implement a size-based cache with an explicit "evict now" step instead of upon every addition.

ekimekim:
But what happens when the max cache size is smaller than the memory needed to store all the currently-active tasks and services? We'd end up with items disappearing before we have a chance to actually report them, and get silent data loss.

2opremio:
It won't happen if the size is big enough (say 10k services and tasks, if
we have more entries than that, there's something clearly wrong).

Even if it happens, it will only affect ECS views, which is preferable to
getting killed by the kernel.

In general, I think it's a good idea to set explicit size limits, because
otherwise the system will set them for you in an uncontrolled fashion.

ekimekim:
PTAL.
I now use gcache with an expiring LRU, with user-controllable settings (default 1MiB, 1 hour). It doesn't "refresh" the expiry time upon use, so there'll be a spike of requests every hour, bu in practice this shouldn't matter.
I've made tests that exercise the reporter code by mocking the `EcsClient`. The client code has no coverage right now. I think this is acceptable considering the complexity of mocking out the AWS api.

ekimekim:
(I've tested manually, and i'm just fixing some linter pedantry right now)

probe/awsecs/client.go
```diff
@@ -1,102 +1,212 @@
...
-// returns a map from deployment ids to service names
-func (c ecsClient) getDeploymentMap(services map[string]*ecs.Service) map[string]string {
-	results := map[string]string{}
-	for serviceName, service := range services {
-		for _, deployment := range service.Deployments {
-			results[*deployment.Id] = serviceName
-		}
+func newECSTask(task *ecs.Task) EcsTask {
+	return EcsTask{
+		TaskARN:           *task.TaskArn,
```
2opremio: Why the `*`s?

2opremio: Oh, sorry, it's `ecs.Task` not `ECSTask`, my bad.

probe/awsecs/client.go
```diff
@@ -1,102 +1,212 @@
...
+			batch = append(batch, arn)
+			if len(batch) == maxServices {
+				group.Add(1)
+				go func(arns []string) {
+					defer group.Done()
+					c.describeServicesBatch(arns)
+				}(batch)
+				count += len(batch)
+				calls++
+				batch = make([]string, 0, maxServices)
```
2opremio: I think it's faster and more descriptive to do `batch := batch[:0]`

2opremio: For instance, there will be no allocations

probe/awsecs/client.go
```diff
@@ -1,102 +1,212 @@
...
+// Returns (input, done) channels. Service ARNs given to input are batched and details are fetched,
+// with full EcsService objects being put into the cache. Closes done when finished.
+func (c ecsClientImpl) describeServices() (chan<- string, <-chan struct{}) {
+	input := make(chan string)
+	done := make(chan struct{})
+
+	log.Debugf("Describing ECS services")
 
+	go func() {
+		const maxServices = 10 // How many services we can put in one Describe command
```
2opremio: Wouldn't all this be equivalent to having a channel with a buffer of >10? Instead of calling `describeServicesBatch()` in the background, you do it in the foreground and let the channel do the buffering for you.

probe/awsecs/client.go
```diff
@@ -106,45 +216,173 @@ func (c ecsClient) getTasks(taskArns []string) (map[string]*ecs.Task, error) {
...
+	results := map[string]string{}
+	unmatched := []string{}
+	for _, taskARN := range taskARNs {
+		taskRaw, err := c.taskCache.Get(taskARN)
+		if err != nil {
+			// this can happen if we have a failure while describing tasks, just pretend the task doesn't exist
+			continue
+		}
+		task := taskRaw.(EcsTask)
+		if !strings.HasPrefix(task.StartedBy, servicePrefix) {
```
2opremio: Would it make sense to add a field in `EcsTask`describing wether it's a service in `newECSTask`?

probe/awsecs/client.go
```diff
@@ -106,45 +216,173 @@ func (c ecsClient) getTasks(taskArns []string) (map[string]*ecs.Task, error) {
...
+		} else {
+			unmatched = append(unmatched, taskARN)
+		}
+	}
+
+	log.Debugf("Matched %d from %d tasks, %d unmatched", len(results), len(taskARNs), len(unmatched))
+	return results, unmatched
+}
+
+func (c ecsClientImpl) ensureTasks(taskARNs []string) {
```
2opremio: `addTasksToCache`? `ensureTasksAreCached`?

probe/awsecs/client.go
```diff
@@ -106,45 +216,173 @@ func (c ecsClient) getTasks(taskArns []string) (map[string]*ecs.Task, error) {
...
-	// do these two fetches in parallel
-	tasks, err := c.getTasks(taskArns)
-	services := <-servicesChan
+func (c ecsClientImpl) makeECSInfo(taskARNs []string, taskServiceMap map[string]string) EcsInfo {
+	// The maps to return are the referenced subsets of the full caches
+	tasks := map[string]EcsTask{}
+	for _, taskARN := range taskARNs {
+		// It's possible that tasks could still be missing from the cache if describe tasks failed.
+		// We'll just pretend they don't exist.
+		if taskRaw, err := c.taskCache.Get(taskARN); err == nil {
```
2opremio: Worth factoring it out in a function (this casting seems to be used at at least a couple of places)? (same for services)

Maybe add a cache interface?

prog/main.go
```diff
@@ -287,6 +289,8 @@ func main() {
 
 	// AWS ECS
 	flag.BoolVar(&flags.probe.ecsEnabled, "probe.ecs", false, "Collect ecs-related attributes for containers on this node")
+	flag.IntVar(&flags.probe.ecsCacheSize, "probe.ecs.cache.size", 1024*1024, "Max size of cached info for each ECS cluster")
+	flag.DurationVar(&flags.probe.ecsCacheExpiry, "probe.ecs.cache.expiry", time.Hour, "How long to keep cached ECS info")
```
2opremio: 👍

2opremio:
Thanks for changing the cache and for the tests. How and for how long did you test this? (Not that I would had been able to make it simpler, but the code is quite involved)

2opremio:
> Ran for about 15h with no rate limiting.

3 clusters, right? With how many instances each?

probe/awsecs/client.go
ekimekim: This will cause the next batch to re-use the same in-memory array as the previous batch. Since we haven't made a copy of the underlying array when passing batch to `describeServicesBatch`, only of the slice (which continues pointing to the same underlying array), this causes subsequent appends to batch to modify entries in the previous batch.

Simplified example: https://play.golang.org/p/FfQf7pMqDK

probe/awsecs/client.go
ekimekim: It's not immediately apparent to me how you'd implement this. How do you block until a channel is full, or has a certain number of elements?
Also, it was an intentional choice to do each describe in the background so that many describe operations could be executing concurrently (my assumption is that any local processing is much faster than any network call)

probe/awsecs/client.go
ekimekim: My instinct is to try to keep the data normalized, ie. just store raw fields and not derived ones. But I could do something like `ecsTask.isServiceManaged()` so that the check itself is in one obvious place (and it's not exactly an expensive check).

ekimekim: done

probe/awsecs/client.go
ekimekim: did the latter

probe/awsecs/client.go
ekimekim: done

ekimekim:
PTAL

Yes, 3 clusters. Each with the defaults for the socks shop demo, which is 3 instances, about 10 or so services with one task per service.

ekimekim:
Note, the code has changed since then, but not in ways that matter to the rate-limiting. And I've since re-tested that it still works, but haven't repeated the 15h test.

probe/awsecs/client.go
2opremio: Good point, thanks

probe/awsecs/client.go
```diff
@@ -106,45 +239,166 @@ func (c ecsClient) getTasks(taskArns []string) (map[string]*ecs.Task, error) {
...
+	servicesRefreshed := map[string]bool{}
+	for _, serviceName := range taskServiceMap {
+		if servicesRefreshed[serviceName] {
+			continue
+		}
+		toDescribe <- serviceName
+		servicesRefreshed[serviceName] = true
+	}
+	close(toDescribe)
+	<-done
```
2opremio: Is done really needed? AFAICT it's not read on the other end.

probe/awsecs/client.go
2opremio: Thanks for the clarifications.

>  it was an intentional choice to do each describe in the background so that many describe operations could be executing concurrently (my assumption is that any local processing is much faster than any network call)

Is it worth the compilation? 

Also, is waiting for 10 services really the right thing? 

The channel needs to explicitly closed by the writer in order to "flush" the services so I don't really see the advantage of using a channel for input. You could pass all the services directly (no input channel), chunk them and spawn go routines. Simpler and more explicit.

ekimekim: > The channel needs to explicitly closed by the writer in order to "flush" the services so I don't really see the advantage of using a channel for input.

The intent was that, with a `ListServices` api call returning pages over time, we could start describing the first 10 before the `ListServices` call was finished (the flush behaviour is only for the last, partial batch). But you're right, it's a huge complication for a negligible gain. Will refactor.

I do still think doing the describes concurrently is worthwhile though. If you have, say, 50 services, and a request takes, say, 100ms round trip, then 5 concurrent requests vs 5 sequential ones is the difference between 100ms and 500ms total wall time.

probe/awsecs/client.go
ekimekim: This is the read here, we're blocking on it being written to or closed inside describeServices. However, by simplifying describeServices to take in args instead of returning an input chan, we will do away with the need for it.

probe/awsecs/client.go
2opremio: I agree. Thanks!

ekimekim:
PTAL

probe/awsecs/client.go
```diff
@@ -129,85 +129,67 @@ func (c ecsClientImpl) getCachedService(serviceName string) (EcsService, bool) {
...
-				batch = make([]string, 0, maxServices)
-			}
-		}
-		if len(batch) > 0 {
-			c.describeServicesBatch(batch)
-			count += len(batch)
-			calls++
-		}
+	// split into batches
+	batches := make([][]string, 0, len(services)/maxServices+1)
```
2opremio: I think it would be event simpler to spawn the goroutines directly (factoring the spawning into a function called inside the loop) instead of accumulating the batches in slices of slices.

2opremio:
LGTM. That last comment is a recommendation only.

probe/awsecs/client.go
ekimekim: I considered it, but found it awkward due to needing to repeat the spawning procedure for the special final case, the partial batch.

2opremio: Yep, that's why I mention factoring the spawning into a function. Your call though.

PR #2067, opened 2016-12-07
Follow redirections in curl when getting k8s resources

2opremio:
Part of 
https://github.com/weaveworks/service/issues/1028


rade:
Why not change the URL to the new location, i.e. https://cloud.weave.works/k8s/scope.yaml ?

peterbourgon:
I defer to @2opremio's initial judgment on -L vs. k8s/scope.yaml and 👍  this to go out as a breakfix.

PR #2070, opened 2016-12-07
Add eBPF connection tracking, without dependencies on kernel headers

### LINKED ISSUE #1168: walking /proc to obtain connections is very expensive

### LINKED ISSUE #1260: Short-lived connections not tracked for containers in shared networking namespaces

### LINKED ISSUE #1961: Remove runtime kernel header dependency from ebpf tracker
In production, a runtime dependency on kernel headers are not acceptable. We should remove it.

Provided a small footprint, one option is to get ebpf bytecode for all minor kernel versions (or even more smartly, for kernel versions API changes).

### LINKED ISSUE #1962: Port ebpf tracker to Go

iaguis:
Based on work from Lorenzo, updated by Iago, Alban, and Alessandro

This PR adds connection tracking using eBPF. This feature is not enabled by default.
For now, you can enable it by launching scope with the following command:
```
```
Scope Probe also falls back on the old /proc parsing if eBPF is not working
(e.g. too old kernel, or pre-compiled eBPF program missing for the current kernel).
This patch allows scope to get notified of every connection event, without relying
on the parsing of /proc/$pid/net/tcp{,6} and /proc/$pid/fd/*, and therefore
improve performance.

We vendor https://github.com/kinvolk/gobpf-elf-loader in Scope to load and
receive information from the ebpf programs. The Scope build fetches the pre-compiled
ebpf programs from https://hub.docker.com/r/kinvolk/tcptracer-bpf/
(see https://github.com/kinvolk/tcptracer-bpf)

At the moment, the eBPF programs are pre-compiled for:
- Arch 4.8.11-1
- fedora-24
- debian-testing
- coreos

The ebpf programs use kprobes on the following kernel functions:
- tcp_v4_connect
- tcp_v6_connect
- inet_csk_accept
- tcp_close

connection tuple but also pid and netns.
Note: the IPv6 events are not plugged in Scope.

conntrack, it also keeps the dead connections for one iteration in order
to report short-lived connections.

state is now done in foreground instead of background, via newForegroundReader().

NAT resolution on connections from eBPF works in the same way as it did on connections
from /proc: by using conntrack. One of the two conntrack instances was removed since
eBPF detects short-lived connections.

The Scope Docker image size comparison:
- weaveworks/scope in current master:  22 MB (compressed),  68 MB
  (uncompressed)
- weaveworks/scope with this patchset: 24 MB (compressed), 70 MB
  (uncompressed)

Fixes #1260 (Short-lived connections not tracked for containers in
shared networking namespaces)
Fixes #1962 (Port ebpf tracker to Go)
Fixes #1961 (Remove runtime kernel header dependency from ebpf tracker)

Supersedes #2024

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,8 @@ CODECGEN_TARGETS=report/report.codecgen.go render/detailed/detailed.codecgen.go
 RM=--rm
 RUN_FLAGS=-ti
 BUILD_IN_CONTAINER=true
-GO_ENV=GOGC=off
-GO_ENV_ARM=$(GO_ENV) CC=/usr/bin/arm-linux-gnueabihf-gcc CGO_ENABLED=1
+GO_ENV=GOGC=off CGO_ENABLED=1
+GO_ENV_ARM=$(GO_ENV) CC=/usr/bin/arm-linux-gnueabihf-gcc
 GO_BUILD_INSTALL_DEPS=-i
 GO_BUILD_TAGS='netgo unsafe'
 GO_BUILD_FLAGS=$(GO_BUILD_INSTALL_DEPS) -ldflags "-extldflags \"-static\" -X main.version=$(SCOPE_VERSION) -s -w" -tags $(GO_BUILD_TAGS)
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,7 @@ GO_BUILD_INSTALL_DEPS=-i
 GO_BUILD_TAGS='netgo unsafe'
 GO_BUILD_FLAGS=$(GO_BUILD_INSTALL_DEPS) -ldflags "-extldflags \"-static\" -X main.version=$(SCOPE_VERSION) -s -w" -tags $(GO_BUILD_TAGS)
 IMAGE_TAG=$(shell ./tools/image-tag)
+EBPF_IMAGE=kinvolk/tcptracer-bpf:semaphore-master-3f7c26b
 
 all: $(SCOPE_EXPORT)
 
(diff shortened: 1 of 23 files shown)


rade:
> Scope fetches the pre-compiled ebpf program from https://hub.docker.com/r/kinvolk/tcptracer-bpf/
(see https://github.com/kinvolk/tcptracer-bpf)

Perhaps include all the binaries in the scope container image? Yes, it adds bloat, but dynamically downloading binaries (on every launch?) is gross.

alban:
> Perhaps include all the binaries in the scope container image? Yes, it adds bloat, but dynamically downloading binaries (on every launch?) is gross.

@rade It is already in the Scope Docker image. It is done at build-time (implemented in Scope's Makefile).

At the moment, the size of the compiled ebpf programs is very small, I think <50KB uncompressed.
https://hub.docker.com/r/kinvolk/tcptracer-bpf/tags/

rade:
Ah, so when you wrote "Scope fetches" you didn't actually mean "Scope" but "The scope build". Also you wrote "Scope fetches the pre-compiled ebpf *program*" (note the singular). Taking into account your comments, I believe that sentence should have read "The scope build fetches the pre-compiled ebpf programs." Correct?

alban:
@rade correct. Sorry for the impreceise wording. I or @iaguis will edit the PR text.

alban:
I closed the two previous iterations of this work, in favor of this PR. For the record, this was:
- iteration 1: #1967 (ebpf using bcc/python)
- iteration 2: #2024 (ebpf using bcc/gobpf)

schu:
Two graphs comparing the performance between master and this branch (by running our test dialer, #2082):

With up to 4K connections (40 dialer container with 100 connections each):

scope-probe-time-4k-conn

With a pool of 100 containers, each holding a random number of connections between 1 and 20:

scope-probe-time-100-containers-with-up-to-20-conn

Makefile
```diff
@@ -22,15 +22,28 @@ CODECGEN_TARGETS=report/report.codecgen.go render/detailed/detailed.codecgen.go
...
+GO_BUILD_FLAGS+=-pkgdir ~
+else
+GO=env $(GO_ENV) go
+endif
+
+NO_CROSS_COMP=unset GOOS GOARCH
+GO_HOST=$(NO_CROSS_COMP); env $(GO_ENV) go
+WITH_GO_HOST_ENV=$(NO_CROSS_COMP); $(GO_ENV)
 IMAGE_TAG=$(shell ./tools/image-tag)
+EBPF_IMAGE=kinvolk/tcptracer-bpf:semaphore-master-475be63
```
schu: Maybe change this to pull by digest as labels are not unique, e.g.

```
EBPF_IMAGE=kinvolk/tcptracer-bpf@sha256:e1ec43eaddff2712909fc61e869b088da73d283359eb470f12d01e3d074549c0`
```

iaguis:
Superseded by https://github.com/weaveworks/scope/pull/2135

PR #2071, opened 2016-12-08
Fix grid-mode-selector styling

### LINKED ISSUE #2056: Graph/table buttons don't resize nicely
For presentation/demos I get a better result by zooming in (<kbd>⌘</kbd>+<kbd>+</kbd>) rather than using contrast mode.

However, when doing that the Graph/table buttons get a strange vertical alignment, with an odd surrounding border.

screen shot 2016-12-05 at 14 09 38

jpellizzari:
Fixes #2056.

Nothing else in the header tries to re-size to accommodate  smaller screens, so neither should the graph/table selector. This sets it to a minimum width to prevent the wrapping behavior.

Before:
screen shot 2016-12-07 at 4 33 10 pm
After:
screen shot 2016-12-07 at 4 33 19 pm

@bowenli Could you review pretty please?

### DIFF
--- a/client/app/styles/main.less
+++ b/client/app/styles/main.less
@@ -1262,10 +1262,19 @@ h2 {
 
   margin-top: 8px;
   margin-left: 8px;
+  min-width: 161px;
 
   &-wrapper {
     pointer-events: all;
     border-color: @background-darker-secondary-color;
+    overflow: hidden;
+  }
+
+  &:first-child,
+  &:last-child {
+    .grid-mode-selector-action {
+      border-radius: 0;
+    }
   }
 
   &-action {


jpellizzari:
I should also mention that this solves an issue where there was one pixel of space between the inner div's border and the other divs border.

client/app/styles/main.less
```diff
@@ -1262,10 +1262,19 @@ h2 {
...
   &-wrapper {
     pointer-events: all;
     border-color: @background-darker-secondary-color;
+    overflow: hidden;
+  }
+
+  &:first-child,
+  &:last-child {
+    .grid-mode-selector-action {
+      border-radius: 0;
```
bowenli: This is a separate issue from 2056, pls make a note.

bowenli:
LGTM

PR #2074, opened 2016-12-08
Correctly instrument websocket handshakes

jml:
Part of weaveworks/monitoring#82

Just vendors a correct version of the library.


tomwilkie:
LGTM

PR #2082, opened 2016-12-12
Add extras/dialer

schu:
`extras/dialer` contains the dialer tool and scripts we @kinvolk currently use for performance tests between master and the eBPF work (#2070).


tomwilkie:
Tools is a subtree of github.com/weaveworks/build-tools.  Please don't modify it in place.

Also, this probably doesn't belong in github.com/weaveworks/build-tools.

schu:
Sorry, didn't notice it's a subtree. Some place else where we could put them?

2opremio:
@schu I would place it under `/extras` adding more contextual information on the README (purpose and usecases)

schu:
@2opremio thanks, updated with 6caf471

extras/dialer/dialer.go
```diff
@@ -0,0 +1,83 @@
...
+			fmt.Println("Error accepting: ", err.Error())
+			os.Exit(1)
+		}
+		// Handle connections in a new goroutine.
+		go handleRequest(conn)
+	}
+
+}
+
+func handleRequest(conn net.Conn) {
```
2opremio: Shouldn't this be simply removed?

2opremio:
Minor comment, otherwise LGTM

extras/dialer/dialer.go
schu: Thanks, removed.

extras/dialer/README.md
```diff
@@ -0,0 +1,46 @@
...
+# every 3 seconds (default 10 seconds) for 3 times (default 60 times)
+sudo ./tools/dialer/time-scope-probe 3 3
+...
+```
+
+## Build dialer container
+
+```
+go build -o bin/dialer
+docker build -t <username>/dialer .
```
alban: I would suggest to choose a name, e.g. `docker build -t dialer .`

Since the sources of dialer are included in this PR, the instructions could be changed to use them instead of running my image `albanc/dialer`. The "docker run" commands below could be changed to `docker run -d albanc/dialer ...`

PR #2086, opened 2016-12-12
Add flag for logging headers

jml:
To help fix weaveworks/monitoring#82 (issue with how we record metrics for websockets), we'd like to be able to get detailed request header information in our logs. This adds a flag to do that.


tomwilkie:
LGTM.  Perhaps also log the output of IsWSRequest?

prog/app.go
```diff
@@ -270,7 +270,11 @@ func appMain(flags appFlags) {
 
 	handler := router(collector, controlRouter, pipeRouter, flags.externalUI)
 	if flags.logHTTP {
-		handler = middleware.LogFailed.Wrap(handler)
+		if flags.logHTTPHeaders {
+			handler = middleware.Log{LogRequestHeaders: true}.Wrap(handler)
+		} else {
+			handler = middleware.LogFailed.Wrap(handler)
+		}
```
tomwilkie:
```golang
handler = middleware.Log{
  LogRequestHeaders: flags.logHTTPHeaders,
  LogFailed: false
}
```

jml:
> Perhaps also log the output of IsWSRequest?

Not a bad idea, but have to make that change to weaveworks/common.

Issue #1982, opened 2016-11-04
Add control icon to top corner of the terminal window


2opremio:
Once you have triggered a control you may not be able to tell what control it was by looking at the terminal window alone.

See for instance: https://github.com/weaveworks/scope/pull/1979

jpellizzari:
@2opremio I'm trying to understand the requirement. When a control is activated, there should be an icon on the terminal to indicate what happened, like this?
![screen shot 2016-11-08 at 1 18 19 pm](https://cloud.githubusercontent.com/assets/2802257/20117559/03e17e1c-a5b6-11e6-92c2-2d3089f8be66.png)
Or maybe:
![screen shot 2016-11-08 at 1 18 29 pm](https://cloud.githubusercontent.com/assets/2802257/20117567/0ac4ab28-a5b6-11e6-8393-8475253badb5.png)

2opremio:
I am not a designer but I was thinking the second option but with the icon
on the top left corner (instead of the top right corner)

jpellizzari:
@2opremio Do you think we should represent all of the controls statuses, or just the ones related to the terminal (attach and exec)? For example, does showing the 'stop' icon make sense?:
<img width="780" alt="screen shot 2016-11-15 at 2 16 24 pm" src="https://cloud.githubusercontent.com/assets/2802257/20309018/43865d98-ab3e-11e6-9e55-5d515414fe1b.png">

2opremio:
I would say just attach and exec. Happy to answer these questions by I am not the expert.  @bowenli @davkal or @pidster are much more knowledgable in UX than I am :)

Issue #1727, opened 2016-07-28
Tighten up Scope url state.


foot:
Loads of nulls, empty strings and empty arrays up there that could be filtered out and treated as defaults.

Could go as far as doing this to the nested objects too, filter out nulls and merged into a default.

PR #2087, opened 2016-12-12
Add control status icon to Terminal header

jpellizzari:
Implements #1982 

This change will indicate the control status of the terminal. For example, if you are using the `exec` control a terminal icon will appear in the upper left (same for the 'attach' command):
screen shot 2016-12-12 at 1 49 48 pm

 The icon is tightly coupled to the human name of the controls: "Attach" and "Exec shell". This will break if we ever change those values. If there is a different value to lookup (like an ID) that might be better.

@foot could you review please?

### DIFF
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -316,6 +316,7 @@ class Terminal extends React.Component {
           <span title="Close" className="terminal-header-tools-item-icon fa fa-close"
             onClick={this.handleCloseClick} />
         </div>
+        {this.getControlStatusIcon()}
         <span className="terminal-header-title">{this.getTitle()}</span>
       </div>
     );
@@ -382,12 +383,26 @@ class Terminal extends React.Component {
       </div>
     );
   }
+  getControlStatusIcon() {
+    const icon = this.props.controlStatus && this.props.controlStatus.get('control').icon;
+    return (
+      <span
+        style={{marginRight: '8px', width: '14px'}}
+        className={classNames('fa', {[icon]: icon})}
+      />
+    );
+  }
 }
 
+function mapStateToProps(state, ownProps) {
+  const controlStatus = state.get('controlPipes').find((pipe) =>
+    pipe.get('nodeId') === ownProps.pipe.get('nodeId')
+  );
+  return { controlStatus };
+}
 
 Terminal.defaultProps = {
   connect: true
 };
 
-
-export default connect()(Terminal);
+export default connect(mapStateToProps)(Terminal);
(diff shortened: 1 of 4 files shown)


client/app/scripts/components/terminal.js
```diff
@@ -382,12 +383,28 @@ class Terminal extends React.Component {
...
     );
   }
+  getControlStatusIcon() {
+    const icon = this.props.controlStatus && this.props.controlStatus.get('raw')
+      ? 'terminal'
+      : 'desktop';
+    return (
+      <span
+        style={{marginRight: '8px', width: '14px'}}
+        className={classNames('fa', {[`fa-${icon}`]: icon})}
```
foot: 👍

foot:
Slick! Like the design.

`raw: true|false` aligning w/ the icons is a bit co-incidental, though its a bummer to add more state it would be nice to record the icon of the control that created that pipe and use that explicitly. If you save it into the controlPipe object next to the raw status it should be de/serialized into the url state for reloads and what not.

jpellizzari:
@foot Updated to use the 'icon' field of the control. Thanks, that's much better.

foot:
Nice.

There's a bit of awkward implicitness in how we do the url state stuff.

As the entire `control` is being saved into the `controlPipe` you get a bit of unused info ( `controlPipe.control.{id,probeId,human,nodeId}`) up in there:

```
{
    "controlPipe": {
        "control": {
            "human": "Exec shell",
            "icon": "fa-terminal",
            "id": "docker_exec_container",
            "nodeId": "6229d32d614b8f7bc0ee51791ddf197de5a9d3c950a31e441082ab9528855f2c;<container>",
            "probeId": "392656d14781da1e",
            "rank": 2
        },
        "id": "pipe-5735362070544949278",
        "nodeId": "6229d32d614b8f7bc0ee51791ddf197de5a9d3c950a31e441082ab9528855f2c;<container>",
        "raw": true,
        "resizeTtyControl": {
            "id": "docker_resize_exec_tty",
            "nodeId": "6229d32d614b8f7bc0ee51791ddf197de5a9d3c950a31e441082ab9528855f2c;<container>",
            "probeId": "392656d14781da1e"
        }
    },
    "gridSortedBy": "label",
    "gridSortedDesc": false,
    "nodeDetails": [
        {
            "id": "6229d32d614b8f7bc0ee51791ddf197de5a9d3c950a31e441082ab9528855f2c;<container>",
            "label": "weavescope-app",
            "topologyId": "containers"
        }
    ],
    "pinnedMetricType": null,
    "pinnedSearches": [],
    "searchQuery": "",
    "selectedNodeId": "6229d32d614b8f7bc0ee51791ddf197de5a9d3c950a31e441082ab9528855f2c;<container>",
    "topologyId": "containers",
    "topologyOptions": {
        "containers": {
            "container-label-filters-group": "all",
            "pseudo": "hide",
            "stopped": "both",
            "system": "both"
        },
        "processes": {
            "unconnected": "hide"
        }
    },
    "topologyViewMode": "topo"
}
```

But actually its not the worst thing in the world, we might need that info in the future. We have another ticket #1727 for tightening up the url state.

LGTM!

PR #2094, opened 2016-12-15
Fix force-push to master

foot:
does it?


2opremio:
Here's what happened for future reference.

@fbarl force-pushed to bbcf184031a8e5047a45d3287440c938719e31c9 when master's tip was at 5ed63de . That caused PRs #2082 ,  #2086 and #2060 to be lost.

We used CircleCI to reconstruct what happened (since it build on every commit to master).

This PR fixes it.

Issue #1991, opened 2016-11-07
Conntrack XML parsing is consuming a lot of CPU


2opremio:
See https://github.com/weaveworks/scope/pull/1985#issuecomment-258826860

2opremio:
@alban Suggests replacing the conntrack command (and in turn parsing it's XML output) by talking to the kernel directly through a Netlink socket.

2opremio:
Found https://github.com/typetypetype/conntrack but the status is not really reassuring ...

2opremio:
Using netlink is going to be a bit more work than I originally expected due to the incomplete Golang support and my almost total ignorace about netlink until yesterday.

* https://github.com/vishvananda/netlink doens't support Conntrack directly (I created https://github.com/vishvananda/netlink/issues/171)
* https://github.com/typetypetype/conntrack/ is in dubious state (I tested it and seems to work though) and doesn't fully support flows either. There's [`Follow()`](https://github.com/typetypetype/conntrack/blob/72c92241b390ad6ce8ca1078268153537901a952/client.go#L103) which reads flows into [Conn](https://github.com/typetypetype/conntrack/blob/master/client.go#L174-L182). However `Conn` [lacks the `Orig` side of the flow](https://github.com/typetypetype/conntrack/blob/72c92241b390ad6ce8ca1078268153537901a952/client.go#L226-L230) and there is no simple or public way to send commands to conntrack.

My plan is to fork https://github.com/typetypetype/conntrack/ to reuse all the parsing but it may be more sensible to bite the bullet and implement https://github.com/vishvananda/netlink/issues/171 (which I think would take a much longer time). @awh Thoughts?

2opremio:
After an offline discussion with @awh I've confirmed that implementing a netlink protocol is a non-trivial task and that we should instead try improving parsing first. I will proceed in the following order, reevaluating performance after each step:

* Try to find a better performing XML library
* Try parsing the textual output of conntrack (i.e. without `-o xml`)
* Use netlink

2opremio:
> Try to find a better performing XML library

I have failed to find an alternative XML library at all. Moving on to parse the conntrack text manually.

tomwilkie:
Huh, I thought codec did it, but it doesn't.

https://godoc.org/github.com/ugorji/go/codec

2opremio:
Parsing the line-based output of conntrack is a pain in the ass because some (at least 4) fields are optional. This means I cannot simply use Scanf. 

[...]

PR #1985, opened 2016-11-04
Release 1.0.0

2opremio:
@davkal @foot @ekimekim @bowenli @jpellizzari @tomwilkie @jml @abuehrle @pidster @errordeveloper This is a major release, please chip in by testing this release branch in the scenarios below and tick off what you tested afterwards (including your name at the end). 

**EDIT:** To make things  easier to test I have pushed the image `weaveworks/scope:release-1.0-cd1e06b`

Thanks!

- [x] Start a 2 node cluster and leave a UI connected, watch for performance regression (Done by @2opremio)
- [x] Run a Scope cluster on ECS (Done by @2opremio)
- [x] Run Scope on https://github.com/microservices-demo/microservices-demo/
- [x] Run Scope on Docker For Mac
- [x] Start and stop Scope in standalone mode
  - [x] Linux
  - [x] docker-machine
- [x] Connect probes from release candidate to Weave Cloud
  - [x] Local scope in probe-only mode connects to Weave Cloud and renders correctly
    - [x] dev
    - [x] prod
  - [x] Controls appear for containers and function correctly (Done by @2opremio)
    - [x] dev
    - [x] prod
  - [x] No new errors in logging or telemetry
    - [x] dev
    - [x] prod

## Things to think about

- Kubernetes Integration
- CPU and memory usage

PR #2095, opened 2016-12-15
Disable XML in conntrack parsing

2opremio:
This attempts to improve the performance problems from #1991 

Parsing the default output of conntrack instead of XML turns out to be more difficult than we thought since some fields are optional and context-based.

I put together this hacky non-optimal parser to see if it's better than XML. If there's a considerable difference I will improve it and write a proper parser, otherwise I may just bite the bullet and fix https://github.com/vishvananda/netlink/issues/171

### DIFF
--- a/probe/endpoint/conntrack.go
+++ b/probe/endpoint/conntrack.go
@@ -54,10 +54,9 @@ type meta struct {
 
 type flow struct {
 	XMLName xml.Name `xml:"flow"`
-	Metas   []meta   `xml:"meta"`
 	Type    string   `xml:"type,attr"`
 
-	Original, Reply, Independent *meta `xml:"-"`
+	Original, Reply, Independent meta `xml:"-"`
 }
 
 type conntrack struct {
@@ -220,17 +219,6 @@ func (c *conntrackWalker) run() {
 	}
 }
 
-func makeEmptyFlow() flow {
-	var f flow
-	metas := make([]meta, 3)
-	f.Metas = metas
-	// TODO: do we really need the direction/protocol type when not using XML?
-	f.Original = &metas[0]
-	f.Reply = &metas[1]
-	f.Independent = &metas[2]
-	return f
-}
-
 func getUntaggedLine(reader *bufio.Reader) (string, error) {
 	// TODO: read bytes?
 	line, err := reader.ReadString('\n')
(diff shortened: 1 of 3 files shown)


2opremio:
After running this in WeaveCloud dev I can see a parsing improvement of 1-(1.74/1.01)*100 ~= 40%

This is a rough estimation and I should make further tests but it's promising.

Before (look for `xml.(*Decoder).unmarshal` taking 1.74s out of 30s)

before


After (look for `decodeStreamedFlow` taking 1.01s out of 30s)

after

2opremio:
I have made some more thorough testing: multiple runs, using a more loaded machine (similar to the one seen on https://github.com/weaveworks/scope/pull/1985#issuecomment-258826860) and making sure I always used probe pods running in the same machine. The results are even better than I thought:  1-(1.25/4.23) = 70% improvement.


Before:

before

After (focusing on `decodeStreamedFlow`):

after_streamed_2


So, I will clean this up, fix the tests and merge it.

probe/endpoint/conntrack.go
```diff
@@ -204,39 +201,113 @@ func (c *conntrackWalker) run() {
...
-		return
-	}
-
-	defer log.Infof("contrack exiting")
+	scanner := bufio.NewScanner(bufio.NewReader(stdout))
+	defer log.Infof("conntrack exiting")
 
-	// Now loop on the output stream
-	decoder := xml.NewDecoder(reader)
+	// Lop on the output stream
```
ekimekim: nit: typo `Lop`->`Loop`

2opremio:
> Have you double-checked this output against various verisons of the conntrack CLI to ensure it will work with older versions?

We control the version (i.e. the CLI included in the Scope container).

> Is this format guarenteed to not change in the future?

It hasn't changed in a long time. The last conntrack-tools release is from 2012 http://conntrack-tools.netfilter.org/

2opremio:
> It hasn't changed in a long time. The last conntrack-tools release is from 2012 http://conntrack-tools.netfilter.org/

That's a lie, http://www.netfilter.org/projects/conntrack-tools/downloads.html says otherwise. So, no guarantees but I would doubt it (it isn't very heavily changed https://git.netfilter.org/conntrack-tools/log/src/conntrack.c )

alban:
> > Have you double-checked this output against various verisons of the conntrack CLI to ensure it will work with older versions?
>
> We control the version (i.e. the CLI included in the Scope container).

@2opremio unfortunately, the output depends not only on the version of conntrack CLI but also on the kernel and on the sysctl options configured (see #2117)

Issue #1190, opened 2016-03-21
"Weave Scope is reachable at..." lies when app address is changed


rade:
```
$ scope version
Weave Scope version b5da2e4
$ scope launch --app.http.address=127.0.0.1:4041
abf9d4d730dec730a62bd11889116d256088b621bafe22630ff6a9fd66912d5e
Weave Scope is reachable at the following URL(s):
  * http://192.168.1.65:4040/
  * http://192.168.48.1:4040/
```

The output is a complete lie; the app is in fact reachable exactly where I told it.

PR #2096, opened 2016-12-16
prog/main: use flags.app.port for default target

schu:
User can control the webserver listen address with `-app.http.address`.
If they do, `scope --mode probe ...` should use the port part of address
in the default target and not ':4040'. Otherwise we encounter a
`dial tcp 127.0.0.1:4040: getsockopt: connection refused` error and
`scope-probe` cannot report.

Signed-off-by: Michael Schubert <michael@kinvolk.io>

### DIFF
--- a/prog/main.go
+++ b/prog/main.go
@@ -344,7 +344,7 @@ func main() {
 	flags.probe.noApp = *noApp || *probeOnly
 
 	// Special case for #1191, check listen address is well formed
-	_, _, err := net.SplitHostPort(flags.app.listen)
+	_, port, err := net.SplitHostPort(flags.app.listen)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


schu:
Not perfect as the `scope` launcher script always prints 4040 as the listening port (https://github.com/weaveworks/scope/blob/master/scope#L178-L180), but allows you to actually use a different `-app.http.address` locally.

2opremio:
Thanks @schu 

> Not perfect as the scope launcher script always prints 4040 as the listening port (https://github.com/weaveworks/scope/blob/master/scope#L178-L180), but allows you to actually use a different -app.http.address locally.

It wasn't perfect before either (i.e. you could already listen on an different address)

2opremio:
Related: https://github.com/weaveworks/scope/issues/1190

PR #2097, opened 2016-12-16
scope launcher script: allow setting SCOPE{_APP,}_CONTAINER_NAME

schu:
Helpful when you want to run multiple instances of scope on your system,
e.g. for testing.

Signed-off-by: Michael Schubert <michael@kinvolk.io>

### DIFF
--- a/scope
+++ b/scope
@@ -12,8 +12,11 @@ fi
 IMAGE_VERSION=${VERSION:-$IMAGE_VERSION}
 SCOPE_IMAGE_NAME=weaveworks/scope
 SCOPE_IMAGE="$SCOPE_IMAGE_NAME:$IMAGE_VERSION"
-SCOPE_CONTAINER_NAME=weavescope
-SCOPE_APP_CONTAINER_NAME=weavescope-app
+# Careful: it's easy to operate on (e.g. stop) the wrong scope instance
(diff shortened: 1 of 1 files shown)


2opremio:
I see how it can be useful but it's a bit dangerous due to the potential inconsistency between env var values in different runs.

I am fine with it as long as we don't document it externally.

scope
```diff
@@ -12,8 +12,8 @@ fi
 IMAGE_VERSION=${VERSION:-$IMAGE_VERSION}
 SCOPE_IMAGE_NAME=weaveworks/scope
 SCOPE_IMAGE="$SCOPE_IMAGE_NAME:$IMAGE_VERSION"
-SCOPE_CONTAINER_NAME=weavescope
-SCOPE_APP_CONTAINER_NAME=weavescope-app
+SCOPE_CONTAINER_NAME="${SCOPE_CONTAINER_NAME:-weavescope}"
```
2opremio: Can you add a brief comment on this and why it should be handled with care?

schu: Done, thanks.

schu:
Note: when ebpf will be enabled, it'll not be possible to start 2 scopes on the same machine because only one kprobe on a kernel function can be installed at a time. Then the vars only can be used to switch  between different images more easily. /cc @alban

PR #2100, opened 2016-12-19
Rename `Table` model to `PropertyList`

fbarl:
As a first step towards #1980, I renamed `Table` to `PropertyList` to clear the room for introducing a *real* generic multi-column table.


probe/overlay/weave.go
```diff
@@ -19,34 +19,34 @@ import (
...
+	WeaveDNSUpstream           = "weave_dns_upstream"
+	WeaveDNSTTL                = "weave_dns_ttl"
+	WeaveDNSEntryCount         = "weave_dns_entry_count"
+	WeaveProxyTableID          = "weave_proxy_table"
+	WeaveProxyStatus           = "weave_proxy_status"
+	WeaveProxyAddress          = "weave_proxy_address"
+	WeavePluginTableID         = "weave_plugin_table"
+	WeavePluginStatus          = "weave_plugin_status"
+	WeavePluginDriver          = "weave_plugin_driver"
+	WeaveConnectionsListPrefix = "weave_connections_list_"
```
2opremio: This will break backwards compatibility with old probes.

render/detailed/summary.go
```diff
@@ -45,19 +45,19 @@ type Column struct {
...
+	LabelMinor    string                `json:"labelMinor"`
+	Rank          string                `json:"rank"`
+	Shape         string                `json:"shape,omitempty"`
+	Stack         bool                  `json:"stack,omitempty"`
+	Linkable      bool                  `json:"linkable,omitempty"` // Whether this node can be linked-to
+	Pseudo        bool                  `json:"pseudo,omitempty"`
+	Metadata      []report.MetadataRow  `json:"metadata,omitempty"`
+	Parents       []Parent              `json:"parents,omitempty"`
+	Metrics       []report.MetricRow    `json:"metrics,omitempty"`
+	PropertyLists []report.PropertyList `json:"propertyLists,omitempty"`
```
2opremio: I don't think this will break backwards compatibility (IFAICR this is only used between app<->UI) but worth double checking.

report/topology.go
```diff
@@ -10,14 +10,14 @@ import (
...
-	MetricTemplates   `json:"metric_templates,omitempty"`
-	TableTemplates    `json:"table_templates,omitempty"`
+	Shape                 string `json:"shape,omitempty"`
+	Label                 string `json:"label,omitempty"`
+	LabelPlural           string `json:"label_plural,omitempty"`
+	Nodes                 `json:"nodes"`
+	Controls              `json:"controls,omitempty"`
+	MetadataTemplates     `json:"metadata_templates,omitempty"`
+	MetricTemplates       `json:"metric_templates,omitempty"`
+	PropertyListTemplates `json:"property_list_templates,omitempty"`
```
2opremio: This will break backwards compatiblity with the probes

fbarl: Same question here, do you think there is a nice alternative to just keeping the old name?

probe/overlay/weave.go
2opremio: I don't see another option (either that or not rename anything)

PR #2101, opened 2016-12-19
Forward OS/Kernel version to checkpoint

2opremio:
Useful to prioritize ebpf testing

Also:
* Make treatment of kernel release and version consistent across Darwin/Linux
* ~~Unify extraction of kernel version/release across Linux/Darwin~~ (Can't be done since OSX doesn't provide the `Uname` syscall)

cc @alban

### DIFF
--- a/probe/host/reporter.go
+++ b/probe/host/reporter.go
@@ -1,6 +1,7 @@
 package host
 
 import (
+	"fmt"
 	"net"
 	"runtime"
 	"sync"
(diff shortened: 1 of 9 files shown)


2opremio:
Related #2070

rade:
We do something like this in weave-net too; would be good to use the same labels and perhaps even code.

2opremio:
Yep, the kernel version label is the same as in
https://github.com/weaveworks/weave/pull/2372

I am not sure about sharing the code with Weave Net. I am reusing the code
in the host reporter.

PR #2105, opened 2016-12-22
Prioritize ecs topologies on initial load if available

### LINKED ISSUE #2042: ECS: prioritize Service/Task view if available
Leftover from #2026 

Just like in K8s, we should prioritize showing the ECS views if available

See https://github.com/weaveworks/scope/blob/master/client/app/scripts/utils/topology-utils.js

The ECS topologies are called `ecs-tasks` and `ecs-services` at the time of writing this.

foot:
Global topo order priority:
 - Check url for topoid
 - Check localStorage for topoid
 - Check topo-utils.js for what to show first!

Fixes #2042


foot:
For a report.json ping simon or fons

PR #2109, opened 2016-12-23
Add support for generic multicolumn tables

### LINKED ISSUE #1980: Add support for multicolumn tables from the probes

fbarl:
Resolves #1980. Prior to this story, we could only render *Node Tables* and *Labels* (renamed to *Property Lists* in this PR) in the node details panel. This PR introduces *Generic Table* as a separate UI component that enables us to render generic data in a multi-column format without associating rows to nodes.

#### Modeling

As @davkal reasoned, it makes sense to have *Property Lists* and *Generic Tables* as distinct components in the UI, so I made them fully independent there, but for backward compatibility reasons, I kept them under the old `Table` model on the backend, which I finally decided to extend like this:

```go
// Column is the type for multi-column tables in the UI.
type Column struct {
	ID       string `json:"id"`
	Label    string `json:"label"`
	DataType string `json:"dataType"`
}

// Row is the type that holds the table data for the UI. Entries map from column ID to cell value.
type Row struct {
	ID      string            `json:"id"`
	Entries map[string]string `json:"entries"`
}

// Table is the type for a table in the UI.
type Table struct {
	ID              string   `json:"id"`
	Label           string   `json:"label"`
	Type            string   `json:"type"`
	Columns         []Column `json:"columns"`
	Rows            []Row    `json:"rows"`
	TruncationCount int      `json:"truncationCount,omitempty"`
}

// TableTemplate describes how to render a table for the UI.
type TableTemplate struct {
	ID        string            `json:"id"`
	Label     string            `json:"label"`
	Prefix    string            `json:"prefix"`
	Type      string            `json:"type"`
	Columns   []Column          `json:"columns"`
	FixedRows map[string]string `json:"fixedRows"`
}
```

where `Columns` is used only with the `multicolumn-table` template type and `FixedRows` in only supported for `property-list`. The main reason I decided not to support fixed rows for generic (multicolumn) tables is that it wasn't clear what the exact use-case might be and how they would be rendered in the UI.

#### Storing

To be consistent with how property lists are being stored on the node, I decided to use the `node.Latest` map with table prefixes to store generic tables as well. The only difference is that since a multi-column table is two-dimensional, the table value is determined not only by *label*, but by (*row*, *column*) pair. Storing whole rows in the `Latest` map was not an option for two reasons:

1. `Latest` is a `string` to `string` map so we can only use it to store strings.
2. I also thought it would be better to use latest info on the cell level instead of row.

To be able to differ between the rows, each of them has to be given an ID (instead of what used to be a label). For backward compatibility with the old reports, property lists label is indeed used as a row ID when sending a table to the UI.

#### Rendering in the UI

*Generic table* is sortable and supports different column types (sent through `Column.DataType` field on the backend). Column headers are the same as for *Node Tables* (they are now in a new component called `NodeDetailsTableHeaders`), and column widths can also be hard-coded on the frontend.

screenshot from 2017-01-06 11-45-50

To see this, run `fixprobe` on [this report](https://github.com/weaveworks/scope/files/689695/report.json.zip)

#### Other Observations

* Right now we iterate once through the whole `node.Latest` map for every table (or a property list) we are extracting, instead of just running through the entries with a given prefix and extracting the fixed row information separately. I haven't looked into the implementation details of our `Map` structure, but maybe it wouldn't be that difficult to add an efficient `ForEachFromTo` kind of method there to run only on selected intervals. That would drop our time complexity of extraction roughly from `O(T * N)` to `O(N + T * log N)` (or `O(N * log N)`, if a lot of entries correspond to fixed rows), where `T` is the number of tables and `N` is the total size of `node.Latest`. Not sure if it's a bottleneck or if it's worth digging into this, but I just wanted to point it out.
* There might other components in `node-details` dir in the UI, like `NodeDetailsInfo`, which are both semantically and implementation-wise similar to `NodeDetailsPropertyList`, so consider refactoring them in some future PR.

### DIFF
--- a/report/table.go
+++ b/report/table.go
@@ -13,98 +13,135 @@ import (
 // MaxTableRows sets the limit on the table size to render
 // TODO: this won't be needed once we send reports incrementally
 const (
-	MaxTableRows          = 20
-	TruncationCountPrefix = "table_truncation_count_"
-	MulticolumnTableType  = "multicolumn-table"
-	PropertyListType      = "property-list"
+	MaxTableRows           = 20
+	TableEntryKeySeparator = "___"
+	TruncationCountPrefix  = "table_truncation_count_"
+	MulticolumnTableType   = "multicolumn-table"
+	PropertyListType       = "property-list"
 )
 
-// AddPrefixTable appends arbitrary rows to the Node, returning a new node.
-func (node Node) AddPrefixTable(prefix string, rows []Row) Node {
-	count := 0
+// WithTableTruncationInformation appends table truncation info to the node, returning the new node.
+func (node Node) WithTableTruncationInformation(prefix string, totalRowsCount int) Node {
+	if totalRowsCount > MaxTableRows {
+		truncationCount := fmt.Sprintf("%d", totalRowsCount-MaxTableRows)
+		node = node.WithLatest(TruncationCountPrefix+prefix, mtime.Now(), truncationCount)
+	}
+	return node
+}
+
+// AddPrefixMulticolumnTable appends arbitrary rows to the Node, returning a new node.
+func (node Node) AddPrefixMulticolumnTable(prefix string, rows []Row) Node {
+	addedRowsCount := 0
 	for _, row := range rows {
-		if count >= MaxTableRows {
+		if addedRowsCount >= MaxTableRows {
 			break
 		}
-		// TODO: Figure a more natural way of storing rows
-		for column, value := range row.Entries {
-			key := fmt.Sprintf("%s %s", row.ID, column)
+		// Add all the row values as separate entries
+		for columnID, value := range row.Entries {
+			key := strings.Join([]string{row.ID, columnID}, TableEntryKeySeparator)
 			node = node.WithLatest(prefix+key, mtime.Now(), value)
 		}
-		count++
+		addedRowsCount++
 	}
-	if len(rows) > MaxTableRows {
-		truncationCount := fmt.Sprintf("%d", len(rows)-MaxTableRows)
-		node = node.WithLatest(TruncationCountPrefix+prefix, mtime.Now(), truncationCount)
-	}
-	return node
+	return node.WithTableTruncationInformation(prefix, len(rows))
 }
 
-// AddPrefixLabels appends arbitrary key-value pairs to the Node, returning a new node.
-func (node Node) AddPrefixLabels(prefix string, labels map[string]string) Node {
-	count := 0
-	for key, value := range labels {
-		if count >= MaxTableRows {
+// AddPrefixPropertyList appends arbitrary key-value pairs to the Node, returning a new node.
+func (node Node) AddPrefixPropertyList(prefix string, propertyList map[string]string) Node {
+	addedPropertiesCount := 0
+	for label, value := range propertyList {
+		if addedPropertiesCount >= MaxTableRows {
 			break
 		}
-		node = node.WithLatest(prefix+key, mtime.Now(), value)
-		count++
+		node = node.WithLatest(prefix+label, mtime.Now(), value)
+		addedPropertiesCount++
 	}
-	if len(labels) > MaxTableRows {
-		truncationCount := fmt.Sprintf("%d", len(labels)-MaxTableRows)
-		node = node.WithLatest(TruncationCountPrefix+prefix, mtime.Now(), truncationCount)
-	}
-	return node
+	return node.WithTableTruncationInformation(prefix, len(propertyList))
 }
 
-// ExtractTable returns the rows to build a table from this node
+// WithoutPrefix returns the string with trimmed prefix and a
+// boolean information of whether that prefix was really there.
+// NOTE: Consider moving this function to utilities.
+func WithoutPrefix(s string, prefix string) (string, bool) {
+	return strings.TrimPrefix(s, prefix), len(prefix) > 0 && strings.HasPrefix(s, prefix)
+}
+
+// ExtractMulticolumnTable returns the rows to build a multicolumn table from this node
+func (node Node) ExtractMulticolumnTable(template TableTemplate) (rows []Row) {
+	rowsMapByID := map[string]Row{}
+
+	// Itearate through the whole of our map to extract all the values with the key
+	// with the given prefix. Since multicolumn tables don't support fixed rows (yet),
+	// all the table values will be stored under the table prefix.
+	// NOTE: It would be nice to optimize this part by only iterating through the keys
+	// with the given prefix. If it is possible to traverse the keys in the Latest map
+	// in a sorted order, then having LowerBoundEntry(key) and UpperBoundEntry(key)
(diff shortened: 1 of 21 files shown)


report/table.go
```diff
@@ -13,54 +13,186 @@ import (
...
+	for key, value := range r.Entries {
+		entriesCopy[key] = value
+	}
+	r.Entries = entriesCopy
+	return r
+}
+
 // Table is the type for a table in the UI.
 type Table struct {
-	ID              string        `json:"id"`
```
2opremio: Just confirming: This DS change doesn't affect backwards compatibility because it's only transferred between UI and app, right?

report/table.go
```diff
@@ -13,54 +13,186 @@ import (
...
+	}
+	return rows
+}
+
+// ExtractTable returns the rows to build either a property list or a generic table from this node
+func (node Node) ExtractTable(template TableTemplate) (rows []Row, truncationCount int) {
+	switch template.Type {
+	case MulticolumnTableType:
+		rows = node.ExtractMulticolumnTable(template)
+	default: // By default assume it's a property list (for backward compatibility).
```
2opremio: 👍

2opremio:
@fbarl Really clean design and code.

Did you test with an old probe, to check there are no backward compatibility issues?

2opremio:
> To be consistent with how property lists are being stored on the node, I decided to use the node.Latest map with table prefixes to store generic tables as well

I've always thought that using the `Latest` map to store the metadata of tables was a bit of a hack (particularly for data only used in tables). However now that we are married (for backwards compatibility) I guess we can't cleanly store the metadata anywhere else.

report/table.go
```diff
@@ -13,54 +13,186 @@ import (
...
+
+	// Gather a sorted list of rows' IDs.
+	ids := make([]string, 0, len(rowsByID))
+	for id := range rowsByID {
+		ids = append(ids, id)
+	}
+	sort.Strings(ids)
+
+	// Return the rows in the sorted order.
+	rows = []Row{}
```
2opremio: You can preallocate the rows with make since you know their length.

report/table.go
```diff
@@ -13,54 +13,186 @@ import (
...
+			// At this point, the row with that ID always exists, so we just update the value.
+			rowsByID[rowID].Entries[columnID] = value
 		}
-		if len(template.Prefix) > 0 && strings.HasPrefix(key, template.Prefix) {
-			label := key[len(template.Prefix):]
-			rows[label] = value
+	})
+
+	// Gather a sorted list of rows' IDs.
+	ids := make([]string, 0, len(rowsByID))
```
2opremio: Alternatively you can do `make([]string, len(rowsByID)` and assign them by key.

2opremio: I find it simpler, but it's a matter of taste, so up to you.

report/table.go
```diff
@@ -13,54 +13,186 @@ import (
...
+	for id := range rowsByID {
+		ids = append(ids, id)
+	}
+	sort.Strings(ids)
+
+	// Return the rows in the sorted order.
```
2opremio: Can't you implement `sort.Less` and `sort.Swap` for Rows and use `sort.Sort` directly?  (It would be cleaner)

report/table.go
```diff
@@ -13,54 +13,186 @@ import (
...
+	// with the given prefix as well as the keys corresponding to the fixed table rows.
+	node.Latest.ForEach(func(key string, _ time.Time, value string) {
+		if label, ok := template.FixedRows[key]; ok {
+			valuesByLabel[label] = value
+		} else if label, ok := WithoutPrefix(key, template.Prefix); ok {
+			valuesByLabel[label] = value
 		}
 	})
+
+	// Gather a sorted list of labels.
```
2opremio: Same here, can't you use `sort.Less` and `sort.Swap`?

rade:
Does this PR happen to address #1735 too?

report/table.go
fbarl: In this particular case it seems easier to me to keep it as it is, since I don't know how to iterate through the hash with `for .. range` with index information. In general, `make([]string, 0, len(rowsByID))` seemed to be a more common pattern around the code I was changing, so I kept it like that everywhere for consistency, even though I'm not particularly biased towards it :)

report/table.go
fbarl: Nice catch!

report/table.go
fbarl: Thanks for the tip, it does look cleaner indeed!

fbarl:
@rade Not really. All the column widths are still hardcoded on the frontend, I just moved that code to a different file :)

report/table.go
fbarl: Yup, sorting the rows at the end made it possible to remove one `for` loop from the code, so it also looks nicer that way.

fbarl:
@fons,

> Did you test with an old probe, to verify there are no backwards compatibility issues?

I tested it on a manually modified report that I linked to above, in my PR description. Out of the three nodes visible in Weave Net, one of them has everything stored in the new format, one of them in the old, and one of them both (even though that case will never happen I think). I didn't know how to test this data dynamically (without loading it from static reports), so maybe we should do that additional check somehow before pushing. What do you think?

> However, now that we are married (for backwards compatibility) I guess we can't cleanly store the metadata anywhere else.

While I was working on this story, I was gradually finding out how big of a constraint backwards compatibility is. Do you have any ideas how we could come around this in the future?

report/table.go
fbarl: Yup, the only field that has changed here is `Rows` and they only get populated through `ExtractTable` that I kept compatible with the old reports. After that, `Table`s get aggregated up and are indirectly only used in `baseNodeSummary` in `render/detailed/summary.go` which I guess only serves the UI?

2opremio:
> so maybe we should do that additional check somehow before pushing. What do you think?

Your test sounds reasonable, but I would also run a standalone probe with an older version (e.g. Scope 1.1.0) to make sure everything renders as it should.

> Do you have any ideas how we could come around this in the future?

Not really (without content negotiation or versioned APIs)

2opremio:
Code looks good, waiting on further backwards compatibility testing to approve.

2opremio:
@fbarl Would you mind (at least) squashing the commits?

fbarl:
> Code looks good, waiting on further backwards compatibility testing to approve.

I tested it with a standalone Scope 1.1.0 as you suggested @fons and good that I did because a small change on the backend still needed to be made for backwards compatibility with old probes to work properly. I also tested it with the new probe and with mixed reports after that, but I found no problems there. So as far as I could tell, everything should be working fine after merging this PR!

2opremio:
Great job @fbarl . LGTM!

PR #2114, opened 2017-01-03
Report persistent connections in states other than ESTABLISHED

### LINKED ISSUE #1972: probe's procspy does not report netcat's half-duplex long-lived connections
The reproducer below creates a container in Kubernetes and it creates two long-lived processes:
- A listening process "nc"
- A connecting process "nc"

The connection and processes are long-lived. But still, the connection is not displayed in the Scope UI. When I inspect /proc manually, I see that the 2 processes exist and are connected to each other.

I am using a Fedora based image to reproduce this bug because I couldn't get "nc" to work correctly in busybox (in busybox, "nc" is a built-in of the shell, and it does not seem to fork a new process when invoked).

```
apiVersion: v1
kind: Pod
metadata:
  name: connectlo
  labels:
    name: connectlo
spec:
  containers:
  - name: connectlo
    image: albanc/toolbox
    args:
      - '/bin/sh'
      - '-c'
      - "MYIP=$(ip addr show eth0 | grep \"inet\\b\" | awk '{print $2}' | cut -d/ -f1) ; echo MYIP=$MYIP ; (/bin/nc -l --keep-open -p 8081 & ) ; sleep 1 ; echo CLIENT ; (echo Hello ; sleep 1000 ; echo 'Bye' ) | /bin/nc $MYIP 8081"
```

I also tried with 127.0.0.1 instead of the eth0's IP address, but it also does not work.

This issue was discussed on https://github.com/weaveworks/scope/pull/1967#issuecomment-257624674. See also https://github.com/weaveworks/scope/pull/1780.

I have not checked the json report.

/cc @2opremio

2opremio:
This aligns the `/proc` connection tracking (persistent connections) with
conntrack (short-lived connections).

Fixes #1972 

EDIT: Also, make `/proc` parsing a bit more robust

### DIFF
--- a/probe/endpoint/procspy/procnet.go
+++ b/probe/endpoint/procspy/procnet.go
@@ -35,7 +35,7 @@ again:
 	b := p.b
 
 	var (
-		sl, local, remote, inode []byte
+		sl, local, remote, state, inode []byte
 	)
 
 	sl, b = nextField(b) // 'sl' column
@@ -46,7 +46,14 @@ again:
 	}
 	local, b = nextField(b)
 	remote, b = nextField(b)
-	_, b = nextField(b) // 'st' column
+	state, b = nextField(b)
+	switch parseHex(state) {
+	// Only process established or half-closed connections
+	case tcpEstablished, tcpFinWait1, tcpFinWait2, tcpCloseWait:
+	default:
+		p.b = nextLine(b)
+		goto again
+	}
 	_, b = nextField(b) // 'tx_queue' column
 	_, b = nextField(b) // 'rx_queue' column
 	_, b = nextField(b) // 'tr' column
--- a/probe/endpoint/procspy/procnet.go
+++ b/probe/endpoint/procspy/procnet.go
@@ -5,6 +5,9 @@ import (
 	"net"
 )
 
+// Used to check whether we are parsing a header line
+var slHeader = []byte("sl")
+
 // ProcNet is an iterator to parse /proc/net/tcp{,6} files.
 type ProcNet struct {
 	b                       []byte
@@ -31,16 +34,16 @@ again:
 	}
 	b := p.b
 
-	if p.b[2] == 's' {
+	var (
+		sl, local, remote, inode []byte
+	)
+
+	sl, b = nextField(b) // 'sl' column
+	if bytes.Equal(sl, slHeader) {
 		// Skip header
 		p.b = nextLine(b)
 		goto again
 	}
-
-	var (
-		local, remote, inode []byte
-	)
-	_, b = nextField(b) // 'sl' column
 	local, b = nextField(b)
 	remote, b = nextField(b)
 	_, b = nextField(b) // 'st' column
--- a/probe/endpoint/procspy/procnet.go
+++ b/probe/endpoint/procspy/procnet.go
@@ -9,17 +9,15 @@ import (
 type ProcNet struct {
 	b                       []byte
 	c                       Connection
-	wantedState             uint
 	bytesLocal, bytesRemote [16]byte
 	seen                    map[uint64]struct{}
 }
 
 // NewProcNet gives a new ProcNet parser.
-func NewProcNet(b []byte, wantedState uint) *ProcNet {
+func NewProcNet(b []byte) *ProcNet {
 	return &ProcNet{
 		b:           b,
 		c:           Connection{},
-		wantedState: wantedState,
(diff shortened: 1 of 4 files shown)


2opremio:
@alban Please confirm whether this fixes the bug since I cannot reproduce following your instructions.

2opremio:
Uhm, I still need to look into a failing test.

probe/endpoint/procspy/procnet.go
```diff
@@ -5,22 +5,23 @@ import (
 	"net"
 )
 
+// Used to check whether we are parsing a header line
+var slHeader = []byte("sl")
```
ekimekim: this would be better as `const`

2opremio: Unfortunately `[]byte("sl")` isn't a constant in go.

2opremio:
> I'm a little concerned about the implications of this behaviour. Connections can sometimes hang around for a long time in weird states.

I was too, until I realized we admit all states coming from conntrack for short-lived connections and it hasn't explicitly caused any problems I know of.

We do [treat TIME_WAIT specially](https://github.com/weaveworks/scope/blob/64f1a5d0f5f25a755c88f581bd427c30b952bb3a/probe/endpoint/conntrack.go#L414-L419) but it seems to be a detail due to the streaming nature of conntrack and it doesn't apply here (it would be good to get confirmation from @tomwilkie since he originally wrote that part of the code).

> I'd want to review all possible connection states and whether we explicitly want ALL of them to actually result in a "connection" being shown. 

Here are all the posible states: http://lxr.free-electrons.com/source/include/net/tcp_states.h

It seems sensible to ignore `TCP_CLOSE` but I am not sure what the impact would be in practice.

> I'd also like to see how this branch looks in scope view on a system under load to confirm it does what we want.

I've done that in a limited way but I think it would make more sense to ask our networking whizzes for tcp states we may not want to show in Scope. @awh @bboreham @brb thoughts?

rade:
I too am surprised that we care about states other than ESTABLISHED, and that this doesn't completely mess up the graphs.

2opremio:
> I too am surprised that we care about states other than ESTABLISHED, and that this doesn't completely mess up the graphs.

We do for half-closed connections, as per the example in https://github.com/weaveworks/scope/issues/1972#issuecomment-270103892 (which you surely have seen)

rade:
> We do, for half-closed connections

Do we really?

rade:
> communication was ongoing between the processes

Right. We do care about CLOSE_WAIT, since apps can still send data in that state. But I am pretty sure ESTABLISHED and CLOSE_WAIT are the only states in which that can happen.

2opremio:
> Right. We do care about CLOSE_WAIT, since apps can still send data in that state. But I am pretty sure ESTABLISHED and CLOSE_WAIT are the only states in which that can happen.

I will give that a try.

bboreham:
I don't think this is the correct fix for #1972.  If the connection is getting half-closed, then that is the problem you need to fix.

I do not think, from an application perspective, connections in CLOSE_WAIT, FIN_WAIT, etc., are of interest.  One side has closed the connection; it does not care what the other side is still trying to do; the connection is finished.

2opremio:
> If the connection is getting half-closed, then that is the problem you need to fix.

It's not a problem in this case since the (admittedly simplistic) example is till operating as expected from the user's perspective.

> the connection is finished.

I disagree, unidirectional communication is happening, which may be intentional (or irrelevant) for a particular use-case.

I agree on the fact that tracking this connections won't be useful in the general case though and we may decide to ignore the nc toy-example.

Not strong opinions.

2opremio:
> > Right. We do care about CLOSE_WAIT, since apps can still send data in that state. But I am pretty sure ESTABLISHED and CLOSE_WAIT are the only states in which that can happen.

> I will give that a try.

This didn't work (at least not out of the box) since not reporting the `FIN_WAIT2` side of the connection caused its corresponding PID to be missing.

In order to draw an edge for persistent connections (and particularly for loopback connections whose addresses are not unique across net namespaces) we need the PID of both processes. 

Each PID is included in each connection side, when means we need report both connection sides in order to be able to draw the edge.

2opremio:
I have limited the tracking to ESTABLISHED, FIN_WAIT_1, FIN_WAIT_2 and CLOSE_WAIT as prompted by https://github.com/weaveworks/scope/issues/1972#issuecomment-270933092

@ekimekim PTAL

Issue #1738, opened 2016-08-01
Table-mode: maintain focus while table is changing


rade:
In table mode, when the table rows change due to sort order, and rows being added/removed, it can be difficult/impossible to focus on a node.

Right now the worst than can happen as a result is that the wrong details panel is brought up. But once we have #1736 bad things can happen.

PR #2115, opened 2017-01-06
Maintain focus on hovered node table rows

fbarl:
Fix for #1738. On hovering over a node in the table view, the row freezes its position until it loses focus. We only do that for the hovered row (and not for the selected one), with a short debounce to prevent excessive re-rendering. We keep the removed node's row in the table as long as it's focused so that the row would have a real freeze effect.

I think the best way to test this PR is by setting short lived nodes in the debug toolbar. To increase the table dynamics, I suggest temporarily dropping the constant in `debug-toolbar.js:232` to `100`.

### DIFF
--- a/client/app/scripts/components/node-details/node-details-table.js
+++ b/client/app/scripts/components/node-details/node-details-table.js
@@ -1,4 +1,3 @@
-import debug from 'debug';
 import React from 'react';
 import classNames from 'classnames';
 import { find, get, union, sortBy, groupBy, concat, debounce, findIndex } from 'lodash';
@@ -117,19 +116,32 @@ function getSortedNodes(nodes, sortedByHeader, sortedDesc) {
 }
 
 
+// By inserting this fake invisible row into the table, with the help of
+// some CSS trickery, we make the inner scrollable content of the table
+// have a minimal height. That prevents auto-scroll under a focus if the
+// number of table rows shrinks.
+function minHeightConstraint(height = 0) {
+  return <tr className="min-height-constraint" style={{height}} />;
+}
+
+
 export default class NodeDetailsTable extends React.Component {
 
   constructor(props, context) {
     super(props, context);
+
     this.state = {
       limit: props.limit || NODE_DETAILS_DATA_ROWS_DEFAULT_LIMIT,
       sortedDesc: this.props.sortedDesc,
       sortedBy: this.props.sortedBy
     };
+    this.focusState = {};
+
     this.updateSorted = this.updateSorted.bind(this);
     this.handleLimitClick = this.handleLimitClick.bind(this);
     this.onMouseLeaveRow = this.onMouseLeaveRow.bind(this);
     this.onMouseEnterRow = this.onMouseEnterRow.bind(this);
+    this.saveTableContentRef = this.saveTableContentRef.bind(this);
     // Use debouncing to prevent event flooding when e.g. crossing fast with mouse cursor
     // over the whole table. That would be expensive as each focus causes table to rerender.
     this.debouncedFocusRow = debounce(this.focusRow, TABLE_ROW_FOCUS_DEBOUNCE_INTERVAL);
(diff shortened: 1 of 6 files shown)


client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -1,12 +1,17 @@
...
+import { find, get, union, sortBy, groupBy, concat, debounce, findIndex } from 'lodash';
 
 import ShowMore from '../show-more';
 import NodeDetailsTableRow from './node-details-table-row';
 import { ipToPaddedString } from '../../utils/string-utils';
+import { moveElement, insertElement } from '../../utils/array-utils';
+import { TABLE_ROW_FOCUS_DEBOUNCE_INTERVAL } from '../../constants/timer';
 
 
+const log = debug('scope:table');
```
davkal: This should be the name of the component, i.e. `scope:node-details-table`

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -208,6 +219,34 @@ export default class NodeDetailsTable extends React.Component {
     this.setState({limit});
   }
 
+  focusRow(rowIndex, node) {
+    this.setState({
+      focusedRowIndex: rowIndex,
+      focusedNode: node
+    });
+    log(`Focused row ${rowIndex}`);
```
davkal: Is that still needed?

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -208,6 +219,34 @@ export default class NodeDetailsTable extends React.Component {
...
+  }
+
+  unfocusRow() {
+    if (this.state.focusedRowIndex) {
+      this.setState({
+        focusedRowIndex: null,
+        focusedNode: null
+      });
+      log('Unfocused row');
```
davkal: Still needed?

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -269,6 +308,22 @@ export default class NodeDetailsTable extends React.Component {
       defaultSortDesc(sortedByHeader);
 
     let nodes = getSortedNodes(this.props.nodes, sortedByHeader, sortedDesc);
+    if (focusedRowIndex && focusedRowIndex < nodes.length) {
+      const nodeRowIndex = findIndex(nodes, node => node.id === focusedNode.id);
+      if (nodeRowIndex >= 0) {
+        // If the focused node still exists in the table, we move it
+        // to the hovered row, keeping the rest of the table sorted.
+        moveElement(nodes, nodeRowIndex, focusedRowIndex);
```
davkal: Let's not mutate `nodes`, but rather return a new copy.

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -208,6 +219,34 @@ export default class NodeDetailsTable extends React.Component {
...
+      this.setState({
+        focusedRowIndex: null,
+        focusedNode: null
+      });
+      log('Unfocused row');
+    }
+  }
+
+  onMouseEnterRow(rowIndex, node) {
+    this.debouncedUnfocusRow.cancel();
```
davkal: That `cancel` could use a comment.

fbarl:
@davkal and I discovered that row focusing was actually not working under some special conditions so I'm still working on this trying to cover all the cases.

fbarl:
@davkal This slightly hacky solution should now work for all cases. PTAL.

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -188,10 +261,11 @@ export default class NodeDetailsTable extends React.Component {
                   colStyles={styles}
                   columns={columns}
                   onClick={onClickRow}
-                  onMouseLeaveRow={onMouseLeaveRow}
-                  onMouseEnterRow={onMouseEnterRow}
+                  onMouseEnter={() => { this.onMouseEnterRow(index, node); }}
```
davkal: Best practice here would be to pass the index and node as props to the row, and then call the props.onMouseEnter from the row with props.node and props.index. Otherwise a new closure is generated on every row render.

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -1,13 +1,15 @@
 import React from 'react';
 import classNames from 'classnames';
-import { find, get, union, sortBy, groupBy, concat } from 'lodash';
+import { find, get, union, sortBy, groupBy, concat, debounce, findIndex } from 'lodash';
```
davkal: `findIndex` is already polyfilled by `babel-polyfill`

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -114,17 +116,36 @@ function getSortedNodes(nodes, sortedByHeader, sortedDesc) {
...
+
     this.updateSorted = this.updateSorted.bind(this);
+    this.handleLimitClick = this.handleLimitClick.bind(this);
+    this.onMouseLeaveRow = this.onMouseLeaveRow.bind(this);
+    this.onMouseEnterRow = this.onMouseEnterRow.bind(this);
+    this.saveTableContentRef = this.saveTableContentRef.bind(this);
+    // Use debouncing to prevent event flooding when e.g. crossing fast with mouse cursor
+    // over the whole table. That would be expensive as each focus causes table to rerender.
+    this.debouncedFocusRow = debounce(this.focusRow, TABLE_ROW_FOCUS_DEBOUNCE_INTERVAL);
+    this.debouncedUnfocusRow = debounce(this.unfocusRow, TABLE_ROW_FOCUS_DEBOUNCE_INTERVAL);
```
davkal: I prefer `blur` as the antonym to `focus`.

Issue #1188, opened 2016-03-21
High node count layouts take too long


davkal:
The lifted node limit in #1186 leads to a new problem as found by @foot:

> 998 new nodes took too long to layout on the highly connected data the debugbar generates. (2-3mins then hit pause and saw it stuck in the layout code). 500nodes took about 20s to layout. Maybe set the limit to 300-500 for now.

Ideas are: (a) new limit, (b) simpler layouts for high node counts, (c) run layouter in the background.

pidster:
See also: https://github.com/weaveworks/scope/issues/1614

fbarl:
What shall we do with this issue now that #2128 has been merged? I feel it's too broad to be closed just yet, but maybe we could at least remove it from the milestone?

PR #2119, opened 2017-01-09
[WIP] New layout for high nodes count

fbarl:


rade:
Is this meant to fix #1188?

fbarl:
@rade Yes, the current changes contain only the prototype of the new layout, but I was also gonna include some common optimizations that would affect both layouts. I will freeze this story, at least for now, and do the current layout optimizations in another branch.

Issue #2123, opened 2017-01-11
Incorrect route is used in getAllNodes() call


jpellizzari:
Moved from https://github.com/weaveworks/service/issues/1053

The leading slash in the `topologyUrlsById` is causing the incorrect route to be requested.

Repro steps:
1. Within cloud, enter some search terms in scope.
2. `JSON.parse()` errors will appear in the console, as a result of a call to /api/topology/containers
3. Notice that the cloud index.html file is served instead of a JSON response.

The request to
https://cloud.weave.works/api/topology/containers
should be
https://cloud.weave.works/api/app/loud-breeze-77/api/topology/containers

rade:
What functionality, if any, does this break? Search seems to work fine in Weave Cloud.

jpellizzari:
Searching triggers a fetch of topologies which is throwing errors in the console. The search functionality itself seems to still work, since it does not rely on the result of the fetch. I'm not sure why searching needs to trigger that request, but its not working.

rade:
> I'm not sure why searching needs to trigger that request,

This is worth investigating.

jpellizzari:
@rade Found it!:

```javascript
    dispatch({ type: ActionTypes.FOCUS_SEARCH });
    // update nodes cache to allow search across all topologies,
    // wait a second until animation is over
    setTimeout(() => {
      getAllNodes(getState, dispatch);
    }, 1200);
```

We discussed a possible improvement in our cloud-prod sync today. I documented it here: https://github.com/weaveworks/service-conf/issues/500

jpellizzari:
Closed by #2124

PR #2124, opened 2017-01-11
Removed leading slash from getAllNodes() api request

jpellizzari:
Fix for #2123 

Removes a leading slash that was causing requests to api/topology/containers to return the index.html file instead of the JSON response. 

@foot Could you review please? This feels like it should be fixed in the output from the back-end, but I'm worried that other stuff (including other services) might be relying on that extra slash being there.


davkal:
For now API urls need to be relative when handled internally.

The backend still returns absolute paths in case someone wants to use Scope's API directly.

LGTM

Issue #2120, opened 2017-01-10
default view fails to select 'application containers only'


rade:
Steps to reproduce:
1. run some docker container, e.g. `docker run -d -ti alpine /bin/sh`
2. fire up a scope
3. open a *private* browsing window and nagivate to localhost:4040

Expected result:
- scope select the default view & selection, which is the container view showing application containers only, and as a result only the single started container is shown
![screenshot_2017-01-10_22-33-24](https://cloud.githubusercontent.com/assets/109109/21828010/14089aa8-d786-11e6-9afb-4484b5c3a14b.png)

Actual result:
- scope shows the container view but with app/system container selection *blank*, as a result of which it shows *all* containers, i.e. the single container we launched plus the scope container.
![screenshot_2017-01-10_22-32-02](https://cloud.githubusercontent.com/assets/109109/21828014/199c9b54-d786-11e6-9d97-e2c70031dfb9.png)

This works as intended in Scope 1.0.0 and is broken in Scope 1.1.0 onwards.

rade:
I've added this to the current m/s since it rather badly affects the initial user experience.

jpellizzari:
Root cause appears to be that a topologies key was changed from "application" to "notsystem" in this commit:

https://github.com/weaveworks/scope/commit/9833a854b1638d8bd3270323c8aa0041016ebb98#diff-6126bda540b57df79f551488091760daR56

jpellizzari:
Merged PR. Closing.

PR #2125, opened 2017-01-12
Change topologies option from 'notsystem' to 'application'

jpellizzari:
Fix for #2120 

A bug was introduced here: https://github.com/weaveworks/scope/commit/9833a854b1638d8bd3270323c8aa0041016ebb98#diff-6126bda540b57df79f551488091760daR56

To test:
1. Click the 'bug' icon in the bottom left of the screen
2. Click the 'Reset your local view state' option
3. When the page reloads, the 'Application' filter should be selected.

### DIFF
--- a/integration/config.sh
+++ b/integration/config.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # NB only to be sourced
 
 set -e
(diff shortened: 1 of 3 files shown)


jpellizzari:
@ekimekim Can you review please?

jpellizzari:
@ekimekim Having some trouble getting integration tests to pass:

```
Launch scope and check it boots, with a spurious host arg
83ecbe3e7aeb8d55334bfd318fe86b45b9a7834afdce033592f2065e0697bf5c
Scope probe started
  * http://10.240.0.20:4040/
Failed to find nodes weavescope after 60 secs
test #1 "curl -s http://host2-6126-0.us-central1-a.scope-integration-tests:4040/api/topology/containers?system=show | jq -r '[.nodes[] | select(.label == "weavescope")] | length'" failed:
	expected "1"
	got "0"
1 of 1 tests failed in 66.873s.
```

That failure doesn't look related to the change I made in this PR. Any thoughts on what might be causing that?

ekimekim:
Hm. I haven't looked into it properly, but at a first glance, I think the following happened:
* Originally, option was system = `show` or `hide`. Test code above works.
* Backend is changed so option is `all`, `system` or `notsystem`. This causes the bug, and the default is still all. Therefore *because* of the bug, test code above still works.
* You fix bug. Test is now broken.

I think the fix is to change the argument to the curl url in that text to value `system`, but i'm not sure what the correct argument key is now? pretty sure it's no longer `system=`.

integration/config.sh
```diff
@@ -41,7 +41,7 @@ has() {
     local host=$2
     local name=$3
     local count=${4:-1}
-    assert "curl -s http://${host}:4040/api/topology/${view}?system=show | jq -r '[.nodes[] | select(.label == \"${name}\")] | length'" "$count"
+    assert "curl -s http://${host}:4040/api/topology/${view}? | jq -r '[.nodes[] | select(.label == \"${name}\")] | length'" "$count"
 }
```
jpellizzari: @2opremio This ended up getting my tests to pass on this branch, but did I remove the purpose of the test?

ekimekim: Hm. Hard to say without carefully reading the whole thing, but I suspect the desired behaviour here is to "show all". So you should pass in the value corresponding to that. If that's the default, what you have here is fine (though nit: if you have no params, just omit the `?` entirely)

jpellizzari:
@2opremio That change to from `s/containerLabelFiltersGroupID = "container-label-filters-group"/systemGroupID = "system"/` looks like it did the trick. Thanks!

PR #2132, opened 2017-01-16
Obtain local pods from kubelet

### LINKED ISSUE #2049: kubernetes reporter is broken on katacoda

2opremio:
Fixes #2049

Supersedes #2122 

@paulbellamy PTAL

### DIFF
--- a/probe/kubernetes/kubelet_test.go
+++ b/probe/kubernetes/kubelet_test.go
@@ -1,0 +1,73 @@
+package kubernetes_test
+
+import (
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/weaveworks/scope/probe/kubernetes"
+)
+
+const kubeletPodsJSONFile = "kubelet_pods.json"
+
+// obtained with jq .items[].metadata.uid < kubelet_pods.json
+var expectedPodUIDs = []string{
+	"af1b5325-d8cf-11e6-84fa-0800278a0c83",
+	"afeb8017-d8cf-11e6-84fa-0800278a0c83",
+	"b57a1ada-d8cf-11e6-84fa-0800278a0c83",
+	"00915ccb-dbd5-11e6-84fa-0800278a0c83",
+	"009a3d10-dbd5-11e6-84fa-0800278a0c83",
+	"fa8320c0-d8b0-11e6-828c-0800278a0c83",
+	"fa95ebae-d8b0-11e6-828c-0800278a0c83",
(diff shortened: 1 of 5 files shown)


probe/kubernetes/kubelet_test.go
```diff
@@ -0,0 +1,73 @@
+package kubernetes_test
```
rade: Why are we putting tests in a different package to the code being tested, and hence have to export functions that otherwise would be local? I don't think that's The Go Way.

2opremio: To be honest, I don't know why and when we started doing this. @paulbellamy any idea?

2opremio: Here's why https://github.com/weaveworks/scope/commit/2b7662a3c6a7cb1b77908e9d169b3f44c651e4d3#commitcomment-20502161

2opremio: I will make the symbols private after merging https://github.com/weaveworks/scope/pull/2136 (which contains https://github.com/weaveworks/build-tools/pull/64 )

2opremio: I was hoping I would do this here, but making those symbols private imply making `reporter_rest.go` not import the kubernetes package, which implies a rather large refactoring.

PR #2134, opened 2017-01-17
Deploy to quay when merging to master

2opremio:
To workaround the 500s obtained from Docker Hub

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -77,6 +77,12 @@ deployment:
           docker push ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:$(./tools/image-tag) &&
           (test -z "${UI_BUCKET_KEY_ID}" || make ui-upload)
         )
+      - | 
+        test -z "${QUAY_USER}" || (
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -77,6 +77,12 @@ deployment:
           docker push ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:$(./tools/image-tag) &&
           (test -z "${UI_BUCKET_KEY_ID}" || make ui-upload)
         )
+      - | 
+        test -z "${QUAY_USER}" || (
+          docker docker login -e '.' -u "$QUAY_USER" -p "$QUAY_PASSWORD" quay.io &&
+          docker tag weaveworks/scope:$(./tools/image-tag) "quay.io/${QUAY_ORGANIZATION}/scope:$(./tools/image-tag)" &&
+          docker push "quay.io/${QUAY_ORGANIZATION}/scope:$(./tools/image-tag)"
```
alban: This will not push the default "latest" tag. Is it on purpose?

2opremio: It was on purpose, we are doing this to mitigate the 500s in WeaveCloud. However, if you have a use for it, I can also incorporate the `:latest` tag.

PR #2068, opened 2016-12-07
Fix linter errors

iaguis:
https://github.com/koalaman/shellcheck/wiki/SC2048

We get a https://github.com/koalaman/shellcheck/wiki/SC2016
but we do mean literal backquotes.

PR #2135, opened 2017-01-17
Add eBPF connection tracking without dependencies on kernel headers

### LINKED ISSUE #1168: walking /proc to obtain connections is very expensive

### LINKED ISSUE #1260: Short-lived connections not tracked for containers in shared networking namespaces

### LINKED ISSUE #1961: Remove runtime kernel header dependency from ebpf tracker

### LINKED ISSUE #1962: Port ebpf tracker to Go

iaguis:
Based on work from Lorenzo, updated by Iago, Alban, Alessandro and
Michael.

This PR adds connection tracking using eBPF. This feature is not enabled by default.
For now, you can enable it by launching scope with the following command:

```
```

This patch allows scope to get notified of every connection event,
without relying on the parsing of /proc/$pid/net/tcp{,6} and
/proc/$pid/fd/*, and therefore improve performance.

We vendor https://github.com/iovisor/gobpf in Scope to load the
pre-compiled ebpf program and https://github.com/kinvolk/tcptracer-bpf
to guess the offsets of the structures we need in the kernel. In this
way we don't need a different pre-compiled ebpf object file per kernel.
The Scope build fetches the pre-compiled ebpf program from
https://hub.docker.com/r/kinvolk/tcptracer-bpf/ (see
https://github.com/kinvolk/tcptracer-bpf). To update to a new version
you can modify the EBPF_IMAGE variable in Makefile.

The ebpf program uses kprobes/kretprobes on the following kernel functions:
- tcp_v4_connect
- tcp_v6_connect
- tcp_set_state
- inet_csk_accept
- tcp_close

connection tuple but also pid and netns.
Note: the IPv6 events are not supported in Scope and thus not passed on.

conntrack, it also keeps the dead connections for one iteration in order
to report short-lived connections.

state is now done in foreground instead of background, via
newForegroundReader().

NAT resolution on connections from eBPF works in the same way as it did
on connections from /proc: by using conntrack. One of the two conntrack
instances is only started to get the initial state and then it is
stopped since eBPF detects short-lived connections.

The Scope Docker image size comparison:
- weaveworks/scope in current master:  22 MB (compressed),  68 MB
  (uncompressed)
- weaveworks/scope with this patchset: 23 MB (compressed), 69 MB
  (uncompressed)

Fixes #1260 (Short-lived connections not tracked for containers in
shared networking namespaces)
Fixes #1962 (Port ebpf tracker to Go)
Fixes #1961 (Remove runtime kernel header dependency from ebpf tracker)

This PR is kernel version/configuration independent so it supersedes https://github.com/weaveworks/scope/pull/2070 which needed a different object file per kernel/kernel config.

Depends on https://github.com/weaveworks/scope/pull/2068

TODO:

* [ ] Gracefully fall back to proc parsing if the user enables ebpf with `--probe.ebpf.connections=true` and the ebpf loading fails for whatever reason. We can fix this in a followup PR
* [ ] tracking won't work if the server socket is on `accept()` at start time : https://github.com/weaveworks/tcptracer-bpf/issues/10


Makefile
```diff
@@ -41,7 +58,11 @@ docker/weave:
 	curl -L git.io/weave -o docker/weave
 	chmod u+x docker/weave
 
-$(SCOPE_EXPORT): $(SCOPE_EXE) $(DOCKER_DISTRIB) docker/weave $(RUNSVINIT) docker/Dockerfile docker/demo.json docker/run-app docker/run-probe docker/entrypoint.sh
+docker/ebpf.tgz: Makefile
+	$(SUDO) docker pull $(EBPF_IMAGE)
+	CONTAINER_ID=$(shell $(SUDO) docker run -d $(EBPF_IMAGE) /bin/false 2>/dev/null || true); $(SUDO) docker export -o docker/ebpf.tgz $${CONTAINER_ID}
```
2opremio: Why not simply export have `ebpf.tgz` as a release artifact of https://github.com/kinvolk/tcptracer-bpf ?

Also,

* I think https://github.com/kinvolk/tcptracer-bpf should be part of the weaveworks organization
* And its code should go through a review

( @rade , thoughts?)

Makefile
```diff
@@ -23,14 +23,31 @@ RM=--rm
 RUN_FLAGS=-ti
 BUILD_IN_CONTAINER=true
 GO_ENV=GOGC=off
-GO=env $(GO_ENV) go
-NO_CROSS_COMP=unset GOOS GOARCH
-GO_HOST=$(NO_CROSS_COMP); $(GO)
-WITH_GO_HOST_ENV=$(NO_CROSS_COMP); $(GO_ENV)
+GO_ENV_ARM=$(GO_ENV) CC=/usr/bin/arm-linux-gnueabihf-gcc
```
2opremio: Shouldn't CC be defined in a new line and part of the `ifeq` below? Also, where is it used? A comment may be in order.

Makefile
```diff
@@ -23,14 +23,31 @@ RM=--rm
...
 GO_BUILD_INSTALL_DEPS=-i
 GO_BUILD_TAGS='netgo unsafe'
 GO_BUILD_FLAGS=$(GO_BUILD_INSTALL_DEPS) -ldflags "-extldflags \"-static\" -X main.version=$(SCOPE_VERSION) -s -w" -tags $(GO_BUILD_TAGS)
+
+ifeq ($(GOOS),linux)
+GO_ENV+=CGO_ENABLED=1
+endif
+
+ifeq ($(GOARCH),arm)
+GO=env $(GO_ENV_ARM) go
```
2opremio: Why not define GO_ENV_ARM directly here? (or use it directly, it only seems to be used in a single place)

Makefile
```diff
@@ -23,14 +23,31 @@ RM=--rm
...
+# The version of go shipped on debian doesn't have some standard library
+# packages for arm and when it tries to install them it fails because it
```
2opremio: It used to work just fine, what packages does it fail with?

circle.yml
```diff
@@ -41,7 +41,7 @@ test:
         parallel: true
     - cd $SRCDIR; make RM= client-lint static:
         parallel: true
-    - cd $SRCDIR; rm -f prog/scope; if [ "$CIRCLE_NODE_INDEX" = "0" ]; then GOARCH=arm make GO_BUILD_INSTALL_DEPS= RM= prog/scope; else GOOS=darwin make GO_BUILD_INSTALL_DEPS= RM= prog/scope; fi:
+    - cd $SRCDIR; rm -f prog/scope; if [ "$CIRCLE_NODE_INDEX" = "0" ]; then GOARCH=arm GOOS=linux make GO_BUILD_INSTALL_DEPS= RM= prog/scope; else GOOS=darwin GOOS=linux make GO_BUILD_INSTALL_DEPS= RM= prog/scope; fi:
```
2opremio: Why do we need this is it due to the backend image change?

docker/Dockerfile
```diff
@@ -6,6 +6,7 @@ RUN echo "http://dl-cdn.alpinelinux.org/alpine/edge/community" >>/etc/apk/reposi
 	apk add --update bash runit conntrack-tools iproute2 util-linux curl && \
 	rm -rf /var/cache/apk/*
 ADD ./docker.tgz /
+ADD ./ebpf.tgz /usr/libexec/scope/
```
2opremio: Is it worth the pain of compressing the file? And .. is it many files? (i.e. it may make things simpler to simply use the plain ebpf object file)

probe/endpoint/conntrack.go
```diff
@@ -463,14 +463,14 @@ func (c *conntrackWalker) handleFlow(f flow, forceAdd bool) {
 
 // walkFlows calls f with all active flows and flows that have come and gone
 // since the last call to walkFlows
-func (c *conntrackWalker) walkFlows(f func(flow)) {
+func (c *conntrackWalker) walkFlows(f func(flow, bool)) {
```
2opremio: Can you please document the new `alive` parameter? (maybe even using that name explicitly in the parameter list if Go allows it).

Makefile
rade: agreed.

alban: > Why not simply export have ebpf.tgz as a release artifact of
> https://github.com/kinvolk/tcptracer-bpf ?

We could do that. But it is useful during development to generate a new image/artifact automatically for each branch / after each PR on tcptracer-bpf.

Who should review tcptracer-bpf?

Makefile
alban: @iaguis told me it is needed for cgo, which we started to use since we use iovisor/gobpf:

https://github.com/iovisor/gobpf/blob/a81bc05bf71ec17bf4c8d388621d2c81bc1f6a6b/elf/elf.go#L210

docker/Dockerfile
alban: It is only one file now (ebpf.o). It used to be one file per kernel version, but not anymore.

If we release the file as part of releases of tcptracer-bpf on github, is there a way to upload that file automatically after a build in the CI?

Makefile
iaguis: That should work, yep. It is used by cgo which is called by the go compiler.

Makefile
iaguis: makes sense

circle.yml
iaguis: I wanted to detect if the OS is linux in the Makefile to add `CGO_ENABLED` or not (I found some problems with cgo and darwin).

docker/Dockerfile
iaguis: it used to be many files but it's just one now, the object file should be enough

alban:
We have tested (manually) on Debian, Ubuntu, Fedora, Arch, CoreOS (beta), Amazon Linux.

Note that Amazon Linux does not have `/sys/kernel/debug` mounted by default. So it should be mounted manually before starting the Docker daemon:
```
$ sudo yum install docker
$ sudo mount -t debugfs debugfs /sys/kernel/debug
$ sudo /etc/init.d/docker start
```

We have not tested on GCE yet.

Makefile
2opremio: For the `goebpf` library, right?

Makefile
2opremio: Why due to cgo? It seems odd. I would appreciate more details on this.

Makefile
2opremio: >  But it is useful during development to generate a new image/artifact automatically for each branch / after each PR on tcptracer-bpf.

Right, but as you said, the same can be done with artifacts (CI can do that automatically)

> Who should review tcptracer-bpf?

I would say @brb / @awh (if they have the time) and myself. I have just created https://github.com/weaveworks/tcptracer-bpf , please create a PR there.

circle.yml
2opremio: But `GOOS=darwin GOOS=linux` doesn't make sense to me ...

docker/Dockerfile
2opremio: Great (I am fine with compressing it if it gets us somewhere, but I doubt it)

probe/endpoint/conntrack.go
2opremio: Also, can you either pick `alive` or `active` to make it consistent?

alepuccetti:
I tried on gc with the image gci but it does not work because bpf is not supported. Error message from strace:
`bpf(BPF_MAP_CREATE, {map_type=BPF_MAP_TYPE_PERF_EVENT_ARRAY, key_size=4, value_size=4, max_entries=1024}, 48) = -1 ENOSYS (Function not implemented)`

These are the information about the gci virtual machine image: 
```
#cat /etc/os-release 

BUILD_ID=8872.47.0
NAME="Container-VM Image"
GOOGLE_CRASH_ID=Lakitu
VERSION_ID=55
BUG_REPORT_URL=https://crbug.com/new
PRETTY_NAME="Google Container-VM Image"
VERSION=55
GOOGLE_METRICS_PRODUCT_ID=26
HOME_URL="https://cloud.google.com/compute/docs/containers/vm-image/"
ID=gci
```

2opremio:
> So it should be mounted manually before starting the Docker daemon

Is it worth asking them to mount it by default?

alban:
> @alban Is it worth asking them to mount it by default?

Ok, asked in that thread: https://forums.aws.amazon.com/thread.jspa?messageID=762683

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,256 @@
...
+	walkConnections(f func(ebpfConnection))
+	initialize()
+	isInitialized() bool
+	stop()
+}
+
+var ebpfTracker *EbpfTracker
+
+// nilTracker is a tracker that does nothing, and it implements the eventTracker interface.
+// It is returned when the useEbpfConn flag is false.
```
2opremio: Why not simply return `nil` and check for it? Or even better, create a common interface for the proc and ebpf trackers and use one or the other based on the configuration?

2opremio: That would also remove the sprinkled `if`s checking whether we are using ebpf or not.

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,256 @@
...
+	initialized bool
+	dead        bool
+
+	openConnections   map[string]ebpfConnection
+	closedConnections []ebpfConnection
+}
+
+func newEbpfTracker(useEbpfConn bool) eventTracker {
+	if !useEbpfConn {
+		return &nilTracker{}
```
2opremio: See comments above.

probe/endpoint/ebpf_linux.go
```diff
@@ -0,0 +1,7 @@
+//+build linux
```
2opremio: I don't think this is needed (the filename determines it should be only built in linux)

probe/endpoint/ebpf.go
2opremio: And it would help choosing the tracker based on the OS (for instance macs can only do proc walking)

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,256 @@
+package endpoint
```
2opremio: (minor) Ideally, this should only be compiled in linux. Not that we care too much about macs these days.

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,256 @@
...
+		log.Errorf("Cannot find BPF object file: %v", err)
+		return &nilTracker{}
+	}
+
+	bpfPerfEvent := bpflib.NewModule(bpfObjectFile)
+	if bpfPerfEvent == nil {
+		return &nilTracker{}
+	}
+	err = bpfPerfEvent.Load()
+	if err != nil {
```
2opremio: Please do `if err := bpfPerfEvent.Load(); err != nil` in a single block

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,256 @@
...
+		log.Errorf("Error loading BPF program: %v", err)
+		return &nilTracker{}
+	}
+
+	bpfPerfEvent.EnableKprobes()
```
2opremio: No error?

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,256 @@
...
+	case EventClose:
+		return "close"
+	default:
+		return "unknown"
+	}
+}
+
+// tcpEvent should be in sync with the struct in the ebpf maps.
+type tcpEvent struct {
+	// Timestamp must be the first field, the sorting depends on it
```
2opremio: Why is the declaration order of the timestamp field relevant? I hope it's not due to serialization constraints ...

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,256 @@
...
+	byteorder.Host.PutUint32(saddrbuf, uint32(event.SAddr))
+	byteorder.Host.PutUint32(daddrbuf, uint32(event.DAddr))
+
+	sIP := net.IPv4(saddrbuf[0], saddrbuf[1], saddrbuf[2], saddrbuf[3])
+	dIP := net.IPv4(daddrbuf[0], daddrbuf[1], daddrbuf[2], daddrbuf[3])
+
+	sport := event.SPort
+	dport := event.DPort
+
+	if typ.String() == "close" || typ.String() == "unknown" {
```
2opremio: Shouldn't this constants be defined at `tcptracer-bpf`?

2opremio: Actually, `String()` is defined here. Why convert to from the event to string? Why not use the enums?

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,256 @@
...
+
+	log "github.com/Sirupsen/logrus"
+	bpflib "github.com/iovisor/gobpf/elf"
+	"github.com/kinvolk/tcptracer-bpf/pkg/byteorder"
+	"github.com/kinvolk/tcptracer-bpf/pkg/offsetguess"
+)
+
+type eventType uint32
+
+// These constants should be in sync with the equivalent definitions in the ebpf program.
```
2opremio: Shouldn't this be defined in `tcptracer-bpf`?

probe/endpoint/ebpf.go
2opremio: Also, shouldn't this be fined in `tcptracer-bpf`?

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,256 @@
...
+		f(connection)
+	}
+	for _, connection := range t.closedConnections {
+		f(connection)
+	}
+	t.closedConnections = t.closedConnections[:0]
+}
+
+func (t *EbpfTracker) run() {
+	if err := offsetguess.Guess(t.reader); err != nil {
```
2opremio: Can't we hide this inside `tcptracer-bpf`?

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,256 @@
...
+			err := binary.Read(bytes.NewBuffer(data), byteorder.Host, &event)
+			if err != nil {
+				log.Errorf("Failed to decode received data: %s\n", err)
+				continue
+			}
+			tcpEventCallback(event)
+		}
+	}()
+
+	pmIPv4, err := bpflib.InitPerfMap(t.reader, "tcp_event_ipv4", channel)
```
2opremio: I think this could also be hidden inside an initialization function inside `tcptracer-bpf`?

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,256 @@
...
+func (t *EbpfTracker) initialize() {
+	t.initialized = true
+}
+
+func (t *EbpfTracker) isInitialized() bool {
+	return t.initialized
+}
+
+func (t *EbpfTracker) stop() {
+	// TODO: stop the go routine in run()
```
2opremio: Shouldn't this have been implemented?

probe/endpoint/procspy/background_reader_linux.go
```diff
@@ -38,6 +38,28 @@ func newBackgroundReader(walker process.Walker) *backgroundReader {
 	return br
 }
 
+func newForegroundReader(walker process.Walker) *backgroundReader {
```
2opremio: This shouldn't be here (either change the file name or place it somewhere else).

probe/endpoint/reporter.go
```diff
@@ -214,3 +235,29 @@ func (r *Reporter) makeEndpointNode(namespaceID string, addr string, port uint16
 func newu64(i uint64) *uint64 {
 	return &i
 }
+
+// procParsingSwitcher make sure that if eBPF tracking is enabled,
+// connections coming from /proc parsing are only walked once.
+func (r *Reporter) procParsingSwitcher() {
+	if r.conf.WalkProc && r.conf.UseEbpfConn {
+		r.conf.WalkProc = false
+		r.ebpfTracker.initialize()
```
2opremio: Shouldn't this be part of `newEbpfTracker`? The initialization should happen in a single place.

Makefile
iaguis: I can't fork it or create a PR because it's empty

Makefile
iaguis: Yes.

Since whenever we call the go binary we use `env $(GO_ENV) go`, the `CC` variable won't be picked up.

We can either leave it this way or do:

```
ifeq ($(GOARCH),arm)
CC=/usr/bin/arm-linux-gnueabihf-gcc
GO=env CC=$(CC) $(GO_ENV_ARM) go
...
```

circle.yml
iaguis: That's a mistake, only the first change should be here.

2opremio:
I am half way through the first pass of the review. However, the interaction between ebpf and the proc trackers is pretty hard to read and understand: (

* The code is sprinkled with conditionals checking if ebpf is enabled, making it stateful.
* The ebpf tracker can be phony.

Can you guys try to find an alternative improving this situation? For instance, create a single tracker interface, asking the proc walker for help when necessary (all the time if ebpf is not available).

prog/probe.go
```diff
@@ -46,25 +46,46 @@ var (
 	dockerEndpoint   = "unix:///var/run/docker.sock"
 )
 
-func check(flags map[string]string) {
```
2opremio: This is unrelated to this PR.

2opremio: That includes the other refactorings in this PR. Please create a different PR if necessary.

2opremio:
Also, I am missing unit tests.

circle.yml
2opremio: 👍 

Regarding `GOOS=linux`, can't you use `go tool dist env` in the Makefile?

probe/endpoint/ebpf.go
alban: tl;dr: it is implicitly required by the gobpf API we are using. We should update the API to make it more explicit... The cast is done here:
https://github.com/iovisor/gobpf/blob/a81bc05bf71ec17bf4c8d388621d2c81bc1f6a6b/elf/perf.go#L195

Longer version:

The kernel implements perf maps without locking to avoid contention between multiple execution of the ebpf program in parallel on multiple CPU. We actually have one perf file descriptor per CPU (each created with the syscall perf_event_open()).

But since each perf fd is independent from each other, we don't have any ordering guarantee between them. Other ebpf tracing programs in bcc currently don't mind that much about the lack of ordering. In practice, events are "very often" ordered.

Scope needs to process the events in chronological order, otherwise connections might appear to stay alive forever (e.g. if in a short-lived connection, a "close" event on one cpu is processed before its corresponding "connect" event on another cpu, the "close" event will be ignored because the connection is not know by Scope yet; then there will be the "connect" event without any following "close").

Perf events are opaque buffers. The perf subsystem on Linux does not handle reordering itself. The perf event struct is just a convention between the ebpf program ([struct tcp_ipv4_event_t](https://github.com/kinvolk/tcptracer-bpf/blob/b021c81f340f16376990ab9566ed1cace8ee9311/tcptracer-bpf.c#L25)) and Scope ([type tcpEvent struct](https://github.com/weaveworks/scope/pull/2135/files#diff-5023f232702c5a1b8177be985db66b08R40)).

However, we implemented the reordering in iovisor/gobpf, which is foreign to that convention. iovisor/gobpf does not know about the struct tcp_ipv4_event_t. So in the current implementation, it just assumes the ebpf program writes a timestamp at the beginning of the event buffer.

alepuccetti:
> Also, I am missing unit tests.

I am working on unit tests on this branch https://github.com/kinvolk/scope/commits/alessandro/ebpf-unit-test, but it is still a WIP. Because ebpf is not enabled by default, we were thinking to do a follow-up PR with the unit tests.

Makefile
iaguis: It seems it's currently not needed. I put it there to fix some errors like the ones mentioned in https://github.com/golang/go/issues/10210 but I can't reproduce them anymore.

circle.yml
iaguis: Yes, I didn't know about that. Thanks!

Makefile
2opremio: Great, let's remove this then

2opremio:
> I am working on unit tests on this branch 

Uhm, I would prefer if unit tests were merged at the same time.

Makefile
2opremio: I wonder why they set that limitation. Anyways, I have created the first commit.

alepuccetti:
I filed a bug report asking to enable kprobe and ebpf syscall on gci https://code.google.com/p/google-compute-engine/issues/detail?id=524

prog/probe.go
iaguis: I introduced this here because our ebpf code made the cyclomatic complexity of probeMain go above the threshold, but yes, it's an unrelated change. Here's the PR for that: https://github.com/weaveworks/scope/pull/2148

probe/endpoint/ebpf.go
iaguis: We didn't implement it because, even if we stop the go routine, it's not enough since we disabled the async proc parser.

We added a TODO to fix this in a follow-up PR (see OP) because:

1. ebpf parsing is optional (need to be enabled explicitly with `--probe.ebpf.connections=true`)
2. if a user enables it we assume they check on the logs whether it works or not
3. if the ebpf tracker works, it's unlikely that it stops working at a later time and if it does, we probaby want it to fail hard

iaguis:
> Can you guys try to find an alternative improving this situation? For instance, create a single tracker interface, asking the proc walker for help when necessary (all the time if ebpf is not available). I am happy to hear other solutions.

We'll find some alternatives

probe/endpoint/ebpf.go
alban: https://github.com/iovisor/gobpf/pull/17

probe/endpoint/ebpf.go
iaguis: I moved the things you mention to weaveworks/tcptracer-bpf#1

probe/endpoint/reporter.go
iaguis: This is not really an initialization, we use it to signal when the initial data from proc parsing was fed to the ebpf tracker so it can take over.

The name is rather unfortunate and we'll try to address this with the redesign of the ebpf tracker.

Makefile
alban: I am also adding logs in CircleCI to make it easier to figure out which version of ebpf.o it is using.

image

Also, I am adding that info in the Scope Docker image to make it easier to debug:

image

2opremio: that object file is elf, right? why not add a symbol or include text-file as a section in the object? http://stackoverflow.com/questions/16349557/does-gcc-have-any-options-to-add-version-info-in-elf-binary-file

2opremio: (I know we are using llvm and not gcc but I am sure there's an equivalente way to do this with the llvm toolchain)

alban: That's more or less what I did :)

I used an elf section to add the following:
```
char _build_id[] SEC("build_id") = CIRCLE_BUILD_URL;
```

compiled with clang with option `-DCIRCLE_BUILD_URL=\"$(CIRCLE_BUILD_URL)\"`

Then, the Scope build and the tcptracer-bpf build will both print the build URL:

```
	sha1sum ebpf/ebpf.o
	objdump -s --section=build_id ebpf/ebpf.o | grep "^ " | xxd -r | tr "\0" "\n"
```

So the build logs on CircleCI will keep that information.

Should we update this PR as long as we go, or address all review tasks and update it in one go? At the moment, @iaguis has commits in another branch to avoid too much traffic on this PR page:
https://github.com/kinvolk/scope/commits/iaguis/conn-perf-ebpf-guess-ci

Do you want other information than `$CIRCLE_BUILD_URL`?
I could add the commit id `$CIRCLE_SHA1` too, although it is visible in the CircleCI webpage already.

2opremio: No need to add anything else. I would say, let's address the problems with this PR here and then incorporate other improvements in different PRs, otherwise we are never going to finish :)

iaguis:
I updated this PR.

It addresses the review comments, including some refactor/redesign.

It is still pending on https://github.com/weaveworks/tcptracer-bpf/pull/1: We need to update the vendoring and the ebpf artifact URL.

cc @alepuccetti, @alban

probe/endpoint/connection_tracker.go
```diff
@@ -0,0 +1,260 @@
...
+			conf:            conf,
+			flowWalker:      newConntrackFlowWalker(conf.UseConntrack, conf.ProcRoot, conf.BufferSize, "--any-nat"),
+			ebpfTracker:     nil,
+			reverseResolver: newReverseResolver(),
+		}
+	}
+	// When ebpf will be active by default, check if it starts correctly otherwise fallback to flowWalk
+	et, err := newEbpfTracker(conf.UseEbpfConn)
+	if err != nil {
+		// TODO: fallback to flowWalker, when ebpf is enabled by default
```
2opremio: Please create an issue for this

probe/endpoint/connection_tracker.go
```diff
@@ -0,0 +1,260 @@
...
+		t.addConnection(rpt, tuple, namespaceID, fromNodeInfo, toNodeInfo)
+	}
+	return nil
+}
+
+func (t *connectionTracker) performEbpfTrack(rpt *report.Report, hostNodeID string) error {
+	t.ebpfTracker.walkConnections(func(e ebpfConnection) {
+		fromNodeInfo := map[string]string{
+			Procspied: "true",
+			EBPF:      "true",
```
2opremio: Have you checked the implications of these tags on the UI side?

Why is Procspied still marked? Backwards compatibility? (If so please add a comment)

2opremio:
Integration tests need to pass on this PR before merging. Let's discuss how to do this on Slack.

Also, I would like a few integration tests for ebpf connection tracking since it's not going to be enabled by default and want some reassurance of it not bitrotting/breaking before we do. At least I would like tests for the happy path, NATting and short-lived connections across containers in the same networking namespace.

Finally, I recall @iaguis mentioned a problem with tracking with sockets in accept state at the moment in which the tracker starts. Has that been addressed?

alban:
I pushed a new version:
- rebased on origin/master
- now that gobpf & tcptracer-bpf PRs are merged upstream, vendor the upstream versions
- it does not use the CircleCI artifact for ebpf but instead uses the vendored version
- includes the unit tests (ebpf_test.go) & refactoring (connection_tracker.go) from @alepuccetti

There are still some cleanup to do in the PR (newForegroundReader in background_reader_linux.go).

probe/endpoint/connection_tracker.go
2opremio: Clarification: I know that we currently make rendering decisions and exceptions due to the limitations of conntrack (pre-ebpf), when `Procspied` being false.

The current code seems to do the right thing, but it should be triple-checked and commented :)

alban:
> Finally, I recall @iaguis mentioned a problem with tracking with sockets in accept state at the moment in which the tracker starts. Has that been addressed?

No, this has not been addressed. I just filed https://github.com/weaveworks/tcptracer-bpf/issues/10 for this.

probe/endpoint/connection_tracker.go
iaguis: https://github.com/weaveworks/scope/issues/2215

probe/endpoint/connection_tracker.go
iaguis: [Non-procspied endpoints are not rendered](https://github.com/weaveworks/scope/blob/c9881a764e2f8a0e7d254b0cfd87ae8b11044f37/render/process.go#L26), we could probably change the name of that function and filter out nodes that are not `Procspied || EBPF`.

2opremio: They are, just not in the process topology. See https://github.com/weaveworks/scope/blob/c9881a764e2f8a0e7d254b0cfd87ae8b11044f37/render/container.go#L103

iaguis: Right.

render/process.go
```diff
@@ -23,7 +23,7 @@ func renderProcesses(rpt report.Report) bool {
 }
 
 // EndpointRenderer is a Renderer which produces a renderable endpoint graph.
-var EndpointRenderer = FilterNonProcspied(SelectEndpoint)
+var EndpointRenderer = FilterNonProcspiedNorEBPF(SelectEndpoint)
```
2opremio: How about naming it `RemoveConntracked`?

2opremio: Now that I see `FilterNonEBPF` exists your name may be better for consistency

Makefile
```diff
@@ -23,13 +23,24 @@ RM=--rm
...
 BUILD_IN_CONTAINER=true
 GO_ENV=GOGC=off
-GO=env $(GO_ENV) go
-NO_CROSS_COMP=unset GOOS GOARCH
-GO_HOST=$(NO_CROSS_COMP); $(GO)
-WITH_GO_HOST_ENV=$(NO_CROSS_COMP); $(GO_ENV)
 GO_BUILD_INSTALL_DEPS=-i
 GO_BUILD_TAGS='netgo unsafe'
 GO_BUILD_FLAGS=$(GO_BUILD_INSTALL_DEPS) -ldflags "-extldflags \"-static\" -X main.version=$(SCOPE_VERSION) -s -w" -tags $(GO_BUILD_TAGS)
+GOOS=$(shell go tool dist env | grep GOOS | sed -e 's/GOOS="\(.*\)"/\1/')
```
2opremio: Won't this break cross-compilation? (we set GOOS to compile for darwin for instance)

iaguis: If it's set externally, `go tool dist env` will respect that:

```
> GOOS=darwin go tool dist env
CC="gcc"
CC_FOR_TARGET="gcc"
GOROOT="/usr/lib/go"
GOBIN="/usr/lib/go/bin"
GOARCH="amd64"
GOOS="darwin"
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOTOOLDIR="/usr/lib/go/pkg/tool/linux_amd64"
```

2opremio: 👍 Makes sense

alepuccetti:
Test:
50 containers create a random number of connections between 1 and 50 each one.
For more information look the [dialer script](https://github.com/weaveworks/scope/blob/master/extras/dialer/dialer).

In three different terminals I run the following:
`./listener 8082`
`./dialer LISTENER_IP:8082 50 1`
`./time-scope-probe`

I run the test on `master-f8cf32c` and `iaguis-conn-perf-ebpf-guess-4ab97f5` back to back in a dedicated VM with four cores on GCE. We use randomly generated numbers for sleep time and the number of connections produced by each container, to reduce the difference between the tests we set the random 
d to `1`.
Before each test, I stop all the dialer containers and start from scratch.

The following graph shows the CPU consumtion for `master-f8cf32c` and `iaguis-conn-perf-ebpf-guess-4ab97f5`.

ebpf-performance

I run the experiment multiple times to ensure that the results are replicable within a reasonable interval.
Also, I notice that master does not show all the dialer containers connected to the server.
In fact, when running on master in the container view, I see containers not connected to the server, while using ebpf all the containers are connected to the server.
This happens because Scope is slower with detecting the connections and consequently showing them.

alepuccetti:
> Also, I notice that master does not show all the dialer containers connected to the server.
In fact, when running on master in the container view, I see containers not connected to the server, while using ebpf all the containers are connected to the server.
This happens because Scope is slower with detecting the connections and consequently showing them.

master:
master

ebpf
ebpf-screenshot

ekimekim:
I'm getting this panic on every report attempt:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x6458f9]

goroutine 180 [running]:
panic(0x1477080, 0xc420010060)
	/usr/lib/go-1.7/src/runtime/panic.go:500 +0x1a1
github.com/weaveworks/scope/probe/endpoint.(*connectionTracker).performWalkProc(0xc4202067d8, 0xc423c62900, 0xc420faa5f0, 0xf, 0xc422a1a078, 0x0, 0x253cb80)
	/go/src/github.com/weaveworks/scope/probe/endpoint/connection_tracker.go:161 +0x59
github.com/weaveworks/scope/probe/endpoint.(*connectionTracker).ReportConnections(0xc4202067d8, 0xc423c62900)
	/go/src/github.com/weaveworks/scope/probe/endpoint/connection_tracker.go:127 +0x135
github.com/weaveworks/scope/probe/endpoint.(*Reporter).Report(0xc420206780, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/weaveworks/scope/probe/endpoint/reporter.go:97 +0x123
github.com/weaveworks/scope/probe.(*Probe).report.func1(0xc4203ca6c0, 0xc4242be600, 0x252e280, 0xc420206780)
	/go/src/github.com/weaveworks/scope/probe/probe.go:155 +0x11c
created by github.com/weaveworks/scope/probe.(*Probe).report
	/go/src/github.com/weaveworks/scope/probe/probe.go:165 +0xec
```
The problem seems to be that it's attempting to `performWalkProc` despite no scanner being set due to `prog/probe.go:165`. This means somehow `connectionTracker.ebpfTracker != nil` even though `conf.UseEbpfConn` is true.

This can happen is if there's an error in creating the tracker (`probe/endpoint/connection_tracker.go:48`).

In this case, when we call `ReportConnections`, we skip both `t.performEbpfTrack` and `t.performFlowWalk`, executing `t.performWalkProc`, which we don't have a scanner for.

It seems to me that we need to pass the processCache in via conf, instead of scanner. Then we can instantiate the scanner or not depending on the internal results of `newConnectionTracker`.

Anyway, this was the error I was getting. It's very confusing - it looks like a whole list of errors happening at once? And they contain a double newline.
```
<probe> ERRO: 2017/02/15 01:58:25.668266 Cannot find BPF object file: failed to init perf map for IPv4 events: error guessing offsets: error initializing tcptracer_status map: unable to update element: argument list too long

<probe> ERRO: 2017/02/15 01:58:25.668280 Error setting up the ebpfTracker, connections will not be reported: failed to init perf map for IPv4 events: error guessing offsets: error initializing tcptracer_status map: unable to update element: argument list too long

```
I'm running arch linux, `Linux inflaton 4.8.13-1-ARCH #1 SMP PREEMPT Fri Dec 9 07:24:34 CET 2016 x86_64 GNU/Linux`

alban:
@ekimekim : thanks for testing.
- panic with performWalkProc: @alepuccetti will address this. Related to this, there is #2243 for implementing a proper fallback later.
- "double newline": @schu did https://github.com/weaveworks/tcptracer-bpf/pull/16
- "unable to update element: argument list too long": @iaguis is working on that via https://github.com/iovisor/gobpf/pull/23 We think this bug happens with more cpu/cores.

ekimekim:
Oh, and unrelated. I saw a ton of these when testing:
```
<probe> ERRO: 2017/02/16 00:41:31.036081 EbpfTracker error: unmatched close event: 172.17.0.1:52988-172.17.0.1:53 pid=27870 netns=4026531957
<probe> ERRO: 2017/02/16 00:41:41.036139 EbpfTracker error: unmatched close event: 172.17.0.1:52992-172.17.0.1:53 pid=27870 netns=4026531957
<probe> ERRO: 2017/02/16 00:41:51.037197 EbpfTracker error: unmatched close event: 172.17.0.1:52998-172.17.0.1:53 pid=27870 netns=4026531957
<probe> ERRO: 2017/02/16 00:41:58.894227 EbpfTracker error: unmatched close event: 192.168.5.115:43306-54.219.186.226:4040 pid=7218 netns=4026531957
<probe> ERRO: 2017/02/16 00:42:01.036348 EbpfTracker error: unmatched close event: 172.17.0.1:53010-172.17.0.1:53 pid=27870 netns=4026531957
<probe> ERRO: 2017/02/16 00:42:06.007529 EbpfTracker error: unmatched close event: 127.0.0.1:35032-127.0.0.1:6784 pid=27870 netns=4026531957
<probe> ERRO: 2017/02/16 00:42:11.036514 EbpfTracker error: unmatched close event: 172.17.0.1:53018-172.17.0.1:53 pid=27870 netns=4026531957
```
I expected a few of these when the process first started due to https://github.com/weaveworks/tcptracer-bpf/issues/10, but these kept more or less constant (one every second or so). But what really piqued my interest is this: They're all from the same pid. And that pid the `scope-probe` process.

alban:
@ekimekim yes, the fd-passing/fork/dup issue is unresolved. On CoreOS with systemd socket-activated sshd, it appeared that all incoming connections belong to pid 1 (systemd) instead of a ssh process because of socket activation. It's difficult to solve in a generic way but for the socket activation case, maybe we could use the proc connector (see also #1931) to be notified of new processes and check if they inherited fds to reassign sockets to their "last" owner. It appears that your test was not a socket activation but something else. What was your scenario?

2opremio:
> the fd-passing/fork/dup issue is unresolved

Is there a GitHub issue reflecting this problem?

alban:
> Is there a GitHub issue reflecting this problem?

@2opremio No. I understood that @ekimekim wanted to file it after this PR is merged. I could also file it if you want. I am not sure whether it would belong to scope or to tcptracer-bpf though.

2opremio:
Please file it

> I am not sure whether it would belong to scope or to tcptracer-bpf though

your call

alban:
Filed: #2253

schu:
@ekimekim regarding the "unmatched close event": those are due to https://github.com/weaveworks/tcptracer-bpf/issues/19

alepuccetti:
Depends on https://github.com/weaveworks/scope/pull/2225

alepuccetti:
I rebased this PR on master. We also added some integration tests for ebpf, more are to come.

Because our company CI has not been set up yet to run integration tests, I ran it on my personal repo, here is the result https://circleci.com/gh/alepuccetti/scope/169.

/cc @2opremio

backend/Dockerfile
```diff
@@ -1,6 +1,8 @@
-FROM golang:1.7.4
+FROM ubuntu:yakkety
+ENV GOPATH /go
+ENV PATH /go/bin:/usr/lib/go-1.7/bin:/usr/bin:/bin:/usr/sbin:/sbin
```
2opremio: Please don't sprinkle the Go version around, use a variable if needed.

integration/107_launch_with_ebpf_test.sh
```diff
@@ -0,0 +1,16 @@
+#! /bin/bash
+
+# shellcheck disable=SC1091
+. ./config.sh
+
+start_suite "Launch scope and check it boots, with ebpf connection tracking enabled"
```
2opremio: This test is not all that useful. It simply checks that the container is created. It may as well be crashing after bootup.

I would say, either make it more useful or remove it.

integration/301_internet_edge_with_ebpf_test.sh
```diff
@@ -0,0 +1,32 @@
+#! /bin/bash
+
+# shellcheck disable=SC1091
+. ./config.sh
+
+start_suite "Test short lived connections from the Internet"
+
+if ! echo "$HOST1" | grep "us-central1-a"; then
```
2opremio: Why is this needed? Why wouldn't it work locally? 

Also, hardcoding the zone is a hack. What if the machines are created somewhere else?

integration/301_internet_edge_with_ebpf_test.sh
```diff
@@ -0,0 +1,32 @@
...
+scope_on "$HOST1" launch --probe.ebpf.connections=true
+docker_on "$HOST1" run -d -p 80:80 --name nginx nginx
+
+do_connections() {
+    while true; do
+        curl -s "http://$HOST1:80/" >/dev/null || true
+        sleep 1
+    done
+}
+do_connections &
```
2opremio: Why bomb the container with connections? Wouldn't one be enough? (they should be in the UI for 15 seconds or so)

integration/301_internet_edge_with_ebpf_test.sh
```diff
@@ -0,0 +1,32 @@
...
+
+wait_for_containers "$HOST1" 60 nginx "The Internet"
+
+has_connection_by_id containers "$HOST1" "in-theinternet" "$(node_id containers "$HOST1" nginx)"
```
2opremio: Isn't there a race condition here? Isn't it possible that the connection hasn't shown up in the app by the time this is run?

integration/311_container_to_container_edge_with_ebpf_test.sh
```diff
@@ -0,0 +1,22 @@
...
+# shellcheck disable=SC1091
+. ./config.sh
+
+start_suite "Test short lived connections between containers, with ebpf connection tracking enabled"
+
+weave_on "$HOST1" launch
+scope_on "$HOST1" launch --probe.ebpf.connections=true
+weave_on "$HOST1" run -d --name nginx nginx
+weave_on "$HOST1" run -d --name client alpine /bin/sh -c "while true; do \
+	wget http://nginx.weave.local:80/ -O - >/dev/null || true; \
```
2opremio: Same comment as in the example above (why so many connections?). Also, if possible please use curl for consistency with the rest of the tests.

integration/107_launch_with_ebpf_test.sh
```diff
@@ -0,0 +1,16 @@
...
+
+scope_on "$HOST1" launch --probe.ebpf.connections=true
+
+wait_for_containers "$HOST1" 60 weavescope
+
+has_container "$HOST1" weavescope
+
+endpoints_have_ebpf "$HOST1"
```
2opremio: No prior actions from the test guarantee that endpoints existing.

2opremio:
> We also added some integration tests for ebpf, more are to come.

I am missing a test checking that connections between containers in the same networking namespace are tracked. I see this as a requirement before merging.

probe/endpoint/connection_tracker.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/weaveworks/scope/probe/process"
+	"github.com/weaveworks/scope/report"
+)
+
+// connectionTrackerConfig are the config options for the endpoint tracker.
+type connectionTrackerConfig struct {
+	HostID       string
+	HostName     string
+	SpyProcs     bool
+	UseConntrack bool
```
2opremio: This is just for short-lived connection tracking, right? Shouldn't `UseConntrack` and `WalkProc` be merged?

probe/endpoint/connection_tracker.go
```diff
@@ -0,0 +1,258 @@
...
+			ebpfTracker:     nil,
+			reverseResolver: nil,
+		}
+		return noopConnectionTracker
+	}
+	// Run conntrack and proc parsing synchronously only once to initialize ebpfTracker
+	seenTuples := map[string]fourTuple{}
+	// Consult the flowWalker to get the initial state
+	if !conf.UseConntrack {
+		log.Warn("conntrack is disabled")
```
2opremio: Why do we need this? If the flag reached this point it should be intentional, shouldn't it?

probe/endpoint/connection_tracker.go
```diff
@@ -0,0 +1,258 @@
...
+	} else if err = IsConntrackSupported(conf.ProcRoot); err != nil {
+		log.Warnf("Not using conntrack: not supported by the kernel: %s", err)
+	} else {
+		existingFlows, err := existingConnections([]string{"--any-nat"})
+		if err != nil {
+			log.Errorf("conntrack existingConnections error: %v", err)
+		} else {
+			for _, f := range existingFlows {
+				tuple := fourTuple{
```
2opremio: Please refactor this with `performFlowWalk` below

probe/endpoint/connection_tracker.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	// Run conntrack and proc parsing synchronously only once to initialize ebpfTracker
+	seenTuples := map[string]fourTuple{}
+	// Consult the flowWalker to get the initial state
+	if !conf.UseConntrack {
+		log.Warn("conntrack is disabled")
+	} else if err = IsConntrackSupported(conf.ProcRoot); err != nil {
+		log.Warnf("Not using conntrack: not supported by the kernel: %s", err)
+	} else {
+		existingFlows, err := existingConnections([]string{"--any-nat"})
```
2opremio: Please embed this into the if branch above:

```go
else if existingFlows, err := existingConnections([]string{"--any-nat"}); err != nil {
	log.Errorf("conntrack existingConnections error: %v", err)
} else {
	for _, f := range existingFlows {
...
}
```

probe/endpoint/connection_tracker.go
```diff
@@ -0,0 +1,258 @@
...
+
+	var processCache *process.CachingWalker
+	var scanner procspy.ConnectionScanner
+	processCache = process.NewCachingWalker(process.NewWalker(conf.ProcRoot))
+	processCache.Tick()
+	scanner = procspy.NewSyncConnectionScanner(processCache)
+	defer scanner.Stop()
+	conns, err := scanner.Connections(conf.SpyProcs)
+	if err != nil {
+		log.Errorf("Error initializing ebpfTracker while scanning /proc, continue without initial connections: %s", err)
```
2opremio: `s/continue/continuing/`

probe/endpoint/connection_tracker.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+	}
+
+	var processCache *process.CachingWalker
+	var scanner procspy.ConnectionScanner
+	processCache = process.NewCachingWalker(process.NewWalker(conf.ProcRoot))
+	processCache.Tick()
+	scanner = procspy.NewSyncConnectionScanner(processCache)
+	defer scanner.Stop()
+	conns, err := scanner.Connections(conf.SpyProcs)
```
2opremio: Please embed this on the if below

probe/endpoint/connection_tracker.go
```diff
@@ -0,0 +1,258 @@
...
+	if err != nil {
+		log.Errorf("Error initializing ebpfTracker while scanning /proc, continue without initial connections: %s", err)
+		et.feedInitialConnectionsEmpty()
```
2opremio: Why is this needed?

Won't it be enough to use `feedInitialConnections` with an empty `conns` argument?

probe/endpoint/ebpf.go
```diff
@@ -0,0 +1,167 @@
...
+func (t *EbpfTracker) feedInitialConnectionsEmpty() {
+	t.fed = true
+}
+
+func (t *EbpfTracker) isFed() bool {
+	return t.fed
+}
+
+func (t *EbpfTracker) stop() {
+	// TODO: stop the go routine in run()
```
2opremio: Why isn't this implemented?

probe/endpoint/connection_tracker.go
2opremio: Also, won't this block probe (preventing it from reporting other information) until the first walk is done? 

This is an issue in very active systems in which it can take a long time.

schu:
@2opremio regarding your comments on the tests:

The added tests are a mostly a copy of existing tests modulo the launch with `probe.ebpf.connections` set
(i.e. `scope_on "$HOST1" launch --probe.ebpf.connections=true`) and the new helper `endpoints_have_ebpf()`. `wget` is also used in 310 and 320 for example. We are happy to address the raised concerns but would prefer to do so as part of a follow-up pull request.

> I am missing a test checking that connections between containers in the same networking namespace are tracked. I see this as a requirement before merging.

We started working on a test but it's not done yet: we can't use `weave_on` here as weave adds additional DNS parameters (e.g. `--hostname`) to the Docker cmdline not compatible with `--net=container:...`: https://docs.docker.com/engine/reference/run/#network-container

integration/107_launch_with_ebpf_test.sh
schu: We will drop 107 (and use `endpoints_have_ebpf()` in the other two tests to check if ebpf is enabled).

2opremio:
> We are happy to address the raised concerns but would prefer to do so as part of a follow-up pull request.

Great, please fill an issue.

> We started working on a test but it's not done yet: we can't use weave_on ..

You don't need to use weave, using a single host should be fine.

integration/301_internet_edge_with_ebpf_test.sh
2opremio: This should be fixed before merging

2opremio:
> Great, please file an issue.

Let me refine this. I am fine with a follow-up for issues requiring a refactoring out of the scope of this PR otherwise they should be fixed here.

1.

probe/endpoint/connection_tracker.go
alepuccetti: `UseConntrack` determine if we use conntrack or not independently from the other way of detecting connections. While `WalkProc` determine if we use proc parsing or not.
I designed `connectionTracker` in a general way, so if ebpf is enabled then the internal logic will decide to not use proc parsing and/or conntrack. However, the configuration comes from the user that set the conntrack flag.

probe/endpoint/connection_tracker.go
alepuccetti: This is executed only once as part of the ebpf tracker initialization, so this will affect only the startup time, we will not execute scanner anymore when using ebpf.

2opremio: > so this will affect only the startup time

That's not acceptable

probe/endpoint/ebpf.go
iaguis: I explained our reasoning here: https://github.com/weaveworks/scope/pull/2135#discussion_r97312319

We can explain it better on the TODO message.

2opremio: Please do

schu:
> I am missing a test checking that connections between containers in the same networking namespace are tracked by ebpf. I see this as a requirement before merging.

We pushed https://github.com/weaveworks/scope/pull/2135/commits/e04365fec422a3163714792b224528b3727c1d18

2opremio:
The CircleCI hook got stuck, mind clicking on rebuild?

iaguis:
> The CircleCI hook got stuck, mind clicking on rebuild?

Done

schu:
>> The CircleCI hook got stuck, mind clicking on rebuild?

> Done

Integration tests don't run on Kinvolk Circle CI; we did for e04365f here: https://circleci.com/gh/schu/scope/173

2opremio:
> Integration tests don't run on Kinvolk Circle CI

I'm confused, this was the goal of a big part of the work in the past weeks (so that Kinvok PRs didn't break master).

If we were to run tests manually in other repositories, then I would had done it and we would have saved a lot of work.

Can we please make sure that whatever org you create PRs from runs integration tests?

schu:
> Can we please make sure that whatever org you create PRs from runs integration tests?

With the changes from https://github.com/weaveworks/scope/pull/2225 (particularly https://github.com/weaveworks/scope/pull/2225/commits/e031728b04b0c935071598c8fafaff8f2a5599d3) it's now possible for every user/org to configure their own Circle CI to run integration tests on GCE (it wasn't possible before). The configuration remains a manual step for every user to do on their own. We have just configured https://circleci.com/gh/kinvolk/scope accordingly and will ping you once the integration tests pass there as well.

schu:
While tests are green now, we saw two failures we can't explain:

* https://circleci.com/gh/kinvolk/scope/740 (`200_clustering_2_test.sh` failed, not related to this pull request)
* https://circleci.com/gh/kinvolk/scope/741 (`312_container_to_container_edge_same_netns_with_ebpf_test.sh` failed, a new test added with this pull request.)

alban:
> https://circleci.com/gh/kinvolk/scope/741
> (312_container_to_container_edge_same_netns_with_ebpf_test.sh failed, a new test added with this pull request.)

It has failed with the error:
> Only 43 endpoints of 49 have ebpf enabled, should be equal

I looked in the report at the 6 endpoints that don't have eBPF=true and they are tagged with the following pids/cmdline:
- 3726 scope-probe
- 1953 /usr/bin/python /usr/share/google/google_daemon/manage_addresses.py
- 5212 /usr/bin/python /usr/share/google/google_daemon/manage_accounts.py
- 3724 scope-app
- 3726 scope-probe
- 1942 /usr/bin/python /usr/share/google/google_daemon/manage_clock_sync.py

So it's not about the connections being tested (in the two containers "nginx" and "client"). It seems to be a race during initialization. I'm trying to understand how it can happen.

schu:
> It seems to be a race during initialization. I'm trying to understand how it can happen.

We think the behavior comes with https://github.com/weaveworks/scope/pull/2135/commits/341054638c41a7001c110b48d23a5792a9446307 and that the endpoint is found when the ebpf tracker is not initialized yet.

iaguis:
We've been running integration tests all morning on [@schu's circleci](https://circleci.com/gh/schu/scope) and we haven't encountered any reds. Also, we couldn't reproduce the issue on our local machines.

By reading our code we shouldn't see **any** connection sith `Procspied`. The only place where nodes are set to be `Procspied: "true"` is in [`performWalkProc()`](https://github.com/kinvolk/scope/blob/e04365fec422a31637147]92b224528b3727c1d18/probe/endpoint/connection_tracker.go#L138) which should never be called if eBPF is enabled because [`t.ebpfTracker` should not be `nil`](https://github.com/kinvolk/scope/blob/e04365fec422a3163714792b224528b3727c1d18/probe/endpoint/connection_tracker.go#L97): here's where we [initialize the tracker](https://github.com/kinvolk/scope/blob/e04365fec422a3163714792b224528b3727c1d18/probe/endpoint/connection_tracker.go#L34).

The only explanation I can think of is that another probe from another test is running at the same time and feeding the scope app, and that's why we get some endpoints that are not `eBPF` but `Procspied`. I tested this scenario locally (starting a scope app and 2 probes, one with eBPF and one without) and I get some nodes with only `Procspied` like in the failed report.

2opremio:
> The only explanation I can think of is that another probe from another test is running at the same time and feeding the scope app, and that's why we get some endpoints that are not `eBPF` but `Procspied`. I tested this scenario locally (starting a scope app and 2 probes, one with eBPF and one without) and I get some nodes with only `Procspied` like in the failed report.

That's scary and should be resolved. Either way this should be resolved before merging.

2opremio:
Also, regarding https://circleci.com/gh/kinvolk/scope/741 , please don't output full reports in tests since they fill the output.

2opremio:
I have tried to test this branch locally and I get panics: https://gist.github.com/2opremio/8aece6a2fba726fabc6b944ab2d9ec92

schu:
Rebased on master.

probe/endpoint/ebpf.go
```diff
@@ -40,11 +43,49 @@ type EbpfTracker struct {
...
 
+var releaseRegex = regexp.MustCompile(`^(\d+)\.(\d+).*$`)
+
+func isKernelSupported() error {
+	release, _, err := host.GetKernelReleaseAndVersion()
+	if err != nil {
+		return err
+	}
+
+	releaseParts := releaseRegex.FindStringSubmatch(release)
```
2opremio: You can use `SScanf` see https://github.com/weaveworks/scope/blob/master/probe/endpoint/procspy/proc_linux.go#L49-L61 , which makes the code much simpler.

Extra points for refactoring the code and making `proc_linux.go` use `GetKernelReleaseAndVersion()` (which should probably be moved to a common package).

docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM alpine:3.5
```
2opremio: The rebase on top of master wasn't done properly. This file shouldn't exist after https://github.com/weaveworks/scope/pull/2284

2opremio:
A quick and rough check (`top -c`) of the CPU consumption in an empty system shows that, when ebpf is enabled, the probe consumes 2x the CPU.

How can you explain this?

With ebpf:

screen shot 2017-03-06 at 15 04 45

Without ebpf:

screen shot 2017-03-06 at 15 11 17


If anything, I would expect the opposite ...

2opremio:
Note that the app CPU consumption varies a bit but the probe consumption is fairly static: ~7% when  enabling ebpf and ~3% without it.

2opremio:
I am seeing a lot of `unmatched close event` errors, just like @ekimekim at https://github.com/weaveworks/scope/pull/2135#issuecomment-280193572

In this case, they are DNS requests and requests to weave net (which is not running):

```
<probe> ERRO: 2017/03/06 14:19:12.590043 EbpfTracker error: unmatched close event: 127.0.0.1:35732-127.0.0.1:6784 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:19:17.608381 EbpfTracker error: unmatched close event: 172.17.0.1:42202-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:19:27.586822 EbpfTracker error: unmatched close event: 172.17.0.1:42206-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:19:37.589443 EbpfTracker error: unmatched close event: 172.17.0.1:42212-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:19:47.592937 EbpfTracker error: unmatched close event: 172.17.0.1:42216-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:19:57.597814 EbpfTracker error: unmatched close event: 172.17.0.1:42222-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:20:07.597878 EbpfTracker error: unmatched close event: 172.17.0.1:42226-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:20:12.597237 EbpfTracker error: unmatched close event: 127.0.0.1:35764-127.0.0.1:6784 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:20:17.598236 EbpfTracker error: unmatched close event: 172.17.0.1:42234-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:20:27.598730 EbpfTracker error: unmatched close event: 172.17.0.1:42238-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:20:37.603551 EbpfTracker error: unmatched close event: 172.17.0.1:42244-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:20:47.606111 EbpfTracker error: unmatched close event: 172.17.0.1:42248-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:20:57.608987 EbpfTracker error: unmatched close event: 172.17.0.1:42254-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:21:07.608991 EbpfTracker error: unmatched close event: 172.17.0.1:42258-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:21:12.591241 EbpfTracker error: unmatched close event: 127.0.0.1:35796-127.0.0.1:6784 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:21:17.614570 EbpfTracker error: unmatched close event: 172.17.0.1:42266-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:21:27.618499 EbpfTracker error: unmatched close event: 172.17.0.1:42270-172.17.0.1:53 pid=3022 netns=4026531957
<probe> ERRO: 2017/03/06 14:21:37.624944 EbpfTracker error: unmatched close event: 172.17.0.1:42276-172.17.0.1:53 pid=3022 netns=4026531957
```

I truly doubt it's #2253 but simply requests to ports where no-one is listening on. In fact, if I try to connect to a port where no-one is listening:


```
ubuntu@ubuntu-yakkety:~$ netcat localhost 9999
```

I get this in the logs:

```
<probe> ERRO: 2017/03/06 14:25:36.558455 EbpfTracker error: unmatched close event: 127.0.0.1:59104-127.0.0.1:9999 pid=5632 netns=4026531957
```

Please correct this

iaguis:
> I am seeing a lot of unmatched close event errors, just like @ekimekim at #2135 (comment)

This is tracked in https://github.com/weaveworks/tcptracer-bpf/issues/19

2opremio:
> This is tracked in https://github.com/weaveworks/tcptracer-bpf/issues/19

Thanks. Should we silence those errors in the meantime? (they are not really errors after all).

docker/Dockerfile
schu: Thanks for catching, removed.

iaguis:
> Thanks. Should we silence those errors in the meantime? (they are not really errors after all).

I think we should do that in the meantime, yes.

alban:
> A quick and rough check (top -c) of the CPU consumption in an empty system shows that, when ebpf is enabled, the probe consumes 2x the CPU.
> 
> How can you explain this?

@2opremio The connection scanner used to feed the ebpf tracker with the initial state was not stopped correctly. I added commit 915eb6995c181cb25d8939b751fb9fee85072b40 which fixes the perf issue. @schu tested that it fixes the problem: `top -c` and `cpustat` show similar results when there are no connections.

2opremio:
Good job, I will test this branch again before the day ends.

alban:
PR updated:
- rebased on master (including #2312 with the changes on the GCE network)
- squashed to a small amount of commits
- includes the new vendoring of tcptracer-bpf with https://github.com/weaveworks/tcptracer-bpf/pull/27
- includes the performance fix mentioned in https://github.com/weaveworks/scope/pull/2135#issuecomment-285013523

Let's see if CircleCI is happy.

Issue #2133, opened 2017-01-17
Investigate using Scope as a UI component


jpellizzari:
It would be super sweet if we could include scope as a dependency in other projects, such as service-ui.

Notes:
* Scope has to be loaded via service-ui's router instance, so having Scope maintain its own render lifecycle is probably not going to work
* `dispatch` and `getState` are used all over the place in components, actions and utils, so those functions from `redux` will need to be intercepted
* `redux-thunk` middleware also has access to `dispatch` and `getState`

rade:
> It would be super sweet

why?

jpellizzari:
@rade This will open the door for lots of integrations with other projects. Scope's application state and user actions, such as controls, will be usable by any project that mounts scope in the UI.

rade:
Do you have an example of such an integration that we are contemplating?

jpellizzari:
Some ideas:

* Lock, release or automate a k8s service from the scope UI
* Show Flux history of a service in Scope's service's details panel
* Change a configuration option (such as a service env variable) and have it propagate back to version control

rade:
got it. thanks.

jpellizzari:
Implemented by #2144

PR #2144, opened 2017-01-21
Allow Scope UI to be installed as a Node module

jpellizzari:
Implements #2133 

This PR exposes a JS API consisting of the Scope `<App />` React component, the application state reducer, and the constant action types to allow for another project to use Scope as a dependency. There is also some shifting around of runtime dependencies that were mis-classified as dev dependencies. 

The largest change is to accommodate for relative URLs for websocket and API requests.

Since NPM requires that the `package.json` file be in the root of the directory, it is not possible to use the existing repository structure as a module. Rather than moving the package.json to the root and changing lots of build configuration settings, it made more sense to push the bundle to S3. There is an `s3-publish` job that will `tar` the 'client' directory and push it to S3. The module can then be added to a projects package.json, like so: 

```json
"dependencies": {
  "weave-scope": "https://s3-us-west-2.amazonaws.com/weaveworks-npm-modules/weave-scope/weave-scope-1.2.0.tgz"
}
```
This is meant to be a proof-of-concept for how to include the package as a dependency. There may need to be changes pending feedback.

@davkal Please review and give feedback, especially regarding the S3 thing.


client/.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 build/
 coverage/
 test/*png
+*.tgz
```
foot: maybe tighten this one up.

foot:
s3 looks fine, we could build and push to npm too but whats the diff really?

LGTM though

davkal:
While I prefer publishing to npm, we should try the S3 route first.

client/app/scripts/utils/web-api-utils.js
```diff
@@ -58,8 +58,14 @@ export function basePathSlash(urlPath) {
...
 
+const API_PATH = basePathSlash(window.location.pathname) === '/'
+  ? 'api'
+  : `/api${window.location.pathname}/api`;
+
 const wsProto = location.protocol === 'https:' ? 'wss' : 'ws';
-const wsUrl = `${wsProto}://${location.host}${basePath(location.pathname)}`;
+export const wsUrl = basePathSlash(window.location.pathname) === '/'
+  ? `${wsProto}://${location.host}${basePath(window.location.pathname)}`
+  : `${wsProto}://${location.host}/api${basePath(window.location.pathname)}`;
```
jpellizzari: I was thinking the `web-api-utils` could use some re-factoring; maybe replace the `reqwest` module with a decorated XHR function that already has the hostname and path stored. We may be able to use the `redux-thunk` .withExtraArgument() to pass down deps to async actions.

client/app/scripts/utils/web-api-utils.js
```diff
@@ -171,13 +177,13 @@ export function getNodeDetails(topologyUrlsById, currentTopologyId, options, nod
   const obj = nodeMap.last();
   if (obj && topologyUrlsById.has(obj.topologyId)) {
     const topologyUrl = topologyUrlsById.get(obj.topologyId);
-    let urlComponents = [topologyUrl, '/', encodeURIComponent(obj.id)];
+    let urlComponents = [API_PATH, '/', trimStart(topologyUrl, '/api'), '/', encodeURIComponent(obj.id)];
```
jpellizzari: We could, but we still have to trim off the leading slash from the topologyUrl like we do in `getAllNodes()`. `urlComponents` used to have a `.substr(1)` on it to remove the leading slash.

jpellizzari:
I re-worked the urls to support our three use cases for the scope-ui: standalone, iframe, and component. I created this ticket as a re-factoring followup: https://github.com/weaveworks/scope/issues/2152

Issue #2031, opened 2016-11-26
`make clean` fails with `rm: ... Permission denied`


rade:
```
make -k clean
env GOGC=off go clean ./...
# Don't actually rmi the build images - rm'ing the .uptodate files is enough to ensure
# we rebuild the images, and rmi'ing the images causes us to have to redownload a lot of stuff.
#  docker rmi weaveworks/scope-ui-build weaveworks/scope-backend-build >/dev/null 2>&1 || true
rm -rf scope.tar .scope_ui_build.uptodate .scope_backend_build.uptodate \
	prog/scope vendor/runsvinit/runsvinit prog/staticui/staticui.go prog/externalui/externalui.go client/build/*.js client/build-external/*.js docker/weave .pkg \
	report/report.codecgen.go render/detailed/detailed.codecgen.go vendor/github.com/ugorji/go/codec/codecgen/bin
rm: cannot remove 'client/build-external/app-bc6345edea5f887d82d4.js': Permission denied
rm: cannot remove 'client/build-external/contrast-app-8bc9d18178cc106cc5cf.js': Permission denied
rm: cannot remove 'client/build-external/terminal-app-1f43e6901bfb8d6d8060.js': Permission denied
rm: cannot remove 'client/build-external/vendors-bcb446fe1ffedd0c7171.js': Permission denied
Makefile:171: recipe for target 'clean' failed
make: *** [clean] Error 1
```

cc @davkal

ekimekim:
I was confused for quite a while since there seemed to be no visible difference between `client/*.js` and `client-external/*.js` - both are generated files owned by root and have been for some time.
I eventually realised - you don't need to be root to delete a file owned by root, you just need to be able to write to its parent directory. And so:
```
$ ls -ld client/build client/build-external
drwxr-xr-x 2 mike mike 12288 Jan 25 14:53 client/build
drwxr-xr-x 2 root root  4096 Jan 25 14:50 client/build-external
```
So the problem is that `client/build-external` is created as a side-effect of the build process, whereas `client/build` is created at checkout because it contains a tracked file (`client/build/favicon.ico`).

Possible solutions:
* Make build-external tracked by adding a tracked file inside it (eg. `.git-track-this-directory`). This would still require everyone with a currently-existing build-external to manually chown or delete it.
* Add a `sudo chown $USER client/build-external` to the makefile
* Add a `sudo rm -f client/build-external/*js` to the makefile

davkal:
[...]

PR #2153, opened 2017-01-26
Fix ownership issues with client/build-external

ekimekim:
When the directory is created during the build process, it is as root.

This prevents us deleting root-owned files inside the directory during make clean.

In client/build/, we get around this because the directory contains a tracked file,
so it's created during git checkout.
We make this the case for build-external by adding a hidden, empty file to track.
This won't fix existing checkouts though, so we also add a fix-up step to make clean.

Fixes https://github.com/weaveworks/scope/issues/2031

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,6 +174,9 @@ ui-build-pkg:
 
 clean:
 	$(GO) clean ./...
+	# build-external was accidentially created as owned by root in earlier versions. This is fixed now,
+	# but we need to fix up old checkouts, since git doesn't track ownership.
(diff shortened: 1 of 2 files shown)


2opremio:
You may consider using a `.gitignore` file, like suggested at http://stackoverflow.com/a/932982/1914440

ekimekim:
done. PTAL?

2opremio:
LGTM, CI failed though

ekimekim:
```
ERROR: unsatisfiable constraints:
  bash (missing):
    required by: world[bash]
```
fixed by https://github.com/weaveworks/scope/pull/2158, so I'll rebase on master to pick it up

PR #2158, opened 2017-01-26
Upgrade container base image to alpine:3.5

ekimekim:
As well as upgrading to latest stable being nice in general,
this was triggered by an issue where we could no longer find the bash package in 3.3

This upgrade also lets us get rid of reliance on the edge (rolling latest) repository
for runit, since 3.5 now contains runit natively.

I've done some basic tests and nothing seems to be broken by the version changes.


jpellizzari:
@ekimekim I was able to build the previously failing docker container and a `make clean && make` finishes successfully. LGTM.

Issue #2154, opened 2017-01-26
Scope stuck "connecting to scope"


tomwilkie:
With this in the console:
```
VM245 vendors-b764d66….js?6ded3cd…:12 Uncaught TypeError: t.filter is not a function
    at f (VM245 vendors-b764d66….js?6ded3cd…:12)
    at i (VM245 vendors-b764d66….js?6ded3cd…:33)
    at f (VM245 vendors-b764d66….js?6ded3cd…:33)
    at l (VM245 vendors-b764d66….js?6ded3cd…:33)
    at VM245 vendors-b764d66….js?6ded3cd…:33
    at dispatch (VM245 vendors-b764d66….js?6ded3cd…:45)
    at VM245 vendors-b764d66….js?6ded3cd…:1
    at VM245 vendors-b764d66….js?6ded3cd…:33
    at Object.success (VM245 vendors-b764d66….js?6ded3cd…:13)
    at Reqwest.o.success._successHandler (VM245 vendors-b764d66….js?6ded3cd…:50)
```

davkal:
Looks like requests to the backend are malformed when used in the service UI: `http://hostname/api/api/app/{org}/api/topology`

Notice the double `api` in the beginning.

Possible introduced by #2144

2opremio:
This happened somewhere between 244bba2 and 29edaeb . I doubt this was the backend. My main suspect is https://github.com/weaveworks/scope/pull/2144

@foot @davkal @jpellizzari @fbarl This is urgent, it broke dev (and then prod)

2opremio:
Interestingly, I did test prod right after the upgrade and Scope was working. Could it be due to caching of the assets?

davkal:
> Interestingly, I did test prod right after the upgrade and Scope was working. Could it be due to caching of the assets?

That could have been masking the issue.

2opremio:
So, IMO, we need to do three things to prevent this from happening in the future:

1. Fix the assets caching problem (if there´s any, I believe there is)
2. Add an integration test minimally testing the UI
3. Be more thorough when testing changes

2opremio:
But let´s of course first figure out what the problem is.

rade:
I thought we had some metrics/alerts for set up to catch this sort of breakage.

2opremio:
> I thought we had some metrics/alerts for set up to catch this sort of breakage.
Assuming the problem comes from the UI, I don´t think so.

davkal:
[...]

PR #2159, opened 2017-01-26
Fix for incorrect iFrame API URLs

jpellizzari:
Should have been included in #2144 

Ensures that scope uses the correct API URLs when running as a dependency, or as an iFrame.


2opremio:
@jpellizzari Does this fully fix https://github.com/weaveworks/scope/issues/2154

jpellizzari:
@2opremio Yes it does.

PR #2161, opened 2017-01-27
fix error when docker daemon is running with user namespace enabled.

QianJin2013:
Hi, i have my docker daemon run with user namespace enabled(All users in containers will be mapped to sub user of specified external user),

When i start `/usr/local/bin/scope`, it reports

1:"Privileged mode is incompatible with user namespaces".

Then modify the script, add `--userns=host` to let it share user namespace with host.

2:"Cannot share the host's network namespace when user namespaces are enabled".

The same reason and solution with 1.

That's all, hope helpful.

### DIFF
--- a/scope
+++ b/scope
@@ -22,6 +22,13 @@ IP_ADDR_CMD="find /sys/class/net -type l | xargs -n1 basename | grep -vE 'docker
     xargs -n1 ip addr show | grep inet | awk '{ print \$2 }' | grep -oE '$IP_REGEXP'"
 WEAVESCOPE_DOCKER_ARGS=${WEAVESCOPE_DOCKER_ARGS:-}
 
+# When docker daemon is running with User Namespace enabled, this tool will run into errors:
+#  "Privileged mode is incompatible with user namespaces" for `docker run --privileged`
+#  "Cannot share the host's network namespace when user namespaces are enabled" for `docker run --net=host`
+# To avoid above errors, use `--userns=host` option to let container use host User Namespace.
+# This option(saved in $USERNS_HOST) will be inserted ONLY IF docker support `--userns` option.
+docker run --help | grep -q -- --userns && USERNS_HOST="--userns=host"
+
 usage() {
     name=$(basename "$0")
     cat >&2 <<-EOF
@@ -160,7 +167,7 @@ create_plugins_dir() {
 
 launch_command() {
     # shellcheck disable=SC2086
-    echo docker run --privileged -d --name="$SCOPE_CONTAINER_NAME" --net=host --pid=host \
+    echo docker run --privileged $USERNS_HOST -d --name="$SCOPE_CONTAINER_NAME" --net=host --pid=host \
         -v /var/run/docker.sock:/var/run/docker.sock \
         -v /var/run/scope/plugins:/var/run/scope/plugins \
         -e CHECKPOINT_DISABLE \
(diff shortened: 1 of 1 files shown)


2opremio:
Thanks for this!

This change assumes the Docker Engine version is >= 1.10 , which is "only" one year old (some distros may not have caught up).

@QianJin2013 Could you conditionally incorporate the `--userns=host` flag only if the docker engine version is >= 1.10 ?

2opremio:
@bboreham You may want to take this into consideration for Weave Net as well.

rade:
TBH, I reckon mandating Docker >=1.10 is fine - Weave Net will do that from the next release onwards anyway.

2opremio:
@rade Will you explicitly be checking the Docker version on launch or is it implicit?

rade:
no idea. @bboreham will know.

QianJin2013:
ok, i will add a check whether or not support --userns=host

2opremio:
@QianJin2013 Hold on, it may not be necessary. It seems almost all users run >= 1.10 anyways

QianJin2013:
@2opremio yeah, not much necessary. 
I anyone want, i will use 
```
USERNS_OPTION_KEY=$(docker run --help |grep -Fo -- --userns)
```
its result will be "--userns" or "", 
then further convert it to --userns=host if not empty. It's no harmful.

alban:
Since `$USERNS_HOST` can be empty and in that case we don't want it to be considered a parameter, I think you will need to add `# shellcheck disable=SC2086` just before using it the second time so that CircleCI can run shellcheck correctly.

See also https://github.com/koalaman/shellcheck/wiki/SC2086

QianJin2013:
@alban OK, added `shellcheck disable=SC2086`, rebased, conflict resolved.

bboreham:
It seems that `--userns=host` was added in Docker 1.11, not in 1.10 - https://github.com/docker/docker/pull/20111

So your current proposed mitigation is workable; the original assumption that since we mandate 1.10 we're ok is not.

QianJin2013:
anyway, for docker which does not support `--userns`, this patch also works without error.

scope
```diff
@@ -22,6 +22,8 @@ IP_ADDR_CMD="find /sys/class/net -type l | xargs -n1 basename | grep -vE 'docker
     xargs -n1 ip addr show | grep inet | awk '{ print \$2 }' | grep -oE '$IP_REGEXP'"
 WEAVESCOPE_DOCKER_ARGS=${WEAVESCOPE_DOCKER_ARGS:-}
 
+docker run --help |grep -q -- --userns && USERNS_HOST="--userns=host"
```
2opremio: Mind adding a short comment on why this is needed?

2opremio:
> anyway, for docker which does not support --userns, this patch also works without error.

Fair enough.

LGTM, modulo a descriptive comment. @QianJin2013 I will merge after you add it.

QianJin2013:
ok, comment added.  Seems too long, i am not good at english, sorry.

scope
```diff
@@ -22,6 +22,13 @@ IP_ADDR_CMD="find /sys/class/net -type l | xargs -n1 basename | grep -vE 'docker
     xargs -n1 ip addr show | grep inet | awk '{ print \$2 }' | grep -oE '$IP_REGEXP'"
 WEAVESCOPE_DOCKER_ARGS=${WEAVESCOPE_DOCKER_ARGS:-}
 
+# When docker daemon is running with User Namespace enabled, this tool will run into errors:
+#  "Privileged mode is incompatible with user namespaces" for `docker run --privileged`
+#  "Cannot share the host's network namespace when user namespaces are enabled" for `docker run --net=host`
+# To avoid above errors, use `--userns=host` option to let container use host User Namespace.
+# This option(saved in $USERNS_HOST) will be inserted ONLY IF docker support `--userns` option.
+docker run --help |grep -q -- --userns && USERNS_HOST="--userns=host"
```
alban: Could you add a space after `|`:
```diff
-docker run --help |grep -q -- --userns && USERNS_HOST="--userns=host"
+docker run --help | grep -q -- --userns && USERNS_HOST="--userns=host"
```

lint on CircleCI would complain otherwise.

The following command would also fix it automatically:
```
shfmt -i 4 -w scope
```

QianJin2013: ok, added.

QianJin2013:
OK, added

PR #2162, opened 2017-01-27
DNSSnooper: Support Dot1Q and limit decoding errors

### LINKED ISSUE #2155: DNSSnooper: error decoding packet: No decoder for layer type Dot1Q
```
[12:48 PM]  
carlpett So, first feedback :slightly_smiling_face: We get a lot of this:
`<probe> ERRO: 2017/01/26 12:46:27.282902 DNSSnooper: error decoding packet: No decoder for layer type Dot1Q`

[12:48 PM]  
Problem related to running on physical hardware?

[1:27 PM]  
tom googles Dot1Q...

[1:27 PM]  
tom
> Dot1Q is an IEEE's open standard, which be used to create trunk connection between switches of different vendors. In ISL the original ethernet frame is not modified, it is encapsulated between an ISL header and an FCS.
```

2opremio:
Fixes #2155

I haven't been able to test this properly though, since I don't know how obtain a Dot1Q interface.

### DIFF
--- a/probe/endpoint/dns_snooper_linux_amd64.go
+++ b/probe/endpoint/dns_snooper_linux_amd64.go
@@ -15,15 +15,17 @@ import (
 )
 
 const (
-	bufSize              = 8 * 1024 * 1024 // 8MB
-	maxReverseDNSrecords = 10000
+	bufSize                 = 8 * 1024 * 1024 // 8MB
+	maxReverseDNSrecords    = 10000
+	maxLogsPerDecodingError = 4
 )
 
 // DNSSnooper is a snopper of DNS queries
 type DNSSnooper struct {
-	stop            chan struct{}
-	pcapHandle      *pcap.Handle
-	reverseDNSCache gcache.Cache
+	stop                chan struct{}
+	pcapHandle          *pcap.Handle
+	reverseDNSCache     gcache.Cache
+	decodingErrorCounts map[string]uint64 // for limiting
 }
 
 // NewDNSSnooper creates a new snooper of DNS queries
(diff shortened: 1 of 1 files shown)


probe/endpoint/dns_snooper_linux_amd64.go
```diff
@@ -203,6 +207,20 @@ func (s *DNSSnooper) run() {
 	}
 }
 
+// handleDecodeError logs errors up to the maximum allowed count
+func (s *DNSSnooper) handleDecodingError(err error) {
+	str := err.Error()
+	count := s.decodingErrorCounts[str]
+	count++
+	s.decodingErrorCounts[str] = count
```
rade: if the error string is variable, e.g. contain some id, or a timestamp, etc, etc, then this becomes a memory leak.

2opremio: I can limit the map size.

2opremio: Ideally we should use a size-limited cache with time-based expiration on keys. That's overkill though. We are running into this problem quite often, so I may consider creating a library for this.

probe/endpoint/dns_snooper_linux_amd64.go
```diff
@@ -203,6 +208,25 @@ func (s *DNSSnooper) run() {
 	}
 }
 
+// handleDecodeError logs errors up to the maximum allowed count
+func (s *DNSSnooper) handleDecodingError(err error) {
+	// prevent potential memory leak
+	if len(s.decodingErrorCounts) > maxDecodingErrorCardinality {
```
ekimekim: optional: maybe have a flag for "reached max cardinality", so the first time we reach it we can emit a final `DNSSnooper: Too many unique error messages, silencing all errors` or similar?

2opremio: If we ever get there we would need to rethink the approach. It's a bit of a hack.

PR #2166, opened 2017-01-30
scope: fix URL

alban:
Symptoms: the Scope URL are printed this way:

```
  * http://10.2.2.1
172.16.28.1
172.16.28.1
192.168.99.1
192.168.35.117
172.16.28.1
172.16.28.1
192.168.122.1:4040/
```

Regression from https://github.com/weaveworks/scope/pull/2068


2opremio:
Thanks @alban !!!

2opremio:
LGTM, I will merge when green

PR #2177, opened 2017-02-01
Node details spinner Chrome display bug fix

### LINKED ISSUE #2173: Details panel spinner looks odd
See video: http://recordit.co/U86TtCFcbA

fbarl:
Fixes #2173.

### DIFF
--- a/client/app/scripts/components/node-details.js
+++ b/client/app/scripts/components/node-details.js
@@ -1,5 +1,6 @@
 import debug from 'debug';
 import React from 'react';
+import classNames from 'classnames';
 import { connect } from 'react-redux';
 import { Map as makeMap } from 'immutable';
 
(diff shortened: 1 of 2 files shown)


client/app/scripts/components/node-details.js
```diff
@@ -73,6 +74,10 @@ class NodeDetails extends React.Component {
   renderLoading() {
     const node = this.props.nodes.get(this.props.nodeId);
     const label = node ? node.get('label') : this.props.label;
+    // NOTE: If we start the fa-spin animation before the node details panel has been
+    // mounted, the spinner is displayed blurred the whole time in Chrome (possibly
+    // caused by a bug having to do with animating the details panel).
+    const spinnerClassName = classNames('fa fa-circle-o-notch', { 'fa-spin': this.props.mounted });
```
jpellizzari: Should we be checking that the `<NodeDetails />` component is mounted instead of its parent?

fbarl: For one thing, it's the `details-card` that gets animated and that animation was supposedly causing the problem, but also I didn't want to copy the mounting logic to `node-details` and having it in two places. Your suggestion seems more *reasonable* somehow, so I'm not sure which approach I like better in the end. I'd say it doesn't matter though, as this code will be gone soon as we switch to @bowenli's new spinners.

jpellizzari: Ok that makes sense. No sense adding state to node-details then.

jpellizzari:
Sorry for the delay on this review @fbarl. LGTM

PR #2180, opened 2017-02-02
Re-applied "Graph layout optimizations"

fbarl:
Brings back the reverted changes from #2128 and introduces a fix for node labels rendering on Firefox.


foot:
LGTM!

- Default zoom for small layouts (e.g. 2-3 nodes) is very high. I understand the zoom is set so that the nodes fit + 10% border. This works great for large layouts. Too accomodate smaller ones we could perhaps `Math.min(calculatedZoomWith10%Border, aNiceAestheticNodeZoom);`

fbarl:
@foot, I did the change you proposed, essentially with `aNiceAestheticNodeZoom = maxZoomScale / 2` and it does look better :)

Issue #2063, opened 2016-12-07
Improve how ECS reporting code handles clusters


ekimekim:
Currently, the code is very naive - it will fetch info for each cluster, but there's some issues I can foresee (untested):
* If two services in different clusters shared a service name, it would get confused since the nodes would have the same id (we should be using ARNs for this, and we do for tasks, but not services) (this is a correctness issue so I'm flagging this as `bug`)
* We currently fetch each cluster's info in sequence, ie. we don't start fetching cluster B until cluster A is fully processed. This could cause bad report building latency should the user have many clusters. (this is solely a performance issue)
* We currently make one ECS connection object per cluster, which prevents connection pooling etc.
* Along with the above, if any one of these ECS connections fails to be created, we give up on all remaining clusters also.

PR #2186, opened 2017-02-03
Rename ECS Service node ids to be cluster;serviceName

ekimekim:
This is important for two reasons:
* It prevents nasty false-equality bugs when two different services from different ECS clusters
  are present in the same report
* It allows us to retrieve the cluster and service name - all the info we need to look up the service -
  using only the node ID. This matters, for example, when trying to handle a control request.

We deal with maintaining backwards compatability with a special case when parsing the IDs.

Partially fixes https://github.com/weaveworks/scope/issues/2063

### DIFF
--- a/report/id.go
+++ b/report/id.go
@@ -77,6 +77,11 @@ func MakeProcessNodeID(hostID, pid string) string {
 	return hostID + ScopeDelim + pid
 }
 
+// MakeECSServiceNodeID produces an ECS Service node ID from its composite parts.
+func MakeECSServiceNodeID(cluster, serviceName string) string {
+	return cluster + ScopeDelim + serviceName
+}
+
 var (
 	// MakeHostNodeID produces a host node ID from its composite parts.
 	MakeHostNodeID = makeSingleComponentID("host")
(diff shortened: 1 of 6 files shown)


ekimekim:
Failing test is due to integration test that's broken on master.

report/id.go
```diff
@@ -204,6 +203,20 @@ func ParseAddressNodeID(addressNodeID string) (hostID, address string, ok bool)
...
 }
 
+// ParseECSServiceNodeID produces the cluster, service name from an ECS Service node ID
+func ParseECSServiceNodeID(ecsServiceNodeID string) (cluster, serviceName string, ok bool) {
+	fields := strings.SplitN(ecsServiceNodeID, ScopeDelim, 2)
+	if len(fields) != 2 {
+		return "", "", false
+	}
+	// In previous versions, ECS Service node IDs were of form serviceName + "<ecs_service>".
+	// For backwards compatibility, we should still return a sensical serviceName for these cases.
```
2opremio: Can you add a backwards compatibility test?

ekimekim: done

ekimekim:
ptal

PR #2189, opened 2017-02-03
don't attempt to make 'make clean' work on old checkouts

### LINKED ISSUE #2188: `make clean` exlodes with "unexpected operator"
```
$ make clean
env GOGC=off go clean ./...
# build-external was accidentially created as owned by root in earlier versions. This is fixed now,
# but we need to fix up old checkouts, since git doesn't track ownership.
if [ "$(stat -c %%U client/build-external)" == "root" ]; then sudo chown "$USER:$USER" client/build-external; fi
/bin/sh: 1: [: %U: unexpected operator
# Don't actually rmi the build images - rm'ing the .uptodate files is enough to ensure
# we rebuild the images, and rmi'ing the images causes us to have to redownload a lot of stuff.
#  docker rmi weaveworks/scope-ui-build weaveworks/scope-backend-build >/dev/null 2>&1 || true
rm -rf scope.tar .scope_ui_build.uptodate .scope_backend_build.uptodate \
	prog/scope vendor/runsvinit/runsvinit prog/staticui/staticui.go prog/externalui/externalui.go client/build/*.js client/build-external/*.js docker/weave .pkg \
	report/report.codecgen.go render/detailed/detailed.codecgen.go vendor/github.com/ugorji/go/codec/codecgen/bin
```

rade:
It's just not worth the extra complication in the Makefile, and the code for it was actually broken.

Also, outdent comment so it doesn't get splattered onto the console.

Fixes #2188.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,12 +174,9 @@ ui-build-pkg:
 
 clean:
 	$(GO) clean ./...
-	# build-external was accidentially created as owned by root in earlier versions. This is fixed now,
-	# but we need to fix up old checkouts, since git doesn't track ownership.
-	if [ "$$(stat -c %%U client/build-external)" == "root" ]; then sudo chown "$$USER:$$USER" client/build-external; fi
-	# Don't actually rmi the build images - rm'ing the .uptodate files is enough to ensure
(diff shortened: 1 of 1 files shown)


ekimekim:
You'll need to rebase to get tests to pass

Issue #2182, opened 2017-02-02
Master branch is failing on CI


fbarl:
Integration tests failing with the message `ERROR: weave requires Docker version 1.10.0 or later; you are running 1.8.2`. See e.g. https://circleci.com/gh/weaveworks/scope/6292

@2opremio (from Slack):
> Probably the tests are using weave:latest
> And latest now requires 1.10 as a minimum

PR #2190, opened 2017-02-03
Embed and require Docker >= 1.10

2opremio:
Fixes https://github.com/weaveworks/scope/issues/2182

Matches https://github.com/weaveworks/weave/pull/2713 in Weave Net


circle.yml
```diff
@@ -4,6 +4,8 @@ general:
       - gh-pages
 
 machine:
+  pre:
+    - curl -sSL https://s3.amazonaws.com/circle-downloads/install-circleci-docker.sh | bash -s -- 1.10.0
```
fbarl: This line is only necessary until CircleCI switches to a newer version of Docker, right? Maybe add a comment to remove it when that happens (not sure how much this step is slowing down the whole build).

2opremio: It is necessary now, because by default it runs docker < 1.10

Issue #2181, opened 2017-02-02
Already closed connections showing up in the containers tab


iaguis:
## Steps to reproduce

1. Start a busybox docker container and connect to the internet:

```
$ docker run -it busybox
# nc google.com 80
```

2. Wait until the edge `container -> internet` is shown in the containers tab.

3. Close the connection e.g. with Ctrl-C

4. Start a second busybox docker container and connect to the internet again:

```
$ docker run -it busybox
# nc google.com 80
```

5. See that the UI shows both containers connected to the internet

If I switch to another tab (e.g. Processes) and back to the Containers one the problem is fixed: only the connection from the second container shows up so I guess it's a problem in the UI layer.

2opremio:
@iaguis Does reloading the UI (i.e. reloading the browser page) make the connection go away?

iaguis:
Yes, reloading the browser page makes the connection go away.

Even changing to another view and going back to "Containers" makes it go away as I say in the OP.

2opremio:
Then it must be a UI problem. Thanks for the confirmation.

jpellizzari:
The root cause of this appears to be the way we are merging updates. The `adjacency` property is not present on the node object after a connection has been removed. When we call `.merge()`, it keeps the old one because the previous state has adjacency, and does not delete the adjacency key.

jpellizzari:
Fixed by #2194

PR #2194, opened 2017-02-03
Change merge to re-assign when nodes are updated

jpellizzari:
Fix for #2181 

Changes the update function in RECEIVE_NODES_DELTA to completely replace the node, rather than merging it with existing properties. This was causing old adjacencies to hang around because they were not being overwritten by updates. 

Since the `adjacency` key is not present in in updates, a merge would not remove they key from the UI state. `Immutable.merge()` only affects keys that are defined in the source object.

This will remove any extra properties that get written by the UI to the `nodes`. As far as I can tell, there is nothing that mutates the values in the `nodes` key. It would be nice to be able to try an `Object.freeze` on that key to throw runtime errors if something tries to add keys or mutate it, but the Immutable lib doesn't support that.

@davkal Please review. 

cc @2opremio

### DIFF
--- a/client/app/scripts/reducers/__tests__/root-test.js
+++ b/client/app/scripts/reducers/__tests__/root-test.js
@@ -500,4 +500,29 @@ describe('RootReducer', () => {
     nextState = reducer(nextState, {type: ActionTypes.SET_RECEIVED_NODES_DELTA});
     expect(nextState.get('gridMode')).toBe(true);
   });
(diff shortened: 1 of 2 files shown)


davkal:
Yup, works great. That was an ambiguity in the BE/FE contract (merge would have worked if the BE sent an empty list). This also should have no side effects on the layouts, because the coords are kept in a separate state object.

LGTM

Issue #2051, opened 2016-12-02
Add controls for ECS views


2opremio:
We should think about what exact controls to add but here go a few ideas:

* Edit replication factor of services (+ and - controls like we do in k8s)
* Edit/view task definition
* Aggregated logs for service/task
* ...

ekimekim:
Comparing to what we currently have for k8s objects:
* Pod: view logs, delete
* Replica Set: increase/decrease size by one
* Deployment: increase/decrease size by one

The scale up/down controls will carry over simply and easily, so I'll concentrate on that first. Deleting tasks can also be a thing, but I don't know if there's a simple way to just "view logs" of the task.
In any case, all these are going to require additional permissions for the IAM role. I'll try to make a point of gracefully failing if they aren't given. Maybe there's some way i can test for the permission and make a "greyed out" control if not present?

Issue #2085, opened 2016-12-12
ECS: Render Taskless Services


2opremio:
Right now (after #2026) we only "discover" ECS services through tasks (we obtain the task list from Docker labels and obtain services associated to those tasks). 

This means we don't obtain information nor render services without associated tasks (e.g. Services with replication factor 0, which is valid, and services whose tasks are not running).

This may or may not be a real problem for the users.

2opremio:
I think that a regular call to [`ListServices`](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ListServices.html) wouldn't add a lot of overhead considering we regularly call `DescribeService` for every service.

ekimekim:
Well, as of #2065 we only describe services that have tasks running on our machine. I like this because it means the number of calls is bounded by the number of things running per machine, whereas describing every service on every machine would be problematic in very large clusters.
Sigh. These rate limits make everything difficult.

Given the current constraints we may need to accept not supporting showing non-running services for now.
Ideally we should document that somewhere.

PR #2197, opened 2017-02-06
Add ECS Service scale up/down controls

ekimekim:
Adds Scale Up / Scale Down controls for ECS services in the same manner as Kubernetes Replica Sets/Deployments.

Note this requires an additional permission "ecs:UpdateService". Documentation will need to be changed to include this.
However I don't know where that particular documentation is kept (perhaps an argument for keeping the documentation in-repo,
so it can be modified in conjunction with code changes instead of as an afterthought).

Partially addresses https://github.com/weaveworks/scope/issues/2051

### DIFF
--- a/probe/awsecs/reporter.go
+++ b/probe/awsecs/reporter.go
@@ -153,7 +153,12 @@ func (r Reporter) Tag(rpt report.Report) (report.Report, error) {
 				ServiceDesiredCount:   fmt.Sprintf("%d", service.DesiredCount),
 				ServiceRunningCount:   fmt.Sprintf("%d", service.RunningCount),
 				report.ControlProbeID: r.probeID,
-			}).WithLatestActiveControls(ScaleUp, ScaleDown))
+			}).WithLatestControls(map[string]report.NodeControlData{
+				ScaleUp: {Dead: false},
+				// We've decided for now to disable ScaleDown when only 1 task is desired,
+				// since scaling down to 0 would cause the service to disappear (#2085)
+				ScaleDown: {Dead: service.DesiredCount <= 1},
+			}))
 		}
 		log.Debugf("Created %v ECS service nodes", len(ecsInfo.Services))
 
--- a/probe/awsecs/reporter.go
+++ b/probe/awsecs/reporter.go
@@ -5,6 +5,8 @@ import (
 	"time"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/weaveworks/scope/common/xfer"
+	"github.com/weaveworks/scope/probe/controls"
 	"github.com/weaveworks/scope/probe/docker"
 	"github.com/weaveworks/scope/report"
 )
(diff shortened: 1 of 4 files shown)


ekimekim:
rebased onto master now that https://github.com/weaveworks/scope/pull/2186 is merged.

ekimekim:
Tests failed due to https://github.com/weaveworks/scope/pull/2220, retrying for now since this obviously isn't happening in most test runs.

probe/awsecs/reporter.go
```diff
@@ -126,10 +149,11 @@ func (r Reporter) Tag(rpt report.Report) (report.Report, error) {
...
 			rpt.ECSService = rpt.ECSService.AddNode(report.MakeNodeWith(serviceID, map[string]string{
-				Cluster:             cluster,
-				ServiceDesiredCount: fmt.Sprintf("%d", service.DesiredCount),
-				ServiceRunningCount: fmt.Sprintf("%d", service.RunningCount),
-			}))
+				Cluster:               cluster,
+				ServiceDesiredCount:   fmt.Sprintf("%d", service.DesiredCount),
+				ServiceRunningCount:   fmt.Sprintf("%d", service.RunningCount),
+				report.ControlProbeID: r.probeID,
+			}).WithLatestActiveControls(ScaleUp, ScaleDown))
```
2opremio: Is `ScaleDown` disabled when the number of tasks is 0? (which is possible)

ekimekim: I figured there's not much point in doing that until https://github.com/weaveworks/scope/issues/2085 is fixed. Happy to do so as a matter of completeness/future-proofing though.

2opremio: Then with more reason. In the current state of things, reducing the number of tasks to 0 it would make the service disappear, right?

Then I would set a minimum number of tasks to 1 (not displaying the minus button if that's the case)

ekimekim: Ack. Is there a way to have the minus button there but greyed out? I know I can set it "dead" but that makes it disappear, not "disabled", iiuc?

2opremio:
Not AFAIK. But to be consistent with the rest (e.g. stop button for
containers) I think making it disappear is the right thing

2opremio:
@ekimekim ping

ekimekim:
PTAL

probe/awsecs/reporter.go
```diff
@@ -153,7 +153,12 @@ func (r Reporter) Tag(rpt report.Report) (report.Report, error) {
 				ServiceDesiredCount:   fmt.Sprintf("%d", service.DesiredCount),
 				ServiceRunningCount:   fmt.Sprintf("%d", service.RunningCount),
 				report.ControlProbeID: r.probeID,
-			}).WithLatestActiveControls(ScaleUp, ScaleDown))
+			}).WithLatestControls(map[string]report.NodeControlData{
+				ScaleUp: {Dead: false},
+				// We've decided for now to disable ScaleDown when only 1 task is desired,
+				// since scaling down to 0 would cause the service to disappear (#2085)
+				ScaleDown: {Dead: service.DesiredCount <= 1},
```
2opremio: You should also check this in the control itself, and not allow rescaling below 1 in case the UI gets out of sync

ekimekim: done (changed it from checking against zero)

ekimekim:
ptal

PR #2204, opened 2017-02-08
Added dynamic contrast-mode toggle

jpellizzari:
Removes the need to load the contrast.html page to get contrast mode. This will be necessary for scope-as-a-component. Clicking the contrast button should toggle the theme immediately, instead of loading a new page. There is a key in the url state that will save the user's preference on page reload.

@fbarl Could you take a look please?


client/app/scripts/utils/router-utils.js
```diff
@@ -19,11 +19,18 @@ function encodeURL(url) {
     .replace(new RegExp(SLASH, 'g'), SLASH_REPLACEMENT);
 }
 
-function decodeURL(url) {
+export function decodeURL(url) {
   return decodeURIComponent(url.replace(new RegExp(SLASH_REPLACEMENT, 'g'), SLASH))
     .replace(new RegExp(PERCENT_REPLACEMENT, 'g'), PERCENT);
 }
 
+export function parseHashState(hash = window.location.hash) {
```
fbarl: Nice! Just a tiny observation that `parseStateHash` sounds like a better name to me but it's up to you :)

client/app/scripts/reducers/root.js
```diff
@@ -704,6 +706,7 @@ export function rootReducer(state = initialState, action) {
       }
       state = state.set('topologyOptions',
         fromJS(action.state.topologyOptions) || state.get('topologyOptions'));
+      state = state.set('contrastMode', action.state.contrastMode);
```
fbarl: Also not sure this one here is necessary..

client/app/scripts/components/footer.js
```diff
@@ -76,7 +79,7 @@ class Footer extends React.Component {
             title={forceRelayoutTitle}>
             <span className="fa fa-refresh" />
           </a>
-          <a className="footer-icon" href={otherContrastModeUrl} title={otherContrastModeTitle}>
+          <a onClick={this.handleContrastClick} className="footer-icon" href="" title={otherContrastModeTitle}>
```
fbarl: Is it necessary to put an explicit `href=""` here? Just looking at the layout refresh `<a>` above which doesn't have it..

client/app/scripts/reducers/root.js
```diff
@@ -721,6 +724,34 @@ export function rootReducer(state = initialState, action) {
...
+      modules.forEach((i) => {
+        const children = require.cache[i] ? require.cache[i].children : [];
+        children.forEach((c) => {
+          delete require.cache[c];
+        });
+        delete require.cache[i];
+      });
+
+      const contrastEnabled = state.get('contrastMode');
+      if (!contrastEnabled || action.enabled) {
```
fbarl: Shouldn't it be `&&` instead of `||` or am I missing something obvious here? :) Also, in line `746` I'd put a similar (symmetrical) condition - I assume the idea here is that `require.ensure` would be executed if and only if `contrastEnabled !== action.enabled`, right?

client/app/scripts/reducers/root.js
```diff
@@ -721,6 +724,34 @@ export function rootReducer(state = initialState, action) {
...
+        require.resolve('../../styles/contrast.scss')
+      ];
+      // Bust the webpack require cache to for a re-download of the stylesheets
+      modules.forEach((i) => {
+        const children = require.cache[i] ? require.cache[i].children : [];
+        children.forEach((c) => {
+          delete require.cache[c];
+        });
+        delete require.cache[i];
+      });
```
fbarl: Webpack magic :) The only thing I hope is that it will work in production as well as it does for me locally ^^

jpellizzari: It __should__. Unfortunately, there is no way to test ahead of time.

client/app/scripts/reducers/root.js
jpellizzari: This says "If contrast mode is currently or if the user explicitly says they want to turn it ondisabled, turn it on". The action is used for page reload only. 

You are right, there should be inverse condition on the normal theme. I will add.

client/app/scripts/utils/router-utils.js
jpellizzari: I was thinking that we are parsing the hash's state, not the state's hash.

client/app/scripts/components/app.js
```diff
@@ -46,6 +46,9 @@ class App extends React.Component {
       getTopologies(this.props.activeTopologyOptions, this.props.dispatch);
     }
     getApiDetails(this.props.dispatch);
+    if (this.props.contrastMode) {
+      this.props.toggleContrastMode(true);
+    }
```
jpellizzari: This only fires once when the application load. It is mainly just there to ensure that if `{ contrastMode: true }` is in the URL it will get toggled.

client/app/scripts/utils/router-utils.js
fbarl: Alright, makes sense!

client/app/scripts/components/app.js
fbarl: Does it not work for you as intented if you comment out these lines? For me it seems to work just fine still and I thought that was because it always gets triggered through the `route` action in `app-actions`.

client/app/scripts/reducers/root.js
fbarl: > "If contrast mode is currently or if the user explicitly says they want to turn it on, turn it on"

My understanding is that this sentence translates to `contrastEnabled || action.enabled`, in which case there's an extra `!` at the beginning of your *if*-clause, but even then I'm not sure why we should care for `contrastEnabled` part in here at all.

The current conditions seem to overlap whenever `contrastEnabled === action.enabled` in which case the contrast styles would always take advantage, which seems a bit arbitrary. The code seems to work just fine though and I'm not sure why :D But I still believe these conditions could be simplified a bit! :)

jpellizzari: Sorry I forgot a word. It should have been "If contrast mode is currently __OFF__ or if the user explicitly says they want to turn it on, turn it on"

jpellizzari:
@fbarl Ok I will keep that in mind about the PR commits. Just means I have to squash before merging, which is something I'm sure I will forget to do on more than one occasion.

client/app/scripts/reducers/root.js
fbarl: That makes sense but I think you still missed a part of my comment :) Let me put it this way:

`contrastEnabled && action.enabled` -> `contrast.scss`
`!contrastEnabled && !action.enabled` -> `contrast.scss`
`!contrastEnabled && action.enabled` -> `contrast.scss`
`contrastEnabled && !action.enabled` -> `main.scss`

3/4 scenarious would result in `contrast` styles being chosen over `main` which feels wrong. I tried ignoring the `contrastEnabled` variable and this seemed to work just fine:

```
if (action.enabled) {
  require.ensure([], () => {
    require('../../styles/contrast.scss');
  });
} else {
  require.ensure([], () => {
    require('../../styles/main.scss');
  });
}
```

Could you perhaps also verify that on your side? Not sure I'm missing some special circumstances like hot-reloading, etc...

jpellizzari: Yes that would work, so long as the user of the API call explicitly passes in `true` to the `toggleContrastMode`. My idea was that there are two scenarios: 
1. The user explicitly wants contrast mode
2. The user wants to toggle the between the two states, and doesn't necessarily care what the current state is. So calling `toggleContrastMode()` without an arg would switch from one to the other.

But then I realized that's stupid, so I simplified it with your recommendation :)

fbarl:
There is one more detail @davkal and I observed that might be nice to address in this PR before you merge - when toggling the contrast mode, some of the components ought to be rerendered so that the JS controlled styles get applied as well. See for example how the edge thickness doesn't change until you do the action that rerenders the graph.

The components in question are basically the ones that are using `isContrastMode` helper: `edge.js`, `node-networks-overlay.js`, `node-shape-stack.js`. You could try simply binding the `contrastMode` global state variable to those components in `mapStateToProps` which should do the trick.

jpellizzari:
@fbarl Nice catch. Here is what my nodes look like now:
screen shot 2017-02-13 at 9 50 34 am

jpellizzari:
@fbarl I was worried about the same thing, and even did an iteration where the top level components passed in `contrastMode` as a prop, but it didn't make sense.

PR #2206, opened 2017-02-08
extras/dialer: allow to seed random

schu:
Helpful if you want to get the same random numbers in subsequent runs
(e.g. to compare results between branches).


schu:
cc @alepuccetti

PR #2210, opened 2017-02-08
Fix for node stack display bug

### LINKED ISSUE #2207: Stacks aren't rendering
I think both of these should be rendered as stacks?

screen shot 2017-02-08 at 15 03 14

fbarl:
Fixes #2207. Forgot to rescale the stack to the `NODE_BASE_SIZE` when doing the graph layout optimizations in #2180.


foot:
LGTM

PR #2220, opened 2017-02-09
tools/sched: Fix unicode encoding issues

ekimekim:
The response.text and json-parsed string values are both unicode strings.
The print statement requires bytes, as per write(2).
Python is dumb and bad at converting between them properly - in the name of being "safe"
it errors if any non-ascii characters are present, so it works until it doesn't.
We fix this by explicitly encoding as utf-8.

### DIFF
--- a/tools/sched
+++ b/tools/sched
@@ -1,20 +1,20 @@
 #!/usr/bin/python
-import sys, string, json, urllib
+import sys, string, urllib
 import requests
 import optparse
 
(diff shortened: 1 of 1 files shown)


2opremio:
`tools` is a subtree of https://github.com/weaveworks/build-tools , please fix is there instead

ekimekim:
ah, dammit. the name caught me out.

Issue #1353, opened 2016-04-19
Nodes layout runner caches node metadata


davkal:
`nodes-layout` has a node cache that copies node data from previous layout runs to the incoming nodes. But it's not just coordinates that are copied, but also node metadata, like labels and metrics.

This came up in a container rename issue #1091, and its PR #1332 which simply re-applies the metadata after the layout run (that should be remedied).

rade:
Is this related to #1829?

davkal:
> Is this related to #1829?

No. The node is augmented with metrics after the layout step.

PR #2221, opened 2017-02-09
Layout rendering dynamic optimizations

fbarl:
In response to #1188:

**Most relevant optimizations**

* Translation/scaling of the nodes is now fully managed on the `NodeContainer` component level, so nodes are only rerendered when they change internally and not when they are repositioned in the layout.
* Caching the metrics that is displayed inside the nodes in the selectors.
* The search matching logic has also been partly moved to selectors and the difference can be felt when searching through the big graphs. Also, excessive recalculations in some special circumstances (e.g. when resfreshing the page with a broad search query) have been eliminated.

**Layout changes**

* Slightly decreased the distances between the nodes in the graph layout (to me it looks better this way).
* Doubled the treshold in the graph complexity heuristics as I believe we can now handle bigger graphs :)

**Other changes**

* Renamed `subLabel` to `labelMinor` everywhere for consistency with the grid mode and the backend naming.

**Some ideas for the future**

* Move `zoomCache` from `NodesChart` to a global state (would also solve #2254).
* Make deep updates to the nodes when receiving delta and select only adjacencies when doing the graph layout (would solve #1353 again).
* Make all selectors *pure*, i.e. dependent only on the global state.


client/app/scripts/charts/node.js
```diff
@@ -55,29 +54,20 @@ class Node extends React.Component {
...
   }
 
-  componentWillReceiveProps(nextProps) {
-    // marks as matched only when search query changes
-    if (nextProps.searchQuery !== this.props.searchQuery) {
-      this.setState({ matched: nextProps.matched });
-    } else {
-      this.setState({ matched: false });
-    }
-  }
```
fbarl: **Note to myself**: Put back some of this code as the `throb` animation doesn't get triggered now.

client/app/scripts/selectors/nodes-chart.js
```diff
@@ -0,0 +1,32 @@
...
+
+export const nodesSelector = createSelector(
+  [
+    allNodesSelector,
+  ],
+  allNodes => allNodes.filter(node => !node.get('filtered'))
+);
+
+
+export const nodeAdjacenciesSelector = createSelector(
```
davkal: With the removal of `returnPreviousRefIfEqual()` when creating the selector, this will always need to create a new node map.

Also, this selector has now merged the adjacency data (id + adjacency) with the other node metadata like metrics. The previous split was to make sure that the layout engine only gets called if the adjacency data changed by being dependent only on the adjacency selector.

client/app/scripts/charts/nodes-chart-elements.js
```diff
@@ -15,18 +14,12 @@ class NodesChartElements extends React.Component {
...
 }
 
-function mapStateToProps(state, props) {
-  return {
-    completeNodes: completeNodesSelector(state, props)
-  };
-}
-
-export default connect(mapStateToProps)(NodesChartElements);
+export default connect()(NodesChartElements);
```
davkal: Probably better to use `React.PureComponent` here than to `connect()`.

client/app/scripts/reducers/__tests__/root-test.js
```diff
@@ -64,7 +64,7 @@ describe('RootReducer', () => {
       }
     ],
     stats: {
-      edge_count: 319,
+      edge_count: 379,
```
davkal: What changed here?

fbarl: I increased the constant in `graphExceedsComplexityThresh` so the definition of a *complex* graph changed.

client/app/scripts/charts/nodes-chart-elements.js
fbarl: Nice! I did the same with all the other components as well, made them `PureComponent` instead of using empty Redux `connect()`.

client/app/scripts/selectors/nodes-chart.js
```diff
@@ -0,0 +1,32 @@
...
+import { Map as makeMap } from 'immutable';
+
+
+const allNodesSelector = state => state.get('nodes');
+
+export const nodesSelector = createSelector(
+  [
+    allNodesSelector,
+  ],
+  allNodes => allNodes.filter(node => !node.get('filtered'))
```
fbarl: Great point - that's definitely something to keep an eye on, so I'll add a comment in the code.

fbarl: Ok, I misunderstood your comment, thinking that every element that passes the filter gets recreated, but if only the returned collection is new, this actually isn't be a problem as long as we use this selector as input to *map selectors*, which implement per-element caching. There are actually a couple of places in the code where this is the case, so I won't put a special comment here.

client/app/scripts/selectors/nodes-chart.js
fbarl: That's a good point. I put an extensive `TODO` in the code that explains how we could exploit new selectors for smarter layout engine caching, but one prerequisite here would be to implement *deep update* of the nodes in `reducers/root.js`, where currently we are just doing:

```
state = state.setIn(['nodes', node.id], fromJS(node))
```

A consequence of that single line is that everything about the node always gets updated, so the case e.g. of node metrics being updated but the adjacencies not being modified actually never happens.

fbarl:
@davkal PTAL, maybe just at the last commit, just to check if I've identified all the points you raised in the comments.

client/app/scripts/charts/node.js
```diff
@@ -93,37 +83,37 @@ class Node extends React.Component {
...
+    const { hovered } = this.state;
 
     const color = getNodeColor(rank, label, pseudo);
     const truncate = !focused && !hovered;
     const labelOffsetY = (showingNetworks && networks) ? 40 : 28;
 
     const nodeClassName = classnames('node', {
+      // NOTE: Having a CSS animation here might not be the best idea.
+      // See https://github.com/weaveworks/scope/issues/2255
+      matched: !matches.isEmpty(),
```
jpellizzari: I'm getting an error where sometimes `matches` is undefined, which causes `matches.isEmpty()` to throw. Maybe check that it exists before calling `isEmpty`.

fbarl: Good catch! I reverted to `const { matches = makeMap() } = this.props`.

fbarl:
Thanks for the review and the benchmarks @jpellizzari! I didn't expect the numbers to be so much better with my PR, but I think some of it is compensated with the slowness of deep comparison of the props (which now happens more often I think).

PR #2224, opened 2017-02-09
Reintroduce probe checkpoint flags for kernel version and OS

2opremio:
Removed unintentionally in https://github.com/weaveworks/scope/pull/2148

Unfortunately this means we won't get kernel version information for the probes until the next Scope release :(


2opremio:
CC @alban @iaguis

PR #2225, opened 2017-02-10
integration tests: fix scripts

alban:
bin/setup-circleci-secrets: fix shell "unbound variable"

Since this shell script uses "set -u", we must not use unset variables directly.

Symptoms:
> https://circleci.com/gh/weaveworks/scope/6384
> bin/setup-circleci-secrets: line 10: SECRET_SCRIPT: unbound variable

Instead, we should use the parameter expansion "+" to test if the variable is set. See:
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

-----

Also, fix GCE --zones parameter

-----

@2opremio Sorry I broke the CircleCI build on master. I am marking this PR as WIP because I want to do more tests.


alban:
@2opremio the integration tests passes on: https://circleci.com/gh/alban/scope/109

Could you try this branch on weaveworks CircleCI config? If it passes as well, I can pick the changes from tools/ and make a pull request on https://github.com/weaveworks/build-tools

2opremio:
> Could you try this branch on weaveworks CircleCI config?

https://circleci.com/gh/weaveworks/scope/6416

alban:
This still failed on weaveworks CircleCI config. I updated this PR with the fix:
```
-    read -r -d '' SECRET_SCRIPT <<EOF
+    # "read" return a non-zero value when reading until EOF, see "help read"
+    read -r -d '' SECRET_SCRIPT <<EOF || true
```

It seems to work for:
- GCE not configured: https://circleci.com/gh/kinvolk/scope/679
- GCE configured: https://circleci.com/gh/alban/scope/110

@2opremio could you try again with the fix on the weaveworks CircleCI config?

alban:
After https://github.com/weaveworks/build-tools/pull/75 gets merged, this PR can be updated with git-subtree.

alban:
The new build-tools bring other changes that require to update circle.yml. I am trying to figure this out.

alepuccetti:
We updated the new build-tools. Because of that circleCI needs to use ubuntu 14.04 as building environment. Also, we fixed few regressions.
Here there is the link to the circleCI with the integration test executed on gce https://circleci.com/gh/alepuccetti/scope/162

This PR does not depend on https://github.com/weaveworks/build-tools/pull/79

/cc @2opremio

alepuccetti:
This PR is a blocker for https://github.com/weaveworks/scope/pull/2135

backend/Dockerfile
```diff
@@ -10,8 +10,10 @@ RUN go get -tags netgo \
 		github.com/golang/lint/golint \
 		github.com/kisielk/errcheck \
 		gopkg.in/mvdan/sh.v1/cmd/shfmt \
+		github.com/fatih/hclfmt \
 		github.com/mjibson/esc \
 		github.com/client9/misspell/cmd/misspell && \
+	chmod a+wr --recursive /usr/local/go/pkg && \
```
2opremio: Why  do we need this?

2opremio:
I am testing this at https://circleci.com/gh/weaveworks/scope/6550

2opremio:
And it failed

alepuccetti:
@2opremio 

> And it failed

> Because of that circleCI needs to use ubuntu 14.04 as building environment

It failed because your WW CircleCI is still using ubuntu 12.04 as building environment, please switch to ubuntu 14.04.

backend/Dockerfile
alepuccetti: This is necessary to allow `go install net` (https://github.com/weaveworks/scope/pull/2225/commits/f850f4486aa37639edd6eb8b86e21e64b60ef9b6)

2opremio:
Just changed it and rebuilt, let's see what happens

2opremio:
Failed again

2opremio:
Rebuilding without cache in case that was the problem ...

2opremio:
nah

alepuccetti:
@2opremio I need to re-push it. rebuilding is not enough in case you change the building env. This is a CircleCI limitation

2opremio:
_Please note that you need to trigger a build by pushing commits to GitHub (instead of rebuilding) to apply the new setting._

Doing that with a phony commit

alepuccetti:
CircleCI fails while performing `lint` because it runs lint on the `.git/` directory. I had this issue before and I fixed it just by rebuild.

2opremio:
@alepuccetti It did fix it but it's flaky, can you make sure to explicitly exclude that directory?

2opremio:
I am merging this to unblock you but please look into the linting problems seen above.

Issue #1804, opened 2016-08-16
Changing between topologies feels slow


davkal:
- the load indicator might take too long to show up?
- if a topo has been seen before, and i switch back to it, i should be able to see it right away.

foot:
Loading indicator only show for initial load at the moment, shouldn't be hard to change..

jpellizzari:
Closed by #2232

PR #2232, opened 2017-02-10
Show loading indicator on topology changes

jpellizzari:
Implements #1804 

When a user changes topologies via click on the topology filter, or by clicking 'SHOW IN <topology>', the loading indicator will show after 1 second. It will be cleared by the first nodes delta data arriving via websockets.

### DIFF
--- a/client/app/scripts/reducers/__tests__/root-test.js
+++ b/client/app/scripts/reducers/__tests__/root-test.js
@@ -499,6 +499,7 @@ describe('RootReducer', () => {
     let nextState = initialState.set('currentTopology', fromJS(topologies[0]));
     nextState = reducer(nextState, {type: ActionTypes.SET_RECEIVED_NODES_DELTA});
     expect(nextState.get('gridMode')).toBe(true);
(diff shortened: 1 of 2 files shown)


jpellizzari:
@simon Could you take a look por favor?

fbarl:
Two comments about the UI:

* When you are in the graph view mode and you click on the topology that's already selected, the loader shows up on top of the graph that is already there. One solution here would be to simply move your added line(s) into the `if`-clause above, so that `nodesLoaded` is set to false only when the topology actually changes. As a side note, I'd also extract all this shared code between `CLICK_SHOW_TOPOLOGY_FOR_NODE` and `CLICK_TOPOLOGY` into a `switchTopology` helper, just because there's a lot of repeated logic.
* Switch to the table mode happens now every time I change back to a big topology (instead of just the first time), which I think can be slightly annoying. This is because of the condition in line `521` which I think we could be fixed easily if we also add a boolean `nodesLoadedFirstTime` or something like that to the global state... or perhaps you can find a more elegant solution.

jpellizzari:
@fbarl Updated

PR #2237, opened 2017-02-14
Add logic to remove non-transferrable state when switching Cloud instances

jpellizzari:
Change needed for https://github.com/weaveworks/service-ui/pull/184

This moves the logic added in https://github.com/weaveworks/service-ui/pull/198 to Scope. In the near future, service-ui will interact with Scope via Scope's public JS API, so manipulating the iframe URL state will not be a viable option for preventing instance-specific state from transferring when the user changes instances.


jpellizzari:
@fbarl Could you take a look please?

client/app/scripts/reducers/root.js
```diff
@@ -721,6 +721,12 @@ export function rootReducer(state = initialState, action) {
       return state.set('showingTroubleshootingMenu', !state.get('showingTroubleshootingMenu'));
     }
 
+    case ActionTypes.CHANGE_INSTANCE: {
+      state = state.set('selectedNodeId', null);
+      state = state.update('nodeDetails', nodes => nodes.clear());
```
fbarl: I think that `controlPipes` should also be cleared here, just by seeing what `closeNodeDetails` does. And then I would actually replace the content of `closeAllNodeDetails` helper with these lines and just use it in this action, as it's all more straightforward and consistent that way.

client/package.json
```diff
@@ -61,6 +61,7 @@
     "eslint-plugin-import": "2.2.0",
     "eslint-plugin-jsx-a11y": "2.2.3",
     "eslint-plugin-react": "6.8.0",
+    "expect": "^1.20.2",
```
fbarl: How were the other tests working without this (e.g. line `530` in `root-test.js`)? :) If we need to have it explicitly mentioned, I'd just remove the `^` as I think we've decided at some point that it's better to lock the versions in Scope.

jpellizzari: I have no idea. I was getting an error in my build and I went to check the version of `expect` we were using, and found that it wasn't there. If any other projects list it as a dependency, we would be able to use it as a side effect.

client/app/scripts/reducers/root.js
jpellizzari: Updated, much better.

client/app/scripts/reducers/root.js
```diff
@@ -721,6 +721,12 @@ export function rootReducer(state = initialState, action) {
       return state.set('showingTroubleshootingMenu', !state.get('showingTroubleshootingMenu'));
     }
 
+    case ActionTypes.CHANGE_INSTANCE: {
+      state = state.set('selectedNodeId', null);
```
fbarl: This line is also redundant now as `closeAllNodeDetails` takes care of it. Otherwise LGTM!

PR #2244, opened 2017-02-15
Fix filtering out uncontained nodes in DNS view

### LINKED ISSUE #2170: Hide Uncontained filter in DNS view not hiding uncontained
Select DNS view, and filter by application containers and then filter out uncontained.  You can see that the uncontained are still appearing. 

screen shot 2017-01-31 at 8 29 09 am

fbarl:
Fixes #2170.

The problem was that no decorators were being applied to `ContainerRenderer` in `ContainerHostnameRenderer` (line `238` in `render/container.go`), so the *Uncontained* node wouldn't be filtered out there.

My solution was to move the `ApplyDecorators` call from `ContainerWithImageNameRenderer` higher up to `ContainerRenderer` so that it also affects line `238`.

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -439,7 +439,10 @@ func (r *Registry) RendererForTopology(topologyID string, values url.Values, rpt
 		}
 	}
 	if len(decorators) > 0 {
-		return topology.renderer, render.ComposeDecorators(decorators...), nil
+		// Here we tell the topology renderer to apply the filtering decorator
+		// that we construct as a composition of all the selected filters.
+		composedFilterDecorator := render.ComposeDecorators(decorators...)
+		return render.ApplyDecorator(topology.renderer), composedFilterDecorator, nil
 	}
 	return topology.renderer, nil, nil
 }
--- a/render/container.go
+++ b/render/container.go
@@ -23,7 +23,7 @@ const (
 // NB We only want processes in container _or_ processes with network connections
 // but we need to be careful to ensure we only include each edge once, by only
 // including the ProcessRenderer once.
-var ContainerRenderer = MakeFilter(
+var ContainerRenderer = ApplyDecorators(MakeFilter(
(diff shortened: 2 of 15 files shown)


render/container.go
```diff
@@ -23,7 +23,7 @@ const (
 // NB We only want processes in container _or_ processes with network connections
 // but we need to be careful to ensure we only include each edge once, by only
 // including the ProcessRenderer once.
-var ContainerRenderer = MakeFilter(
+var ContainerRenderer = ApplyDecorators(MakeFilter(
```
2opremio: ContainerRenderer is used at other places.(e.g. `ContainerWithImageNameRenderer`), so I guess this will cause the decorators to be applied multiple times, right? Can we avoid that?

For instance we could only apply the decorators at the toplevel.

In fact, a general strategy for this would be desirable (i.e. make sure that Apply decorators is only applied once at the top level, assuming that's correct and semantically equivalent to what we do)

2opremio: Maybe it would be enough to only invoke ApplyDecorators in api_topologies, at the top level?

app/api_topologies.go
```diff
@@ -439,7 +439,10 @@ func (r *Registry) RendererForTopology(topologyID string, values url.Values, rpt
 		}
 	}
 	if len(decorators) > 0 {
-		return topology.renderer, render.ComposeDecorators(decorators...), nil
+		// Here we tell the topology renderer to apply the filtering decorator
+		// that we construct as a composition of all the selected filters.
+		composedFilterDecorator := render.ComposeDecorators(decorators...)
+		return render.ApplyDecorator(topology.renderer), composedFilterDecorator, nil
```
2opremio: 👍

fbarl:
@2opremio PTAL. I had to fix a couple of failing tests and I also tested filtering directly on `RendererForTopology` function.

PR #2256, opened 2017-02-17
Load contrast stylesheet 

jpellizzari:
Part of https://github.com/weaveworks/service-ui/pull/184

Instead of loading a separate contrast.html page when the user clicks the contrast button, load a different stylesheet that will override the normal scope styles.

A window level variable is written to the index.html file. Adds a webpack plugin to accomplish this. From the code:
```
/**
 * Change the Scope UI theme from normal to high-contrast.
 * In prod, this will inject a stylesheet into <head> and override the styles.
 * In dev, it will force a webpack hot-reload.
 *
 * A window-level variable is written to the .html page during the build process that contains
 * the filename (and content hash) needed to download the file.
 */
```

@fbarl New approach to the previous PR. PTAL


jpellizzari:
@fbarl Sorry about the false-start. This should be ready for you review now.

Things I tested:
* Contrast mode switches in hot-reload and static mode, ie `npm run build && NODE_ENV=production npm start`
* The node svgs adapt correctly to contrast-mode changes
* Reloading the page with `contrastMode: true` in the URL state behaves correctly

client/app/scripts/actions/app-actions.js
```diff
@@ -664,6 +677,10 @@ export function route(urlState) {
       state.get('nodeDetails'),
       dispatch
     );
+
+    if (urlState.contrastMode) {
+      dispatch(toggleContrastMode(true));
```
fbarl: I think `setContrastMode` might actually be a better name for this action (as I associate toggling with switching to the opposite state that wouldn't need any arguments).

client/app/scripts/contrast-compiler.js
```diff
@@ -0,0 +1,46 @@
...
+        // Find the name of the contrast stylesheet and save it to a window variable.
+        const { css, publicPath } = htmlPluginData.assets;
+        const contrast = findAsset(css, 'contrast-theme');
+        const normal = findAsset(css, 'style-app');
+        // Convert to JSON string so they can be parsed into a window variable
+        const themes = JSON.stringify({ normal, contrast, publicPath });
+        // Append a script to the end of <head /> to evaluate before the other scripts are loaded.
+        const script = `<script>window.__WEAVE_SCOPE_THEMES = JSON.parse('${themes}')</script>`;
+        const [head, end] = htmlPluginData.html.split('</head>');
+        htmlPluginData.html = head.concat(script).concat(end);
```
fbarl: When I run `npm run build` locally, I get the following `index.html` file:

```
<!doctype html>
<html class="no-js"><head><link href="style-app-4134b6394afe6c264f1e.css?c044bfbe96884908980f" rel="stylesheet"></head>
  <head>
    <meta charset="utf-8">
    <title>Weave Scope</title>
  <script>window.__WEAVE_SCOPE_THEMES = JSON.parse('{"normal":"style-app-4134b6394afe6c264f1e.css?c044bfbe96884908980f","contrast":"style-contrast-theme-fd61e9572aa66a409ca1.css?c044bfbe96884908980f","publicPath":""}')</script>
  <body>
    <!--[if lt IE 10]>
      <p class="browsehappy">You are using an <strong>outdated</strong> browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
    <![endif]-->
    <div class="wrap">
      <div id="app"></div>
    </div>
  <script type="text/javascript" src="vendors-098b95ea8f29c00c4a2f.js?c044bfbe96884908980f"></script><script type="text/javascript" src="app-4134b6394afe6c264f1e.js?c044bfbe96884908980f"></script></body>
</html>
```

So it actually contains two `<head>` tags, with the second one missing a closure `</head>`. Funny enough, it works just fine on my Chrome, but I still think it should be fixed before going to prod :)

client/app/scripts/utils/contrast-utils.js
```diff
@@ -1,7 +1,22 @@
...
+export function loadTheme(theme = 'normal') {
+  if (window.__WEAVE_SCOPE_THEMES) {
+    // Load the pre-built stylesheet.
+    const stylesheet = window.__WEAVE_SCOPE_THEMES[theme];
+    const head = document.querySelector('head');
+    const link = document.createElement('link');
+    link.rel = 'stylesheet';
+    link.href = `${window.__WEAVE_SCOPE_THEMES.publicPath}${stylesheet}`;
+
+    head.appendChild(link);
```
fbarl: It would be nice if we'd also remove the previous theme here, as right now every toggle adds one `<link>` tag to the html, so by toggling a lot we end up with something like this:

```
<link rel="stylesheet" href="style-contrast-theme-fd61e9572aa66a409ca1.css?c044bfbe96884908980f">
<link rel="stylesheet" href="style-app-4134b6394afe6c264f1e.css?c044bfbe96884908980f">
<link rel="stylesheet" href="style-contrast-theme-fd61e9572aa66a409ca1.css?c044bfbe96884908980f">
<link rel="stylesheet" href="style-app-4134b6394afe6c264f1e.css?c044bfbe96884908980f">
<link rel="stylesheet" href="style-contrast-theme-fd61e9572aa66a409ca1.css?c044bfbe96884908980f">
<link rel="stylesheet" href="style-app-4134b6394afe6c264f1e.css?c044bfbe96884908980f">
...
```

client/webpack.production.config.js
```diff
@@ -114,22 +115,17 @@ module.exports = {
     new ExtractTextPlugin('style-[name]-[chunkhash].css'),
     new HtmlWebpackPlugin({
       hash: true,
-      chunks: ['vendors', 'contrast-app'],
-      template: 'app/html/index.html',
-      filename: 'contrast.html'
-    }),
-    new HtmlWebpackPlugin({
-      hash: true,
       chunks: ['vendors', 'terminal-app'],
```
fbarl: I assume that right now the `contrast-theme` doesn't affect the `terminal-app` in any way, but do you think it would make sense to add it here now, so that we don't have to think about it in the future?

fbarl:
Also, sorry for the conflicts with master, they're due to my layout optimizations PR. Just double check for `isContrastMode` helper calls when resolving them.

client/webpack.production.config.js
jpellizzari: I would say let's wait until we need to use it in the terminal before adding.

client/app/scripts/utils/contrast-utils.js
jpellizzari: This could introduce a race condition where the new sheet isn't loaded before the old one is removed. I guess we could listen for an `onload` event on that `<link />`, but I don't know that the re-paint would be done when that fires?

I don't think there is any harm in leaving that old stylesheet there. Most users will only toggle once.

client/webpack.production.config.js
fbarl: Alright, makes sense

Issue #2236, opened 2017-02-13
Add CSRF headers to requests 


2opremio:
The ajax module will look for `window.__WEAVEWORKS_CSRF_TOKEN` variable and add a header called `X-CSRF-Token` to all requests.

jpellizzari:
Implemented by #2260

PR #2260, opened 2017-02-20
Consolidate API requests into single helper; added CSRF header

jpellizzari:
Implements #2236 .

Consolidates all the API requests into a single `doRequest` function and adds a CSRF token when it exists. 

This also fixes a long-standing issue of HTML content being sent to JSON requests and not throwing an HTTP error. API requests will now have a `content-type: json` header instead of accepting any content response. A parsing or type error would be thrown further down in the stack when text content was served to a JSON API request, leading to confusing error messages.


jpellizzari:
@2opremio You should be able to test your work with this branch. CC @fbarl to get a second pair of eyes :)

client/.eslintrc
```diff
@@ -9,6 +9,7 @@
   "rules": {
     "no-debugger" : 1,
     "comma-dangle": 0,
+    "no-underscore-dangle": 0,
```
fbarl: I think we decided to keep this rule back when I was upgrading `eslint-config-airbnb`: https://github.com/weaveworks/scope/pull/2058

However, if you think `__WEAVEWORKS_CSRF_TOKEN` should be an exception, I'd much rather add `/* eslint no-underscore-dangle: 0 */` to `web-api-utils.js` than disable this rule globally.

client/app/html/index.html
```diff
@@ -5,6 +5,7 @@
     <title>Weave Scope</title>
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <script language="javascript">window.__WEAVEWORKS_CSRF_TOKEN = "$__CSRF_TOKEN_PLACEHOLDER__";</script>
```
fbarl: Don't quite understand where `$__CSRF_TOKEN_PLACEHOLDER__` gets replaced with the actual token, but could we maybe make this string a constant since it's used in two places on frontend?

fbarl:
Left only two styling comments, otherwise the code looks good to me, but I'd let @2opremio test the behaviour as I'm not so sure how it works :)

2opremio:
Successfully tested at https://github.com/weaveworks/service/pull/1104#issuecomment-281111611

client/app/html/index.html
2opremio: The Weave Cloud http frontend query/replaces it

2opremio: See https://github.com/weaveworks/service/pull/1104

jpellizzari: @fbarl index.html can't access our constants because it would have to load the bundle first.

client/.eslintrc
jpellizzari: This rule is poorly implemented. It triggers on any property, even ones that are defined by external modules which we don't control. If there was a case where a third party module used a leading underscore, it would not pass linting.

However, I got the same feedback from two different people, so I will switch it back and do a local override.

Issue #1918, opened 2016-10-10
Deploy step rebuilds UI on circle


tomwilkie:
See https://circleci.com/gh/weaveworks/scope/5465

It shouldn't do that.

ekimekim:
This bit us. Some third party lib changed in between the original build and the rebuild, so only the rebuild was uploaded but the original build's hash was baked into the index.html, breaking dev.

PR #2266, opened 2017-02-22
circle.yml: Fix ui upload step so it doesn't build twice

ekimekim:
There are two source checkouts in circle, and we were building in the wrong one
(ie. not the one where the earlier build happened). This was wasteful at best
and buggy at worst due to build differences.

Fixes https://github.com/weaveworks/scope/issues/1918


foot:
LGTM after circle passes!

PR #2268, opened 2017-02-22
Fix edge directionality in graph layout

### LINKED ISSUE #2267: graph not rendered top-down, despite lack of cycles
Here's what Weave Cloud Dev shows for the 'cortex' service:

screenshot_2017-02-22_09-29-43

The distributor and ingester are both rendered below consul, even though they connect *to* consul.

There are no cycles in the directed graph except for the consuls and memcaches connecting to themselves, so there is no intrinsic reason why this couldn't be laid out top-down:

1. querier, ruler, distributor
2. ingester
3. consul, memcached, dynamo, s3
4. The Internet

fbarl:
Fixes #2267.

layout-fixed

The problem was that we were not making a difference between *source* and *target* nodes when constructing edges for the layout engine (for the purpose of not rendering the same edge twice).

~~**Note:** The bidirectional edges are now represented as two directed edges, both in the layout engine and when rendering. So they are actually rendered as two overlapping edges, although that's currently not visible. I intentionally left it like that so that we could make them non-overlapping if we want or have an easier time drawing the arrows later on.~~

### DIFF
--- a/client/app/scripts/utils/__tests__/layouter-utils-test.js
+++ b/client/app/scripts/utils/__tests__/layouter-utils-test.js
@@ -1,0 +1,41 @@
+import { fromJS } from 'immutable';
+
+describe('LayouterUtils', () => {
+  const LayouterUtils = require('../layouter-utils');
+
+  describe('initEdgesFromNodes', () => {
+    const f = LayouterUtils.initEdgesFromNodes;
+
+    it('should return map of edges', () => {
+      const input = fromJS({
+        a: { adjacency: ['b', 'c'] },
+        b: { adjacency: ['a', 'b'] },
+        c: {}
+      });
+      expect(f(input).toJS()).toEqual({
+        'a-b': { id: 'a-b', source: 'a', target: 'b', value: 1 },
+        'a-c': { id: 'a-c', source: 'a', target: 'c', value: 1 },
+        'b-a': { id: 'b-a', source: 'b', target: 'a', value: 1 },
+        'b-b': { id: 'b-b', source: 'b', target: 'b', value: 1 },
+      });
+    });
+  });
+
+  describe('collapseMultiEdges', () => {
+    const f = LayouterUtils.collapseMultiEdges;
+
+    it('should return collapsed multi-edges', () => {
+      const input = fromJS({
+        'a-b': { id: 'a-b', source: 'a', target: 'b' },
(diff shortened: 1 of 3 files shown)


rade:
> The bidirectional edges are now represented as two directed edges, both in the layout engine and when rendering. So they are actually rendered as two overlapping edges, although that's currently not visible. I intentionally left it like that so that we could make them non-overlapping if we want or have an easier time drawing the arrows later on.

Hmm. Not sure about that. I would expect bidirectional edges to be rendered as one edge with two arrowheads.

If it's relatively easy to have one edge for bidirectional edges, let's do that. If not, then what you have is ok.

rade:
btw, was this always broken, or is it a regression?

fbarl:
@rade Alright, I modified the code so that it doesn't render both edges, but I'm still sending them both to the layout engine so that the rank assignment would be more accurate. I cleaned up that code and added some comments and unit tests.

I think I broke edge directions in one of my optimizations stories, so I don't think it was there for too long. However, I think we were always ignoring one direction in case of bidirectional edges, which is something this story also fixes. There is a chance this fix will improve stability and accuracy of some rendered graphs slightly.

rade:
> I think I broke edge directions in one of my optimizations stories,

Did that breakage make it 1.2? If so, we *may* want to release a 1.2.1 with a fix. cc @2opremio

client/app/scripts/utils/__tests__/layouter-utils-test.js
```diff
@@ -0,0 +1,41 @@
+import { fromJS } from 'immutable';
+
+describe('LayouterUtils', () => {
+  const LayouterUtils = require('../layouter-utils');
```
jpellizzari: `import` syntax would keep this more consistent

fbarl:
@rade Fortunately the breakage didn't make it into any of the releases as it was introduced on *master* only 2 days ago when merging #2221 (line 19 in [nodes-chart-layout.js](https://github.com/weaveworks/scope/pull/2221/files#diff-bb937cf1c7e428785c32dca6802273bc)).

client/app/scripts/utils/__tests__/layouter-utils-test.js
fbarl: Agree

jpellizzari:
LGTM. Nice to have tests for this!

PR #2272, opened 2017-02-23
Add loading indicator on topology option change

jpellizzari:
Missed in #2232 

Adds the loading indicator if changing topology options is taking a long time. Otherwise, its a blank screen.

@fbarl Mind taking a look?


jpellizzari:
@fbarl Nice catch. I added `clearNodes` to `ROUTE_TOPOLOGY`, since that is checking for a topology change. As far as I can tell, the `OPEN_WEBSOCKETS` action is just clearing the cache, so no need to apply the "loading" logic. I left that one as is.

PR #2275, opened 2017-02-23
Move zoomCache to the global state

### LINKED ISSUE #2254: Graph view zoom states get reset when switching to table view
To reproduce:

1. View any topology in the graph mode
2. Zoom in
3. Switch to table mode
4. Switch back to graph mode
5. Notice how the zoom is not preserved

The problem is that `zoomCache` is currently a part of `NodesChart` component which gets destroyed when switching to table view. A solution might be to move `zoomCache` to the global state.

fbarl:
Fixes #2254.

**Other improvements**

* Made all the selectors depend only on the global state, following @foot's suggestion in https://github.com/weaveworks/scope/pull/2128#issuecomment-276683955.
* As a consequence, `NodesChart` component is only in charge of zooming now and doesn't contain any delicate logic (which was important to make sure a global `zoomCache` behaves right).
* Viewport dimensions are also stored on the global state and get updated through a debounced action, thus making `forceRelayout` more unnecessary under certain conditions, as recalculations and rerenderings got smarter and more efficient with selectors.
* Made the positioning of the graph layout take into account all the margins in a more consistent manner.
* Unified the graph complexity heuristic functions, so that the animations are turned off if and only if the initial switch to the table view happens.


fbarl:
I added one more commit with two cheap performance gains:

* `activeTopologyZoomCacheKeyPath` has been made a selector (together with `activeTopologyOptionsSelector`), to prevent unnecessary recalculations of `activeLayoutCachedZoomSelector` when zooming/panning.
* Fixed the return value of `focusedNodesIdsSelector` in case of no selected nodes (the previous one was causing unnecessary recalculations of layout nodes/edges).

@davkal Please take another look at the behaviour quickly before I merge, the changes in the code were very straightforward so I don't think you need to go through them again. Thanks!

Issue #2276, opened 2017-02-23
Docker client 1.10 used in Scope, contains vulnerable sqlite


pidster:
The Docker Store container image scanning process detects a vulnerability in Scope 1.2.1, on the layer that adds `docker.tgz`.

    ADD file:a05c12c8c298349550c5495b9573ca3a0d78e5f1d6173968f6c7ed7e0951b8d3 in /

I believe this corresponds to [Dockerfile L7](https://github.com/weaveworks/scope/blob/master/docker/Dockerfile#L7):

    ADD ./docker.tgz /

`sqlite3 3.8.5` is detected as having 3 critical and 1 major vulnerabilities (CVE-2015-3414, CVE-2015-3415, CVE-2015-3416 and CVE-2016-6153).

2opremio:
Triggered by #2262

PR #2278, opened 2017-02-23
Embed a copy of weaveutil instead of calling via docker

bboreham:
Scope calls the `weave` script, which checks to see if it can exec `weaveutil` directly and, if not, execs it via a Docker image `weaveexec`.  This PR downloads that image and copies the binary into the Scope image, which saves all the indirection and overhead of running Docker every time.

This mechanism is used when `weave ps` is run, which is every few seconds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ SCOPE_UI_BUILD_UPTODATE=.scope_ui_build.uptodate
 SCOPE_BACKEND_BUILD_IMAGE=$(DOCKERHUB_USER)/scope-backend-build
 SCOPE_BACKEND_BUILD_UPTODATE=.scope_backend_build.uptodate
 SCOPE_VERSION=$(shell git rev-parse --short HEAD)
+WEAVENET_VERSION=1.9.0
 DOCKER_VERSION=1.10.3
 DOCKER_DISTRIB=.pkg/docker-$(DOCKER_VERSION).tgz
 DOCKER_DISTRIB_URL=https://get.docker.com/builds/Linux/x86_64/docker-$(DOCKER_VERSION).tgz
@@ -38,7 +39,7 @@ $(DOCKER_DISTRIB):
 	curl -o $(DOCKER_DISTRIB) $(DOCKER_DISTRIB_URL)
 
 docker/weave:
-	curl -L git.io/weave -o docker/weave
+	curl -L https://github.com/weaveworks/weave/releases/download/v$(WEAVENET_VERSION)/weave -o docker/weave
 	chmod u+x docker/weave
 
 $(SCOPE_EXPORT): $(SCOPE_EXE) $(DOCKER_DISTRIB) docker/weave $(RUNSVINIT) docker/Dockerfile docker/demo.json docker/run-app docker/run-probe docker/entrypoint.sh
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -38,10 +39,16 @@ $(DOCKER_DISTRIB):
 	curl -o $(DOCKER_DISTRIB) $(DOCKER_DISTRIB_URL)
 
 docker/weave:
-	curl -L git.io/weave -o docker/weave
+	curl -L https://github.com/weaveworks/weave/releases/download/v$(WEAVENET_VERSION)/weave -o docker/weave
 	chmod u+x docker/weave
 
-$(SCOPE_EXPORT): $(SCOPE_EXE) $(DOCKER_DISTRIB) docker/weave $(RUNSVINIT) docker/Dockerfile docker/demo.json docker/run-app docker/run-probe docker/entrypoint.sh
+docker/weaveutil:
+	docker pull weaveworks/weaveexec:$(WEAVENET_VERSION)
```
2opremio: Couldn't this full block be replaced by `docker run --rm weaveworks/weaveexec:$(WEAVENET_VERSION) cat /usr/bin/weaveutil  > docker/weaveutil` ?

2opremio: In fact: `docker run --rm weaveworks/weaveexec:$(WEAVENET_VERSION) cat /usr/bin/weaveutil > $@`

2opremio: Apparently I cannot do this because Docker 1.10.2 doesn't like running the weave-exec container. Making it fail in CircleCI

```
vagrant@vagrant-ubuntu-wily-64:~/scope$ docker run  weaveworks/weaveexec:1.9.0 true
docker: Cannot connect to the Docker daemon. Is the docker daemon running on this host?.
See 'docker run --help'.
vagrant@vagrant-ubuntu-wily-64:~/scope$ docker version
Client:
 Version:      1.10.2
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   c3959b1
 Built:        Mon Feb 22 21:40:35 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.10.2
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   c3959b1
 Built:        Mon Feb 22 21:40:35 2016
 OS/Arch:      linux/amd64
vagrant@vagrant-ubuntu-wily-64:~/scope$ docker run busybox echo hello
hello
```

2opremio: It's not 1.10 specific. @brb can reproduce with 1.13.1 https://weaveworks.slack.com/archives/net/p1487936232008709

2opremio:
Paves the way for #2276

2opremio:
Also, solves a problem from a user in #weave-users: https://weaveworks.slack.com/archives/weave-users/p1487283019002156

> Hey guys. I'm seeing some odd behavior with weavescope on Kubernetes. At one point in my cluster, there were over 4000 dead `weaveexec` containers, and it was growing. It seemed that something was launching `weaveexec` on the masters about every minute. After poking around a bit, I killed the weave-scope-agent DaemonSet, and that noticed that the weaveexec count stopped. To confirm, I recreated the DaemonSet, and the count started increasing again. Is this normal behavior? What exactly does weaveexec do?

PR #2284, opened 2017-02-27
Create cloud agent image

### LINKED ISSUE #2279: Invert default of '--probe.docker' to match typical usage
When launching the Scope probe, we set the following:

    --probe.docker=true

In the majority of cases, it's set to `true`. We should flip the default from `false` and reduce the config params needed to launch a typical probe.

https://github.com/weaveworks/scope/blob/master/prog/main.go#L283

### LINKED ISSUE #2283: Docker Store cloud-agent:  don't run 'runsvinit'
Per the Docker Store feedback/report, Docker do not approve of `runsvinit`.

> Appears to use runsvinit (process manager utility).  The processes are spawned from the same binary within the image.  Please consider if there are architecture changes that could reduce this extra layer of indirection.

See also: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/each-container-should-have-only-one-concern

2opremio:
Fixes #2283 and includes the cloud-agent image in our CI and release pipelines.

To run the new cloud-agent image, just do

```
RELEASE=latest
TOKEN=shaiuhfquhqwofjwoisncasbifus
docker run --privileged  --net=host --pid=host -v /var/run/docker.sock:/var/run/docker.sock -v /var/run/scope/plugins:/var/run/scope/plugins -e CHECKPOINT_DISABLE weaveworks/cloud-agent:$RELEASE --probe.token=$TOKEN
```

(Edit: it also fixes #2279 )


rade:
> `docker run --privileged  --net=host --pid=host -v /var/run/docker.sock:/var/run/docker.sock -v /var/run/scope/plugins:/var/run/scope/plugins -e CHECKPOINT_DISABLE weaveworks/cloud-agent:$TOKEN --probe.docker=true --probe.token=$TOKEN`

I think you meant `weaveworks/cloud-agent:$RELEASE`.

Also, the plugins and checkpoint stuff can be omitted.

2opremio:
> I think you meant `weaveworks/cloud-agent:$RELEASE`.
Yes, well spotted

> Also, the plugins and checkpoint stuff can be omitted.

Up to whoever writes the docs ( @pidster ? ). I think it's worth mentioning it as optional though.

rade:
Couldn't we simply modify the existing entrypoint.sh file to only do the runsvinit dance when running both the app and probe?

That would work for the normal scope too.

2opremio:
> Couldn't we simply modify the existing entrypoint.sh file to only do the runsvinit dance when running both the app and probe?

I want to keep the agent image as lean and simple as possible to avoid certification problems in the future.

That said, that doesn't discard doing what you suggest separately.

docker/Dockerfile.scope
```diff
@@ -0,0 +1,10 @@
+FROM weaveworks/cloud-agent
+RUN apk add --update runit && \
+	rm -rf /var/cache/apk/*
+ADD ./demo.json /
+ADD ./weave ./weaveutil /usr/bin/
```
rade: that last line is already in the parent image.

docker/Dockerfile.cloud-agent
```diff
@@ -0,0 +1,10 @@
+FROM alpine:3.5
+MAINTAINER Weaveworks Inc <help@weave.works>
+LABEL works.weave.role=system
+WORKDIR /home/weave
+RUN apk add --update bash conntrack-tools iproute2 util-linux curl && \
+	rm -rf /var/cache/apk/*
+ADD ./docker /usr/local/bin/
+ADD ./weave ./weaveutil /usr/bin/
+COPY ./scope /home/weave/
+ENTRYPOINT ["/home/weave/scope", "--mode=probe", "--probe-only"]
```
rade: could we add `--probe.docker=true` too, until we fix #2279?

2opremio: Sure, good call

circle.yml
```diff
@@ -75,12 +76,16 @@ deployment:
       - |
         test -z "${DOCKER_USER}" || (
           docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS &&
-          (test "${DOCKER_ORGANIZATION:-$DOCKER_USER}" == "weaveworks" || (
-            docker tag weaveworks/scope:latest ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:latest &&
-            docker tag weaveworks/scope:$(./tools/image-tag) ${DOCKER_ORGANIZATION:-$DOCKER_USER}/scope:$(./tools/image-tag)
+          (test "${DOCKER_ORGANIZATION:-$DOCKER_USER}" != "weaveworks" && (
```
fbarl: Can you explain briefly why `== "weaveworks" ||` was changed to `!= "weaveworks" &&`?

docker/Dockerfile.scope
```diff
@@ -0,0 +1,9 @@
+FROM weaveworks/cloud-agent
+RUN apk add --update runit && \
+	rm -rf /var/cache/apk/*
+ADD ./demo.json /
+COPY ./runsvinit ./entrypoint.sh /home/weave/
```
fbarl: Does `/home/weave` need to be copied again in `Dockerfile.scope` because `RUN apk add --update runit` modifies it?

2opremio: `/home/weave/` is the destination

circle.yml
2opremio: It's equivalent, it just makes it easier to read (for me at least). I read it as, _if the organization is not `weaveworks` then ...._

bin/release
```diff
@@ -161,15 +161,18 @@ draft() {
...
     if [ "$RELEASE_TYPE" = 'PRERELEASE' ]; then
         echo "== Tagging and pushing images on docker hub as user $DOCKERHUB_USER"
-        $SUDO docker tag "$DOCKERHUB_USER"/scope "$DOCKERHUB_USER/scope:$VERSION"
-        $SUDO docker push "$DOCKERHUB_USER/scope:$VERSION"
-        $SUDO docker tag "$DOCKERHUB_USER/scope:$VERSION" "$DOCKERHUB_USER/scope:latest_release"
-        $SUDO docker push "$DOCKERHUB_USER/scope:latest_release"
+        for IMAGE in $IMAGES; do
+            $SUDO docker tag "$DOCKERHUB_USER/$IMAGE" "$DOCKERHUB_USER/$IMAGE:$VERSION"
+            $SUDO docker push "$DOCKERHUB_USER/$IMAGE:$VERSION"
+            $SUDO docker tag "$DOCKERHUB_USER/$IMAGE:$VERSION" "$DOCKERHUB_USER/scope:latest_release"
```
fbarl: `scope:latest_release` -> `$IMAGE:latest_release`?

2opremio: Good catch! thanks

circle.yml
fbarl: Makes sense and I agree that it's more readable :)

fbarl:
How about adding `cloud-agent.tar` to `.gitignore`?

2opremio:
> How about adding `cloud-agent.tar` to `.gitignore`?

Good point, will do

PR #2285, opened 2017-02-28
When k8s present, allow filtering of containers by namespace

### LINKED ISSUE #1504: Filter by kubernetes namespace in CONTAINERS view
With scope-0.15 it's possible to filter by namespace in the PODS view which is great when having lots of namespaces. From there however it's difficult to get to the right container as all containers are listed in the sidebar, not just the ones belonging to the filtered namespace:

scope-ns-issue

A namespace filter in the CONTAINERS view would allow to directly jump into the container and e.g. open a shell.

@2opremio suggested making the namespace filter global

ekimekim:
To facilitate this, we replace the existing functionality of updateFilters which
sets k8s topologies to have the filters [namespace, managed], to instead append the namespace filter
to any existing. This lets it apply to both k8s and container topologies without overwriting existing
container filters. We instead set the managed filter in the static definition.

This however has the side effect that the ordering of the namespace filter and the managed filter
in k8s topologies has been reversed, so it reads:
```
Show Unmanaged | Hide Unmanaged
foo | bar | default | baz | All Namespaces
```
instead of:
```
foo | bar | default | baz | All Namespaces
Show Unmanaged | Hide Unmanaged
```

Fixes #1504

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -80,21 +80,23 @@ func updateFilters(rpt report.Report, topologies []APITopologyDesc) []APITopolog
 		ns = append(ns, namespace)
 	}
 	sort.Strings(ns)
-	for i, t := range topologies {
-		if t.id == podsID || t.id == servicesID || t.id == deploymentsID || t.id == replicaSetsID {
-			topologies[i] = updateTopologyFilters(t, []APITopologyOptionGroup{
-				kubernetesFilters(ns...), unmanagedFilter,
-			})
+	if len(ns) > 0 { // We only want to apply k8s filters when we have k8s-related nodes
+		for i, t := range topologies {
+			if t.id == containersID || t.id == containersByImageID || t.id == containersByHostnameID || t.id == podsID || t.id == servicesID || t.id == deploymentsID || t.id == replicaSetsID {
+				topologies[i] = mergeTopologyFilters(t, []APITopologyOptionGroup{
+					kubernetesFilters(ns...),
+				})
+			}
 		}
 	}
 	return topologies
 }
 
-// updateTopologyFilters recursively sets the options on a topology description
-func updateTopologyFilters(t APITopologyDesc, options []APITopologyOptionGroup) APITopologyDesc {
-	t.Options = options
+// mergeTopologyFilters recursively merges in new options on a topology description
+func mergeTopologyFilters(t APITopologyDesc, options []APITopologyOptionGroup) APITopologyDesc {
+	t.Options = append(t.Options, options...)
 	for i, sub := range t.SubTopologies {
-		t.SubTopologies[i] = updateTopologyFilters(sub, options)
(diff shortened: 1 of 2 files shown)


app/api_topologies.go
```diff
@@ -80,21 +80,24 @@ func updateFilters(rpt report.Report, topologies []APITopologyDesc) []APITopolog
 		ns = append(ns, namespace)
 	}
 	sort.Strings(ns)
-	for i, t := range topologies {
-		if t.id == podsID || t.id == servicesID || t.id == deploymentsID || t.id == replicaSetsID {
-			topologies[i] = updateTopologyFilters(t, []APITopologyOptionGroup{
-				kubernetesFilters(ns...), unmanagedFilter,
-			})
+	if len(ns) > 0 {
+		// We only want to apply k8s filters if any k8s info is present
```
2opremio: I would rephrase this into _We only want to apply k8s filters for k8s-relevant topologies_ (since you may be showing the filters even without k8s-tracked nodes)

app/api_topologies.go
```diff
@@ -80,21 +80,24 @@ func updateFilters(rpt report.Report, topologies []APITopologyDesc) []APITopolog
 		ns = append(ns, namespace)
 	}
 	sort.Strings(ns)
-	for i, t := range topologies {
-		if t.id == podsID || t.id == servicesID || t.id == deploymentsID || t.id == replicaSetsID {
-			topologies[i] = updateTopologyFilters(t, []APITopologyOptionGroup{
-				kubernetesFilters(ns...), unmanagedFilter,
```
2opremio: Why are you removing the unmanagedFilter from here and then adding it manually?

render/filters.go
```diff
@@ -293,7 +293,14 @@ func IsNotPseudo(n report.Node) bool {
 // IsNamespace checks if the node is a pod/service in the specified namespace
 func IsNamespace(namespace string) FilterFunc {
 	return func(n report.Node) bool {
-		gotNamespace, _ := n.Latest.Lookup(kubernetes.Namespace)
+		try_keys := []string{kubernetes.Namespace, docker.LabelPrefix + "io.kubernetes.pod.namespace"}
```
2opremio: Is `"io.kubernetes.pod.namespace"` is used at render/filters.go too, please use a common constant and/or a function to extract it from a node.

2opremio:
@ekimekim ping

app/api_topologies.go
ekimekim: The purpose of the `len(ns) > 0` is to not show the k8s filters if there's no k8s-tracked nodes. I'll attempt to clarify.

app/api_topologies.go
ekimekim: To avoid accidentally applying it to the container topologies.

render/filters.go
ekimekim: done

ekimekim:
ptal

PR #2289, opened 2017-03-01
Removed missed merge text and made terminology consistent

abuehrle:
There was some merge text on the install page that I have deleted @2opremio

### DIFF
--- a/site/how-it-works.md
+++ b/site/how-it-works.md
@@ -70,7 +70,7 @@ Scope can also be used to feed reports to Weave Cloud. With Weave Cloud, you can
 
 To register with the service, go to [https://cloud.weave.works](https://cloud.weave.works), enter your email address and wait for the confirmation email that contains your service token.
 
-The cloud token is also accessible from the Weave Cloud main page after you've logged in:
+The cloud token is accessible from the Weave Cloud main page after you've logged in:
 
 ![Weave Cloud Token](images/weave-cloud-token.png)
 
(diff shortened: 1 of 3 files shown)


site/introducing.md
```diff
@@ -55,7 +55,7 @@ Controlling Scope containers goes beyond simple lifecycle operations. Scope can
 
 Weave Scope can visualize your app within:
 
- * Docker Datacenter
+ * Docker Datacenters
```
2opremio: Why? Docker Datacenter is a product https://www.docker.com/products/docker-datacenter

abuehrle: Didn't realize that.  I didn't know what that was tbh. Will change that.

site/how-it-works.md
```diff
@@ -70,12 +70,9 @@ Scope can also be used to feed reports to Weave Cloud. With Weave Cloud, you can
 
 To register with the service, go to [https://cloud.weave.works](https://cloud.weave.works), enter your email address and wait for the confirmation email that contains your service token.
 
-![`Weave Cloud` Signup](images/weave-cloud-signup.png)
+The cloud token is accessible from the Weave Cloud main page after you've logged in:
```
2opremio: This won't be true very soon, once we rework the landing page. Can you be unspecific about it? (or mention the Scope/Explore section)

site/introducing.md
2opremio: As a general recommendation: please look up things before changing them :)

site/how-it-works.md
abuehrle: I will wait for that change, since this is effects _all of the docs_ and not just scope.

2opremio: OK, your call

PR #2290, opened 2017-03-01
Add logic to turn off network requests when Scope dismounts

jpellizzari:
Needed for https://github.com/weaveworks/service-ui/pull/184.

When navigating to another page in Weave Cloud, Scope will be unmounted by `react-router`. This change will turn off topology polling and websocket connections when Scope dismounts.

### DIFF
--- a/client/app/scripts/utils/router-utils.js
+++ b/client/app/scripts/utils/router-utils.js
@@ -90,21 +90,20 @@ export function updateRoute(getState) {
 
 
 export function getRouter(dispatch, initialState) {
+  let mergedState = initialState;
   // strip any trailing '/'s.
   page.base(window.location.pathname.replace(/\/$/, ''));
 
+  const storageState = storageGet(STORAGE_STATE_KEY);
+  if (storageState) {
+    window.location.hash = `!/state/${storageState}`;
+    const parsedState = JSON.parse(decodeURL(storageState));
+    mergedState = Object.assign(initialState, parsedState);
+  }
+
   page('/', () => {
     // recover from storage state on empty URL
-    const storageState = storageGet(STORAGE_STATE_KEY);
-    if (storageState) {
-      // push storage state to URL
-      window.location.hash = `!/state/${storageState}`;
-      const parsedState = JSON.parse(decodeURL(storageState));
-      const mergedState = Object.assign(initialState, parsedState);
-      dispatch(route(mergedState));
-    } else {
-      dispatch(route(initialState));
-    }
+    dispatch(route(mergedState));
   });
 
   page('/state/:state', (ctx) => {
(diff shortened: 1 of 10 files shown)


jpellizzari:
@davkal If you could review when you get a chance. I pushed up a commit on `service-ui` that does a LocalModuleProxy to a local Scope repo. You can test this change along with [the scope as a component PR](https://github.com/weaveworks/service-ui/pull/184) by doing:

`SCOPE_PATH=/path/to/scope PROXY_SCOPE_COMPONENT=true npm start`

in the service-ui repo.

jpellizzari:
Added some extra files to support doing partial imports, ie:
`import reducer from 'weave-scope/reducer'`

I originally attempted to avoid polluting the app/scripts directory with more files using a bash script...
```
#! /bin/bash
split -l 1 app/scripts/index.js ../tmp/module__
for file in `ls ../tmp/module__*`; do
  contents=$(cat $file)
  module_name=$(echo $contents | grep -oE 'exports\.(.*?\s)' | sed 's/exports\.//')
  filename="build-pkg/$(echo "$module_name.js" | sed 's/ //')"
  rm $file
  touch "$filename"
  echo $contents > "$filename"
done
```
...but that will not work with `LocalModuleProxy`.

client/app/scripts/utils/web-api-utils.js
```diff
@@ -348,3 +352,18 @@ export function getPipeStatus(pipeId, dispatch) {
...
+export function stopTopologyPolling() {
+  clearTimeout(topologyTimer);
+  topologyTimer = 0;
+}
+
+export function teardownWebsockets() {
+  clearTimeout(reconnectTimer);
+  if (socket) {
+    socket.onerror = null;
+    socket.onclose = null;
```
davkal: I'd also set `onmessage` and `onopen`, and finally `socket` itself to `null`.

client/app/scripts/utils/router-utils.js
```diff
@@ -90,21 +90,20 @@ export function updateRoute(getState) {
...
 export function getRouter(dispatch, initialState) {
+  let mergedState = initialState;
   // strip any trailing '/'s.
   page.base(window.location.pathname.replace(/\/$/, ''));
 
+  const storageState = storageGet(STORAGE_STATE_KEY);
+  if (storageState) {
+    window.location.hash = `!/state/${storageState}`;
+    const parsedState = JSON.parse(decodeURL(storageState));
+    mergedState = Object.assign(initialState, parsedState);
```
davkal: Not a fan of `...`?

jpellizzari: I guess I'm just old fashioned that way. Fixed.

jpellizzari: Also worth noting that Scope didn't have the `transform-object-rest-spread` babel plugin. Also worth noting that the service-ui eslint loader looks at the `.eslint` file of the Scope repo instead of its own (probably a bug).

PR #2292, opened 2017-03-02
Update node libraries

fbarl:
I feel the time has come once again to update the libraries in `package.json` - I already had it a couple of times that a library I wanted to try out was requiring a newer version of some library that we were already using.

#### Highlights

* *Webpack 2* launched a stable release a couple of months ago, so I thought we might want to start using some nice features it offers (e.g. tree shaking). I followed the [migration notes](https://webpack.js.org/guides/migrating/) as carefully as I could but there is a chance I missed something, so it would be nice to test it well locally before we try pushing it to dev.
* Major [Reselect update](https://github.com/reactjs/reselect/releases/tag/v3.0.0) introduces some optimizations.
* Minor [React update](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html).

#### Other notes

* Since *React 15.5*, `React.PropTypes` have been deprecated and https://github.com/reactjs/prop-types should be used instead. I've created an issue #2496 that should be resolved shortly after this PR is merged.
* I upgraded `materialize-css` only to version `0.98.1` as `0.98.2` introduces errors with relative path to *Roboto* fonts that are included in `_base.scss`. We might want to get rid of Webfonts altogether soon and drop the `materialize-css` dependency altogether.
* `json-loader` that is [not required in Webpack 2](https://webpack.js.org/guides/migrating/#json-loader-is-not-required-anymore) has been dropped from the config files in this PR, but that's causing `service-ui` throw some warnings when ran with the proxied version of scope. One solution is to add the following lines to `service-ui` webpack config, but that will become altogether unnecessary when we upgrade `service-ui` to Webpack 2:

```
module: {
  rules: [
    {
      test: /\.json/,
      loader: "json-loader"
    }
  ]
}
```

#### New linting rules

* [no-multi-assign](http://eslint.org/docs/rules/no-multi-assign)
* [react/no-array-index-key](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md)
* [react/require-default-props](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md)


fbarl:
@jpellizzari Would you mind taking a look?

The whole PR is very small, most of the code changes are in the Webpack config files. I tested it against all the *local* scenarios that I could think of, but don't feel super confident saying no errors would appear when pushing it into dev, so I wouldn't mind a second pair eyes going through the migration changes :)

jpellizzari:
BTW to get rid of the netlifly check error, just rebase master

fbarl:
> BTW to get rid of the netlifly check error, just rebase master

Thanks, it seems to have worked!

> Looks like sparkline.js still has a call to React.PropTypes, which is logging a warning to console. 

Yup, I created a separate issue #2496 for it, as I didn't want to run into conflict with your #2487. Will address it today.

Issue #2293, opened 2017-03-02
'Max call stack exceeded' error


jpellizzari:
Steps to reproduce: 

1. ~~Load https://frontend.dev.weave.works~~ (see url below)
2. Check the JS console

Expected:
Everything works

Actual: 
Max call stack error and super slowness

jpellizzari:
Initial hunch is that this is causing problems: 
https://github.com/weaveworks/scope/commit/4f24d9bba8743bf389798f16fe56905d67ed6d93#diff-5897bc396ca616224f48758015a6ece1R215

jpellizzari:
Seems to be localized to this URL state:

https://frontend.dev.weave.works/app/proud-wind-05#!/state/{"controlPipe":{"id":"pipe-823277968531583873","nodeId":"02c4fa6bc09fef2e97e0ba68f2a56f5159e861399c951f7c953badc1b8152a63;<container>","raw":true,"resizeTtyControl":{"id":"docker_resize_exec_tty","probeId":"3616cce1df0a057e","nodeId":"02c4fa6bc09fef2e97e0ba68f2a56f5159e861399c951f7c953badc1b8152a63;<container>"},"control":{"probeId":"3616cce1df0a057e","nodeId":"02c4fa6bc09fef2e97e0ba68f2a56f5159e861399c951f7c953badc1b8152a63;<container>","id":"docker_exec_container","human":"Exec shell","icon":"fa-terminal","rank":2}},"topologyViewMode":"topo","nodeDetails":[{"id":"02c4fa6bc09fef2e97e0ba68f2a56f5159e861399c951f7c953badc1b8152a63;<container>","label":"tom","topologyId":"containers"}],"pinnedMetricType":"CPU","pinnedSearches":[],"searchQuery":"","selectedNodeId":"02c4fa6bc09fef2e97e0ba68f2a56f5159e861399c951f7c953badc1b8152a63;<container>","gridSortedBy":"host_scope_version","gridSortedDesc":null,"topologyId":"containers","topologyOptions":{"processes":{"unconnected":"hide"},"containers":{"system":"all","stopped":"running","pseudo":"hide"},"pods":{"namespace":"scope","pseudo":"hide"}},"contrastMode":true}

jpellizzari:
The root cause appears to be the redux `dispatch` going crazy. When you log `action.type` to console:
[...]

PR #2294, opened 2017-03-02
Removed contrastMode from url state

jpellizzari:
Temporary fix for #2293 

Removes contrastMode from the URL state, which means its state will be lost on page refresh. Temporary fix for a nasty bug; the permanent solution will be to prevent a `route` action from happening on a browser `replace`, or something to that effect.


jpellizzari:
Merging to fix the bug, but leaving #2294 open

rade:
> leaving #2294 open

I've opened a fresh bug instead.

Issue #2242, opened 2017-02-14
View showing hierarchal relationship of things consuming resources


pidster:
As a user I want to:
- see resource utilisation (memory/CPU/file descriptor) metrics for each Host, so that I can compare how containers on different hosts are consuming resources
- see resource utilisation vs limits vs capacity on each Host, so that I can tune the limits assigned per container
- visualise a hierarchical relationship between hosts, containers & processes (and threads?) so I can so that I can quickly identify the source of a (problem caused by resource utilisation).

Types of resources for consideration:
- CPU
- RAM
- File handles
- Disk space

pidster:
See also: http://www.brendangregg.com/blog/2017-02-06/flamegraphs-vs-treemaps-vs-sunburst.html

fbarl:
Here are my initial [mockups](https://docs.google.com/presentation/d/1nQwy5UoICwPNSnzAcQscethsecLEKBHZ9oFzVRO_jic) accompanied by a [design doc](https://docs.google.com/document/d/1421CytU3c6BisQsyy1xVL2OhCgBPtFABq7D6AH7cut4).

I think I've come up with a layout that would be able to capture all the important aspects mentioned above **per resource**, but showing all the hierarchical layers with the info for all the resources in one view seemed like a difficult task, so I'd propose having a heuristics that would combine all the resources info into one metric for the *general view* (as opposed to selecting particular *resource views*).

There are two very similar layout variations in my mockups: _flame graph_ and _sunburst_. The central question that should decide between the two would be how important the edges i.e. node connections are for the resource view (I'd personally slightly prefer keeping them and make it a _sunburst_ layout).

@davkal @jpellizzari @bowenli @pidster @2opremio I'd appreciate your feedback with some comments/questions/suggestions and perhaps giving your opinion on the questions I raised in the docs.

monadic:
+1 for the flame graph

fbarl:
I just made separate slides with [slightly refined mockups](https://docs.google.com/presentation/d/123NzxsPJDUFtsTxiH0kRFj6uc7Bq3CPnv4vK7i9XPgQ/edit#slide=id.p), focusing only on the flame graph variation that we decided to go with. Also, check out the changes to the [ideas doc](https://docs.google.com/document/d/1421CytU3c6BisQsyy1xVL2OhCgBPtFABq7D6AH7cut4/edit) that are now reflecting all the feedback I got, as well as with some notes on what backend can provide the backend part of the story.

[...]

PR #2296, opened 2017-03-03
Initial version of the resource view

fbarl:
First step towards resolving #2242.

Pretty much copying the layout from the [slides](https://docs.google.com/presentation/d/123NzxsPJDUFtsTxiH0kRFj6uc7Bq3CPnv4vK7i9XPgQ/edit#slide=id.g20ee7cf307_0_38) with some metrics that are already supported by backend:

* Metrics: *CPU*, *Memory*
* Layers: *Hosts -> Containers -> Processes*
* Resource capacity only shown for *Hosts*

##### Changes

* Replaced `gridMode` global state boolean with `topologyViewMode` string (which can now have values `topo`, `grid`, `resource`).
* Moved bunch of stuff to selectors:
  * `availableCanvasMetrics` and `pinnedMetric`,  to simplify the transition between topologies in the resource view
  * Canvas margins, which are now customizable per view mode
  * Generalized viewport/canvas selectors
  * All the node networks logic
* Zooming logic:
  * Extracted all of it (with caching) into a special component, as now we'll have zooming both in the resource view and the graph view
  * Separated zooming limits from zoom state and made them always reflect the layout to ensure e.g. we can always zoom out enough to view the whole layout, even when the layout grows a lot comparing to the initial state
* Moved metric selector to the top, just below the view mode selector.
* Contrast mode and saving to SVG are supported.

##### Action items (by priority)

1. *Processes* metrics given by the backend are inconsistent with the *Containers* metrics, especially in terms of maximum utilization i.e. capacity. In this PR I'm using a hack to fix the view on the frontend, but a more correct solution would require making backend send consistent data from a single source of truth *-> create an issue*
2. The resource view layout is currently kept static by design, i.e. it's not being periodically updated like the other two view modes, since it would require more work on polling and we also have to see how to make dynamic view relatively stable when zoomed in *-> create an issue*
3. No filtering of the nodes in the resource view is currently supported:
  a. Search bar should be shown and functioning *-> create an issue*
  b. Topology options should be shown and working *-> create an issue*
4. Always cap zoom state to the zoom limits (not as straightforward as it sounds, requires some overriding of the cached state) *-> create an issue*
5. Ideally, there should always be a valid metric pinned in the resource view. If we e.g. pin a metric in a certain topology and then switch to a topology in which that metric is not there (while staying in the resource view the whole time), we'll get a blank view. Instead, we should for example select the first available metric, but that's not that simple to do, as we only get available metrics when the nodes are loaded, so we'd actually be modifying the url state twice which messes up the browser history.

##### Other UI/optimizations ideas - see [the mockups](https://docs.google.com/presentation/d/123NzxsPJDUFtsTxiH0kRFj6uc7Bq3CPnv4vK7i9XPgQ/edit#slide=id.g213a58afff_0_0)

* Add a horizontal resource scale info below the resource layers to have a sense of the zoom level; also consider adding a small relative zoom level text somewhere, something like "*x 3.2*"
* Instead of rendering very thin resource boxes, show a `+ 31 more` sort of info at their place - would render less resource boxes which would make zooming smoother
* Consider having the relative consumption in the resource boxes with capacity increase from *left to right* instead of *bottom to top*


rade:
cedfc8b6c6938bb998799593c91e7e2024eeed2a seems unrelated to this PR. Furthermore, it appears to contain changes - the logo.js update - unrelated to its supposed purpose.

fbarl:
@rade In that commit I actually reverted `logo.js` that I did earlier in the PR, so the net code diff contains minimal changes in that file. In the previous commits I broke saving to SVG and as I was fixing it, I noticed it didn't work for the table mode, which required two lines more to fix - do you think I should revert it and have a separate issue for that?

rade:
> create a separate issue for that

just do a separate PR for disabling SVG saving in table mode.

davkal:
Looks great, neat, and informative! Some UI nits:

Initial zoom:
The initial zoom on an incognito window was too far in:
screen shot 2017-03-22 at 13 21 38
At that point it was hard to tell that I had to zoom out to see this:
screen shot 2017-03-22 at 13 22 00

Layers:
I find it confusing that hosts and containers disappear when I load the process topology. What is the benefit?

Small boxes:
There is no interaction that allows me to figure out what a certain box is:
screen shot 2017-03-22 at 13 31 56
There is no hover and no click to give me more info.

Blinking "Select a metric":
It's clear why that is there, but only to me as a developer of Scope. I prefer sensible defaults instead, the empty screen is a confusing surprise.

Wording:
"consumed" means that something is gone. Maybe "used"?

fbarl:
I have addressed all the UI feedback I got for this story:

@davkal,

> The initial zoom on an incognito window was too far in.

This was fixed automatically as I made the resource view for each topology have *HOSTS* as the base layer. Since that topology has no filters, all hosts are always loaded at once and since they are the base layer, they determine max width of the layout, so it doesn't get resized after it's rendered once.

> I find it confusing that hosts and containers disappear when I load the process topology. What is the benefit?

As we agreed, I (temporarily) made all the topologies show all three layers, so there is actually no difference in what they are displaying now (and therefore I'm using the same zoom caching key across all the topologies in the resource view now).

> There is no hover and no click to give me more info.

I added a tooltip with a short summary.

> Blinking "Select a metric"

I hardcoded *CPU* and *Memory* as the only metrics for the resource view, which enabled me to always pin a default metric in the resource view when none (or a different one) was previously pinned.

> Wording

Changed *consumed* to *used*.

@2opremio,

I moved the metric selector to the top, just below the view mode selector in all the views, making it more consistent. If you want, you can take a final look at the resource view behaviour.

fbarl:
@davkal @foot I'm going to merge this PR in a couple of hours, with all the commits squashed. Be free to take a final look (or peek through the code :)) in the meantime.

PR #2300, opened 2017-03-04
parse --probe-only as intended

### LINKED ISSUE #2299: weavecloud-agent w/o token attempts to talk to local app
```
$ docker run --privileged -d --name=weavecloud-agent --net=host --pid=host -v /var/run/docker.sock:/var/run/docker.sock weaveworks/cloud-agent
a6e15b98221ceb697d619933091d215d0af7ca84dc3f336261b49e83c967433d
$ docker logs weavecloud-agent 
time="2017-03-04T11:11:07Z" level=info msg="publishing to: 127.0.0.1:4040" 
<probe> INFO: 2017/03/04 11:11:07.526733 command line args: --logtostderr=true --mode=probe --probe-only=true --probe.docker=true
...
<probe> ERRO: 2017/03/04 11:11:08.528537 Error fetching app details: Get http://127.0.0.1:4040/api: dial tcp 127.0.0.1:4040: getsockopt: connection refused
<repeats every 10 seconds>
```
That only happens when launching the weavecloud-agent without a service token, which is not documented. Still, it's surprising behaviour.

rade:
Fixes #2299.

### DIFF
--- a/docker/Dockerfile.cloud-agent
+++ b/docker/Dockerfile.cloud-agent
@@ -7,4 +7,4 @@ RUN apk add --update bash conntrack-tools iproute2 util-linux curl && \
 ADD ./docker /usr/local/bin/
 ADD ./weave ./weaveutil /usr/bin/
 COPY ./scope /home/weave/
-ENTRYPOINT ["/home/weave/scope", "--mode=probe", "--probe-only", "--probe.docker=true"]
+ENTRYPOINT ["/home/weave/scope", "--mode=probe", "--no-app", "--probe.docker=true"]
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.cloud-agent
```diff
@@ -7,4 +7,4 @@ RUN apk add --update bash conntrack-tools iproute2 util-linux curl && \
 ADD ./docker /usr/local/bin/
 ADD ./weave ./weaveutil /usr/bin/
 COPY ./scope /home/weave/
-ENTRYPOINT ["/home/weave/scope", "--mode=probe", "--probe-only", "--probe.docker=true"]
+ENTRYPOINT ["/home/weave/scope", "--mode=probe", "--no-app", "--probe.docker=true"]
```
2opremio: Why the change?

2opremio:
Minor comment (out of curiosity). Otherwise LGTM

docker/Dockerfile.cloud-agent
rade: See the comment in the commit.

Issue #2286, opened 2017-02-28
Support for importing reports


2opremio:
A user came up with the following usecase:

> freezing’ Scope’s view, so that I can ZIP up a static web site

We devs do this all the time for debugging by exporting the Scope report and loading it with a fixprobe. However, this is really tedious for end users.

I see different alternatives:
1. Implement the fixprobe functionality in the UI. The UI can temporarily act as a fixprobe while the report is "imported". Some things would need to be resolved though: When to stop forwarding the report? What about other probes reporting at the same time? We may need to ask the app to ignore them temporarily, which won't be easy for weave cloud.
2. Implement `scope launch --diagnose=report.json`. Which launches an app with a fixed probe.

While (1) is more flexible (it works with existing instances) it's also much more complicated, so I would go for (2).

An issue for both (1) and (2) is forward compatibility of the standalone app with newer report schemas (which we have never worried about). We can choose to ignore that problem for now.

rade:
We already *almost* can do (2), by invoking the scope binary directly, e.g.
```
prog/scope --mode=app --app.collector=file:/home/matthias/Downloads/report.json 
```
works, but the same doesn't work via a `scope launch` because the file isn't available inside the container. A new launch flag, along the lines suggested above, and parsed by the script could easily take care of mounting the file.

PR #2301, opened 2017-03-04
report playback

rade:
Now you can launch the scope app with something like

./prog/scope --mode=app --weave=false --app.collector=file:///tmp/reports

and if the specified dir contains reports with filenames in the form `<timestamp>.{msgpack|json}[.gz]`
e.g. "1488557088545489008.msgpack.gz", then these reports are replayed in a loop at a sequence and speed determined by the timestamps.

### DIFF
--- a/app/benchmark_internal_test.go
+++ b/app/benchmark_internal_test.go
@@ -22,7 +22,7 @@ func loadReport() (report.Report, error) {
 		return fixture.Report, nil
 	}
 
-	c, err := NewFileCollector(*benchReportFile)
+	c, err := NewFileCollector(*benchReportFile, 0)
 	if err != nil {
 		return fixture.Report, err
 	}
--- a/app/collector.go
+++ b/app/collector.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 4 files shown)


rade:
This is really handy for benchmarking of the standalone app.

@awh grabbed 30s of Weave Cloud production reports for me, which I fed to a scope app running on my laptop and restricted to a single core (i.e. `GOMAXPROCS=1`). I then ran
```
for i in $(seq 1 100); do echo $i $(date); curl -o /dev/null -s http://localhost:4040/api/topology; done
```
to see at what rate I can grab the complete topology - something the UI tries to do every 5s. It's taking an average of 9s, with the CPU pegged. Ouch. And bear in mind that this isn't even reading & parsing reports - that's only done once, at startup.

2opremio:
@rade Have you explicitly tested that the demo service doesn't break with these changes? (i.e. that collecting a single file didn't break). 

For extra points you could supply a unit test for it (using https://github.com/weaveworks/common/blob/master/fs/fs.go for mocking)

rade:
> Have you explicitly tested that the demo service doesn't break with these changes? (i.e. that collecting a single file didn't break).

Yes.

> For extra points you could supply a unit test for it

I thought about it, but the pain/gain equation doesn't work out, especially considering that this really isn't a mainstream user-facing feature.

2opremio:
> I thought about it, but the pain/gain equation doesn't work out, especially considering that this really isn't a mainstream user-facing feature.

It may very well end up used for https://github.com/weaveworks/scope/issues/2286 but, fair enough.

Issue #872, opened 2016-01-28
Add arrow heads to edges


tomwilkie:
Feedback from Darren at Rancher.

davkal:
Arrow heads will only look nice, if they end at the outer border of a node. Otherwise the inner node dot will be crowded. For proper arrow offset to end at the border, we need to calculate those endpoints. It's easy for circles, but more tricky for rects with rounded corners.

pidster:
If edges are directional it makes sense to visualise this for the user to see.
If arrowheads in contact with the node are hard, a V just before the end might work.
Would be worth trying it out.

tomwilkie:
Edges are directional.

jml:
I didn't realize that "down" meant "connects to". Adding arrowheads (or some sort of symbol for indicating direction) could be quite helpful.

rade:
> Arrow heads will only look nice, if they end at the outer border of a node.

As a (temporary) alternative we could render arrow heads at the mid-point of an edge.

rade:
This can be closed, right?

Issue #875, opened 2016-01-29
Make it clear which nodes have incoming connections and which nodes have outgoing connections when you have a node selected


tomwilkie:
Something like this (when viewing the detail panel):

<img width="718" alt="screen shot 2016-01-28 at 17 11 10" src="https://cloud.githubusercontent.com/assets/444037/12664129/2a6c0dae-c5e2-11e5-8bfa-90edc41702e1.png">

I think this will make it super obvious what containers / process etc something is talking to, and what container / process is talking to it.

tomwilkie:
If you combine this with some kind of semantic colouring, this might be interesting way to answer the question "why is this container sick".

What do you think @davkal @foot @paulbellamy?

paulbellamy:
I would be more in favour of this if "incoming" vs "outgoing" was more than a guess, but even determining what that means is tricky.

tomwilkie:
conntrack'd connections have direction: #539

davkal:
I like it, though arrow offset will have to be calculated differently for all new shapes.

tomwilkie:
This also came up with Darren from Rancher.

paulbellamy:
> conntrack'd connections have direction: #539

But most connections are not conntrack'd (only short-lived ones are), no?

tomwilkie:
Actually almost all connections are conntracked, the set of ones detected
by proc spy and not by conntrack is almost nil.

tomwilkie:
Edge direction is now even more meaningful: we get them from conntrack.  @davkal do you think this is something we could have a go at (even without arrow heads)?

pidster:
+1 for adding a visual indicator for direction, to edges.

rade:
How is this different from #872?

tomwilkie:
> How is this different from #872?

This is talking about the "selected" mode, where as #872 is in the general canvas.

Also this suggestion is about changing the location of the nodes (incoming at the top, outgoing at the bottom) when you select a node, whereas #872 is just about adding a direction indicator for edges on canvas.

rade:
> Also this suggestion is about changing the location of the nodes (incoming at the top, outgoing at the bottom) when you select a node

I like that, and it's something that could be done independently from tackling the thorny/controversial issue of adding arrow heads.

jpellizzari:
Related to #1636

jpellizzari:
[...]

Issue #1636, opened 2016-07-04
make topology view edges more useful


rade:
Edges in the topology view are presently very underutilised. We could hang some additional info and controls off them.

Some ideas...
- show IPs and ports on hover
- gather/show network level metrics
- show L7 metrics (see #1631)

rade:
I am guessing that (1) could be accomplished purely in the UI. Correct @davkal @foot ?

jpellizzari:
I have some concepts ready for design feedback:

https://docs.google.com/presentation/d/1AHfwIn2PZtYmpt4XEEiWhnCN77CUaLR-BnZstiLBQNo/edit

@2opremio, @ekimekim , @davkal , @foot , @bowenli, @fbarl  please feel free to leave comments

jpellizzari:
The consensus from the feedback was to use arrows to indicate connection relationships, and to arrange nodes in the 'node details' view with client connections on top, and server connections on bottom (slide 3 reversed).

jpellizzari:
First pass at arrows:
![screen shot 2017-02-27 at 3 34 52 pm](https://cloud.githubusercontent.com/assets/2802257/23385026/5d2358d0-fd02-11e6-90a6-fa801ae446ee.png)

rade:
Nice!

This is for the radial "node selected" view, right? In that view, I think it would be helpful to try to orient the flow from top to bottom, as suggested in #875 (though still placing nodes at equal distance on a circle, perhaps leaving a gap between inbound, inbound+outbound, outbound; this does mean some flows can point in the wrong direction, but at least it's not a complete jumble).

Have we decided yet whether to also show arrow-heads on mouseover? I think we should.

2opremio:
Feature request from user:

> Is there a way (or could there be a way) of allowing me to attach
> metadata to the edges? A typical tag might be “HTTPS / TLS1.2”, and is
> clearly information that Weave wouldn’t know (my assumption, anyway)?

jpellizzari:
@2opremio I take it we can't know for sure what protocol a connection is using without packet inspection; maybe we can assume protocols based on common ports?

jpellizzari:
@rade Yes, that is the radial view.

> perhaps leaving a gap between inbound, inbound+outbound, outbound;

Yes, I was planning on doing something like that. Using the positions of the nodes to add meaning.

>  Have we decided yet whether to also show arrow-heads on mouseover?

[...]

PR #2317, opened 2017-03-09
Add edge arrows

jpellizzari:
Implements #872 and ~~#875~~:
screen shot 2017-03-09 at 11 35 57 am

This deviates from the original design in that the edges do not terminate at the arrow. In order to shorted the edge to match the position of the arrow, we would need to subtract some length from the edge and re-calculate the `path` of the edge.


rade:
Does this also show the arrows in the normal view, on mouseover, as [discussed](https://github.com/weaveworks/scope/issues/1636#issuecomment-283088506)?

jpellizzari:
@rade I had forgotten about that, but it was fairly easy to add:
screen shot 2017-03-09 at 1 10 37 pm

rade:
Nice!

I think this is good enough to close #872 and #875, though I suggest filing a follow up issue to adjust node position based on connection direction in the radial view, as per https://github.com/weaveworks/scope/issues/875#issuecomment-230283677 and https://github.com/weaveworks/scope/issues/1636#issuecomment-282954123.

rade:
Just to check, the arrow(s) also appear on mouseover of an edge, right? (right now this just highlights the edge)

jpellizzari:
They will appear when the edge is in a 'highlighted' state:
`const shouldRenderMarker = (focused || highlighted) && (source !== target);`

Where marker is the arrow.

jpellizzari:
> adjust node position based on connection direction in the radial view

I had a version of that implemented, where I placed all clients one side and all servers on another. It didn't make a lot of sense with real data, as most nodes either had one or the other. You ended up with a bunch of wasted space on one side.

I do think, however, that it will be possible to keep the nodes in their current radial shape, but place client and server nodes next to each other in the radial shape.

rade:
> I do think, however, that it will be possible to keep the nodes in their current radial shape, but place client and server nodes next to each other in the radial shape.

I am not sure what you mean by that, but the basic idea in https://github.com/weaveworks/scope/issues/1636#issuecomment-282954123 is to draw the radial shape exactly as now, with exactly as many "slots", but when it comes deciding which node to put in which slot, we bias positioning near the top for clients, near the bottom for servers, and left/right for client/servers.

A further improvement on that would be to *maybe* add a small gap between the resulting groups.

Anyway, all this should be subject of a follow-up issue. Merge, merge, merge :) (post review, of course)

jpellizzari:
Hmm, I thought I had this working but it turns out that I got a false positive on the bidirectional connections between the 'consul' containers. There may be an issue with the way the `<path />` elements are rendered by D3 that is messing up the behavior of the `markerStart` attribute.

One thing that would make this implementation much simpler would be if we remove the  `collapseMultiEdges` behavior. Then two edges get rendered for bidirectional connections and the arrows just work. 

@fbarl Do you have a sense for how much of a render improvement we get from that optimization? Do you think it would be okay to remove it for a quick win on these edge arrows? 

The alternative will be a complex custom solution in which we have to calculate the edge direction to get a polygon to point the way, rather than having the browser SVG engine render it for us.

rade:
I suggest you gather some stats on what percentage of edges get collapsed. Take something like Weave Cloud Prod, and check the process, container, pod (all namespaces), service (all namespaces) and host views.

I suspect the host view is going to have a very high ratio. And it's already looking like a bowl of spaghetti.

Perhaps don't show *any* arrows for bidirectional connections?

fbarl:
> Perhaps don't show any arrows for bidirectional connections?

This sounds like a good idea, but I'm worried about two potential downsides:
  1. The bidirectional edges will be drawing less attention than the directed ones (while in my opinion, it should be either the same or more)
  2. Will it look like a bug, as if the arrow hasn't been drawn by mistake?

Maybe both of these would be resolved by positioning the nodes well in the radial view though...

> One thing that would make this implementation much simpler would be if we remove the `collapseMultiEdges` behavior. Then two edges get rendered for bidirectional connections and the arrows just work.

From the performance perspective, I'd say go ahead and remove it, I think the effects will mostly be negligible. However, even if the two edges would be drawn perfectly on top of one another, I think there would be cases when they would look slightly different - e.g. on hover, since the semi-transparent blue border would be applied twice. Because of these tiny effects, I think it would be really preferable to draw always max one edge between a pair of nodes, but if it's really proving to be hacky and difficult to make it work with collapsed edges, I'd give my vote for the quick win :)

jpellizzari:
Some stats on the number of edges being collapsed (bidrectionalEdges / totalEdges):

Hosts: .24
Containers (all): .03
Pods (all namespaces): .02
Processes: .03
Services: .02

As expected the hosts view has the most bidrectional edges. The rest have only a few.

rade:
That's from Weave Cloud Prod?

Here's another idea: Could you "uncollapse" edges when you want to show arrows on them? Or is that either hard or does not help solve the problem?

jpellizzari:
I thinks this is ready for review now. @fbarl would you mind taking a look? 

My thoughts on this is that its an 80/20 solution that's not perfect, but I think the upside is worth it. The graph is a lot more informative now. I removed the opacity on the edges so that two edges laying on top of each other are indistinguishable from single edges. The hover background is slightly more opaque on bidirectional edges, but I think I only noticed because I knew it was there.

Some samples:
screen shot 2017-03-10 at 11 40 42 am
screen shot 2017-03-10 at 11 46 45 am

I would like this to be more extensible, but the SVG `<marker />` element is fairly restrictive. Some alternatives would be adding waypoints to our path elements and trying to use `markerMid`, but that might involve digging into the `dagre` library.

rade:
so you got bidirectional edges working properly after all?

jpellizzari:
> That's from Weave Cloud Prod?
@rade Yes thats a report from prod

> "uncollapse" edges when you want to show arrows on them

Since the arrows appear/disappear on mouseover, we would need to re-render on each those events, so I don't think it would be worth it.

jpellizzari:
@rade yes, but I cheated by just rendering two edges for bidirectional connections

rade:
Oh ok. Do you have a before/after of the Weave Cloud Prod hosts spaghetti bowl? I am curious how much worse it looks.

jpellizzari:
The edges themselves haven't changed. You won't even notice the bidirectional edges because they follow the same path and lay on top of each other. We removed them to save cycles on renders because you can't see them.

Prod:
screen shot 2017-03-10 at 11 59 50 am

This branch:
screen shot 2017-03-10 at 11 57 33 am

rade:
> You won't even notice the bidirectional edges because they follow the same path and lay on top of each other

Aha! So once again then... Merge, merge, merge :) (post review, of course)

client/app/scripts/selectors/nodes-chart-graph.js
```diff
@@ -3,7 +3,7 @@ import { createSelector, createStructuredSelector } from 'reselect';
 import { Map as makeMap } from 'immutable';
 import timely from 'timely';
 
-import { initEdgesFromNodes, collapseMultiEdges } from '../utils/layouter-utils';
+import { initEdgesFromNodes } from '../utils/layouter-utils';
```
fbarl: Since `collapseMultiEdges` is not gonna be used anymore, I'd just remove it from `layouter-utils`, together with the test.

client/app/scripts/charts/nodes-chart.js
```diff
@@ -84,12 +84,29 @@ class NodesChart extends React.Component {
...
+              id="end-arrow"
+              viewBox="1 0 10 10"
+              refX={markerOffset}
+              refY="3.5"
+              markerWidth={markerSize}
+              markerHeight={markerSize}
+              orient="auto"
+            >
+              <polygon className="link" points="0 0, 10 3.5, 0 7" />
+            </marker>
```
fbarl: Maybe here you'd want to use `Marker`, as defined in `charts/marker.js`? Since only `markerOffset` and `markerSize` are variable, I'd just rename it to `EndArrowMarker` and put `children`, `id` and `viewBox` attributes inside there so that here you can simply have `<EndArrowMarker offset={markerOffset} size={markerSize} />`

jpellizzari: Whoops, that file was supposed to be deleted. I don't think this needs to be a component, since there is only one instance of `<marker />` at the moment.

client/app/scripts/selectors/nodes-chart-graph.js
jpellizzari: Fixed

jpellizzari:
@fbarl 
> When you hover on a node in big graphs, some of its outgoing nodes might be outside of the viewport in which case you don't see the arrow at all. Have you thought about showing both incoming & outgoing arrows close to the selected/hovered node instead of close to the nodes they point at? I think it might also look nicer in the circular view.

This was brought up during the design phase. On big topologies, I think it is better to see the arrow closest to the target node. It allows for a faster understanding of the way connectivity is working. I think the highlighted edge that disappears off screen is enough to indicate that another connection is happening at an unseen node.

> Currently the selected node label can obscure some arrows in the circular view, which is especially true in the contrast mode where there the labels are fully opaque. Not sure what can be done there, but maybe I'd increase the label transparency in that case as edge arrows should be fully visible in the circular view.

Increasing the label transparency would make the text clash with the arrow. Ideally, we would be able to detect if the two were on top of one another and act accordingly. Since the edges and nodes are rendered separately, we can't get the `bondingRect` to know if a collision is happening. Also, we would have to save each node and edge as a ref and interact with the DOM directly ☹️ .

That said,  I would like to defer this to a future iteration and try to fade out the selected node label on edge hover maybe.

> I'm not sure #875 should be closed before the nodes in the circular view are reordered so that the incoming ones are always above the outgoing ones. Or did we decide that we're not implementing this in another story?

I think that's fair. Updated the PR description.

Issue #2215, opened 2017-02-08
Fall back gracefully to proc parsing eBPF is enabled and it fails


iaguis:
In the current implementation of the eBPF connection tracking (https://github.com/weaveworks/scope/pull/2135), we can't fall back gracefully to proc parsing if the eBPF loading fails for whatever reason. Instead, scope starts but it won't show any connections.

Since it's experimental and not enabled by default, it is fine for now. However, this is not ideal and we should fix it.

~Depends on #2135~

cc @alban @alepuccetti

PR #2326, opened 2017-03-13
scope: use same Docker args for early dry run

schu:
`launch` first does a `scope --dry-run` to parse and validate Scope
cmdline arguments before starting the main Scope container and
processes. Use the same Docker args for the dry run to make sure the
container has the same privileges and mount points. Otherwise valid
`$WEAVESCOPE_DOCKER_ARGS` might fail here.

### DIFF
--- a/scope
+++ b/scope
@@ -165,14 +165,20 @@ create_plugins_dir() {
         "$SCOPE_IMAGE" -c "mkdir -p /var/run/scope/plugins"
 }
 
-launch_command() {
+docker_args() {
     # shellcheck disable=SC2086
-    echo docker run --privileged $USERNS_HOST -d --name="$SCOPE_CONTAINER_NAME" --net=host --pid=host \
+    echo --privileged $USERNS_HOST --net=host --pid=host \
         -v /var/run/docker.sock:/var/run/docker.sock \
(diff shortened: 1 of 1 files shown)


schu:
We wan't to do something like 

```
 WEAVESCOPE_DOCKER_ARGS="-v /dev/net:/sys/kernel/debug/tracing" ./scope launch -probe.ebpf.connections=true
```

to test the fallback for #2215 Currently, that's not possible due to dry run with `$WEAVESCOPE_DOCKER_ARGS`

schu:
cc @alban

2opremio:
I am thinking that, instead, it may be better to simply dry-run with the exact same docker arguments as launch (privileges and all). In that way we will detect more problems early. Thoughts?

schu:
> I am thinking that, instead, it may be better to simply dry-run with the exact same docker arguments as launch (privileges and all). In that way we will detect more problems early. Thoughts?

Sound ok to me, will prepare a fixup to do that.

schu:
Updated.

2opremio:
LGTM, please update the PR title

schu:
Done.

Issue #2337, opened 2017-03-15
scope URLs do not restore context when copy-pasted


rade:
I am quite sure this used to work...

1. fire up a local scope
2. find something interesting to look at, e.g. do a search in the table mode of the process topology
3. fire up an anonymous browser session
4. in that session, find something else to look at, eg. a do a different search, in the graph mode of the container view
5. copy-paste the URL from the first session over the 2nd session and hit return

Nothing happens.

By contrast, if you copy-paste the URL right after step 3, then the proper context is restored.

Perhaps the state stored in the cookie / elsewhere somehow takes precedence over the URL state?

The same problem occurs in Weave Cloud dev/prod.

jpellizzari:
Most likely related to this commit: https://github.com/weaveworks/scope/commit/d77502079466624caff88fa1c60431f4e7bfcdcc#diff-7dc7875cc44ad2eec50c55c6932e6ec2

PR #2340, opened 2017-03-16
Fixed url state copy/paste bug

jpellizzari:
Fix for https://github.com/weaveworks/scope/issues/2337

Reverted changes made here: https://github.com/weaveworks/scope/commit/d77502079466624caff88fa1c60431f4e7bfcdcc#diff-7dc7875cc44ad2eec50c55c6932e6ec2

### DIFF
--- a/client/app/scripts/utils/router-utils.js
+++ b/client/app/scripts/utils/router-utils.js
@@ -90,20 +90,21 @@ export function updateRoute(getState) {
 
 
 export function getRouter(dispatch, initialState) {
-  let mergedState = initialState;
   // strip any trailing '/'s.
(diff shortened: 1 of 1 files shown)


rade:
What was the motivation behind the original change? Or was it accidental?

jpellizzari:
The motivation behind the original change was to prevent scope re-routing when switching to flux or prom within Weave Cloud. The `page` module conflicts with the `react-router` module, since they are both trying to do the same thing.

Since scope has been rolled back to an iframe, this change should restore the behavior until I can figure out how to work around it. It's also worth noting that the `page` module is no longer maintained, and is also causing problems with deep linking.

Issue #2341, opened 2017-03-16
Bidrectional edges only show one arrow on mouse over


jpellizzari:
Reproduce:
1. Hover over a bidirectional edge (NOT the node)

Expected:
Two arrows

Actual: 
Only one arrow (because you are only hovering over one; the other is laying underneath it)

jpellizzari:
Closed by #2342

PR #2342, opened 2017-03-17
Fix for inaccurate bidrectional edges on edge hover

jpellizzari:
Fix for #2341 

Fixes a bug where bi-drectional arrows don't show up when mousing over the edge. Mousing over the node has always worked, but edge mouseover did not because only one edge is hover-able.

screenshot from 2017-03-16 21-23-13

@fbarl Could you review please?

### DIFF
--- a/client/app/scripts/reducers/root.js
+++ b/client/app/scripts/reducers/root.js
@@ -421,7 +421,10 @@ export function rootReducer(state = initialState, action) {
       // highlight edge
       state = state.update('highlightedEdgeIds', (highlightedEdgeIds) => {
         highlightedEdgeIds = highlightedEdgeIds.clear();
-        return highlightedEdgeIds.add(action.edgeId);
+        highlightedEdgeIds = highlightedEdgeIds.add(action.edgeId);
+        const opposite = action.edgeId.split(EDGE_ID_SEPARATOR).reverse().join(EDGE_ID_SEPARATOR);
+        highlightedEdgeIds = highlightedEdgeIds.add(opposite);
+        return highlightedEdgeIds;
       });
 
       return state;
--- a/client/app/scripts/charts/__tests__/node-layout-test.js
+++ b/client/app/scripts/charts/__tests__/node-layout-test.js
@@ -1,5 +1,7 @@
 import { fromJS, Map } from 'immutable';
 
+import { constructEdgeId as edge } from '../../utils/layouter-utils';
+
 const makeMap = Map;
 
 describe('NodesLayout', () => {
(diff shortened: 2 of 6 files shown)


client/app/scripts/reducers/root.js
```diff
@@ -421,7 +421,12 @@ export function rootReducer(state = initialState, action) {
       // highlight edge
       state = state.update('highlightedEdgeIds', (highlightedEdgeIds) => {
         highlightedEdgeIds = highlightedEdgeIds.clear();
-        return highlightedEdgeIds.add(action.edgeId);
+        highlightedEdgeIds = highlightedEdgeIds.add(action.edgeId);
+        if (action.bidirectional) {
+          const opposite = action.edgeId.split(EDGE_ID_SEPARATOR).reverse().join(EDGE_ID_SEPARATOR);
+          highlightedEdgeIds = highlightedEdgeIds.add(opposite);
+        }
+        return highlightedEdgeIds;
```
fbarl: I'm looking at how we're updating the highlighted edges list in `ENTER_NODE` action below:

```
// all neighbour combinations because we dont know which direction exists
highlightedEdgeIds = highlightedEdgeIds.union(adjacentNodes.flatMap(adjacentId => [
  [adjacentId, nodeId].join(EDGE_ID_SEPARATOR),
  [nodeId, adjacentId].join(EDGE_ID_SEPARATOR)
]));
```
There we don't care about bidirectionality and we add the reversed edge IDs even when those edges don't exist. Given how `highlightedEdgeIds` is currently used, that shouldn't be a problem. Maybe you can take the same simple approach here and just do something like:

```
highlightedEdgeIds = highlightedEdgeIds.clear();
highlightedEdgeIds = highlightedEdgeIds.add(action.edgeId);

const oppositeEdgeId = action.edgeId.split(EDGE_ID_SEPARATOR).reverse().join(EDGE_ID_SEPARATOR);
return highlightedEdgeIds = highlightedEdgeIds.add(oppositeEdgeId);
```
Then `action.bidirectional` wouldn't be used and you could remove half of the code from this PR. What do you think?

fbarl: Suggesting that because keeping track of bidirectionality seems kinda unnecessary if it doesn't reflect the way edges are rendered.

jpellizzari: Hmm okay so my original thought would be not to add a bunch of extra edges that are not bidirectional to the `highlightedEdgeIds` list. It would be confusing to someone looking at the state tree and seeing a bunch of edges that don't exist.

But I guess the precedent has already been set and that would make this simpler.

jpellizzari:
@fbarl Simplified with the approach you laid out.

fbarl:
> Hmm okay so my original thought would be not to add a bunch of extra edges that are not bidirectional to the `highlightedEdgeIds` list. It would be confusing to someone looking at the state tree and seeing a bunch of edges that don't exist.

I absolutely agree the only proper way to do it would be not to include non-existent IDs there, but that's also something that could be done without bidirectionality information (if we intersect with the list of all edge IDs). That's what I plan to do in the next refactoring round by moving both `highlightedEdgeIds` and `highlightedNodeIds` into selectors. That will make this whole thing simpler :)

fbarl:
I just discovered an issue that hopefully didn't leave a big mark so far, but is relevant for this story. Having `EDGE_ID_SEPARATOR` set to a simple dash isn't the best policy as host IDs can contain dashes, so an example edge we get looks something like:

`ip-172-20-2-209;<host>-ip-172-20-1-176;<host>`

Then a big mess happens when `split(EDGE_ID_SEPARATOR).reverse().join(EDGE_ID_SEPARATOR)` is executed on such a string. :) That resulted in this PR fix having no effect when I tested it on the *Hosts* topology.

A simple solution would be to set `EDGE_ID_SEPARATOR` to something that we can be fairly sure is never gonna be a part of a node ID. Perhaps `-----`? :D

jpellizzari:
Nice catch! Updated the edge ID separator to several dashes

PR #2353, opened 2017-03-20
Fix help panel fields max height

### LINKED ISSUE #2352: Help panel height overflow in Containers view
help

fbarl:
Fixes #2352.

### DIFF
--- a/client/app/styles/_base.scss
+++ b/client/app/styles/_base.scss
@@ -1650,7 +1650,7 @@
           // Notes: Firefox gets a bit messy if you try and bubble
           // heights + overflow up (min-height issue + still doesn't work v.well),
           // so this is a bit of a hack.
-          max-height: "calc(100vh - 160px - 160px - 160px)";
+          max-height: calc(100vh - 160px - 160px - 160px);
         }
       }
     }


fbarl:
Eh, iMachines and scrollbars :/ We currently have the *overflow* setting at `auto`, which works as it should both on my Chrome and Firefox. I talked to @foot about it and apparently there is a setting you can change to always show scrollbars (http://osxdaily.com/2011/08/03/show-scroll-bars-mac-os-x-lion/). I assume you have it off, because scrollbars were displayed nicely on Simon's mac.

Three ideas that wouldn't require OS X users to change their settings:
  1. Setting `overflow-y: scroll` might force the scrollbars visibility, but honestly it looks very bad on my Chrome when there's no overflow.
  2. Doing some quick CSS magic (hopefully without Javascript), to somehow signify when there is more content below or above (e.g. white shadows) that would work on all platforms. Then, if we would want a consistent UI, it would be good to hide the existing scrollbars on Chrome, but that seems hacky too.
  3. Using an external library or our own component for scrolling - seems like an overkill for this issue.

@davkal All in all, I'd prefer to just merge it as it is (after you confirm it works well by turning on the scrollbar settings on your machine), but I'd also be willing to consider 2 and 3, although perhaps in a separate story?

davkal:
I'd prefer a show-more link, that visibly expands the container, somehow carries a number of entries, and thereby makes it easy to deduce how many items one should see (or scroll through). But yeah, different story.

Issue #2357, opened 2017-03-20
'scope launch' fails on OSX


jpellizzari:
`scope launch` fails with this error:
```
docker: Error response from daemon: Mounts denied: 
The path /var/run/scope/plugins
is not shared from OS X and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> File Sharing.
See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
```

Most likely a result of an early ` --dry-run` before the `check_docker_for_mac` block is triggered. Looks like it is related to this: https://github.com/weaveworks/scope/commit/5f85256055d5eeb7a620d814aa4e448eb60f1560

jpellizzari:
Closed by #2358

PR #2358, opened 2017-03-20
Changed --dry-run command to execute after OSX setup

jpellizzari:
Fix for #2357 

Moves the `--dry-run` to after OSX setup is completed.

### DIFF
--- a/scope
+++ b/scope
@@ -203,6 +203,13 @@ print_app_endpoints() {
     done
 }
 
+dry_run() {
+    # Do a dry run of scope in the foreground, so it can parse args etc
+    # avoiding the entrypoint script in the process.
+    # shellcheck disable=SC2046
+    docker run --rm --entrypoint=/home/weave/scope $(docker_args) "$SCOPE_IMAGE" --dry-run "$@"
+}
+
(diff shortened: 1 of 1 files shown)


scope
```diff
@@ -260,6 +255,10 @@ case "$COMMAND" in
             print_app_endpoints "localhost"
             exit
         fi
+        # Do a dry run of scope in the foreground, so it can parse args etc
+        # avoiding the entrypoint script in the process.
+        # shellcheck disable=SC2046
+        docker run --rm --entrypoint=/home/weave/scope $(docker_args) "$SCOPE_IMAGE" --dry-run "$@"
```
rade: This ends up omitting the dry run altogether on d4m.

scope
```diff
@@ -182,6 +182,7 @@ launch_command() {
 }
 
 launch_docker4mac_app_command() {
+    dry_run
```
rade: You don't really want to invoke a side-effecting command here, seeing as this function is simply meant to return a command (as the name suggests).

So I'd put this just before the `check_not_running` calls, which is the same place as in the non-d4m case.

PR #2359, opened 2017-03-20
Fluent Billing Emitter

paulbellamy:
Ignore the commits, they're to be squashed.

* Adds a billing-client package (to be extracted, probably), and `app/multitenant/billing_emitter`, which emit json to a fluentd instance running on `localhost:24225`

## TODO:

- [x] More work around reliability
- [ ] Add tests
- [x] Squash commits

### DIFF
--- a/app/multitenant/billing_emitter.go
+++ b/app/multitenant/billing_emitter.go
@@ -1,0 +1,116 @@
+package multitenant
+
+import (
+	"crypto/sha256"
+	"encoding/base64"
+	"flag"
+	"strings"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	billing "github.com/weaveworks/billing-client"
+	"golang.org/x/net/context"
+
+	"github.com/weaveworks/scope/app"
+	"github.com/weaveworks/scope/report"
+)
+
+// BillingEmitterConfig has everything we need to make a billing emitter
+type BillingEmitterConfig struct {
+	Enabled         bool
+	DefaultInterval time.Duration
+	UserIDer        UserIDer
+}
+
+// RegisterFlags registers the billing emitter flags with the main flag set.
+func (cfg *BillingEmitterConfig) RegisterFlags(f *flag.FlagSet) {
+	f.BoolVar(&cfg.Enabled, "app.billing.enabled", false, "enable emitting billing info")
+	f.DurationVar(&cfg.DefaultInterval, "app.billing.default-publish-interval", 3*time.Second, "default publish interval to assume for reports")
+}
+
+// BillingEmitter is the billing emitter
+type BillingEmitter struct {
+	app.Collector
+	BillingEmitterConfig
+	billing *billing.Client
+}
+
+// NewBillingEmitter changes a new billing emitter which emits billing events
+func NewBillingEmitter(upstream app.Collector, billingClient *billing.Client, cfg BillingEmitterConfig) (*BillingEmitter, error) {
+	return &BillingEmitter{
+		Collector:            upstream,
+		billing:              billingClient,
+		BillingEmitterConfig: cfg,
+	}, nil
+}
+
+// Add implements app.Collector
+func (e *BillingEmitter) Add(ctx context.Context, rep report.Report, buf []byte) error {
+	now := time.Now().UTC()
+	userID, err := e.UserIDer(ctx)
+	if err != nil {
+		return err
+	}
+	rowKey, colKey := calculateDynamoKeys(userID, now)
+
+	containerCount := int64(len(rep.Container.Nodes))
+	interval := e.reportInterval(rep)
+	hasher := sha256.New()
+	hasher.Write(buf)
+	hash := "sha256:" + base64.URLEncoding.EncodeToString(hasher.Sum(nil))
+	amounts := billing.Amounts{billing.ContainerSeconds: int64(interval) * containerCount}
+	metadata := map[string]string{
+		"row_key": rowKey,
+		"col_key": colKey,
+	}
+
+	err = e.billing.AddAmounts(
+		hash,
+		userID,
+		now,
+		amounts,
+		metadata,
+	)
+	if err != nil {
+		log.Errorf("Failed emitting billing data: %v", err)
+		return err
+	}
+
+	return e.Collector.Add(ctx, rep, buf)
+}
+
+// reportInterval tries to find the custom report interval of this report. If
+// it is malformed, or not set, it returns false.
+func (e *BillingEmitter) reportInterval(r report.Report) time.Duration {
+	var inter string
+	for _, c := range r.Process.Nodes {
+		cmd, ok := c.Latest.Lookup("cmdline")
+		if !ok {
+			continue
+		}
+		if strings.Contains(cmd, "scope-probe") &&
+			strings.Contains(cmd, "probe.publish.interval") {
+			cmds := strings.SplitAfter(cmd, "probe.publish.interval")
+			aft := strings.Split(cmds[1], " ")
+			if aft[0] == "" {
+				inter = aft[1]
+			} else {
+				inter = aft[0][1:]
+			}
+
+		}
+	}
+	if inter == "" {
+		return e.DefaultInterval
+	}
+	d, err := time.ParseDuration(inter)
+	if err != nil {
+		return e.DefaultInterval
+	}
+	return d
+}
+
+// Close shuts down the billing emitter and billing client flushing events.
+func (e *BillingEmitter) Close() error {
+	return e.billing.Close()
+}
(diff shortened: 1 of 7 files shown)


tomwilkie:
> to be extracted, probably

yes please.

app/multitenant/billing_emitter.go
```diff
@@ -0,0 +1,127 @@
...
+		Help:      "Time in seconds spent emitting billing info.",
+		Buckets:   prometheus.DefBuckets,
+	}, []string{"method", "status_code"})
+)
+
+func init() {
+	prometheus.MustRegister(billingRequestDuration)
+}
+
+// BillingEmitterConfig has everything we need to make an AWS collector.
```
tomwilkie: > AWS collector.

?

tomwilkie: Oh I see - this implements and wraps an existing collector...

The comment should say that.

app/multitenant/billing_emitter.go
```diff
@@ -0,0 +1,127 @@
...
+	billing         *billing.Client
+}
+
+// NewBillingEmitter is the charging-mechanism
+func NewBillingEmitter(upstream app.Collector, cfg BillingEmitterConfig) (app.Collector, error) {
+	billingClient, err := billing.New()
+	return &billingEmitter{
+		Collector:       upstream,
+		userIDer:        cfg.UserIDer,
+		defaultInterval: cfg.DefaultInterval,
```
tomwilkie: Personal preference is to embed the cfg struct in the component, means adding less code when adding more config.

app/multitenant/billing_emitter.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return err
+	}
+	rowKey, colKey := calculateDynamoKeys(userID, time.Now())
+
+	containerCount := int64(len(rep.Container.Nodes))
+	interval, ok := reportInterval(rep)
+	if !ok {
+		interval = e.defaultInterval
+	}
```
tomwilkie: Nit: I would make `reportInterval` a method on `billingEmitter` and do the defaulting within there.

app/multitenant/billing_emitter.go
```diff
@@ -0,0 +1,127 @@
...
+	hasher := sha256.New()
+	hasher.Write(buf)
+	hash := "sha256:" + base64.URLEncoding.EncodeToString(hasher.Sum(nil))
+	timestamp := time.Now().UTC()
```
tomwilkie: Nit: unify the now with L62?

app/multitenant/billing_emitter.go
```diff
@@ -0,0 +1,127 @@
...
+	amounts := map[string]int64{
+		billing.ContainerSeconds: int64(interval) * containerCount,
+	}
+	metadata := map[string]string{
+		"row_key": rowKey,
+		"col_key": colKey,
+	}
+
+	err = instrument.TimeRequestHistogram(ctx, "Billing.AddAmounts", billingRequestDuration, func(_ context.Context) error {
```
tomwilkie: Nit: If you rolled the instrumentation into the billing client, it could be common across all users...

app/multitenant/billing_emitter.go
```diff
@@ -0,0 +1,127 @@
...
+
+func init() {
+	prometheus.MustRegister(billingRequestDuration)
+}
+
+// BillingEmitterConfig has everything we need to make an AWS collector.
+type BillingEmitterConfig struct {
+	UserIDer        UserIDer
+	DefaultInterval time.Duration
+}
```
tomwilkie: We've got a pattern in Cortex for putting the registration of flags on a method on the config struct - helps keep the names & default values close to where they are used.  You might like it: https://github.com/weaveworks/cortex/blob/master/chunk/chunk_cache.go#L71

billing-client/client.go
```diff
@@ -0,0 +1,157 @@
...
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/fluent/fluent-logger-golang/fluent"
+)
+
+const (
+	// TODO: Figure out good values for these
+	maxBufferedEvents = 1024
+	retryDelay        = 500 * time.Millisecond
```
tomwilkie: These should be configuration as flags, with a Config struct etc.

billing-client/client.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// FluentHostPort just points to localhost on the expected billing port. The
+	// billing fluentd instance *must* be run in each pod wishing to report
+	// billing events.
+	FluentHostPort = "localhost:24225"
```
tomwilkie: Ditto this should be a flag.

app/multitenant/billing_emitter.go
```diff
@@ -0,0 +1,127 @@
...
+		billing:         billingClient,
+	}, err
+}
+
+func (e *billingEmitter) Add(ctx context.Context, rep report.Report, buf []byte) error {
+	userID, err := e.userIDer(ctx)
+	if err != nil {
+		return err
+	}
+	rowKey, colKey := calculateDynamoKeys(userID, time.Now())
```
tomwilkie: I forget: does the report contain a "generated at" timestamp that we should be using?

billing-client/client.go
```diff
@@ -0,0 +1,157 @@
...
+		UniqueKey:          uniqueKey,
+		InternalInstanceID: internalInstanceID,
+		OccurredAt:         timestamp,
+		Amounts:            amounts,
+		Metadata:           metadata,
+	}
+
+	select {
+	case <-c.stop:
+		// TODO: Should this be a panic? We just threw away billing data!
```
tomwilkie: Probably want to add a prometheus metric here to count the number we throw away?

Also, why two select loops?  Isn't one enough?

billing-client/client.go
```diff
@@ -0,0 +1,157 @@
...
+			if err == nil {
+				return nil
+			}
+			log.Errorf("Billing: failed to log event: %v: %v, retrying in %v", e, err, retryDelay)
+			time.Sleep(retryDelay)
+		}
+	}
+}
+
+func (c *Client) Close() error {
```
tomwilkie: Should the collector be calling this?

billing-client/client.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	select {
+	case c.events <- e: // Put event in the channel unless it is full
+		return nil
+	default:
+		// full
+	}
+	// TODO: Big angry log message here, cause we're throwing away billing data
+	return fmt.Errorf("Reached billing event buffer limit (%d), discarding event: %v", maxBufferedEvents, e)
```
tomwilkie: A metric for number of dropped would be good...

billing-client/client.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+
+	// flush remaining events
+	for done := false; !done; {
+		select {
+		case event := <-c.events:
+			c.post(event)
+		default:
+			done = true
```
tomwilkie: You could just return here...

billing-client/client.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+}
+
+func (c *Client) post(e Event) error {
+	for {
+		select {
+		case <-c.stop:
+			return fmt.Errorf("Billing: failed to log event: %v", e)
```
tomwilkie: Won't this interact badly with the flushing in `loop` - ie won't it always fail to flush anything after being stopped?

tomwilkie:
Bunch of comments, mostly optional nits.  Looks good otherwise.

billing-client/event.go
```diff
@@ -0,0 +1,34 @@
...
+	var records []map[string]interface{}
+	for t, v := range e.Amounts {
+		records = append(records, map[string]interface{}{
+			"unique_key":           e.UniqueKey + ":" + t,
+			"internal_instance_id": e.InternalInstanceID,
+			"amount_type":          t,
+			"amount_value":         v,
+			"occurred_at":          e.OccurredAt,
+			"metadata":             e.Metadata,
+		})
```
tomwilkie: I'd be more comfortable is this was a struct, although I realise that doesn't really offer much protection...

app/multitenant/billing_emitter.go
paulbellamy: ah comment is wrong. will fix.

app/multitenant/billing_emitter.go
paulbellamy: not afaik

billing-client/event.go
paulbellamy: msgpack (and therefore fluentd) requires it to be a `map[string]interface{}`

paulbellamy: will add a comment explaining that

billing-client/client.go
paulbellamy: first is the ckeck if the stop chan is still open, second is to send the event. go... :(

PR #2361, opened 2017-03-20
Clear client polling and nodes state on dismount 

jpellizzari:
Needed for https://github.com/weaveworks/service-ui/pull/358

Prevents additional polling when Scope is dismounted while running as a component. Also cleans up old nodes to show the correct loading screen when mounted.


davkal:
Works great! LGTM

PR #2348, opened 2017-03-17
Revert "Merge pull request #2285 from weaveworks/mike/k8s-ns-in-container-view"

ekimekim:
This reverts commit 76ddc75fb8180011202ab1494e722758b43d2a86, reversing
changes made to 3ade2933eb7722ef4c8f2fe66eab9a58802daea9.

We are rolling this back for now because it's causing a bug where sub-topologies
would have ~3000 repeated cases of the k8s filters, causing performance issues clientside.

PR #2362, opened 2017-03-20
When k8s present, allow filtering of containers by namespace (take 2)

ekimekim:
Re-applies https://github.com/weaveworks/scope/pull/2285 (reverted in https://github.com/weaveworks/scope/pull/2348)
and fixes the bug(s) that prompted the revert, which turned out to be a matter of accidentially mutating
objects in a variety of ways, some of which were introduced by https://github.com/weaveworks/scope/pull/2285 but
others that weren't, but were exposed as a problem due to changing the operation from a set to an append.

Fixes https://github.com/weaveworks/scope/issues/1504

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -81,6 +81,7 @@ func updateFilters(rpt report.Report, topologies []APITopologyDesc) []APITopolog
 	}
 	sort.Strings(ns)
 	if len(ns) > 0 { // We only want to apply k8s filters when we have k8s-related nodes
+		topologies = append([]APITopologyDesc{}, topologies...) // Make a copy so we can make changes safely
 		for i, t := range topologies {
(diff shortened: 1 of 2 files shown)


app/api_topologies.go
```diff
@@ -80,22 +80,27 @@ func updateFilters(rpt report.Report, topologies []APITopologyDesc) []APITopolog
 		ns = append(ns, namespace)
 	}
 	sort.Strings(ns)
-	for i, t := range topologies {
-		if t.id == podsID || t.id == servicesID || t.id == deploymentsID || t.id == replicaSetsID {
-			topologies[i] = updateTopologyFilters(t, []APITopologyOptionGroup{
-				kubernetesFilters(ns...), unmanagedFilter,
-			})
+	if len(ns) > 0 { // We only want to apply k8s filters when we have k8s-related nodes
```
rade: I'd invert that, returning `topologies` in the branch, make it operate on `len(namespace)` and move it before all this `ns` stuff.

PR #2351, opened 2017-03-20
Remove LatestMap, to reduce memory allocation

bboreham:
Previously, classes like `StringLatestMap` were implemented as a shim over `LatestMap`.
This meant that the memory allocations were:
[`ps.Map` leaf] -> [`interface{}` indirection] -> `LatestEntry` -> [`interface{}` indirection] -> [string contents]

Removing `LatestMap` and storing the entry via pointer rather than value gives us instead:

[`ps.Map` leaf] -> `stringLatestEntry` -> [string contents]

Also the `LatestEntry` was copied one more time on entry into the map. Thus this PR removes 50% of the memory allocations needed to create this particular structure, per entry.

(not counting the allocation for the map key)

PR #2364, opened 2017-03-21
Map decode optimisations

bboreham:
This is branched after #2351, so look at that one first.

Here I apply the the map-reading code taken from `LatestMap` to other uses of `ps.Map`; this removes the indirection via a Go map and consequent copying.

Also call `CodecDecodeSelf()` instead of `Decode()` where possible: this avoids a runtime type lookup, so goes a little faster.  I had to get a little funky with providing an implementation for the parser to run before the real codec code is generated; that is the purpose of `dummySelfer`.

Bonus: having less recursion through `Decode()` makes it easier to interpret profiles.


2opremio:
What's the performance improvement?

bboreham:
Same test as in #2351:

With this PR, commit de97a72661ee3dc082bc3f7a0565aa73818acf65:

```
$ go test -run TestDecode
ok  	github.com/weaveworks/scope/report	1.197s
ok  	github.com/weaveworks/scope/report	1.363s
ok  	github.com/weaveworks/scope/report	1.200s
ok  	github.com/weaveworks/scope/report	1.308s
```

Compare against 1.652:
That's about a 27% speed-up, fastest to fastest.

bboreham:
Note CI failed because you run the unit-tests before running the code-generation step.

2opremio:
Uhm, could the new code generation dependency have messed up coveralls?

bboreham:
The problem is that you weren't testing the right code before.
The coverage of `report.codecgen.co` is at 37%, and it wasn't included before, so that drags the average down.
Arguably this is meaningless, since most of the code paths in that file are for situations we don't expect.  But I don't know what you can do about that.

2opremio:
OK, fair enough

bboreham:
Rebased so we don't see the two commits already merged.  Sorry, should have done that earlier.

bboreham:
Tests failed sometimes with a "panic: Not run on same source!" message from `cover`:
- https://circleci.com/gh/weaveworks/scope/6857
- https://circleci.com/gh/weaveworks/scope/6858
and succeeded sometimes:
- https://circleci.com/gh/weaveworks/scope/6866
- https://circleci.com/gh/weaveworks/scope/6848

I suspect this is caused by `codecgen` embedding a random number into the code, and this can be different between the two CircleCI lanes.
So I pushed another commit fixing this number for CircleCI builds.

Also rebased just now.

PR #2366, opened 2017-03-21
endpoint/ebpf: implement stop

schu:
Since d60874aca8dee7d4d0c1a1c28f3394e60a5c553b `connectionTracker` can
fallback when the `EbpfTracker` died. Hence we only have to stop the
`tracer` in `stop()`.

This commit is also a fixup for d60874aca8dee7d4d0c1a1c28f3394e60a5c553b
where we do a gentle fallback but never actually stop the tracer to stop
polling.


2opremio:
Please bump https://github.com/weaveworks/tcptracer-bpf in this PR to include https://github.com/weaveworks/tcptracer-bpf/pull/31

PR #2386, opened 2017-03-24
Decode reports from a byte buffer

bboreham:
This is speculative: empirically it goes faster but theoretically it could make things worse.

So I would like to try it out by merging into master hence into dev, and if we don't like it we can revert.

### DIFF
--- a/report/benchmark_internal_test.go
+++ b/report/benchmark_internal_test.go
@@ -1,0 +1,40 @@
+package report
+
+import (
+	"flag"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+var (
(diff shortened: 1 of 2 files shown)


rade:
> So I would like to try it out by merging into master hence into dev, and if we don't like it we can revert.

The cleaner way to do this is to de-automate the relevant services in flux, and then deploy an image from this branch. cc @2opremio who has done this a few times.

2opremio:
Yep, I would rather do that before merging. @bboreham mind giving it a try? I can guide you. It basically entails:

1. Modifying flux's config in dev to stop deploying scope automatically
2. Deploying you image manually.

PR #2394, opened 2017-03-24
Add version to Scope UI module S3 path

jpellizzari:
Adds the Scope short commit hash to the path to module path on S3. This will require the `service-ui` app to explicitly declare which version of Scope should be installed. This introduces a (currently) manual step of changing the Scope version in `service-ui` and committing. The upside is an easy-to-understand commit log and predictable rollbacks.


jpellizzari:
@ekimekim Re-purposed this PR to just add an a version directory so that Scope can be called in by its commit hash.

Issue #2393, opened 2017-03-24
Search doesn't consider nodes of unloaded topologies


fbarl:
When doing a generic query (landing on the *PODS* topology), we only get that topology matched by the query:
![search-no-match](https://cloud.githubusercontent.com/assets/1216874/24308478/53292628-10c8-11e7-9b68-ae9330db202b.png)

Then when switching to a different topology, both of them are matched:
![search-no-match-2](https://cloud.githubusercontent.com/assets/1216874/24308525/7a83dc22-10c8-11e7-9aca-d1fb2d9d5703.png)

This seems to imply that only the loaded nodes are matched, as opposed to all the nodes across all topologies, as it should be by design of `getAllNodes` method.

*Note:* This only happens on dev and cloud for some reason, not locally for some reason.

fbarl:
Fixed by #2395.

Issue #2391, opened 2017-03-24
Resource view not rendering anything on dev


fbarl:
![not-rendering](https://cloud.githubusercontent.com/assets/1216874/24307101/241307e6-10c3-11e7-9939-d88497ad0953.png)

Seems to be related to #2393.

fbarl:
Fixed by #2395.

PR #2395, opened 2017-03-24
Using `doRequest` when fetching nodes of multiple topologies

fbarl:
It seems to make #2393 work (at least with the local UI servers that proxy to the dev); also let's see if it does anything to #2391.

### DIFF
--- a/client/app/scripts/utils/web-api-utils.js
+++ b/client/app/scripts/utils/web-api-utils.js
@@ -1,6 +1,5 @@
 import debug from 'debug';
 import reqwest from 'reqwest';
-import trimStart from 'lodash/trimStart';
 import defaults from 'lodash/defaults';
 import { Map as makeMap } from 'immutable';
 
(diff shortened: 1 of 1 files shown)


client/app/scripts/utils/web-api-utils.js
```diff
@@ -167,11 +166,8 @@ function getNodesForTopologies(getState, dispatch, topologyIds, topologyOptions
     .filter((_, topologyId) => topologyIds.contains(topologyId))
     .reduce((sequence, topologyUrl, topologyId) => sequence.then(() => {
       const optionsQuery = buildOptionsQuery(topologyOptions.get(topologyId));
-      // Trim the leading slash from the url before requesting.
-      // This ensures that scope will request from the correct route if embedded in an iframe.
-      return fetch(`${trimStart(topologyUrl, '/')}?${optionsQuery}`);
+      return doRequest({ url: `${getApiPath()}${topologyUrl}?${optionsQuery}` });
```
jpellizzari: Was the `trimStart` causing issues with the URL path?

fbarl: No, `trimStart` alone wasn't the problem as it didn't help when I simply removed it. Maybe it's about the additional CSRF thingy that `doRequest` does?

PR #2398, opened 2017-03-27
Keep the resource metric pinned on mouse out

### LINKED ISSUE #2397: Pinned metric temporarily not displayed on mouse leave
no-pinned

Introduced by #2296 (currently only on dev).

fbarl:
Fixes #2397.

### DIFF
--- a/client/app/scripts/components/metric-selector.js
+++ b/client/app/scripts/components/metric-selector.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 
 import { selectMetric } from '../actions/app-actions';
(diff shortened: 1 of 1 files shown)


foot:
I think this logic is encapsulated in https://github.com/weaveworks/scope/blob/master/client/app/scripts/components/metric-selector.js#L16 but I'm not sure why its not working there.

Still getting an env going...

fbarl:
Good catch @foot, let me look into it

foot:
Oh! Perhaps `props.pinnedMetric` is not being passed in.

foot:
LGTM

PR #2399, opened 2017-03-27
Add a test that checks if reports with data round-trip

bboreham:
Previously the only roundtrip test was for an empty report. This test has fake data similar to that found in real reports.

`Node.Metrics` does not round-trip exactly: it is tagged with `omitempty` so when decoded an empty map turns into the zero value.  A `DeepEqual` workaround is added so the test passes.
Perhaps it would be better to adjust the decode so the field is not nil after decoding.

This PR is a precursor to tests which check backwards compatibility: for instance that the binary data encoded by version 1.2 still decodes to the expected structure.

### DIFF
--- a/report/marshal_test.go
+++ b/report/marshal_test.go
@@ -5,8 +5,11 @@ import (
 	"compress/gzip"
 	"reflect"
 	"testing"
+	"time"
 
+	"github.com/weaveworks/common/mtime"
 	"github.com/weaveworks/scope/report"
+	s_reflect "github.com/weaveworks/scope/test/reflect"
 )
 
 func TestRoundtrip(t *testing.T) {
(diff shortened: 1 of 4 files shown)


2opremio:
@bboreham Mind fixing the linting error?

bboreham:
I did the `DeepEqual` thing with a tag, seems neater.

2opremio:
@bboreham Mind fixing the linter errors?

rade:
@bboreham IIRC you felt that the `DeepEqual` thing still is rather much of a hack.

I don't really want this PR to sit around here forever. So please make a call whether to proceed or close.

bboreham:
No, I didn't say that.  Have rebased and removed the "tests which check backwards compatibility: for instance that the binary data encoded by version 1.2 still decodes to the expected structure", which was too ugly.

bboreham:
I would like this test to be added.  Have rebased on latest master.

PR #2405, opened 2017-03-28
Add check for old options

jpellizzari:
Adds a check to ensure that breaking states are not loaded from localStorage or from the URL. Topology option values were changed from a `String` type to an `Array` type. If a string is loaded, the UI will break.

This can be removed after a reasonable amount of time has gone by (a few weeks maybe?).


jpellizzari:
Merging to ensure that Scope doesn't get rolled out without this during our daily deploy.

Issue #2379, opened 2017-03-23
Integration test 311 shows failures in eBPF tracker initialization


alban:
The integration test `integration/311_container_to_container_edge_with_ebpf_test.sh` sometimes fails.

Example: https://circleci.com/gh/weaveworks/scope/6881

The ebpf offset guessing work fails at different steps. Examples:

> <probe> WARN: 2017/03/23 07:37:09.109943 Error setting up the eBPF tracker, falling back to proc scanning: failed to init perf map for IPv4 events: error guessing offsets: in valid guessing state while guessing source port, got checking expected checked, status.offset_saddr: 4

> <probe> WARN: 2017/03/23 08:33:39.356490 Error setting up the eBPF tracker, falling back to proc scanning: failed to init perf map for IPv4 events: error guessing offsets: in valid guessing state while guessing destination address IPv6, got checking expected checked, status.offset_saddr: 4

So far, only reproducible on CircleCI by repetitively run the tests.

Some more details in https://github.com/weaveworks/tcptracer-bpf/pull/30

iaguis:
This is fixed by #2411. We can close it now.

PR #2411, opened 2017-03-29
Update tcptracer-bpf and re-enable test 311

iaguis:
We found out we [were losing kretprobes sometimes](https://github.com/weaveworks/tcptracer-bpf/issues/24) because maxactive was set too low. This problem was more apparent in our GCE test environment because the kernel was configured with `CONFIG_PREEMPT` not set and we're running single-core VMs (see https://github.com/weaveworks/tcptracer-bpf/issues/24 for more details).

Unfortunately, we can't set the maxactive explicitly from userspace. @alban [submitted a kernel patch to allow this](https://lkml.org/lkml/2017/3/28/629).

This PR bumps tcptracer-bpf to include a [workaround](https://github.com/weaveworks/tcptracer-bpf/pull/33) for this issue in the guess-offsets phase.

Last week we [disabled integration test 311](https://github.com/weaveworks/scope/issues/2379) but we expect it to not fail anymore so we're also re-enabling it.


iaguis:
I'm currently running the re-enabled test in a loop.

iaguis:
I ran the test around 100 times and didn't see any failures.

PR #2415, opened 2017-03-30
Renameamifile

abuehrle:
@2opremio renamed the file and also removed the h1 heading as it shows up duplicate in WP. 

The table in markdown is not displaying correctly, so have to manually line this up in WP. Hopefully this will not be a problem on the new site.


abuehrle:
I've added a screen capture for now for the AMI table (just for the WP site).

PR #2214, opened 2017-02-08
Fix rounded square node shadow radius

fbarl:
Fixes #2212.

PR #2418, opened 2017-03-31
Increase cloud node thickness in graph view

### LINKED ISSUE #2417: Cloud node border too thin comparing to other nodes
cloud

fbarl:
Fixes #2417.

All nodes shapes are now rendered from the same template which unifies the styles across all shapes. The polygon splines are now pre-cached, but that only brought about 10% rendering performance improvement which is unfortunately almost negligible in practice, since most time is still being spent on calculating the layout, not rendering it.

Node shadow now has its own border, which so we don't need to make manual corrections for shapes like clouds or rectangles (see #2214).


foot:
It works!
screen shot 2017-04-04 at 14 13 45

client/app/scripts/charts/node-shape-square.js
```diff
@@ -1,47 +0,0 @@
...
-    <g className={className}>
-      {hasMetric && getClipPathDefinition(clipId, height)}
-      {highlighted && <rect className="highlighted" {...rectProps(NODE_SHAPE_HIGHLIGHT_RADIUS)} />}
-      <rect className="border" stroke={color} {...rectProps(NODE_SHAPE_BORDER_RADIUS)} />
-      <rect className="shadow" {...rectProps(NODE_SHAPE_SHADOW_RADIUS, 0.85)} />
-      {hasMetric && <rect
-        className="metric-fill"
-        clipPath={`url(#${clipId})`}
-        style={metricStyle}
-        {...rectProps(NODE_SHAPE_SHADOW_RADIUS, 0.85)}
```
foot: The rounded corner radiiiiiii!!!

foot: #2214

client/app/styles/_variables.scss
```diff
@@ -32,8 +32,8 @@ $terminal-header-height: 44px;
 
 $node-highlight-fill-opacity: 0.1;
 $node-highlight-stroke-opacity: 0.4;
-$node-highlight-stroke-width: 2;
-$node-border-stroke-width: 6;
+$node-highlight-stroke-width: 0.04;
+$node-border-stroke-width: 0.12;
```
foot: We had issues before when working w/ very small units, (firefox maybe?)

foot:
sans the border radius LGTM :)

client/app/styles/_variables.scss
fbarl: Yup, tried it and it works fine - I think the issue with small units only concerned the fonts i.e. text, no?

client/app/scripts/charts/node-shape-square.js
fbarl: Rewrote this a bit so that we don't need to do any manual corrections like that anymore :)

fbarl:
@foot PTAL.

p.s. I also noticed that the metric fill clipping was off after first commit - that's been corrected now as well!

client/app/scripts/utils/metric-utils.js
```diff
@@ -2,24 +2,19 @@ import { includes } from 'lodash';
...
-import { NODE_BASE_SIZE, NODE_SHAPE_DOT_RADIUS } from '../constants/styles';
 import { formatMetricSvg } from './string-utils';
 import { colors } from './color-utils';
 
-export function getClipPathDefinition(clipId, height) {
+export function getClipPathDefinition(clipId, height, radius) {
   return (
     <defs>
-      <clipPath id={clipId} transform={`scale(${NODE_BASE_SIZE})`}>
+      <clipPath id={clipId} transform={`scale(${4 * radius})`}>
```
foot: What is the `4` for?

foot:
Nice work. clouds look so good now. LGTM!

We can leave this for another day but it would be cool to change the "highlight" to use the border trick too. :)

client/app/scripts/utils/metric-utils.js
fbarl: Meh, due to the way `height` is calculated in `getMetricValue`, this value has to be twice what you'd normally expect. Let me see if there's a quick fix to that, otherwise I'll just add a comment.

fbarl:
Unfortunately, the same trick wouldn't work with the `highlight` layer - the essential reason why my change worked for the border shadow is because now two different elements (`border` and `shadow`) are of very similar radius, which is not the case with `highlight`.

I'd say the highlight layer is off in two cases - cloud shapes and stacks. The former could probably be fixed by moving its center a bit lower (or by redefining the path itself), while the latter seems more tricky - I have an idea how to make it perfect with SVG masks, but that could be a bit overengineered and perhaps not really worth the effort. I'll add an issue for it though.

Issue #2392, opened 2017-03-24
Resource view - mark the zoom context better


fbarl:
Some ideas from [the updated mockups](https://docs.google.com/presentation/d/123NzxsPJDUFtsTxiH0kRFj6uc7Bq3CPnv4vK7i9XPgQ/edit#slide=id.g213a58afff_0_0):

* Add a horizontal metrics scale below the base resource layer to know what the resource box width stands for at the current zoom level.
* Do not render very thin resource boxes in have a text like `+ 31 more` in their place - this would change dynamically as we zoom in and out.
* Consider putting a zoom scale info somewhere, sort of `zoom x 3.2`.

Issue #2389, opened 2017-03-24
Resource view - reconcile inconsistencies between metrics of different topologies


fbarl:
As @2opremio has predicted, having multiple sources of truths for the nodes' metrics between *hosts*, *containers* and *processes* topologies is making some processes in the resource view seem as if they were consuming more resources than their parent containers.

![resources](https://cloud.githubusercontent.com/assets/1216874/24304698/9d3878a8-10ba-11e7-9a4d-1c9436d5a65f.png)

To prevent resource boxes rendering overflows (and overlaps) in such cases, a temporary hack is used that renormalizes the total length of the overflown children to perfectly fit on top of their parent. This is just a visual fix though and it doesn't affect the displayed metric info in any way.

A proper solution would probably require backend to draw all the metrics info from a single source so that:

* the total resource utilization of children could never exceed the parent's
* resource capacities info was (more) consistent between different topologies

rade:
The underlying problem is #1133.

PR #2028, opened 2016-11-24
Update D3 to version 4.4.0

fbarl:
Hopefully nothing should look different from how it is at http://frontend.dev.weave.works/. Most of the changes were just about importing from particular `d3` modules or renaming the methods, but there were also some non-trivial changes:

* Zooming behavior changed a lot so I had to change some logic in `nodes-chart.js`
* I removed the rotation factor from our _hexagon_ and _heptagon_ shapes and also increased their line tension to match the previous looks
* `d3.round` is deprecated so I added our own `round` helper in `math-utils.js`

PR #2420, opened 2017-04-03
Zoom level indicator

fbarl:
Addresses the first suggestion from #2392 - in the resource view, a small horizontal resource scale is displayed from which we can see how deeply we've zoomed. The problem of having zoomed to deeply into a wide box and not seeing any changes when scrolling was brought to me a couple of times.

##### Notes

* ~The resource boxes of zero width are hidden from the resource view now, which fixes the zoom limit being overflown in the CPU view.~ #2499
* We display percentages differently on the zooming scale (from how they are shown on the nodes) - in particular, we use a flexible point precision, with up to 4 digits after the decimal point (as 2 were not precise enough for deep zooms).
* To get a satisfying number rounding policy for zoom scale percentages, I dug up the old `round` helper from #2028 that was removed in the meantime. 

##### ~Additional changes~

* ~Made the displayed metric info reflect the adjustment in resource boxes width. Now the data we display is fully consistent with the layout, although the information is not really always that accurate. Makes #2389 less obvious, but doesn't really fix it.~ #2499
* ~The way we calculate viewport dimensions has been adjusted for consistency between running Scope UI independently as a part of Weave Cloud (previously the margins were a bit off because of the difference in the nav bar height between the two).~ #2472


davkal:
The scale helps a lot, but I found two issues:

The numbers dont add up: 130 + 77 > 160
screen shot 2017-04-07 at 14 32 20

And zooming in on the center makes the upper boxes in the upper band disappear (whereas if I zoom in on the far left, they keep getting displayed)
screen shot 2017-04-07 at 14 35 42
VS (both scales around 128MB
screen shot 2017-04-07 at 14 35 09

client/app/scripts/utils/math-utils.js
```diff
@@ -18,3 +18,10 @@
 export function modulo(i, n) {
   return ((i % n) + n) % n;
 }
+
+// Does the same that the deprecated d3.round was doing.
+// Possibly imprecise: https://github.com/d3/d3/issues/210
+export function round(number, precision = 0) {
```
davkal: What's wrong with `Number.prototype.toFixed(precision)`?

fbarl: The fact that it outputs a string with a fixed precision :) What we want to do here is to not show the trailing zeros, treating the `precision` as maximal precision instead of fixed. So for example, `2.34.toFixed(4) === "2.3400"` while `formatFlexiblePrecision(2.34) === "2.34"`. On the other hand `11.01234.toFixed(4) === formatFlexiblePrecision(11.01234) === "11.0123"`. It seems at first that this could be achieved by `Number.prototype.toPrecision` though, but that one's not good either as it starts rounding up more than just the decimal part when the values get very big.

fbarl:
@davkal PTAL. I draw only a simple small scale in the bottom-right corner now, according to your feedback and our discussion that followed (I also tried it closer to the resource boxes, but didn't look that good to me).

> The numbers not adding up is a blocker, because it diminishes trust.

I adjusted the displayed metrics info (last commit) so hopefully the trust will not be diminished now :)

davkal:
LGTM. This can be merged as is.

I still find the two axis confusing (eg., expressing used memory both in x and y dimensions). Also, when zooming into the center part, the boxes to the left still disappear and you have no way of knowing that they are there. I am aware that other flame graphs suffer the same problems, but maybe indicators could help: either a thumbnail overview with a rectangle representing the visible part, or thin-lined boxes to the left that afford a panning action to get them into view.

fbarl:
Thanks for reviewing @davkal.

I agree with both of your points, let me try to summarize my thoughts on them:

> I still find the two axis confusing (eg., expressing used memory both in x and y dimensions).

This currently only affects the *Hosts* layer, but in the future it might affect any layer with the notion of resource capacity, so it'd be good to decide on the final looks sooner rather than later. Here are some ideas that I had:

1. Drop the notion of capacity and show only the resource utilization for every node.
    * *PROS*: full design consistency between the layers, simpler and more intuitive layout
    * *CONS*: the capacity information actually seems more important to show for *hosts* than the resource consumption as the latter can be roughly approximated by the stack of containers above it
2. Make the above the default and then have a sort of check-box *Show capacities* that would switch to the view as it is now.
    * *PROS*: serves the more intuitive layout first, doesn't remove any information from the resource view
    * *CONS*: your comment would still apply to the layout with capacities, e.g. if someone lands there first
3. Make *hosts* resource boxes look the same as *containers*/*processes* ones, but let their width stand for capacity instead of resource utilization.
    * *PROS*: showing arguably the most useful information while keeping the design consistent
    * *CONS*: inconsistent semantics between the layers, hosts would feel *empty*

*Note:* Show the bar for *hosts* horizontally instead of vertically maybe sounds like a good idea but it actually looked quite messy when I tried it.

I don't really have any other ideas that wouldn't require major changes to the layout. However, I got this comment more than once so I quickly prototyped another layout idea in another experimental branch. Looks more or less like this:

new-design-proposal

So basically all the layers are collapsed into one and the nodes are *embedded* into their parents instead of being shown on top of them. The trigger for me was thinking how point 3. above leaves *hosts* empty, so I thought of filling them with some information.

> Also, when zooming into the center part, the boxes to the left still disappear and you have no way of knowing that they are there. I am aware that other flame graphs suffer the same problems, but maybe indicators could help: either a thumbnail overview with a rectangle representing the visible part, or thin-lined boxes to the left that afford a panning action to get them into view.

1. Thin-lined boxes on the sides would break the existing zooming isometry and would not be consistent with the zooming scale info, so I'm not a huge fan of that idea (unless we modify this PR to show only the zooming factor instead of displaying the scale).
2. Some sort of pseudo rectangle that gives hint about what is outside of the viewport sounds better to me and is definitely something I could implement in another story.
3. Alternatively, a trick with very little impact that could hint there is some content behind the screen would be to extend the text info background for the resource box. Compare the two pictures below:

text-0
text-1

---

In conclusion, I'm not sure whether it makes sense to actually merge this PR, especially if we are going to do some changes to the current layout, so I'd freeze it a bit before I get more feedback on what I wrote above I think.

fbarl:
Dropped in favor of a more generic zoom control/indicator #2513.

PR #2423, opened 2017-04-04
Fix the click origin of the details panel in the table mode

### LINKED ISSUE #1754: Table-mode: origin of details panel is not where clicked
In a grid row, click on a middle column, observe how the details panel is created over the first column? Would be great to have it originate where clicked to not confuse locality.

fbarl:
Fixes #1754.

The click event target is now used instead of the row DOM element.


foot:
Nicely done, LGTM

PR #2435, opened 2017-04-10
Fix disabling Scope in the ECS AMI

2opremio:


peterdeka:
i think that's `rm /etc/init/scope.conf`

ekimekim:
concept lgtm, double-check the path as per @peterdeka

PR #2091, opened 2016-12-13
ecs: Make service a parent of task

ekimekim:
Fixes https://github.com/weaveworks/scope/issues/2041

PR #2436, opened 2017-04-10
Reduce the number of places topologies are explicitly listed

ekimekim:
This is a series of changes meant to make it easier to modify the list of topologies
by inferring information or using iterative APIs instead of explicitly referencing each
topology by name, as well as some misc other refactoring. A short list:

* Explicitly list topology fields in report.Report methods less often
* Refactor how information is looked up in render/detailed so that common functionality to all topologies is not repeated
* Refactor Map2Parent, MapContainer2Pod and MapContainer2Task into one function

Much more could be done, but this is a reasonable first pass.

As a side note, I found it very confusing
that the codebase refers to both report topologies (Process, Container, etc) and in-UI topologies (aka views,
eg. container, container by image, container by dns name) as "topologies", in many cases
using the two (different!) "topologyID"s immediately next to each other with no clear distinction.
I have taken to labelling them "topology" vs "apiTopology" in new code.

Note to reviewer: This PR will likely be easiest understood by reading through each commit in order.

### DIFF
--- a/report/report.go
+++ b/report/report.go
@@ -177,49 +177,49 @@ func MakeReport() Report {
 	}
 }
 
+// TopologyMap gets a map from topology names to pointers to the respective topologies
+func (r *Report) TopologyMap() map[string]*Topology {
+	return map[string]*Topology{
+		Endpoint:       &r.Endpoint,
+		Process:        &r.Process,
+		Container:      &r.Container,
+		ContainerImage: &r.ContainerImage,
+		Pod:            &r.Pod,
+		Service:        &r.Service,
+		Deployment:     &r.Deployment,
+		ReplicaSet:     &r.ReplicaSet,
+		Host:           &r.Host,
+		Overlay:        &r.Overlay,
+		ECSTask:        &r.ECSTask,
+		ECSService:     &r.ECSService,
+	}
+}
+
 // Copy returns a value copy of the report.
 func (r Report) Copy() Report {
-	return Report{
-		Endpoint:       r.Endpoint.Copy(),
-		Process:        r.Process.Copy(),
-		Container:      r.Container.Copy(),
-		ContainerImage: r.ContainerImage.Copy(),
-		Host:           r.Host.Copy(),
-		Pod:            r.Pod.Copy(),
-		Service:        r.Service.Copy(),
-		Deployment:     r.Deployment.Copy(),
-		ReplicaSet:     r.ReplicaSet.Copy(),
-		Overlay:        r.Overlay.Copy(),
-		ECSTask:        r.ECSTask.Copy(),
-		ECSService:     r.ECSService.Copy(),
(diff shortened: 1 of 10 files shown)


report/report.go
```diff
@@ -234,37 +234,33 @@ func (r Report) Topologies() []Topology {
...
+	f(&r.Container, &o.Container)
+	f(&r.ContainerImage, &o.ContainerImage)
+	f(&r.Pod, &o.Pod)
+	f(&r.Service, &o.Service)
+	f(&r.Deployment, &o.Deployment)
+	f(&r.ReplicaSet, &o.ReplicaSet)
+	f(&r.Host, &o.Host)
+	f(&r.Overlay, &o.Overlay)
+	f(&r.ECSTask, &o.ECSTask)
+	f(&r.ECSService, &o.ECSService)
```
rade: You could use `TopologyMap` in here, invoking it on both `r` and `o`, and then iterating over one of the maps and fetching from the other. That would avoid repeating the topology list here.

ekimekim: I did that initially - it's still visible in the git history. I changed it back due to perf concerns - merge and copy are *very* common operations and I don't want to add the gc burden of another string map created and discarded each tine.

That may be premature optimization though - ideally we should compare perf with and without https://github.com/weaveworks/scope/commit/9f0f120bc5d0effa4878819629b5d73c9c7f2a9b

rade: ok. if you leave it as is then that deserves a comment.

ekimekim:
I'm rolling https://github.com/weaveworks/scope/pull/2091 into this PR, since changes introduced here eliminate the special case that made things work without https://github.com/weaveworks/scope/pull/2091. Easier than waiting until that change hits master.

ekimekim:
@2opremio since I don't have anyone around who can review this and I need it merged today, I'm just going to merge now. Could you please review when you're back?

report/report.go
```diff
@@ -234,37 +234,27 @@ func (r Report) Topologies() []Topology {
...
-		Service:        r.Service,
-		Deployment:     r.Deployment,
-		ReplicaSet:     r.ReplicaSet,
-		Host:           r.Host,
-		Overlay:        r.Overlay,
-		ECSTask:        r.ECSTask,
-		ECSService:     r.ECSService,
-	}[name]
-	return t, ok
+	if t, ok := r.TopologyMap()[name]; ok {
```
bboreham: adding an `if` here doesn't seem to be necessary

ekimekim: We need to return `*t` in one case, and an empty `Topology{}` in the other, as the value returned by `r.TopologyMap()[name]` is a `*Topology`, not a `Topology`.

ekimekim:
right, apologies for the messiness between commits, I did try to iron those out but must've missed some.

PR #2437, opened 2017-04-12
gzip: change compression level to the default

alban:
We want the middle ground between a small compression size, a fast
compression time and a fast decompression time.

Tests suggest that the default compression level is better than the
maximum compression level: although the reports are 4% bigger and
decompress slower, they compress 33% faster.

See discussion on https://github.com/weaveworks/scope/issues/1457#issuecomment-293288682

### DIFF
--- a/app/router.go
+++ b/app/router.go
@@ -150,7 +150,7 @@ func RegisterReportPostHandler(a Adder, router *mux.Router) {
 		// a.Add(..., buf) assumes buf is gzip'd msgpack
 		if !isMsgpack {
 			buf = bytes.Buffer{}
-			rpt.WriteBinary(&buf, gzip.BestCompression)
+			rpt.WriteBinary(&buf, gzip.DefaultCompression)
 		}
 
 		if err := a.Add(ctx, rpt, buf.Bytes()); err != nil {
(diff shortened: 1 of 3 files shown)


alban:
@bboreham what do you think of this change?

alban:
@ekimekim: PTAL

ekimekim:
Out of interest, did you test any other compression algorithms? For example, lz4 is generally my goto for fast compression/decompression, but I don't know how badly it would increase compressed size.

ekimekim:
(feel free to merge as-is without answering, it was just for my own curiosity)

bboreham:
The problem with moving to a different compression scheme is that we'd have to change both client and server, and maintain backwards-compatibility in the server. 

With gzip, the same server code will decode data encoded at any level of compression.

alban:
> Out of interest, did you test any other compression algorithms? For example, lz4 is generally my goto for fast compression/decompression, but I don't know how badly it would increase compressed size.

I tested other compression algorithms but outside of Scope and I didn't try lz4:
https://github.com/weaveworks/scope/issues/1457#issuecomment-293618444

I used a msgpack generated by Scope though.

> (feel free to merge as-is without answering, it was just for my own curiosity)

@ekimekim I don't have merge rights, so I'll let you do it.

PR #2444, opened 2017-04-14
Add very basic Docker Swarm service topology

ekimekim:
This is done in a very hacky way - we look at the container labels to identify the parent service,
and that's it. We can't see services that aren't running on that machine, and the only service info
we have is the name.

### DIFF
--- a/app/api_topologies_test.go
+++ b/app/api_topologies_test.go
@@ -40,7 +40,7 @@ func TestAPITopology(t *testing.T) {
 	if err := decoder.Decode(&topologies); err != nil {
 		t.Fatalf("JSON parse error: %s", err)
 	}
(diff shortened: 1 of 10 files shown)


ekimekim:
@2opremio since I don't have anyone around who can review this and I need it merged today, I'm just going to merge now. Could you please review when you're back?

app/api_topologies_test.go
```diff
@@ -50,7 +50,7 @@ func TestAPITopology(t *testing.T) {
 		}
 
 		// TODO: add ECS nodes in report fixture
```
2opremio: adjust comment?

2opremio:
post-LGTM

PR #2445, opened 2017-04-14
allow disabling of weaveDNS advertising/lookup

rade:
...while keeping other weave integrations, by setting `--weave.hostname=`

Addresses a suggestion in #1550.

### DIFF
--- a/prog/app.go
+++ b/prog/app.go
@@ -282,8 +282,8 @@ func appMain(flags appFlags) {
 		}
 	})
 
-	// Periodically try and register out IP address in WeaveDNS.
-	if flags.weaveEnabled {
+	// Periodically try and register our IP address in WeaveDNS.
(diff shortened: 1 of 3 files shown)


rade:
There is a change in behaviour here, beyond the explicit intention of this PR,  in what value we attempt to register/lookup in DNS. This is a) is unlikely to be relied on by existing users, and b) hard to eliminate:

| weave.hostname | {app,probe}.weave.hostname | master | branch |
| - | - | - | - |
| unset | unset | scope.weave.local | scope.weave.local |
| unset | = | | scope.weave.local |
| =  | unset | scope.weave.local | |
| unset | =y | y | y |
| =x | unset | x | x |
| =  | = | | |
| =x | = | x | x |
| =  | =y | y | y |
| =x | =y | x | y |

PR #2447, opened 2017-04-17
scope script: Allow 'user' part of image name to be given by DOCKERHUB_USER env var

ekimekim:
This allows alternate scope images to be used, and mirrors the option available for weave net.


rade:
LGTM

PR #2452, opened 2017-04-18
Add docker swarm Stack selector ala k8s namespace selector

ekimekim:
We have to introduce the kinda hacky concept of a 'No Stack' stack
to reconcile it with the idea of a 'default' k8s namespace. This is important
because swarm services without a stack don't have the same docker labels as ones that do.
Curiously, they still have what appears to be a stack name 'prefix' on their names,
but I can't isolate that name anywhere easily so they'll just have to make do.

I basically copy-pasted updateFilters to make this work, todo go back and refactor
to not duplicate 90% of the code.

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -12,6 +12,7 @@ import (
 	"github.com/gorilla/mux"
 	"golang.org/x/net/context"
 
+	"github.com/weaveworks/scope/probe/docker"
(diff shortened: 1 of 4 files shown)


rade:
> I basically copy-pasted updateFilters to make this work, todo go back and refactor
to not duplicate 90% of the code.

Please file a techdebt issue for that.

LGTM.

Issue #1011, opened 2016-02-23
Merge common parts of process walker and connection walker 


2opremio:
See https://github.com/weaveworks/scope/issues/812#issuecomment-187690962

At the very least they are both going though `/proc/PID/fd/*` after #729

schu:
I read over the code and don't see any obvious common parts between the two walkers:

While both do `fs.ReadDirNames(path.Join(w.procRoot, filename, "fd"))`, `pidWalker` already uses the given `CachingWalker` to gather the namespace info (i.e. the data is already available). In `walkNamespace`, `fs.ReadDirNames` is used very specifically to associate sockets with processes.

Am I missing something?

alban:
probe/process/walker_linux.go calls [ReadDirNames](https://golang.org/pkg/os/#File.Readdirnames) but it actually doesn't care about the names and only reports the number of files:
```
  openFiles, err := fs.ReadDirNames(path.Join(w.procRoot, filename, "fd"))
...
  OpenFilesCount: len(openFiles),
```

`ReadDirNames` calls [syscall.Getdents](https://golang.org/pkg/syscall/#Getdents) and [syscall.ParseDirent](https://golang.org/pkg/syscall/#ParseDirent). The parsing and string allocations could be optimized since we don't care about the names.

I don't think that would change the performances much, but we could try.

alban:
With [readdirnames.go](https://gist.github.com/alban/5a6fedd06432c25e253e69f6ffe2c170), I ran 2000 iterations over `/proc/*/fd/`:
- 12s using Golang's ReadDirNames
- 8.6s using my new implementation without parsing the filenames, just counting the entries

The code takes a couple of assumption on procfs:
1. Linux always generates the "." and ".." directories.
2. procfs' getdents does not returns missing files with inode=0

That's a 28% gain on my laptop with 340 processes running. It's better than I thought. I don't know how much it would help Scope (Scope does not do that many iterations over `/proc/*/fd`). And that would be more code to maintain.

/cc @ekimekim

PR #2456, opened 2017-04-21
proc walker: optimize open file counter

alban:
Golang's ReadDirNames is expensive and better avoided when we don't care
about the names.

See discussion in https://github.com/weaveworks/scope/issues/1011

I tested the performances by looking at the Scope cpu usage with `top -d 10`:
- on master versus this branch
- on my laptop with additional processes versus on my laptop with [200 additional test processes with 500 open files each](https://gist.github.com/alban/009763ef2794f5733aa5463e94f8cdba)

| Scope cpu usage | without test processes | with 200 test processes |
| :-----: | :-----: | -----: |
| master branch | 15% | 24% |
| with this PR | 14% | 20% |

TODO:
- [x] vendor https://github.com/weaveworks/common/pull/30 when it's ready

### DIFF
--- a/probe/process/walker_linux.go
+++ b/probe/process/walker_linux.go
@@ -102,7 +102,7 @@ func (w *walker) Walk(f func(Process, Process)) error {
 			continue
 		}
 
-		openFiles, err := fs.ReadDirNames(path.Join(w.procRoot, filename, "fd"))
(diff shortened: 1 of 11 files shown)


alban:
@2opremio I updated the vendoring, but not to HEAD, as described on https://github.com/weaveworks/common/pull/34#issuecomment-298615877

PTAL

Issue #2460, opened 2017-04-24
Track mixpanel events through Service UI


fbarl:
As @davkal mentioned, it would be nice to get some Scope metrics into our Mixpanel. For privacy reasons, we are only interested in tracking the events when Scope is used through Weave Cloud, in which case we could simply re-use the `window.mixpanel` handle set in the Service UI. Two approaches come to my mind:

1. Forward the mixpanel context/helpers to Scope component explicitly, e.g. `<Scope />` -> `<Scope trackMixpanelEvent={trackMixpanelEvent} />` in the Service UI. Good React patterns would then oblige us to forward the mixpanel stuff all the way down to the sub-components where they would be used which seems like a lot of forwarding.
2. Use exactly the same helpers like `trackMixpanelEvent` in the Scope UI and call them from there directly (as the `window.mixpanel` dependency will be available). The duplicated code wouldn't be that big and we could always extract it into the commons later on.

davkal:
I'm fine with copying (and the check for a global `window.mixpanel`). Could yield some surprise events to someone else who integrates Scope as a component into their mixpanel-ed app 😉

fbarl:
@jpellizzari pointed out to me that `process.env.WEAVE_CLOUD` could be used inside Scope to make sure we're inside Weave Cloud when using mixpanel, so everything should be fine with the second approach then!

jpellizzari:
@fbarl The 'namespace' middleware now supports passing in a dependencies object that will be usable inside actions. It can be used to plumb 'services' through to actions; I already added a `registry` service to the dependencies:

https://github.com/weaveworks/service-ui/blob/master/client/src/main.jsx#L31

The registry is used by an action here:
https://github.com/weaveworks/service-ui/blob/master/client/src/actions/index.js#L695

You could create an action within Scope that will have access to that object, like so:

```javascript
export function trackMixpanelEvent(id, ...params) {
  return (dispatch, getState, deps) => {
    if (process.env.WEAVE_CLOUD && deps) {
      deps.mixpanel.someFunctionCall(id, params);
    }
  }
}
```
The mixpanel service would be created in service-ui and passed all the way through to Scope, avoiding the duplicate code.

fbarl:
@jpellizzari Thanks for the tip, this looks very cool, except that I'm not sure if having a state action for handling mixpanel events is the best way. After all, this 'action' wouldn't really be modifying the state and I'm not sure if it will add any overhead to the computation of global state. Will try it out though!

[...]

PR #2462, opened 2017-04-24
Added mixpanel tracking for some basic events

fbarl:
Addresses #2460 by making use of Service UI's configured `window.mixpanel` variable directly.

##### Tracked events

* `scope.layout.refresh.click`
* `scope.layout.selector.click`
* `scope.layout.selector.keypress`
* `scope.metric.selector.pin.click`
* `scope.metric.selector.pin.next.keypress`
* `scope.metric.selector.pin.previous.keypress`
* `scope.metric.selector.unpin.click`
* `scope.metric.selector.unpin.keypress`
* `scope.node.click`
* `scope.node.relative.click`
* `scope.search.query.change`
* `scope.search.query.pin`
* `scope.topology.option.click`
* `scope.topology.selector.click`

Most of them have the active topology and view mode (layout) props attached to them to have some context of where the events are coming from. I checked it on the mixpanel dashboard and it seems to work nicely!

##### Other changes

* Replaced the `hitEnter` action with `pinSearch` which gets called directly from the `search.js`, since we only need to handle the *ENTER* key press from the search bar.
* Disable <kbd>r</kbd> shortcut when the resource view is disabled.

### DIFF
--- a/client/app/scripts/components/search.js
+++ b/client/app/scripts/components/search.js
@@ -3,13 +3,15 @@ import { connect } from 'react-redux';
 import classnames from 'classnames';
 import { debounce } from 'lodash';
 
-import { blurSearch, doSearch, focusSearch, toggleHelp } from '../actions/app-actions';
+import { blurSearch, doSearch, focusSearch, pinSearch, toggleHelp } from '../actions/app-actions';
 import { searchMatchCountByTopologySelector } from '../selectors/search';
 import { isResourceViewModeSelector } from '../selectors/topology';
 import { slugify } from '../utils/string-utils';
+import { parseQuery } from '../utils/search-utils';
 import { isTopologyEmpty } from '../utils/topology-utils';
 import { trackMixpanelEvent } from '../utils/tracking-utils';
 import SearchItem from './search-item';
+import { ENTER_KEY_CODE } from '../constants/key-codes';
 
 
 function shortenHintLabel(text) {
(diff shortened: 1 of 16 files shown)


client/app/scripts/components/search.js
```diff
@@ -71,15 +75,36 @@ class Search extends React.Component {
...
-    this.setState({value});
+    this.setState({ value });
     this.doSearch(inputValue);
   }
 
+  handleKeyUp(ev) {
+    // If the search query is parsable, pin it when ENTER key is hit.
+    if (ev.keyCode === ENTER_KEY_CODE && parseQuery(this.props.searchQuery)) {
+      trackMixpanelEvent('scope.search.query.pin', {
+        query: this.props.searchQuery,
```
foot: User may be searching for service/process names which might be sensitive. While it would be interesting to know the type of query too (metrics vs property vs just a string), I don't think we should send the query itself.

client/app/scripts/components/search.js
```diff
@@ -71,15 +75,36 @@ class Search extends React.Component {
...
+  }
+
   handleFocus() {
     this.props.focusSearch();
   }
 
   doSearch(value) {
+    if (value !== '') {
+      trackMixpanelEvent('scope.search.query.change', {
+        query: value,
```
foot: ditto, don't send user's info

client/app/scripts/components/search.js
fbarl: Good point, I'll remove it!

client/app/scripts/components/search.js
fbarl: Yes, I'll remove it here too!

fbarl:
@foot I addressed your comment and also added some tracking for `keypress` events (related to the already implemented `click` ones). Also, there was a tiny bug there that I fixed - pressing <kbd>r</kbd> when resource view was not available would still pin the first metric if none was pinned :)

All these changes were rather small so I'll merge this PR after the tests pass without your second look at it.

PR #2465, opened 2017-04-25
[PROTOTYPE] Resource view - replace flame graph with a sort of tree map

fbarl:
Here's how my hacked prototype looks like:

Overview:
tree-map-1

Zoomed in:
tree-map-2
tree-map-3
tree-map-4

@pidster @bowenli @jpellizzari @2opremio Before reading the rationale below, I'd be curious to get your fresh (and unbiased) opinion about the screenshots above. Do they seem more/less intuitive and informative to you than the current flame graph layout?

----

Following @davkal's feedback in https://github.com/weaveworks/scope/pull/2420#issuecomment-295706842, the discussion followed how the current resource view could be improved. To sum up, the concerns were:

* Inconsistent display of resource boxes between *hosts* and *containers*/*processes*
* The concern that *hosts* metrics spreading along both axis (horizontal for capacity, vertical for consumption) might at first raise more questions than it answers
* The idea of processes running *inside* containers and containers runinning *inside* hosts could be made visually more explicit
* The current view can look confusingly empty on higher zoom levels:

zoomed-in

Now, simply dropping the vertical bar for *hosts* resource boxes addresses the first two concerns, but it also brings about the question of how not to lose the information about hosts resource consumption. As far as my intuition goes, the sum of all the children containers should give a good approximation of the resource consumption of their parent hosts. That sort of agregated approach also gets rid of current some data discrepancy between the layers. So why not simply put container boxes into their parent hosts (and then do the same for processes)? That immediatelly addresses the third point above and improves on the last point. 

As can be seen from the screenshots on the top, the *HOSTS*, *CONTAINERS*, *PROCESSES* tabs were also removed from the left and their information was appended to the node labels instead.

### DIFF
--- a/client/app/scripts/actions/app-actions.js
+++ b/client/app/scripts/actions/app-actions.js
@@ -22,6 +22,7 @@ import {
 import { getCurrentTopologyUrl } from '../utils/topology-utils';
 import { storageSet } from '../utils/storage-utils';
 import { loadTheme } from '../utils/contrast-utils';
+import { resourceViewAvailableSelector } from '../selectors/resource-view/layout';
 import {
   availableMetricTypesSelector,
   nextPinnedMetricTypeSelector,
@@ -31,7 +32,6 @@ import {
 import {
   activeTopologyOptionsSelector,
   isResourceViewModeSelector,
-  resourceViewAvailableSelector,
 } from '../selectors/topology';
 import {
   GRAPH_VIEW_MODE,
--- a/client/app/scripts/components/nodes-resources/node-resources-layer.js
+++ b/client/app/scripts/components/nodes-resources/node-resources-layer.js
@@ -15,13 +15,13 @@ class NodesResourcesLayer extends React.Component {
     const { layerVerticalPosition, topologyId, transform, layoutNodes } = this.props;
     let height;
     if (topologyId === 'hosts') height = 400;
-    if (topologyId === 'containers') height = 335;
-    if (topologyId === 'processes') height = 270;
+    if (topologyId === 'containers') height = 355;
+    if (topologyId === 'processes') height = 310;
 
     let y;
     if (topologyId === 'hosts') y = -400;
-    if (topologyId === 'containers') y = -345;
-    if (topologyId === 'processes') y = -290;
+    if (topologyId === 'containers') y = -360;
+    if (topologyId === 'processes') y = -320;
 
     return (
       <g className="node-resources-layer">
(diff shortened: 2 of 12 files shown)


pidster:
Knee jerk: 
- less border lines, they're just distracting
- colours are still awful

2opremio:
> Before reading the rationale below, I'd be curious to get your fresh (and unbiased) opinion about the screenshots above. 

Whilst more compact, I it makes the object hierarchy much less clear (or at least less explicit). 

It immediately made me think of disk consumption treemaps though  (http://www.derlien.com/ ) which I guess is a good thing.

Also, the coloring, while consistent with other views, is really confusing. I think that there is little value in showing different colors when there can be nesting (and potentially color clashes between objects of different hierarchy levels).

> Inconsistent display of resource boxes between hosts and containers/processes

We should fix the metrics consistency regardless of how we present them.

> The concern that hosts metrics spreading along both axis (horizontal for capacity, vertical for consumption) might at first raise more questions than it answers

I find it very useful.

> The idea of processes running inside containers and containers runinning inside hosts could be made visually more explicit

I agree

> The current view can look confusingly empty on higher zoom levels

I think that this will hardly happen in practice (for cpu and memory) and even then, I don't see it as a big problem.

jpellizzari:
@fbarl My initial reaction is that its more difficult to read. I prefer the clear labels that the current resource view has.

pidster:
I've had a little time to think about this, and:

> The idea of processes running inside containers and containers runinning inside hosts could be made visually more explicit

This tree-map style certainly does that.
I quite like this, on reflection.

fbarl:
Thanks for the feedback everyone!

@pidster, @2opremio Just to get confusing colors distraction out of the way of our judgement, I made them primarily depend on the layer (orange hosts, green containers and blue processes choice was quite arbitrary), but they still vary a bit between different ranks within the layer (although we might want to remove rank distinction altogether). Some screenshots:

nicer-overview
nicer-zoom-0
nicer-zoom-1

As you can see, borders and colors were generally smoothened following @pidster's feedback.

What I will try to do this week is to get some *unbiased* feedback from the people who haven't seen neither of the two layouts yet. My impression is that we might all be slightly biased towards the current flame graph view as we already got used to it - personally, I started liking this *tree-map* view more and more the more I was looking at it :)

bowenli:
I would urge you to present in the context of the problem you are trying to solve. This helps people evaluate the design. I'm not really convinced that a graph is the right way to approach this.

Pre-biased feedback:

Main questions I have:
1. What the top level data points I want to see? What am I scanning for?
2. What data points am I doing comparisons against? What data points am I NOT comparing?
3. How do I know when to drill down?

42bc47c6-2f58-11e7-881a-3955ac9ca12d

davkal:
Let's get this sorted. I propose an arbitrary deadline of June 9th to come to either have this ready for review or closed.

PR #2471, opened 2017-04-25
Fixed view mode switching keyboard shortcuts

### LINKED ISSUE #2470: Shortcuts T and G throw a Javascript error
Happens because they trigger `toggleGridMode` helper, which is not there anymore and should be removed.

fbarl:
Fixes #2470 and introduces the following keyboard shortcuts:

<kbd>g</kbd> - switch to graph layout
<kbd>t</kbd> - switch to table layout
<kbd>r</kbd> - switch to resource layout

All of them do nothing if we're already in that layout (as toggling doesn't make sense now that there are 3 different modes).

### DIFF
--- a/client/app/scripts/components/app.js
+++ b/client/app/scripts/components/app.js
@@ -12,8 +12,20 @@ import Status from './status';
 import Topologies from './topologies';
 import TopologyOptions from './topology-options';
 import { getApiDetails, getTopologies } from '../utils/web-api-utils';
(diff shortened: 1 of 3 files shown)


fbarl:
> Add the shortcut to the shortcuts help window (<kbd>?</kbd>)

Good catch!

> <kbd>q</kbd> can still be used to unpin a metric in the resource view.

That's been fixed in #2468.

PR #2473, opened 2017-04-25
Calculate viewport dimensions from the scope-app div

### LINKED ISSUE #2472: Viewport height not calculated correctly inside Weave Cloud
This is how the bottom part of Table View looks like in standalone Scope UI:
table-ok

And this is what we get for the same version of Scope UI running inside of Weave Cloud:
table-bad

The scrollbar overflow indicates there is something wrong with how the viewport height is calculated. In fact, since we get it from `window.height`, the margins might be correct only when Scope covers the whole screen. That's not the case with Weave Cloud because of the top navigation bar and it might also not be the case generally when showing Scope in an iFrame.

fbarl:
Fixes #2472 by calculating the viewport dimensions from the `scope-app` div bounding rectangle instead of `window.width` & `window.height` (the changes were pulled out from PR #2420).

### DIFF
--- a/client/app/scripts/components/nodes.js
+++ b/client/app/scripts/components/nodes.js
@@ -39,10 +39,10 @@ class Nodes extends React.Component {
 
     this.setDimensions = this.setDimensions.bind(this);
     this.handleResize = debounce(this.setDimensions, VIEWPORT_RESIZE_DEBOUNCE_INTERVAL);
-    this.setDimensions();
   }
 
   componentDidMount() {
+    this.setDimensions();
     window.addEventListener('resize', this.handleResize);
   }
 
@@ -73,7 +73,15 @@ class Nodes extends React.Component {
   }
 
   setDimensions() {
-    this.props.setViewportDimensions(window.innerWidth, window.innerHeight);
+    // Use the `scope-app` div to determine the viewport rectangle, because `window` gives
+    // wrong values for scope container inside Weave Cloud (on dev & prod), because of the
+    // top navigation toolbar that is counted in, even though it's not part of Scope.
+    const viewport = document.getElementsByClassName('scope-app')[0].getBoundingClientRect();
+    // Not sure why we need to subtract `left` and `top`, but that gives us correct values.
+    this.props.setViewportDimensions(
+      viewport.width - viewport.left,
+      viewport.height - viewport.top
+    );
   }
 }
 
--- a/client/app/scripts/components/nodes.js
+++ b/client/app/scripts/components/nodes.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import { debounce } from 'lodash';
 
 import NodesChart from '../charts/nodes-chart';
 import NodesGrid from '../charts/nodes-grid';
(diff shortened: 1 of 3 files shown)


client/app/scripts/components/nodes.js
```diff
@@ -73,7 +73,15 @@ class Nodes extends React.Component {
   }
 
   setDimensions() {
-    this.props.setViewportDimensions(window.innerWidth, window.innerHeight);
+    // Use the `scope-app` div to determine the viewport rectangle, because `window` gives
+    // wrong values for scope container inside Weave Cloud (on dev & prod), because of the
+    // top navigation toolbar that is counted in, even though it's not part of Scope.
+    const viewport = document.getElementsByClassName('scope-app')[0].getBoundingClientRect();
```
foot: bit icky to reach up the cmp hierarchy to grab `scope-app`, maybe move `setDimensions` up to `App`?

client/app/scripts/components/nodes.js
```diff
@@ -73,7 +73,15 @@ class Nodes extends React.Component {
...
+    // Not sure why we need to subtract `left` and `top`, but that gives us correct values.
```
foot: It seems to be because of the height: 100% in _base.scss. I think we can remove that?

```

.scope-app, .terminal-app {
  -webkit-font-smoothing: antialiased;
  background: $body-background-color;
  bottom: 0;
  color: $text-color;
  font-family: $base-font;
  font-size: 13px;
  height: 100%;
  --- snip ---
```

client/app/scripts/components/nodes.js
fbarl: Absolutely, makes more sense!

client/app/scripts/components/nodes.js
fbarl: Removing that height didn't really help, but it's true that we should set the `height` of the `scope-app` correctly in Weave Cloud, rather than trying to approximate its position from within Scope UI.

Subtracting `left` and `top` like I did above happened to work in Weave Cloud, but wouldn't necessarily give good results when someone would randomly position Scope iFrame within their app.

So a proper solution is to `setViewportDimensions` directly from `width` and `height` here and to correct the `scope-app` height in the `service-ui` to account for the navigation toolbar, like this:

```
.scope-app {
  height: calc(100% - #{$top-toolbar-height});
}
```

I will add this piece of code to Service UI when I include these latest changes to the Scope UI (including this story) so that they are both pushed at once.

fbarl:
@foot PTAL

foot:
I don't think this fixes #2472 anymore. Tested with service-ui@master:

`SERVICE_HOST=frontend.dev.weave.works PROXY_SCOPE_COMPONENT=1 SCOPE_PATH=/Users/simon/weave/scope npm start`

I (think) I can still fix this by fiddling w/ the height though, what problems did you have with it?
 This is the change I made (done inline as scss doesn't work w/ PROXY_SCOPE_COMPONENT for now).

```diff
diff --git a/client/app/scripts/components/app.js b/client/app/scripts/components/app.js
index 6977c0f1..19eddfdc 100644
--- a/client/app/scripts/components/app.js
+++ b/client/app/scripts/components/app.js
@@ -151,7 +151,7 @@ class App extends React.Component {
     const isIframe = window !== window.top;

     return (
-      <div className="scope-app" ref={this.saveAppRef}>
+      <div className="scope-app" style={{height: 'auto'}} ref={this.saveAppRef}>
         {showingDebugToolbar() && <DebugToolbar />}

         {showingHelp && <HelpPanel />}
```

fbarl:
Ups, I forgot to update the description, sorry for that. Basically this PR is only the first step towards fixing it, the second being adding a line or two to `service-ui` stylesheets.

What I suggested in https://github.com/weaveworks/scope/pull/2473#discussion_r113448181 was to add `height: calc(100% - #{$top-toolbar-height})` but your `height: auto` would also work and it's more elegant, so I'd go with that.

The fact remains that it needs to be added in a separate PR, so I think I'll just do it when updating Scope UI (later today).

fbarl:
@foot Mea culpa again, your `height: auto` solution in Scope stylesheets works fine, so nothing needs to be added to `service-ui` stylesheets, this PR alone should resolve #2472!

foot:
lgtm!

PR #2486, opened 2017-04-28
eliminate excessive calls to mtime.Now()

rade:


rade:
> Shouldn't the performance gain be negligible 

I said that much in the comment: "This isn't going to noticeably improve performance"

Issue #2478, opened 2017-04-26
Proof-of-concept service integration


jpellizzari:
_From @jpellizzari on April 26, 2017 18:34_

We spent lots of time getting Scope to run as a component in Weave Cloud in order to enable integrations between Deploy and Monitor. We should do an initial integration to iron out any kinks before moving on to more complex integrations.

We should start with a simple one, like showing if a k8s service has a new image available in the Scope details panel.

_Copied from original issue: weaveworks/service-ui#484_

jpellizzari:
_From @rade on April 26, 2017 20:26_

This got me thinking... we could have a "flux on canvas" - analogous to metrics on canvas - which shows the flux state of services (automated, locked, up-to-date, etc).

re simplest integration: just link from the service/deployment detail panel to the corresponding flux service screen and vice versa.

PR #2487, opened 2017-05-01
Add service images widget

jpellizzari:
Implements #2478, depends on https://github.com/weaveworks/service-ui/pull/495

Displays information about the status of images for a given kubernetes service. This is an initial proof-of-concept to use a reference for doing more complex integrations between services.
screen shot 2017-05-01 at 10 22 32 am
screen shot 2017-05-01 at 10 24 08 am

The link is clickable and will navigate to Flux using the service-ui router that is passed to it via context.

Introduces a `<CloudFeature />` component that changes the context from Scope to service-ui, allowing for components to access any part of the state tree.


jpellizzari:
@fbarl Please review and let me know if you have any questions on how it works. The best way to test is to use `PROXY_SCOPE_COMPONENT=true` in the service-ui repo with this branch checked out.

rade:
> The link is clickable

Where exactly in the flux ui does it go? AFAIK the flux ui only has link destinations for an entire *service*, not for *service + image*.

jpellizzari:
@rade It takes you to the `/flux/:orgId/services/:serviceName`. The hover effect triggers when the entire list body is moused-over, to indicate that you are clicking on a single entity. The list of images are not individually clickable.

rade:
> The list of container images is not individually clickable.

Ah! But you have underlined "authfe" and "logging" items in your example. Do these go to the container detail panels? Or are they not clickable? (which, IMO, would be visually misleading, since every other underlined item is a link). Also, the blue "New image(s) available" text looks clickable since blue is a very commonly used for links. Like here in github :)

jpellizzari:
> But you have underlined "authfe" and "logging" items in your example. Do these go to the container detail panels?

I wrestled with the right way to represent this for a while. We indicate navigation with underlines, but   clicking on either container image will take you to the same place in Flux. The compromise I came up with is to indicate that either is clickable, then on mouse-over indicate that they are indistinguishable.

The screenshot in the first comment is the :hover state of the component. The non-hover state is a white background.

I agree its an inconsistency. We could also go for what I call "the big dumb button" approach to indicate navigation:
screen shot 2017-05-01 at 12 08 04 pm


> Also, the blue "New image(s) available" text looks clickable since blue is a very commonly used for links. Like here in github :)

Those are clickable as part of the body of the list. The blue text is just to differentiate the text from the normal state.

rade:
> We could also go for what I call "the big dumb button"

That might be better - it avoids any confusion as to what is/isn't clickable, and what clicking will actually do. We could perhaps make it smaller, and place it on the same level as the heading (but right-aligned), so as to save precious vertical space and be less disruptive to the overall look and function of the details panel.

jpellizzari:
Updated after feedback:
screen shot 2017-05-01 at 4 31 24 pm

rade:
On which detail panels is this info displayed? I've looked at the code and can't work it out :(

Flux operates on k8s *deployments*, of which services can be seen as a specialisation. So the info should be shown in both.

rade:
> So the info should be shown in both.

We could even additionally show the info on pods and even containers, when these are part of a deployment.

client/app/scripts/reducers/root.js
```diff
@@ -84,6 +84,7 @@ export const initialState = makeMap({
   viewport: makeMap(),
   websocketClosed: false,
   zoomCache: makeMap(),
+  serviceImages: makeMap({ fetching: false })
```
fbarl: 1. This is just an idea, but I think it might be cleaner to keep the whole the *service* state in one `service` namespace vs. having specific fields (like `serviceImages`). That would make a clear cut in the state tree between the internal and Weave Cloud parts as we continue adding more integrations.
2. I think both `fetching` and `errors` states should be used *per-service* instead of globally (if I already fetched the images for the service A and I'm about to fetch them for service B, I don't want the node details of A to switch back to the fetching state).

To sum up, I'd suggest sort of the following structure: `service -> images -> [serviceId] -> { isFetching, errors, data }`. What do you think?

client/app/scripts/reducers/root.js
```diff
@@ -741,6 +742,23 @@ export function rootReducer(state = initialState, action) {
...
+    }
+
+    case ActionTypes.RECEIVE_SERVICE_IMAGES: {
+      state = state.setIn(['serviceImages', 'fetching'], false);
+
+      if (action.errors) {
+        return state.setIn(['serviceImages', 'errors'], action.errors);
+      }
+
+      const service = find(action.services, s => s.ID === action.serviceId);
```
fbarl: In the spirit of keeping the reducers thin and straightforward, I'd just pick out and forward the service object from `getImagesForService` action instead of here. Then (together with the suggestion above), this whole action would simply be:

```
return state.mergeIn(['service', 'images', action.service.ID], {
  isFetching: false,
  errors: action.errors,
  data: action.service.Containers
});
```

Just an idea, up to you to decide!

client/app/scripts/reducers/__tests__/root-test.js
```diff
@@ -678,4 +678,33 @@ describe('RootReducer', () => {
       constructEdgeId('def456', 'abc123')
     ]);
   });
+  it('receives images for a service', () => {
```
fbarl: Nice to have a test for this!

client/app/scripts/components/node-details/node-details-image-status.js
```diff
@@ -0,0 +1,94 @@
...
+class NodeDetailsImageStatus extends React.PureComponent {
+  constructor(props, context) {
+    super(props, context);
+
+    this.handleServiceClick = this.handleServiceClick.bind(this);
+  }
+
+  componentDidMount() {
+    if (this.props.serviceName) {
+      this.props.getImagesForService(this.props.params.orgId, this.props.serviceName);
```
fbarl: It's a bit confusing that what we call `serviceName` here is called `serviceId` in `getImagesForService`, so I'd just decide on one name..

client/app/scripts/components/node-details/node-details-image-status.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  handleServiceClick() {
+    this.props.router.push(`/flux/${this.props.params.orgId}/services/${encodeURIComponent(this.props.serviceName)}`);
```
fbarl: Don't know why this is passing linter, the line is too long, maybe parse it a bit? :)

client/app/scripts/components/node-details/node-details-image-status.js
```diff
@@ -0,0 +1,94 @@
...
+import map from 'lodash/map';
+import { CircularProgress } from 'weaveworks-ui-components';
+
+import { getImagesForService } from '../../actions/app-actions';
+
+const topologyWhitelist = ['services'];
+
+function getNewImages(images, currentId) {
+  // Assume that the current image is always in the list of all available images.
+  // Should be a safe assumption...
```
fbarl: Sounds reasonable to me, but just in case - @foot?

client/app/scripts/components/node-details/node-details-image-status.js
```diff
@@ -0,0 +1,94 @@
...
+                  View in Deploy
+              </a>
+            </div>
+          }
+
+        </div>
+        {fetching && <CircularProgress />}
+        {!fetching && errors && <p>Error: {JSON.stringify(map(errors, 'message'))}</p>}
+        {!errors && !fetching && !service && 'No service images found'}
+        {!errors && !fetching && service &&
```
fbarl: These set of conditions looks like hard to go through (and more importantly, to immidiately see whether they are complementary and complete), maybe use some ternary ` ? : ` or subrendering helpers?

client/app/scripts/components/node-details/node-details-image-status.js
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import find from 'lodash/find';
+import map from 'lodash/map';
+import { CircularProgress } from 'weaveworks-ui-components';
+
+import { getImagesForService } from '../../actions/app-actions';
+
+const topologyWhitelist = ['services'];
```
fbarl: Any specific reason? Minimal impact of the story?

fbarl:
> On which detail panels is this info displayed? I've looked at the code and can't work it out :(

@rade, it should be displayed on the bottom of node details of each node in *Services* topology. The important thing is to test it in combination with https://github.com/weaveworks/service-ui/pull/495 which hasn't been merged yet. I did it by running `SCOPE_PATH=~/Projects/src/github.com/weaveworks/scope SERVICE_HOST=frontend.dev.weave.works PROXY_SCOPE_COMPONENT=true yarn start` from that `service-ui` branch locally (you'd probably want to change `SCOPE_PATH` though).

rade:
>  it should be displayed on the bottom of node details of each node in `Services` topology

That's what I guessed. My point was that it should *at least* also be shown in the Deployment topology.

client/app/scripts/reducers/root.js
jpellizzari: 1. I don't think we will necessarily keep everything Weave Cloud related in a single key. It's possible we will read/update existing things all over the Scope state tree.

2. A user would only ever see the one details panel at a time, and therefore only one loading state is necessary, but I agree that would be more correct; I will fix that.

client/app/scripts/components/node-details/node-details-image-status.js
jpellizzari: This view only makes sense on the 'services' k8s topology currently. Also, it relies on certain metadata being present on the node, which may not be there on other topologies.

It would be possible to also add this for individual containers, but it would involve a more tenuous integration where we somehow look up which service a container belongs to and then look up images from there.

client/app/scripts/reducers/root.js
jpellizzari: Yup that makes sense. That action should do the lookup instead. Will fix.

jpellizzari:
@fbarl I updated this branch with your feedback, PTAL. I also added the image status widget to the 'deployments' view.

client/app/scripts/components/node-details/node-details-image-status.js
```diff
@@ -0,0 +1,122 @@
...
+      <div className="images">
+        {containers.map((container) => {
+          const statusText = getNewImages(container.Available, container.Current.ID).length > 0
+            ? <span className="new-image">New image(s) available</span>
+            : 'Image up to date';
+
+          return (
+            <div key={container.Name} className="wrapper">
+              <div className="node-details-table-node-label">{container.Name}</div>
+              <div className="node-details-table-node-label">{statusText}</div>
```
fbarl: Just a tiny styling comment - I'd rename this to `node-details-table-node-value` and apply `text-align: right` to it (`statusText`), think it looks a bit better that way:

right-styling

Issue #2461, opened 2017-04-24
process walker perfs: optimize readLimits and readStats


alban:
I am testing Scope with a load of 100 processes running [my openfiles program](https://gist.github.com/alban/009763ef2794f5733aa5463e94f8cdba), so that should generate 50.000 opened files.

I get the following pprof data:

![scope-pprof-proc-walker](https://cloud.githubusercontent.com/assets/27575/25340565/7606f75a-2906-11e7-94ab-2c9d811528b9.png)

For the proc walker, that's:
- 1.12s in syscall.Syscall. Hopefully, https://github.com/weaveworks/scope/pull/2456 will help with that
- 0.79s in readLimits (string parsing)
- 0.61s in readStats (string parsing)

There is room for improving the string parsing. For readStats, I wrote an alternative implementation of the parser: https://gist.github.com/alban/7bdf6ced1352d8efbc6c22a64eabe7e3 and I get the following improvements:

```
$ sudo ./read-stats-limits 
time (old): 8.345651064s
time (new): 5.330234374s
```

I initially tried with `fmt.Sscanf` but unfortunately, the performance was bad. My new implementation follows the pattern explained on http://stackoverflow.com/questions/31333353/faster-input-scanning

alban:
The `/proc/<pid>/limits` files are cached (by probe/process/cache.go, with a timeout of 30s) so they are not read from procfs each time. But only the unparsed version is cached, and it is parsed each time in `readLimits()`. Ditto for `/proc/<pid>/cmdline` and `/proc/<pid>/comm`.

We should try to cache the parsed version instead. The cache timeout could be increased too: processes should rarely change their "ulimit" and their "cmdline"/"comm" (if done at all, this is usually done only one time at startup).

The constant `statsTimeout` is declared but unused: the stats are not cached in master.

alban:
I compared:
- Branch master
- My branch https://github.com/kinvolk/scope/commits/alban/perf-proc-walker containing:
  - optimizing ReadDirNames
  - optimizing readStats parsing
  - optimizing readLimits parsing
  - caching parsed values instead of full files

for 4 minutes with the following commands:
```
$ sudo ./scope launch -probe.ebpf.connections --probe.http.listen :4041
$ go tool pprof http://localhost:4041/debug/pprof/profile?seconds=240
web
```

The scenario is:
- 100 processes with 500 file descriptors each (50.000 opened files)
- 7 idle Docker containers (nginx & busybox)
[...]

PR #2491, opened 2017-05-02
process walker perfs: optimize readLimits and readStats

alban:
- optimize readStats
- optimize readLimits
- cache limits and cmdline/name after parsing

Fixes https://github.com/weaveworks/scope/issues/2461


probe/process/walker_linux.go
```diff
@@ -32,53 +61,107 @@ func readStats(path string) (ppid, threads int, jiffies, rss, rssLimit uint64, e
...
-	rssPages, err = strconv.ParseUint(splits[23], 10, 64)
-	if err != nil {
-		return
+	for spaceCount := 0; pos < len(buf) && spaceCount < procStatFieldThreads-procStatFieldSysJiffies; pos++ {
+		if buf[pos] == ' ' {
+			spaceCount++
+		}
+	}
+	for ; pos < len(buf) && buf[pos] != ' '; pos++ {
+		threads = threads*10 + int(buf[pos]-'0')
```
2opremio: This pattern is repeated very often, please factor this out in a function

2opremio:
@alban I don't doubt that the code is much faster now but it's very hard to read. Please try to refactor it. I think that separating the parsing of integers from the space-counting code would already be an improvement.

Also, are you sure that parsing ints manually provides an improvement? I wouldn't be surprised if the Golang library wasn't optimized in this respect.

alban:
The issue with [strconv.Atoi](https://golang.org/pkg/strconv/#Atoi) or [strconv.ParseInt](https://golang.org/pkg/strconv/#ParseInt) is that they don't stop parsing when finding a non-digit such as a space and they return an error.

The previous code was using [strings.Fields](https://golang.org/pkg/strings/#Fields) to allocate and build a string compatible with [strconv.Atoi](https://golang.org/pkg/strconv/#Atoi) (containing only the digits without extra spaces). The pprof graphs showed that it was [strings.Fields](https://golang.org/pkg/strings/#Fields) taking the cpu rather than [strconv.Atoi](https://golang.org/pkg/strconv/#Atoi) but we cannot use Atoi without Fields.

`fmt.Sscanf("%d "...)` is able to work with extra spaces but it's also slower.

I'll try to refactor this.

alban:
Code refactored and rebased.

Now that #2456 is merged in master, I ran performance tests again to see if this PR was still bringing better perfs. I used the test scenario described in https://github.com/weaveworks/scope/issues/2461#issuecomment-297036674

## Master branch

Up to 95660e3da735634d566325a43ab33c4380b8d346 (so #2456 is included)

pprof-branch-master

## This PR

pprof-branch-2491

## Conclusion

The function `github.com/weaveworks/scope/probe/process.(*walker).Walk` takes respectively:
- 15.32s
- 12.91s (16% faster) 

Overall, the graphs' titles say respectively:
- 36.37s
- 31.32s (19% faster overall, probably coincidental)

probe/process/walker_linux.go
```diff
@@ -32,53 +89,86 @@ func readStats(path string) (ppid, threads int, jiffies, rss, rssLimit uint64, e
...
+		return 0, nil
+	}
+	pos += len(delim)
+
+	for pos < len(content) && content[pos] == ' ' {
+		pos++
+	}
+
+	var softLimit uint64
+	for ; pos < len(content) && buf[pos] != ' '; pos++ {
```
2opremio: Can't you use `parseUint64WithSpaces()` here too?

probe/process/walker_linux.go
```diff
@@ -18,12 +20,67 @@ type walker struct {
...
+	for spaceCount := 0; *pos < len(*buf) && spaceCount < nSpaces; *pos++ {
+		if (*buf)[*pos] == ' ' {
+			spaceCount++
+		}
+	}
+}
+
+// parseIntWithSpaces is similar to strconv.ParseInt but stops parsing when
+// reading a space instead of returning an error
+func parseIntWithSpaces(buf *[]byte, pos *int) (ret int) {
```
2opremio: Can't you simply use parseUint64WithSpaces and cast to int instead?

alban: Updated.

probe/process/walker_linux.go
alban: Updated.

PR #2505, opened 2017-05-08
Fix arrow heads positioning

### LINKED ISSUE #2328: arrow head misplaced
screenshot from 2017-03-13 21-00-46

happens in both ff and chrome.

fbarl:
Fixes #2328 in two steps:

* For edges that are not loops, we're now adding two new endpoints instead of overriding the existing ones, so that the edges enter the nodes in a more *tangential* manner as opposed to having a high curvature.
* Also increased the layout engine's perceived `NODE_SIZE_FACTOR` slightly to the same effect.

Before:
arr-prev

Now:
arr-now

### DIFF
--- a/client/app/scripts/charts/edge-container.js
+++ b/client/app/scripts/charts/edge-container.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { Motion, spring } from 'react-motion';
 import { Map as makeMap } from 'immutable';
 import { line, curveBasis } from 'd3-shape';
-import { each, omit, times } from 'lodash';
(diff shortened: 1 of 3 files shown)


foot:
Nice, code looks good. I tested w/ the debugbar and still got a few floating arrow heads.

screen shot 2017-05-11 at 10 56 04

fbarl:
@foot PTAL. I made the following changes that should result in even more correct drawing of edges:

* Came up with a better strategy for selecting/fixing the waypoints.
* Increased the waypoints cap from `8` to `10`.

foot:
Nice! LGTM

PR #2506, opened 2017-05-08
api_topologies: Don't put namespace filters on containers-by-dns/image

ekimekim:
This should never have been added. The IsNamespace filter will always return false for these nodes.
It isn't even clear what this filter would mean, even if implemented, since those nodes don't map to a single
container and therefore not to a single namespace.

This bug was introduced in https://github.com/weaveworks/scope/pull/2285 and is very problematic
because it means the default view on those topologies is blank if the 'default' k8s namespace is present.


rade:
It may be worth releasing a 1.3.1 for this.

2opremio:
> It may be worth releasing a 1.3.1 for this.

Agreed

PR #2507, opened 2017-05-09
ebpf connection tracker: perf map fixes

alban:
This adds fixes in the perf map that receives the tcp events from the ebpf probes:
- gobpf: use the correct timestamp function instead of assuming the timestamp is at offset zero in the struct
- gobpf: fix loop on events that could drop events
- gobpf: add a go channel to notify lost events
- gobpf: make the size of the ring buffers configurable instead of using 8 page (32KB)
- tcptracer-bpf: change the size of the ring buffer "maps/tcp_event_ipv4" to 256 pages (1MB)
- tcptracer-bpf: use the go channel to notify lost events
- scope: add a callback on lost events: log the error and stop the ebpf tracker correctly. So if it ever happens, we will know about it.

This is to address the address the following failure that happened a few weeks ago:
```
<probe> ERRO: 2017/04/13 07:34:17.299127 tcp tracer received event with timestamp 677884430530033 even though the last timestamp was 677884430530738. Stopping the eBPF tracker.
<probe> WARN: 2017/04/13 07:34:17.337241 ebpf tracker died, gently falling back to proc scanning
```

-----

This is a "work in progress" because this is vendoring unmerged branches in gobpf and tcptracer-bpf.
- [x] https://github.com/iovisor/gobpf/pull/43
- [x] https://github.com/iovisor/gobpf/pull/44
- [x] https://github.com/weaveworks/tcptracer-bpf/pull/37

-----

I tried to reduce the size of the perf ring buffers to one page (4KB) in order to test it. I made lots of connections in parallel in 4 terminals (but on the same cpu to stress one perf ring buffer):

```
for i in $(seq 1 10000) ; do echo -n "$i " ; taskset --cpu-list 2 wget -O /dev/null http://172.17.0.2 2>/dev/null; done
```

Then, I checked that the fallback worked correctly:
```
<probe> ERRO: 2017/05/09 09:17:40.798823 tcp tracer lost 7 events. Stopping the eBPF tracker
<probe> WARN: 2017/05/09 09:17:41.701317 ebpf tracker died, gently falling back to proc scanning
```

-----

/cc @iaguis @2opremio

### DIFF
--- a/probe/endpoint/ebpf.go
+++ b/probe/endpoint/ebpf.go
@@ -82,7 +82,7 @@ func newEbpfTracker() (eventTracker, error) {
 		return nil, fmt.Errorf("kernel not supported: %v", err)
 	}
 
-	t, err := tracer.NewTracer(tcpEventCbV4, tcpEventCbV6)
+	t, err := tracer.NewTracer(tcpEventCbV4, tcpEventCbV6, lostCb)
 	if err != nil {
 		return nil, err
 	}
@@ -117,6 +117,12 @@ func tcpEventCbV4(e tracer.TcpV4) {
 
 func tcpEventCbV6(e tracer.TcpV6) {
 	// TODO: IPv6 not supported in Scope
+}
+
+func lostCb(count uint64) {
+	log.Errorf("tcp tracer lost %d events. Stopping the eBPF tracker", count)
+	ebpfTracker.dead = true
+	ebpfTracker.stop()
 }
 
 func (t *EbpfTracker) handleConnection(ev tracer.EventType, tuple fourTuple, pid int, networkNamespace string) {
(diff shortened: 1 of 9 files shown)


alban:
I rebased and revendored gobpf + tcptracer-bpf since PRs have been merged.

I'm keeping the "WIP" title because I'd like to test again to make sure nothing was wrong in the rebase/revendor/git-subtrees changes.

alban:
I tried to quickly create plenty of connections to nginx on several cpus with this script (in several terminals):
```
cpu=0
for i in $(seq 1 10000) ; do
  echo -n "$i "
  taskset --cpu-list $cpu wget -O /dev/null http://172.17.0.2 2>/dev/null
done
echo
```
And I didn't notice any problems in the logs.

However, chrome quickly took all the cpu and crashed:

chrome-crash

2opremio:
Mind creating a separate ticket for this?

2opremio:
(It's not a real scenario, but it's worth keeping track of it)

alban:
> Mind creating a separate ticket for this?

https://github.com/weaveworks/scope/issues/2517

alban:
@2opremio PTAL

PR #2513, opened 2017-05-10
Canvas zoom control

### LINKED ISSUE #2321: zoom control not obvious
@lukemarsden watched a user struggling to work out how to zoom in scope. They used the browser zoom, which only zooms everything *except* the graph. And they didn't realise that mousewheel zoomed at all.

I wonder whether a) browser zoom should zoom everything, and) we should add some google maps style zoom control in the bottom right corner.

fbarl:
Resolves #2321 and makes #2420 redundant.

##### Changes

* Added zoom control in the bottom right corner for graph and resource views.
* Zoom is disabled (and the control is hidden) when there is no content displayed.
* The zoom control is automatically managed within `ZoomableCanvas`.

##### Note

* I had some problems running this with local `service-ui` proxying, as `~rc-slider/dist/rc-slider.css` wouldn't be found from in Service UI. Running `npm install rc-slider` inside the `service-ui/client` folder did it for me, but I wonder why this dependency wasn't pulled from the local Scope's `package.json`. Maybe it has to do with Webpack 2 vs. Webpack 1 (@foot had a similar issue with `xterm`). @jpellizzari I tried fixing this but couldn't figure it out - do you have any ideas what might be causing the issue?


jpellizzari:
@fbarl Re: CSS not being proxied: you stumbled on one of the limitations of our LocalModuleProxy. It only works if all dependencies are present in the service-ui. I haven't checked to see if webpack 2 exposes any new APIs that we can use to make that a nicer integration.

PR #2518, opened 2017-05-11
handle fdinstall events from tcptracer-bpf (aka "accept before kretprobe" issue)

alban:
During the EbpfTracker initialization, find all the processes that are currently blocked on an `accept()` syscall. `feedInitialConnections()` will use `t.tracer.AddFdInstallWatcher()` to subscribe to [fd_install](https://github.com/torvalds/linux/blob/v4.11/fs/file.c#L625-L630) events. When a fd_install event is received, synthesise an `accept` event with the connection tuple and the network namespace (from `/proc`).

TODO:
- [x] re-vendor tcptracer-bpf (when https://github.com/weaveworks/tcptracer-bpf/pull/39 is merged)
- [x] tcptracer-bpf event struct should have a field for `fd` instead of reusing `netns`
- [x] handle multi-threaded programs that run blocking `accept()` syscalls (look in `/proc/*/tasks/*/wchan` instead of `/proc/*/wchan`)
- [x] find the network namespace (currently hard-coded)
- [x] find the connection tuple (currently hard-coded).
- [x] handle IPv4-mapped IPv6 addresses
- [x] fix lint / tests
- [x] integration test

This is the Scope part of https://github.com/weaveworks/tcptracer-bpf/pull/39
Issue: https://github.com/weaveworks/tcptracer-bpf/issues/10


probe/process/walker_linux.go
```diff
@@ -166,6 +170,16 @@ func (w *walker) readCmdline(filename string) (cmdline, name string) {
 	return
 }
 
+func IsProcInAccept(procRoot, filename string) (ret bool) {
```
2opremio: s/filename/pid/

2opremio:
I thought you were going to extend the proc connection walker to also obtain TCP sockets in LISTEN state. I am a bit concerned of the performance impact of checking `pid/wchan`.

2opremio:
> handle multi-threaded programs that run blocking accept() syscalls (look in /proc/*/tasks/*/wchan instead of /proc/*/wchan)

Won't /proc/pid/wchan also include the content of the tasks?

alban:
> I thought you were going to extend the proc connection walker to also obtain TCP sockets in LISTEN state.

That was what I wanted to do initially but then I realised that the set of processes that have sockets in the LISTEN state is bigger than the set of process blocked on the `accept` syscall. On my laptop, I have only one process blocked on the `accept` syscall (the test program `busybox nc -l 192.168.35.216 -p  8181`). But I have more listening sockets (9 just for the initial network namespace `sudo netstat -anpel|grep tcp.*LISTEN|wc -l`)

>  I am a bit concerned of the performance impact of checking `pid/wchan`.

This is done only one time during the initialization of the EbpfTracker when Scope starts.

> You get this (network namespace) for free out of the proc connection walker.

This part is not during the initialization but when we receive the fd_install event from tcptracer-bpf. Other events (connect, accept, close) includes the network namespace because the kprobe for them is on a kernel function with the `struct sock`. But we don't get the network namespace from the fd_install event because we don't have the `struct sock`, only the `struct file`.

While it should be possible to get the `struct sock` from the `struct file` in general, we don't have any "guess offset" algorithm to find out where to look in the `struct file`. I thought it was not worth adding complexity in the "guess offset" algorithm because we can get the network namespace from userspace and we need to do that only one time per process that were blocked on an `accept` syscall at the time of Scope startup.

> Won't /proc/pid/wchan also include the content of the tasks?

No, it seems to only report the function of the main thread. For example:
```
$ for i in  /proc/29241/task/*/wchan  ; do cat $i ; echo ; done
futex_wait_queue_me
futex_wait_queue_me
futex_wait_queue_me
futex_wait_queue_me
futex_wait_queue_me
futex_wait_queue_me
ep_poll
ep_poll
futex_wait_queue_me
futex_wait_queue_me
futex_wait_queue_me
futex_wait_queue_me
futex_wait_queue_me
futex_wait_queue_me
futex_wait_queue_me
futex_wait_queue_me
$ cat /proc/29241/wchan ; echo
futex_wait_queue_me
```

2opremio:
> That was what I wanted to do initially but then I realised that the set of processes that have sockets in the LISTEN state is bigger than the set of process blocked on the accept syscall. On my laptop, I have only one process blocked on the accept syscall (the test program busybox nc -l 192.168.35.216 -p 8181). But I have more listening sockets (9 just for the initial network namespace sudo netstat -anpel|grep tcp.*LISTEN|wc -l)

Would this impact performance or also accuracy? (If the impact on performance is just a map lookup with a larger map then I guess it's OK)


> This is done only one time during the initialization of the EbpfTracker when Scope starts.

Good point.

alban:
> Would this impact performance or also accuracy? (If the impact on performance is just a map lookup with a larger map then I guess it's OK)

Only performances, and the performance with the `wchan` files should be better since we would receive less false positive "fd_install" events.

2opremio:
> Only performances, and the performance with the wchan files should be better since we would receive less false positive "fd_install" events.

But the false positives won't impact performance so much since it's a map, which has a fast access.

On the other hand, accessing wchan seems to complicate the code.

2opremio:
I would also like to see some tests for this.

vendor/manifest
```diff
@@ -1460,10 +1460,10 @@
 		},
 		{
 			"importpath": "github.com/weaveworks/tcptracer-bpf",
-			"repository": "https://github.com/weaveworks/tcptracer-bpf",
+			"repository": "https://github.com/kinvolk/tcptracer-bpf",
 			"vcs": "git",
-			"revision": "b715a3b635b8d9c4a096bbd6009826b57fe64c38",
-			"branch": "master",
+			"revision": "3b09ef1351d865c0b2519e1486f6c88db3905780",
+			"branch": "alban/fdinstall",
```
alban: The vendoring was generated by updating the "repository", "revision" and "branch" fields and running the command:
```
gvt restore --filter=github.com/weaveworks/tcptracer-bpf
```
using `gvt` patched with https://github.com/FiloSottile/gvt/pull/93

alban:
I added an integration test (314_container_accept_before_kretprobe_test.sh) and it passes now.

2opremio:
> But the false positives won't impact performance so much since it's a map, which has a fast access.
>
> On the other hand, accessing wchan seems to complicate the code.

@alban Are you finally discarding the option of obtaining the sockets in LISTEN state from the proc walker?

alban:
> But the false positives won't impact performance so much since it's a map, which has a fast access.

> @alban Are you finally discarding the option of obtaining the sockets in LISTEN state from the proc walker?

Yes

The process of getting an otherwise-missed accept event should happen only one time for processes that were blocked in the accept() syscall. Once the accept() syscall returns, we don't need that mechanism with fd_install events anymore. So we need a way to know when we can stop the monitoring of fd_install events.

If we monitor the fd_install events for all processes that own a socket in the LISTEN state, we don't know when to stop that monitoring because servers normally keep a socket in the LISTEN state even after a connection has been established (for the purpose of accepting future connections).

So that would be receiving an fd_install event for all servers, during their lifetime. We also receive an install event for file descriptors that are not a socket because the filtering on the file descriptor kind is done in userspace. So using wchan, we minimize that.

2opremio:
Fair enough

integration/314_container_accept_before_kretprobe_test.sh
```diff
@@ -0,0 +1,37 @@
+#! /bin/bash
+
+# shellcheck disable=SC1091
+. ./config.sh
+
+start_suite "Test accept before kretprobe, see https://github.com/weaveworks/tcptracer-bpf/issues/10"
+
+weave_on "$HOST1" launch
+
+# Launch the server before Scope
```
2opremio: Please elaborate why (e.g. _to make sure it calls accept() and the socket is in LISTEN mode before the eBPF kprobes are installed_).

integration/314_container_accept_before_kretprobe_test.sh
```diff
@@ -0,0 +1,37 @@
...
+	done | nc server.weave.local 8080"
+
+wait_for_containers "$HOST1" 60 server client
+
+has_container "$HOST1" client
+
+list_containers "$HOST1"
+list_connections "$HOST1"
+
+has_connection containers "$HOST1" client server
```
2opremio: Parallel remark: I am having my doubts that this does what it should, see https://github.com/weaveworks/scope/pull/2527

integration/314_container_accept_before_kretprobe_test.sh
alban: Updated:
```
+# Launch the server before Scope to make sure it calls accept() before Scope's
+# kretprobe on the accept function is installed. We use busybox' nc instead of
+# Alpine's nc so that it blocks on the accept() syscall.
```

integration/config.sh
```diff
@@ -43,14 +43,15 @@ scope_end_suite() {
 list_containers() {
     local host=$1
     echo "Listing containers on ${host}:"
-    curl -s "http://${host}:4040/api/topology/containers?system=show" | jq -r '.nodes[] | select(has("metadata")) | .metadata[] | select(.id == "docker_image_name") | .value'
+    curl -s "http://${host}:4040/api/topology/containers?system=show" | jq -r '.nodes[] | select(has("metadata")) | { "image": .metadata[] | select(.id == "docker_image_name") | .value, "label": .label, "id": .id} | .id + " (" + .image + ", " + .label + ")"'
```
2opremio: This is starting to get out of hand ... but I don't have a better suggestion

probe/endpoint/ebpf.go
```diff
@@ -125,6 +133,99 @@ func lostCb(count uint64) {
...
 
+func tupleFromPidFd(pid int, fd int) (tuple fourTuple, netns string, ok bool) {
+	// read /proc/$pid/ns/net
+	//
+	// probe/endpoint/procspy/proc_linux.go supports Linux < 3.8 but we
+	// don't need that here since ebpf-enabled kernels will be > 3.8
+	netNamespacePath := fmt.Sprintf("/proc/%d/ns/net", pid)
+	var statNsFile syscall.Stat_t
+	err := fs.Stat(netNamespacePath, &statNsFile)
+	if err != nil {
```
2opremio:
```
if err := ... ; err != nil {
  ...
}
```

probe/endpoint/ebpf.go
```diff
@@ -125,6 +133,99 @@ func lostCb(count uint64) {
...
+		log.Debugf("proc file %q disappeared before we could read it", netNamespacePath)
+		return fourTuple{}, "", false
+	}
+	netns = fmt.Sprintf("%d", statNsFile.Ino)
+
+	// find /proc/$pid/fd/$fd's ino
+	fdFilename := fmt.Sprintf("/proc/%d/fd/%d", pid, fd)
+	var statFdFile syscall.Stat_t
+	err = fs.Stat(fdFilename, &statFdFile)
+	if err != nil {
```
2opremio: same here

probe/endpoint/ebpf.go
```diff
@@ -125,6 +133,99 @@ func lostCb(count uint64) {
 	ebpfTracker.stop()
 }
 
+func tupleFromPidFd(pid int, fd int) (tuple fourTuple, netns string, ok bool) {
```
2opremio: Can you factor out the common parts for this function with the proc walker?
* Reading net NS
* Reading tcp and tcp6 files

2opremio: There is a common package for this sort of stuff

integration/314_container_accept_before_kretprobe_test.sh
alban: Is it worth adding a second test, doing the same but without Weave Net to also test this without NAT?

2opremio: Nah, I think it's OK with NAT.

We want to have tests with and without NAT for ebpf in general, not specifically for accept() (if we don't have them already), but that should be done in a separate PR.

probe/endpoint/ebpf.go
alban: done

probe/endpoint/ebpf.go
alban: done

probe/endpoint/ebpf.go
alban: done

alban:
I rebased, re-vendored tcptracer-bpf and factorized the 2 /proc readings.

I also pre-fetch the busybox image in the integration tests on GCE since we have a test using busybox now, following the same pattern as the other images.

Let's see if the tests pass on CircleCI.

2opremio:
LGTM (on green)

alban:
green

PR #2522, opened 2017-05-16
show all namespaces by default

rade:
There is nothing special about the "default" namespace, and selecting it by default means we only show a partial view of the system, which is hard to spot, especially for novice users.

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -50,13 +50,9 @@ var (
 )
 
 // namespaceFilters generates a namespace selector option group based on the given namespaces
-func namespaceFilters(namespaces []string, defaultNamespace, noneLabel string) APITopologyOptionGroup {
-	options := APITopologyOptionGroup{ID: "namespace", Default: "", SelectType: "union", NoneLabel: noneLabel}
+func namespaceFilters(namespaces []string, noneLabel string) APITopologyOptionGroup {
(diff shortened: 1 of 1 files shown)


rade:
I have tested this against a report from dev, so with some k8s in it, but not docker swarm. If somebody could check that please, that would be great.

app/api_topologies.go
```diff
@@ -50,13 +50,9 @@ var (
 )
 
 // namespaceFilters generates a namespace selector option group based on the given namespaces
-func namespaceFilters(namespaces []string, defaultNamespace, noneLabel string) APITopologyOptionGroup {
-	options := APITopologyOptionGroup{ID: "namespace", Default: "", SelectType: "union", NoneLabel: noneLabel}
+func namespaceFilters(namespaces []string, noneLabel string) APITopologyOptionGroup {
+	options := APITopologyOptionGroup{ID: "namespace", Default: "none", SelectType: "union", NoneLabel: noneLabel}
```
rade: The `"none"` here took me a while to figure out; it is a dependency on a constant in the UI code.

I initially had `""` and then `noneLabel`, neither of which worked.

ekimekim: That is a bug in the UI code then. `""` should be correct.

ekimekim: This means that the UI is currently bugged in cases where there's no default namespace?

Issue #1444, opened 2016-05-06
Add view for DaemonSets


2opremio:
Just like we have for PODs, Services, ReplicaSets and Deployments.

rade:
We are now running quite a few DaemonSets in Weave Cloud, which are currently invisible.

Related, I do wonder whether we need a hybrid view that shows everything at the "highest level of abstractions", i.e. show both Services and DaemonSets, plus individual pods that do not belong to either.

rade:
I think we should have *one* view for deployments and daemonsets. AFAIK a pod can never be part of more than one of these. Call that view 'replicas', perhaps. The different types should of course be represented by visually distinct shapes.

The idea here is that we want to look at our system at a level of abstraction that collapses replicas, w/o having to care about / choose / filter-by the specific kind of abstraction that gave rise to the replicas.

I think ultimately we should include replicasets in this view as well, *if they are not part of a deployment*. But we can add that later; this would go in the direction of some of the hybrid view ideas from #552.

ekimekim:
I concur. We can preserve the existing behaviour by adding a multi-select
option group for deployments/daemonsets/'bare replicasets'/'bare pods',
with the default being 'All Types'

rade:
> We can preserve the existing behaviour by adding a multi-select

TooManyKnobsException. YAGNI.

Issue #552, opened 2015-10-10
Consider a hybrid view showing containers & processes


tomwilkie:
Lots* of people seem to run mixed containerise and non-containerise workloads, for instance we've seen at least three people run their load balancers outside of containers.

Right now it will show up as uncontained, but I've been think for a while a smarter hybrid view will be useful (ie show processes if they are not in a container, containers if they are. And visually differentiate.)

Thoughts?

pidster:
I think a smart / hybrid view is a really good idea.

rade:
> visually differentiate

Shapes have taken care of that already.

+1 to the overall idea. In fact I wonder whether a container-only view is even necessary, i.e. perhaps we should simply replace the current container view with the hybrid view.

2opremio:
Here's another user needing the hybrid view: https://weaveworks.slack.com/archives/scope-public/p1459362185000321

<img width="1109" alt="screen shot 2016-03-30 at 11 32 04 am" src="https://cloud.githubusercontent.com/assets/2362916/14153150/7d91e4c2-f66b-11e5-8e36-0fc62b44ae38.png">

2opremio:
Report from the user: https://weaveworks.slack.com/files/guang/F0WL9MS86/api_report.json.gz

gy8:
+1

rade:
This doesn't just apply to processes and containers, but any topology.

We could go for a more radical solution: *replace* the individual topology views with a single hybrid view, and combine that with hierarchical navigation #719.

Such view would show you everything in existence, at the highest level of abstraction in which it has a representation, and the hierarchical navigation would allow you to "zoom in" and see a *fragment* of the space at a lower level. E.g. if we have just three processes, one of which is in a container and the others are in a container each, in a pod each, in the same service, then we'd show a container and a service, and hierarchical nav would allow the user to "expand" the service.

IMO this would make it easier for users to "get their bearings" and retain context while exploring their system.

There are a few topologies that do not fit into the above...
- topologies that are at the same level of abstraction but can share nodes from lower levels. IIRC this is true of some k8s topos, e.g. the same pod could be part of a service  I guess we could just show all of them. See also https://github.com/weaveworks/scope/issues/1444#issuecomment-249333960
[...]

PR #2526, opened 2017-05-16
Add daemonsets

ekimekim:
Fixes https://github.com/weaveworks/scope/issues/1444

This adds collection of daemonset info to the probe,
and two new views to the app. One showing only daemonsets,
one showing deployments + daemonsets.

This is too many views (see discussion in #1444), but was useful for initial debugging - we can remove the daemonsets view before merging.

We need to decide how we are going to present this new 'combined view' to the user in a way that makes it obvious what's going on. For now I've labelled it 'Replica groups' but that's not good enough.

### DIFF
--- a/probe/kubernetes/daemonset.go
+++ b/probe/kubernetes/daemonset.go
@@ -18,7 +18,7 @@ const (
 // DaemonSet represents a Kubernetes daemonset
 type DaemonSet interface {
 	Meta
-	Selector() labels.Selector
+	Selector() (labels.Selector, error)
 	GetNode() report.Node
 }
 
@@ -35,13 +35,12 @@ func NewDaemonSet(d *extensions.DaemonSet) DaemonSet {
 	}
 }
 
-func (d *daemonSet) Selector() labels.Selector {
+func (d *daemonSet) Selector() (labels.Selector, error) {
 	selector, err := unversioned.LabelSelectorAsSelector(d.Spec.Selector)
 	if err != nil {
-		// TODO(paulbellamy): Remove the panic!
-		panic(err)
+		return nil, err
 	}
-	return selector
+	return selector, nil
 }
 
 func (d *daemonSet) GetNode() report.Node {
--- a/probe/kubernetes/daemonset.go
+++ b/probe/kubernetes/daemonset.go
@@ -1,0 +1,53 @@
+package kubernetes
+
+import (
+	"fmt"
+
+	"k8s.io/kubernetes/pkg/api/unversioned"
+	"k8s.io/kubernetes/pkg/apis/extensions"
+	"k8s.io/kubernetes/pkg/labels"
+
+	"github.com/weaveworks/scope/report"
+)
+
+// These constants are keys used in node metadata
+const (
+	MisscheduledReplicas = "kubernetes_misscheduled_replicas"
+)
+
+// DaemonSet represents a Kubernetes daemonset
+type DaemonSet interface {
+	Meta
+	Selector() labels.Selector
+	GetNode() report.Node
+}
+
+type daemonSet struct {
+	*extensions.DaemonSet
+	Meta
+}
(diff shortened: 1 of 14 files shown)


probe/kubernetes/daemonset.go
```diff
@@ -0,0 +1,53 @@
...
+		DaemonSet: d,
+		Meta:      meta{d.ObjectMeta},
+	}
+}
+
+func (d *daemonSet) Selector() labels.Selector {
+	selector, err := unversioned.LabelSelectorAsSelector(d.Spec.Selector)
+	if err != nil {
+		// TODO(paulbellamy): Remove the panic!
+		panic(err)
```
2opremio: Can you propagate the error instead of the TODO? :)

render/pod.go
```diff
@@ -96,6 +96,28 @@ var ReplicaSetRenderer = ConditionalRenderer(renderKubernetesTopologies,
...
+			),
+			SelectDaemonSet,
+		),
+	),
+)
+
+// ReplicaGroupRenderer is a Renderer which combines the daemonset and deployment views
+// in (for now) a very naive way.
+var ReplicaGroupRenderer = MakeReduce(
+	DeploymentRenderer,
```
2opremio: I am not sure I get this. The ReplicaGroup view will also include DaemonSets? Don't we have separate view for that?

That's confusing ...

render/pod.go
```diff
@@ -96,6 +96,28 @@ var ReplicaSetRenderer = ConditionalRenderer(renderKubernetesTopologies,
...
+	DaemonSetRenderer,
```
2opremio: So, the ReplicaGroup view will also include DaemonSets? Why have a separate view for DaemonSets alone then?

Also, I am not sure I like the name (ReplicaGroup) or the entities it groups. 

I think it would make sense to group controllers together (that is, ReplicaSets, ReplicationControllers, DaemonSets, PetSets) in a view and name it  "Controllers". However, Deployments live one level of abstraction higher from controllers (they are on top of ReplicaSets) which also makes the name confusing.

2opremio:
Done with the review: Minor code comment and a fundamental design question.

I don't mind going ahead with this a refining it later, but as a kubernetes user I find it a bit odd.

ekimekim:
(addressing design question at the top level instead of at the specific comment)

Yeah, I agree. The view is confusing and right now badly named. `Controllers` is a better name (though I'm unsure if it's a good name either). I added the daemonset view itself because it was trivial to do and let me ensure it was working correctly.

The `ReplicaGroup` view was an attempt at addressing concerns in #1444, and yeah I'm looking for ideas on how to make this less confusing.

As I see it, we have three options. In my order of preference:
1. Drop the combined view from this PR and merge it with just the seperate views, as was the original intent of #1444. We can then work on a better, non-rushed version of the combined view.
2. Merge this PR as-is, with both daemonset view and combined view, and iterate on these issues in a followup.
3. Delay merging this PR (which adds any support for daemonsets at all) until the combined issue is resolved.

2opremio:
I would go for (1) as well. @rade ?

rade:
go for (1), but make sure there is a follow up, in particular peeling off bits from #552, specifically replacing (nearly) all existing views with a view that shows you everything in existence, at the highest level of abstraction in which it has a representation.

ekimekim:
I've removed the combined view parts and otherwise rebased fixups. PTAL.

2opremio:
Looks good. There is still a comment on a TODO

ekimekim:
Rebased and resolved conflicts with https://github.com/weaveworks/scope/pull/2528.
Now propagates errors from getting label selectors for all types, not just daemonsets.
PTAL.

2opremio:
I am exceptionally merging this myself (and not the PR creator) to move forward with the release

PR #2535, opened 2017-05-23
Enable eBPF tracking by default

### LINKED ISSUE #2441: enable ebpf connection tracking by default
Currently the ebpf connection tracking from #2135 needs to be enabled explicitly with `--probe.ebpf.connections=true`. Once we have gained some confidence in the correctness and stability of this feature, we should enable it by default.

2opremio:
Fixes #2441

We already had an integration test for the proc fallback (313), so I just adjusted the ebpf flag according to its default value and changed tests explicitly testing ebpf-tracking to explicitly test tracking without ebpf.

### DIFF
--- a/integration/301_internet_edge_without_ebpf_test.sh
+++ b/integration/301_internet_edge_without_ebpf_test.sh
@@ -3,10 +3,10 @@
 # shellcheck disable=SC1091
 . ./config.sh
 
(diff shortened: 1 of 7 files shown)


2opremio:
@alban PTAL (ignore the netlify error)

PR #2539, opened 2017-05-25
Simplify connection tracker init and fix procfs scan fallback

rade:


2opremio:
LGTM (on green).

Much easier to read. Thanks

rade:
ouch @2opremio, I think I found a bug...

I am pretty sure in https://github.com/weaveworks/scope/commit/7497c7d4323a054d1c06cd642f233e7086750927 you should also have changed https://github.com/weaveworks/scope/blob/master/probe/endpoint/connection_tracker.go#L100

And guess what, I picked the wrong version in this PR. Will get that fixed. Is that worthy of a 1.4.x release? Perhaps not since ebpf is not enabled by default, so the above code won't be exercised in most installs.

2opremio:
Good catch!

> Is that worthy of a 1.4.x release? Perhaps not since ebpf is not enabled by default, so the above code won't be exercised in most installs.

I agree, it only happens when explicitly enabling ebpf and when it fails. Coincidentaly I was thinking about how to force this situation in an integration test for https://github.com/weaveworks/scope/pull/2535

PR #2546, opened 2017-05-29
Use map helpers

rade:
Refactor various piece of code that use `ps.Map` to use the functions in map_helpers.go.

### DIFF
--- a/report/counters.go
+++ b/report/counters.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"fmt"
 	"reflect"
-	"sort"
 
 	"github.com/ugorji/go/codec"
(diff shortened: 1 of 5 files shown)


rade:
Q: `NodeSet.ForEach` processes entries in sorted order. Is that actually necessary?

2opremio:
> NodeSet.ForEach processes entries in sorted order. Is that actually necessary?

I recall this was done to present a stable ordering in the serialized JSO. I don’t remember any other reasons. So, we could simply order the keys in toIntermediate  and stop ordering them for the rest

2opremio:
Much cleaner, thanks

PR #2549, opened 2017-05-31
Aarch64 support

adivjoseph:
Tested on Odroid-C2 Ubuntu 16.04, OD1000 Ubuntu 17.04 and MacchiatoBin Ubuntu 16.04. Test application Weaveworks Socks portal.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,21 @@
 .PHONY: all deps static clean client-lint client-test client-sync backend frontend shell lint ui-upload
 
+#add multi arch support for Dockerfiles; x86_64, aarch64, ...
+ARCH=$(shell uname -m) 
+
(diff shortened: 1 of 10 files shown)


rade:
We've been through all this with Weave Net. See the [makefile](https://github.com/weaveworks/weave/blob/master/Makefile). That is neater and also adds manifests, which are missing from this PR.

adivjoseph:
I like the template method, i will rework the PR. Also I will use the Docker distrib model.

2opremio:
Fantastic, thanks!

adivjoseph:
I have made an updated PR that does not require the proliferation of Dockerfiles.
Hence this PR is being removed. Thanks for the good feedback.

Issue #2525, opened 2017-05-16
UI does not treat "" as 'none' in topologyOptions


rade:
Feeding a `APITopologyOptionGroup{ID: "namespace", Default: "", SelectType: "union", NoneLabel: noneLabel}` to the UI, does not select "All Namespaces" in the UI, e.g. you end up with

![screenshot from 2017-05-16 17-46-20](https://cloud.githubusercontent.com/assets/109109/26117806/99f5e1c0-3a5f-11e7-87c2-89a627f11c9b.png)

By contrast, feeding a `Default: "none"` works as expected. @ekimekim believes this is a bug/wart.

See https://github.com/weaveworks/scope/pull/2522#discussion_r116689283

ekimekim:
More correctly, `none` is being treated as a special case value when it should not. Union-type values are comma-seperated lists, with `""` indicating the empty list. A value of `"none"` should mean `["none"]`, not `[]`.

Also, I believe this error to be relevant:
```
app-2ee0f4f….js?4d6e899…:41 Uncaught TypeError: Cannot read property 'includes' of undefined
    at t.value (app-2ee0f4f….js?4d6e899…:41)
    at f._renderValidatedComponentWithoutOwnerOrContext (vendors.js?4d6e899…:25)
    at f._renderValidatedComponent (vendors.js?4d6e899…:25)
    at f.performInitialMount (vendors.js?4d6e899…:25)
    at f.mountComponent (vendors.js?4d6e899…:25)
    at Object.mountComponent (vendors.js?4d6e899…:11)
    at d.mountChildren (vendors.js?4d6e899…:25)
    at d._createInitialChildren (vendors.js?4d6e899…:25)
    at d.mountComponent (vendors.js?4d6e899…:25)
    at Object.mountComponent (vendors.js?4d6e899…:11)
```
I get this error when I try to set default = `""`, and the option group in question doesn't display at all. I believe what is happening is that, since the value is empty and the go struct is annotated with `omitempty`, it is not sending the key at all. This is then causing a type error in the JS code since the expected key is not present.
Possible solutions:
* Remove the omitempty annotation
* Have the UI understand that omitted means `""`

I'd prefer the former as it keeps the API semantics simpler.

PR #2552, opened 2017-06-01
add k8s combined view

ekimekim:
This adds a 'combined' kubernetes view that shows each pod at the 'highest level of abstraction', as follows:
	* Any daemonsets the pod is in
	* Any deployments the pod is in
	* Any replica-sets the pod is in that do not map to a deployment
	* If none of the above, the bare pod itself

Along the way, it fixes https://github.com/weaveworks/scope/issues/2525 by removing the omitempty for the default value, further generalises render.Map2Parent, and introduces a new Renderer type for combining a set of nodes with a set containing the same nodes with more information, but with the output set of nodes only being the original set (but with the extra information), unlike Reduce which takes a union.

Still to do:
	* How does the table view work in this setup?
	* We want to assign shapes to the different topologies present to differentiate them graphically

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -197,6 +197,19 @@ func MakeRegistry() *Registry {
 		},
 	}
 
+	k8sCombinedTypeFilter := APITopologyOptionGroup{
+		ID:         "grouptype",
+		Default:    "",
+		SelectType: "union",
+		NoneLabel:  "All Types",
+		Options: []APITopologyOption{
+			{Value: report.Deployment, Label: "Deployments", filter: render.IsTopology(report.Deployment), filterPseudo: false},
+			{Value: report.DaemonSet, Label: "Daemonsets", filter: render.IsTopology(report.DaemonSet), filterPseudo: false},
+			{Value: report.ReplicaSet, Label: "Replica sets", filter: render.IsTopology(report.ReplicaSet), filterPseudo: false},
+			{Value: report.Pod, Label: "Pods", filter: render.IsTopology(report.Pod), filterPseudo: false},
+		},
+	}
+
 	// Topology option labels should tell the current state. The first item must
 	// be the verb to get to that state
 	registry.Add(
@@ -274,7 +287,7 @@ func MakeRegistry() *Registry {
 			parent:      podsID,
 			renderer:    render.KubeCombinedRenderer,
 			Name:        "combined",
-			Options:     []APITopologyOptionGroup{unmanagedFilter},
+			Options:     []APITopologyOptionGroup{unmanagedFilter, k8sCombinedTypeFilter},
 			HideIfEmpty: true,
 		},
 		APITopologyDesc{
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -29,9 +29,7 @@ const (
 	containersByImageID    = "containers-by-image"
 	podsID                 = "pods"
 	replicaSetsID          = "replica-sets"
-	deploymentsID          = "deployments"
-	daemonsetsID           = "daemonsets"
-	kubeCombinedID         = "kube-combined"
+	kubeControllersID      = "kube-controllers"
 	servicesID             = "services"
 	hostsID                = "hosts"
 	weaveID                = "weave"
@@ -121,7 +119,7 @@ func updateKubeFilters(rpt report.Report, topologies []APITopologyDesc) []APITop
 	sort.Strings(ns)
 	topologies = append([]APITopologyDesc{}, topologies...) // Make a copy so we can make changes safely
 	for i, t := range topologies {
-		if t.id == containersID || t.id == podsID || t.id == servicesID || t.id == deploymentsID || t.id == replicaSetsID || t.id == daemonsetsID || t.id == kubeCombinedID {
+		if t.id == containersID || t.id == podsID || t.id == servicesID || t.id == replicaSetsID || t.id == kubeControllersID {
 			topologies[i] = mergeTopologyFilters(t, []APITopologyOptionGroup{
 				namespaceFilters(ns, "All Namespaces"),
 			})
@@ -197,7 +195,7 @@ func MakeRegistry() *Registry {
 		},
 	}
 
-	k8sCombinedTypeFilter := APITopologyOptionGroup{
+	k8sControllersTypeFilter := APITopologyOptionGroup{
 		ID:         "grouptype",
 		Default:    "",
 		SelectType: "union",
@@ -205,8 +203,6 @@ func MakeRegistry() *Registry {
 		Options: []APITopologyOption{
 			{Value: report.Deployment, Label: "Deployments", filter: render.IsTopology(report.Deployment), filterPseudo: false},
 			{Value: report.DaemonSet, Label: "Daemonsets", filter: render.IsTopology(report.DaemonSet), filterPseudo: false},
-			{Value: report.ReplicaSet, Label: "Replica sets", filter: render.IsTopology(report.ReplicaSet), filterPseudo: false},
-			{Value: report.Pod, Label: "Pods", filter: render.IsTopology(report.Pod), filterPseudo: false},
 		},
 	}
 
@@ -267,27 +263,11 @@ func MakeRegistry() *Registry {
 			HideIfEmpty: true,
 		},
 		APITopologyDesc{
-			id:          deploymentsID,
+			id:          kubeControllersID,
 			parent:      podsID,
-			renderer:    render.DeploymentRenderer,
-			Name:        "deployments",
-			Options:     []APITopologyOptionGroup{unmanagedFilter},
-			HideIfEmpty: true,
-		},
-		APITopologyDesc{
-			id:          daemonsetsID,
-			parent:      podsID,
-			renderer:    render.DaemonSetRenderer,
-			Name:        "daemonsets",
-			Options:     []APITopologyOptionGroup{unmanagedFilter},
-			HideIfEmpty: true,
-		},
-		APITopologyDesc{
-			id:          kubeCombinedID,
-			parent:      podsID,
-			renderer:    render.KubeCombinedRenderer,
-			Name:        "combined",
-			Options:     []APITopologyOptionGroup{unmanagedFilter, k8sCombinedTypeFilter},
+			renderer:    render.KubeControllerRenderer,
+			Name:        "controllers",
+			Options:     []APITopologyOptionGroup{unmanagedFilter, k8sControllersTypeFilter},
 			HideIfEmpty: true,
 		},
 		APITopologyDesc{
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -31,6 +31,7 @@ const (
 	replicaSetsID          = "replica-sets"
 	deploymentsID          = "deployments"
 	daemonsetsID           = "daemonsets"
+	kubeCombinedID         = "kube-combined"
 	servicesID             = "services"
 	hostsID                = "hosts"
 	weaveID                = "weave"
@@ -120,7 +121,7 @@ func updateKubeFilters(rpt report.Report, topologies []APITopologyDesc) []APITop
 	sort.Strings(ns)
 	topologies = append([]APITopologyDesc{}, topologies...) // Make a copy so we can make changes safely
 	for i, t := range topologies {
-		if t.id == containersID || t.id == podsID || t.id == servicesID || t.id == deploymentsID || t.id == replicaSetsID || t.id == daemonsetsID {
+		if t.id == containersID || t.id == podsID || t.id == servicesID || t.id == deploymentsID || t.id == replicaSetsID || t.id == daemonsetsID || t.id == kubeCombinedID {
 			topologies[i] = mergeTopologyFilters(t, []APITopologyOptionGroup{
 				namespaceFilters(ns, "All Namespaces"),
 			})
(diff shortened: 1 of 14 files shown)


ekimekim:
cc @rade

rade:
I've given this a whirl by feeding it 30s of recent production reports. When selecting the kube-system namespace, all the scope probes are shown as individual pods rather than a single daemonset. So something isn't working right.

ekimekim:
That would be because something's wrong with the probes in prod, they are reporting no daemonsets:
```
$ jq '.DaemonSet' </tmp/report.json 
{
  "shape": "heptagon",
  "label": "daemonset",
  "label_plural": "daemonsets",
  "nodes": {}
}
```

I'm willing to bet this is because they're running an old version.
Yep. `f3a9b61c` in prod.

Dev is running `0ee3a3fa` which is the commit that introduced daemonsets to the probe. And indeed it works with that report:
screenshot

rade:
D'oh; I should have figured that out myself.

Mind doing a rebase?

ekimekim:
I'm having a lot of trouble understanding how the table code works, but it seems that currently it's combining auto-detected info from all listed nodes to pick a list of columns. This is probably a reasonable good-enough behaviour for now. Just waiting on @jpellizzari for shapes code - in the meantime @rade could you review?

jml:
[Scope Kubernetes View](https://trello.com/c/iJzYERG7/10-scope-kubernetes-view)

jpellizzari:
@ekimekim This commit will add the shapes to the UI: https://github.com/weaveworks/scope/commit/7aa6ebdb33518c14257354787150b8299c73e61f

Do you want to cherry-pick that into your branch?

ekimekim:
done.

client/app/scripts/charts/node.js
```diff
@@ -14,27 +14,27 @@ import { NODE_BASE_SIZE } from '../constants/styles';
 import NodeShapeStack from './node-shape-stack';
 import NodeNetworksOverlay from './node-networks-overlay';
 import {
-  NodeShapeCloud,
   NodeShapeCircle,
```
jpellizzari: Doesn't this have more sides than all the other shapes? ;)

ekimekim: I did try to put Cloud last in all places, guess I missed one.

ekimekim: oh, derp, I misread your comment. Thought you were talking about the cloud one, not circle. Yes, haha, a circle is uniform convergent to the limit as n->infinity of an n-sided regular polygon. It's also 1-sided if you consider a side to be a smooth line (ie. a curve with no 'corners').

rade:
Feedback on current functionality:

- We had decided that replicasets should go completely; currently they are still shown - as a separate topology, as part of the 'controllers' topology, as children of deployments, as parents of pods.
- I cannot see the new shapes; everything's still a heptagon
- 'controllers' should be the default view, instead of 'services', since it typically gives a more complete view of the system
- the table mode for 'controllers' should contain a column indicating the type (i.e. 'deployment', 'daemonset', ...)
- as discussed, we should enable searching by type, which in turn would allow us to...
- remove the type filter

ekimekim:
Feedback on current functionality:

   - We had decided that replicasets should go completely; currently they
   are still shown - as a separate topology, as part of the 'controllers'
   topology, as children of deployments, as parents of pods.

todo. removing replica sets entirely is non trivial


   - I cannot see the new shapes; everything's still a heptagon

This comes from the report, you won't see new shapes using prod reports.


   - 'controllers' should be the default view, instead of 'services', since
   it typically gives a more complete view of the system

TODO and trivial


   - the table mode for 'controllers' should contain a column indicating
   the type (i.e. 'deployment', 'daemonset', ...)

TODO, may be tricky since Topology isn't a Latest key.


   - as discussed, we should enable searching by type, which in turn would
   allow us to...
   - remove the type filter

latter is trivial once former is done. former is frontend I believe, i'll
get jordan on it

rade:
> removing replica sets entirely is non trivial

In which case let's leave them as a separate topology for now, and not show them as part of the 'controllers' topology.

To be clear, getting rid of them *is* important, but we can do it as a follow-up issue.

report/report.go
```diff
@@ -29,9 +29,12 @@ const (
 
 	// Shapes used for different nodes
 	Circle   = "circle"
+	Triangle = "triangle"
 	Square   = "square"
-	Heptagon = "heptagon"
+	Pentagon = "pentagon"
 	Hexagon  = "hexagon"
+	Heptagon = "heptagon"
+	Octogon  = "octogon"
```
rade: It's called 'oct*a*gon'.

This is wrong everywhere.

report/report.go
```diff
@@ -155,7 +158,7 @@ func MakeReport() Report {
 			WithLabel("host", "hosts"),
 
 		Pod: MakeTopology().
-			WithShape(Heptagon).
+			WithShape(Octogon).
```
rade: pods must be heptagons, since that's the k8s logo.

report/report.go
```diff
@@ -167,11 +170,11 @@ func MakeReport() Report {
 			WithLabel("deployment", "deployments"),
 
 		ReplicaSet: MakeTopology().
-			WithShape(Heptagon).
+			WithShape(Pentagon).
```
rade: I'd make that a triangle since it's the most weird-looking shape and we are going to get rid of replicasets soon.

ekimekim:
It's also the most distinct. I want to avoid having hexagons, heptagons and
octegons on the same screen since they're hard to tell apart at a glance.

rade:
>  I want to avoid having hexagons, heptagons and octegons on the same screen since they're hard to tell apart at a glance.

You are not going to be able to avoid that once we add stateful sets and jobs, though I'd be happy to use triangles for the latter.

report/report.go
ekimekim: done

report/report.go
ekimekim: done

report/report.go
ekimekim: done

rade:
> In which case let's leave [replica sets] as a separate topology for now, and *not show them as part of the 'controllers' topology*.

The 2nd part hasn't been done. Looking at the code, that should simplify things a fair bit.

ekimekim:
Addressed everything but the type search, type filter, and type name in table view. working on the latter now.

render/pod.go
```diff
@@ -58,56 +68,58 @@ var PodServiceRenderer = ConditionalRenderer(renderKubernetesTopologies,
...
+// KubeControllerRenderer is a Renderer which combines all the 'controller' topologies.
+// We first map pods to all possible things they can map to, then we map again to
+// deployments since some things (replica sets) can map to those (the rest are passed through
+// unchanged).
+// Pods with no controller are mapped to 'Unmanaged'
+// We can't simply combine the rendered graphs of the high level objects as they would never
+// have connections to each other.
+// We combine with all the full topologies using ReduceFirstOnly, which keeps the same
+// set of nodes but merges in the full data from the other renderers.
+var KubeControllerRenderer = ConditionalRenderer(renderKubernetesTopologies,
```
rade: Isn't this missing the single metric propagation, carried out elsewhere with `PropagateSingleMetrics(report.Pod)`?

Won't the use of `MakeReduceFirstOnly` mean that this won't show controllers with no pods?

render/detailed/summary.go
```diff
@@ -258,13 +258,24 @@ func podNodeSummary(base NodeSummary, n report.Node) (NodeSummary, bool) {
 	return base, true
 }
 
+var podGroupNodeTypeName = map[string]string{
+	report.Deployment: "Deployment",
+	report.DaemonSet:  "Daemon Set",
+	report.ReplicaSet: "Replica Set",
```
ekimekim: yep, good catch. no need for the special minor label if it's not going to share a topo with other types

render/pod.go
ekimekim: Correct. I think that's what I thought the intent was when I first wrote it - that it only shows the highest abstraction of all active pods.
I might take a crack at just rewriting this whole thing...

rade: Thinking ahead to a world where probes don't include ReplicaSets, and instead list Deployments as direct parents of pods, it occurs to me that we should somehow transform the pod topology to add deployments as parents.

We can do that in the pod renderer. Or even as early as the report reading.

The rest then ought to be straightforward.

rade:
> Addressed everything but the type search, type filter, and type name in table view. working on the latter now.

I'd be happy to defer the first two to a follow-up PR.

ekimekim:
So it turns out that `table_templates` doesn't affect what I thought it does - it creates extra 'tables' in the details view, it doesn't affect the table view. The table view is exclusively populated from the same metadata as the main part of the details view. So we can have the column as we want, but only at the cost of the same data `Type: Daemon Set` showing up in the details panel. Which is probably a good thing, tbh.

rade:
> So we can have the column as we want, but only at the cost of the same data `Type: Daemon Set` showing up in the details panel. Which is probably a good thing, tbh.

Agreed!

ekimekim:
Ok, I'm locking off this branch. @rade please give final review then we can merge. Meanwhile I'll start getting rid of Replica Sets entirely in a seperate branch.

render/render.go
```diff
@@ -69,6 +69,40 @@ func (r *Reduce) Stats(rpt report.Report, dct Decorator) Stats {
...
+	return result
+}
+
+// Stats implements Renderer
+func (r *ReduceFirstOnly) Stats(rpt report.Report, dct Decorator) Stats {
+	// Since we're just taking more detailed data from the other renderers,
+	// we probably only care about the filter stats for the 'main' renderer.
+	return r.first.Stats(rpt, dct)
+}
+
```
rade: we don't need any of this anymore.

app/api_topologies.go
```diff
@@ -196,6 +195,17 @@ func MakeRegistry() *Registry {
 		},
 	}
 
+	k8sControllersTypeFilter := APITopologyOptionGroup{
+		ID:         "grouptype",
+		Default:    "",
```
rade: When I test this branch, the type filter has nothing selected. Unlike, say, the namespace filter, which has "All Namespaces" selected.

I suspect that's because the above ought to be `Default: "none"`. Even though you say this branch fixes #2525. (in which case, btw, I would expect existing filters, such as the namespace filter, to make use of this functionality - which suggests this change would be better off in its own PR)

probe/kubernetes/daemonset.go
```diff
@@ -48,5 +48,6 @@ func (d *daemonSet) GetNode() report.Node {
 		DesiredReplicas:      fmt.Sprint(d.Status.DesiredNumberScheduled),
 		Replicas:             fmt.Sprint(d.Status.CurrentNumberScheduled),
 		MisscheduledReplicas: fmt.Sprint(d.Status.NumberMisscheduled),
+		NodeType:             "Daemon Set",
```
ekimekim: We can add the value in during rendering, though it's not obvious how we can also cleanly add in the metadata_template value so that it actually displays.
It's a fundamental part of scope's design that considerable amounts of display logic comes directly from the scope probe, which means old reports will use old display settings. If that's something we want to move away from, we can, but it will take time, and brings with it complications eg. displaying values that simply don't exist in older reports.

ekimekim: I'm going to leave this as-is for now, and we can revisit later

ekimekim:
Addressed everything except the continuing saga of https://github.com/weaveworks/scope/issues/2525, which I'm looking into now.

render/pod.go
```diff
@@ -109,16 +107,14 @@ func renderParents(childTopology string, parentTopologies []string, noParentsAct
 	}
 	// weird append form is to accomplish MakeReduce(subrenderer, selectors...) since that form isn't allowed
 	return MakeReduce(append(
-		[]Renderer{
+		selectors,
```
rade: you forgot to remove the comment about the "weird append".

ekimekim:
I've split off the https://github.com/weaveworks/scope/issues/2525 fixes into https://github.com/weaveworks/scope/pull/2633

ekimekim:
Rebased and resolved minor conflict. PTAL?

ekimekim:
Note it's a known issue that the type selector will have that minor display bug until https://github.com/weaveworks/scope/pull/2633 lands

rade:
LGTM but still needs benchmarking as per https://github.com/weaveworks/scope/pull/2552#pullrequestreview-44701362. Also please *selectively* squash some of the commits.

ekimekim:
Perf results. Note I adapted the methodology a bit - my numbers were tiny and highly variable, so I did 6 runs: The first I discarded on the premise that it would prime any caches that may affect the result. I then took the top and bottom results and discarded them as outliers, and recorded the second-top and second-bottom. ie. I recorded the 20th and 80th percentiles. All times are in milliseconds.

type | before | after
---- | ------ | -----
processes|79 - 117|49 - 124
containers|151 - 215|148 - 185
pods|31 - 51|36 - 45
services|22 - 45|26 - 45
hosts|17 - 29|11 - 30

Based on the ranges of the results and the high variability, I don't think any of the changes are significant.

ekimekim:
@rade final lgtm to merge?

ekimekim:
or not, since there seems to be a test failing that `make tests` isn't hitting locally...will investigate tomorrow.

rade:
> Perf results. Note I adapted the methodology a bit - my numbers were tiny and highly variable, so I did 6 runs: The first I discarded on the premise that it would prime any caches that may affect the result.

All you are testing with this approach is memoization. You must restart scope after every single test. See #2613 for better instructions.

rade:
I've run the perf tests...

|                  |  old |  new |
| ------------------  | ------ | ------ |
| processes        | 0.59 | 0.59 |
| containers       |  1.6 |  1.6 |
| pods             |  1.5 |  1.5 |
| deployments      |  1.5 |      |
| kube-controllers |      |  1.5 |
| services         |  1.5 |  1.5 |
| hosts            |  2.0 |  2.1 |

hosts is marginally slower. Rest is the same. LGTM. Just the test failures to go then.

ekimekim:
Looks like the broken test is a problem on the version of master I happened to rebase onto. I've just rebased onto a newer version of master and it should work this time.

PR #2554, opened 2017-06-01
Use 127.0.0.1 instead of localhost, more

bboreham:
We do this in case that name resolves to something else.
This was done in #2103 for one case; don't see why we wouldn't do the same everywhere.

### DIFF
--- a/extras/fixprobe/main.go
+++ b/extras/fixprobe/main.go
@@ -19,7 +19,7 @@ import (
 
 func main() {
 	var (
-		publish         = flag.String("publish", fmt.Sprintf("localhost:%d", xfer.AppPort), "publish target")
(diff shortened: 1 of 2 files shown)


tools/socks/main.go
```diff
@@ -23,11 +23,11 @@ const (
 	pacfile = `
 function FindProxyForURL(url, host) {
 	if(shExpMatch(host, "{{.HostMatch}}")) {
-		return "SOCKS5 localhost:8000";
+		return "SOCKS5 127.0.0.1:8000";
```
2opremio: Please make these changes in tools (although I don't think it's needed in this case since we only use this internally).

bboreham: I removed it from this PR

PR #2556, opened 2017-06-01
Use Kubernetes node name to filter pods if possible

bboreham:
We can get the node name via the downward api, if the Scope probe is run as a pod. This needs to be done in the yaml.

### DIFF
--- a/probe/kubernetes/reporter.go
+++ b/probe/kubernetes/reporter.go
@@ -111,11 +111,12 @@ type Reporter struct {
 	probe           *probe.Probe
 	hostID          string
 	handlerRegistry *controls.HandlerRegistry
+	nodeName        string
 	kubeletPort     uint
 }
 
 // NewReporter makes a new Reporter
-func NewReporter(client Client, pipes controls.PipeClient, probeID string, hostID string, probe *probe.Probe, handlerRegistry *controls.HandlerRegistry, kubeletPort uint) *Reporter {
+func NewReporter(client Client, pipes controls.PipeClient, probeID string, hostID string, probe *probe.Probe, handlerRegistry *controls.HandlerRegistry, nodeName string, kubeletPort uint) *Reporter {
 	reporter := &Reporter{
 		client:          client,
 		pipes:           pipes,
@@ -123,6 +124,7 @@ func NewReporter(client Client, pipes controls.PipeClient, probeID string, hostI
 		probe:           probe,
 		hostID:          hostID,
 		handlerRegistry: handlerRegistry,
+		nodeName:        nodeName,
 		kubeletPort:     kubeletPort,
 	}
 	reporter.registerControls()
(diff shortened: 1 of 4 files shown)


2opremio:
To be effective, this requires an associated change in launch-generator for the next release of Scope (CC @errordeveloper )

probe/kubernetes/reporter.go
```diff
@@ -108,18 +108,20 @@ type Reporter struct {
...
-func NewReporter(client Client, pipes controls.PipeClient, probeID string, hostID string, probe *probe.Probe, handlerRegistry *controls.HandlerRegistry, kubeletPort uint) *Reporter {
+func NewReporter(client Client, pipes controls.PipeClient, probeID string, hostID string, probe *probe.Probe, handlerRegistry *controls.HandlerRegistry, nodeName string, kubeletPort uint) *Reporter {
 	reporter := &Reporter{
 		client:          client,
 		pipes:           pipes,
 		probeID:         probeID,
 		probe:           probe,
 		hostID:          hostID,
 		handlerRegistry: handlerRegistry,
+		nodeName:        nodeName,
```
2opremio: This assumes node names cannot change ... I guess that's true but I intentionally didn't cache it before just in case. Can you confirm?

bboreham: According to the [source](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L87) no `Name` of a Kubernetes object can be updated.

`Node` embeds that kind of `Name` [here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/pkg/api/v1/types.go#L3498)

2opremio: Fair enough.

2opremio:
@bboreham Can you please fix the tests?

probe/kubernetes/reporter.go
rade: *nine* arguments!?!? I think it's time for a `ReporterConfig`.

probe/kubernetes/reporter.go
```diff
@@ -111,18 +111,20 @@ type Reporter struct {
 	probe           *probe.Probe
 	hostID          string
 	handlerRegistry *controls.HandlerRegistry
+	nodeName        string
 	kubeletPort     uint
 }
 
 // NewReporter makes a new Reporter
-func NewReporter(client Client, pipes controls.PipeClient, probeID string, hostID string, probe *probe.Probe, handlerRegistry *controls.HandlerRegistry, kubeletPort uint) *Reporter {
+func NewReporter(client Client, pipes controls.PipeClient, probeID string, hostID string, probe *probe.Probe, handlerRegistry *controls.HandlerRegistry, nodeName string, kubeletPort uint) *Reporter {
```
rade: *nine* arguments?!? I think it's time for a `ReporterConfig`.

rade: happy to tackle that post merge.

Issue #2551, opened 2017-05-31
no container connections when `probe.proc.spy=false`


rade:
When setting `probe.proc.spy=false`, no container connections are displayed.

My test setup consist of two machines:

1. my desktop, running the scope app & probe
2. a VM, running the scope probe

I tried three variations:

### scope connections
Just checking whether the connection between the probes and app are showing. They don't. @2opremio says that in the absence of proc info scope tries to identify containers with an IP address, which won't work here because scope runs in the host netns and hence doesn't have  an IP of its own

### plain containers
on the desktop run: `docker run --rm -p 1122:1122 -ti alpine nc -lkp 1122`
on the VM run: `docker run --rm -ti alpine nc <desktop-ip> 1122`

The connection gets established and anything I type gets echoed across, but the connection does not show up in scope.

### containers on weave net
on the desktop run
```
$ weave launch
$ eval $(weave env)
$ docker run --rm --name=foo -ti alpine nc -lkp 1122
```
on the VM run
```
$ weave launch <desktop-ip>
$ eval $(weave env)
$ docker run --rm -ti alpine nc foo 1122
```
The connection gets established and anything I type gets echoed across, but the connection does not show up in scope.

rade:
when repeating the experiment with containers running on the same host, with the nc client connecting to the server's docker IP, the connection does show up. It does *not* show up when going via the port published on the host IP.

rade:
> with containers running on the same host, with the nc client connecting to the server's docker IP, the connection does show up.

That's true even when running with `probe.proc.spy=true`!

rade:
> That's true even when running with `probe.proc.spy=true`!

I have raised #2558 for that.

rade:
I've figured out what's going on here...

The `ContainerRenderer` produces nodes in three different ways:

1. projection from the process renderer, mapping processes to containers
2. projection from the endpoint topology, mapping IPs to containers
3. selection of everything in the "container" topology

Crucially, the 2nd projection *ignores endpoints with `procspied=true`*, under the assumption that such endpoints will be mapped via the 1st projection. That assumption is false...

For the process topology to map an endpoint to a process, that endpoint needs to have a PID. But it is `probe.proc.spy=true` (help text: "associate endpoints with processes") that does that. Whereas `procspied` gets set even when an endpoint was created from a simple scan of `/proc/net/tcp`, which does not produce PIDs. That's exactly what happens when `probe.proc.spy=false`.

[...]

PR #2559, opened 2017-06-03
simplify connection join

rade:
Refactor and rename `ShortLivedConnectionJoin`, simplify some filters, and fix a bug along the way: we weren't filtering out eBPF-tracked endpoints (it's not clear what harm that caused, but at the very least it's a consistency fix to complement the process renderer).

This PR is best reviewed per-commit; there is very little going on here besides some moving and renaming and github does a poor job with the consolidated diff.


render/container.go
```diff
@@ -35,23 +34,31 @@ var ContainerRenderer = MakeFilter(
...
 
-		// This mapper brings in short lived connections by joining with container IPs.
-		// We need to be careful to ensure we only include each edge once.  Edges brought in
-		// by the above renders will have a pid, so its enough to filter out any nodes with
-		// pids.
-		ShortLivedConnectionJoin(SelectContainer, MapContainer2IP),
+		// This mapper brings in connections by joining with container
+		// IPs.  We need to be careful to ensure we only include each
+		// edge once.  Edges brought in by the above renders will have
+		// a pid, so its enough to filter out any nodes with pids.
```
2opremio: Since we are re-writing the comment:
* It's not obvious to the reader why filtering out PIDs will ensure we get no duplicates (i.e. some explanation about PID-based connections vs conntrack may be in order)
* Most importantly, it's not obvious why duplicates are a problem (i.e. merge would settle it  ... except for counts)

render/container.go
```diff
@@ -35,23 +34,31 @@ var ContainerRenderer = MakeFilter(
...
 
 		SelectContainer,
 	),
 )
 
+var mapEndpoint2IP = MakeMap(
+	endpoint2IP,
+	// We drop endpoint nodes with pids, as they will be joined to
+	// containers through the process topology, and we don't want to
+	// double count edges.
```
2opremio: And ... I just show this, maybe this is enough.

render/container.go
```diff
@@ -66,80 +73,67 @@ func ShortLivedConnectionJoin(r Renderer, toIPs func(report.Node) []string) Rend
...
-			MakeMap(
-				endpoint2IP,
-				SelectEndpoint,
-			),
-		),
-	))
+	// We also allow for joining on ip:port pairs.  This is useful for
+	// connections to the host IPs which have been port mapped to a
+	// container can only be unambiguously identified with the port.
+	// So we need to emit two nodes, for two different cases.
```
2opremio: Maybe mention that we need to do this because in some cases connection tracking doesn't provide PIDs?

render/container.go
2opremio: To be precise, you are not dropping nodes with PIDs

render/container.go
2opremio: As above, you are not really dropping endpoint nodes with PIDs ...

2opremio:
It looks good code-wise. I would suggest refining the comments a bit more.

rade:
I had tried updating the comments before but in the process discovered #2551 and didn't want to plaster a big "THIS IS ALL WRONG AND DOESN'T MAKE ANY SENSE" sign all over this PR.

So instead I've now just removed/corrected the most blatant falsehoods, i.e. the fact that we aren't actually filtering based on pid presence but procspied/eBPF-tracked.

rade:
This PR has resulted in an unexpected and significant performance improvement in producing the host and k8s topologies.

Steps to reproduce:
1. grab a recent ["prod on dev"](https://frontend.dev.weave.works/app/nameless-night-79) report via the 'Save raw data as JSON' item in the 'bug' menu
2. feed this as a static report to scope by running the executable rather than the container
```
 prog/scope --mode=app --weave=false --app.collector=file:/home/matthias/tmp/prod-on-dev-report.json 
```
3. `time curl -so /dev/null localhost:4040/api/topology/hosts`

Results (times in seconds):

|            |  before | after |
| ---------- | --- | --- |
| processes  | 0.3 | 0.3 |
| containers | 1.8 | 1.8 |
| pods       | 2.5 | 1.5 |
| services   | 2.4 | 1.4 |
| hosts      | 2.6 | 1.6 |

This happens because previously the `ConnectionJoin`s from the k8s and container topologies had distinct `mapEndpoint2IP` mappers, whereas now they are shared via a global variable. Consequently memoization can eliminate the multiple executions of these mappers which results from the embedding of the container renderers in the k8s and host topology renderers.

PR #2561, opened 2017-06-05
parallel reduce

rade:
Execute reducer branches in parallel.

This reduces the time it takes to fetch topologies. Using the same test methodology is as in https://github.com/weaveworks/scope/pull/2559#issuecomment-306142823, running on my feeble laptop...

|            |  before | after |
| ---------- | --- | --- |
| processes  | 0.3 | 0.3 |
| containers | 1.8 | 1.6 |
| pods       | 1.5 | 1.3 |
| services   | 1.4 | 1.2 |
| hosts      | 1.6 | 1.2 |

(all results in seconds)

### DIFF
--- a/render/render.go
+++ b/render/render.go
@@ -39,11 +39,25 @@ func MakeReduce(renderers ...Renderer) Renderer {
 
 // Render produces a set of Nodes given a Report.
 func (r *Reduce) Render(rpt report.Report, dct Decorator) report.Nodes {
-	result := report.Nodes{}
-	for _, renderer := range *r {
-		result = result.Merge(renderer.Render(rpt, dct))
+	l := len(*r)
+	switch l {
+	case 0:
+		return report.Nodes{}
 	}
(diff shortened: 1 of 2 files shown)


render/render.go
```diff
@@ -39,11 +39,25 @@ func MakeReduce(renderers ...Renderer) Renderer {
...
-	result := report.Nodes{}
+	l := len(*r)
+	switch l {
+	case 0:
+		return report.Nodes{}
+	}
+	c := make(chan report.Nodes, l)
 	for _, renderer := range *r {
-		result = result.Merge(renderer.Render(rpt, dct))
+		renderer := renderer // Pike!!
```
2opremio: :(

2opremio:
I am really curious as to how this performs in busy machines with many cores, merging reports from many probes.

2opremio:
Really small patch for such a performance improvement BTW.

PR #2562, opened 2017-06-05
Back off upon errored kubernetes api requests

### LINKED ISSUE #1009: k8s: Back-off retrying to connect to Kubernetes API
Right now the kube probe tries to connect two times, then waits for 10s and tries again. I think we should back-off a bit.

rndstr:
fixes #1009

### DIFF
--- a/probe/kubernetes/client.go
+++ b/probe/kubernetes/client.go
@@ -7,8 +7,9 @@ import (
 	"sync"
 	"time"
 
-	log "github.com/Sirupsen/logrus"
 	"github.com/weaveworks/common/backoff"
+
+	log "github.com/Sirupsen/logrus"
 	"k8s.io/kubernetes/pkg/api"
 	"k8s.io/kubernetes/pkg/apis/extensions"
 	"k8s.io/kubernetes/pkg/client/cache"
--- a/probe/kubernetes/client.go
+++ b/probe/kubernetes/client.go
@@ -1,12 +1,14 @@
 package kubernetes
 
 import (
+	"fmt"
 	"io"
 	"strconv"
 	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/weaveworks/common/backoff"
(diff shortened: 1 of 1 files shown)


probe/kubernetes/client.go
```diff
@@ -56,14 +58,39 @@ type client struct {
...
+	case <-stopCh:
+		return
+	default:
+	}
+
+	var err error
+	wait := resyncPeriod
+	for {
+		func() {
+			defer runtime.HandleCrash()
```
2opremio: Is this needed?

From the HandleCrash documentation: 

```
// TODO: remove this function. We are switching to a world where it's safe for
// apiserver to panic, since it will be restarted by kubelet.
```

2opremio:
@rndstr Do you think you could reuse `github.com/weaveworks/common/backoff` ?

probe/kubernetes/client.go
rndstr: The `vendor/`ed version doesn't have this comment yet and I presume it only concerns apimachinery since it has restarts in place. But I've heard we are okay if scope crashes so I removed it, thanks!

rndstr:
@2opremio thanks for the suggestion, looks much cleaner now. I also stumbled over an inconsistency regarding logging errors in the backoff code, fix in the common repo: https://github.com/weaveworks/common/pull/42

probe/kubernetes/client.go
```diff
@@ -1,12 +1,14 @@
...
 
 import (
+	"fmt"
 	"io"
 	"strconv"
 	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/weaveworks/common/backoff"
```
2opremio: Please move this to a separate block above.

As a convention we create 3 import blocks:

1. Standard library imports
2. Weaveworks imports
3. External imports

See for instance https://github.com/weaveworks/scope/blob/master/probe/overlay/weave.go

PR #2567, opened 2017-06-06
Fix prog/main_test.go

rndstr:
Is excluded from tests and does not compile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,10 +179,12 @@ tmp/weave-scope.tgz: $(shell find client/app -type f) $(SCOPE_UI_BUILD_UPTODATE)
 else
 
 client/build/index.html:
-	cd client && npm run build
+	test "true" = "$(SCOPE_SKIP_UI_ASSETS)" && { mkdir -p client/build; } || \
+		{ cd client && npm run build; }
 
 client/build-external/index.html:
-	cd client && npm run build-external
+	test "true" = "$(SCOPE_SKIP_UI_ASSETS)" && { mkdir -p client/build-external; } || \
+		{ cd client && npm run build-external; }
 
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -179,11 +179,11 @@ tmp/weave-scope.tgz: $(shell find client/app -type f) $(SCOPE_UI_BUILD_UPTODATE)
 else
 
 client/build/index.html:
-	test "true" = "$(SCOPE_SKIP_UI_ASSETS)" && { mkdir -p client/build; } || \
+	test "true" = "$(SCOPE_SKIP_UI_ASSETS)" && mkdir -p client/build || \
 		{ cd client && npm run build; }
 
 client/build-external/index.html:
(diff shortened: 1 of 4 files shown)


rndstr:
Not sure what the consensus is on testing private interfaces; if "okay" then I can re-add and fix it

2opremio:
Good catch, sigh. I think we should re-add it.

rndstr:
@2opremio ok if I use github.com/stretchr/testify/assert? It's already in `vendor/`

2opremio:
Sure

rndstr:
Executing any backend tests in `prog/` requires the "scope ui asset providers" (`staticui` & `externalui`) to be generated. These in turn have a [dependency on the client app](https://github.com/weaveworks/scope/blob/7d8448c586ea2743815558fab70b2f593bbc0fd5/Makefile#L84-L85) which then needs to be fully built. For this to work we would need to move the `scope-ui-backend` docker image to a node based image (or additionally install node && npm). Which is in my opinion not worth the trouble till we actually do have asset-delivery tests.

Makefile
```diff
@@ -179,10 +179,12 @@ tmp/weave-scope.tgz: $(shell find client/app -type f) $(SCOPE_UI_BUILD_UPTODATE)
 else
 
 client/build/index.html:
-	cd client && npm run build
+	test "true" = "$(SCOPE_SKIP_UI_ASSETS)" && { mkdir -p client/build; } || \
```
rndstr: I'm pretty sure the first expression could be improved?

2opremio:
> Which is in my opinion not worth the trouble till we actually do have asset-delivery tests.

I agree. I didn't know this was going to open a can of worms.

Makefile
2opremio: I've never seen braces used like this. Is it for subprocess execution? What's the difference between `{ mkdir -p client/build; }` and `( mkdir -p client/build; )`?

rndstr: You are right, there is no difference in what is executed, I removed them. In the subsequent line I think keeping it makes sense for improved readability.

rndstr:
@2opremio I was wrong with `Edit2`, should work as intended & ready for review :)

tools/test
```diff
@@ -74,7 +74,7 @@ fail=0
 
 if [ -z "$TESTDIRS" ]; then
     # NB: Relies on paths being prefixed with './'.
-    TESTDIRS=($(git ls-files -- '*_test.go' | grep -vE '^(vendor|prog|experimental)/' | xargs -n1 dirname | sort -u | sed -e 's|^|./|'))
+    TESTDIRS=($(git ls-files -- '*_test.go' | grep -vE '^(vendor|experimental)/' | xargs -n1 dirname | sort -u | sed -e 's|^|./|'))
```
bboreham: This change needs to be made in the build-tools repository

rndstr: I stumbled over this: https://github.com/weaveworks/build-tools/pull/103/files#diff-098f6bcd4621d373cade4e832627b4f6R81

Is this sufficient?

bboreham: Yes, by mistake that PR included your change.

PR #2575, opened 2017-06-08
Add app capabilities to /api endpoint

2opremio:
The first capability is whether there is report persistence, needed to conditionally enable time controls in the UI  ( https://github.com/weaveworks/scope/pull/2524 )

### DIFF
--- a/common/xfer/constants.go
+++ b/common/xfer/constants.go
@@ -14,6 +14,8 @@ const (
 	ScopeProbeVersionHeader = "X-Scope-Probe-Version"
 )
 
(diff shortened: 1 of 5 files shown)


common/xfer/constants.go
```diff
@@ -14,12 +14,15 @@ const (
 	ScopeProbeVersionHeader = "X-Scope-Probe-Version"
 )
 
+const ReportPersistenceCapability = "report_persistence"
```
rndstr: comment missing (see circleci)

PR #2582, opened 2017-06-11
fix error when docker DAEMON is running with user namespace enabled.

jjqq2013:
Hi, i love to use weaveworks. This is pull request is almost same as https://github.com/weaveworks/scope/pull/2161

I begin to use Docker for Mac, and set daemon using user namespace map: 
screen shot 2017-06-11 at 17 14 37

Then `scope launch` will report error:
```
mkdir: can't create directory '/var/run/scope/': Permission denied
```

The reason is very obvious because when it internally runs
```
docker run --rm --entrypoint=/bin/sh \
        -v /var/run:/var/run \
        "$SCOPE_IMAGE" -c "mkdir -p /var/run/scope/plugins"
```
It actually is using remapped root:root(`10000:10000`), instead of normal `0:0`.

So by same way, it inserted `--userns=host` option to the above run command, then it works fine.

Hope you can merge it.

Finally, i suggest that everyone should pay more attention about user namespace enabled daemon, and use it. This really rocks, it maps non root user to root in container, maybe this will be the default daemon mode in future.

### DIFF
--- a/scope
+++ b/scope
@@ -165,7 +165,8 @@ create_plugins_dir() {
     # is not shared from OS X and does not belong to the system."
     # In any case, creating /var/run/scope/plugins on Mac OS would not work, as domain
     # sockets do not cross VM boundaries. We need this directory to exits on the VM.
-    docker run --rm --entrypoint=/bin/sh \
+    # shellcheck disable=SC2086
+    docker run $USERNS_HOST --rm --entrypoint=/bin/sh \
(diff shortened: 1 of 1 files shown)


2opremio:
Thanks for this. Please add a shellcheck exception `# shellcheck disable=SC2086` since the variable is not quoted.

jjqq2013:
OK, added

PR #2584, opened 2017-06-12
Moved highlighted nodes/edges info to selectors

fbarl:
A small refactoring PR which gets rid of `highlightedNodeIds`/`highlightedEdgeIds` fields in the global state and uses selectors to calculate them from adjacencies and `mouseOverNodeId`/`mouseOverEdgeId` fields instead.

With this change, all the nodes/edges decorators are now seen as computed properties, either through selectors or through decorators in components. That enables use to move forward with stories like #2432, where we need to deal with more complex logic of node states.

The performance wasn't visibly affected with these changes.


fbarl:
@davkal If the recalculation wasn't happening on every `state.get('nodes')` change, then it's likely that it was a bug all along as new nodes wouldn't be highlighted even when they should. That makes for another good case for this PR that I completely forgot about :)

PR #2585, opened 2017-06-12
Move pause button back to footer

fbarl:
Since Time Travel is currently feature-flagged, we don't want the Pause button to be a part of that component, as that would make it unavailable e.g. in production. Once Time Travel will be exposed to everyone, we might reconsider moving the pause button back there.

Additionally, I changed the name of the flag from `timeline-control` to `time-travel` as it corresponds better to the new naming in the UI.

After this PR is merged, I'd update the Scope version in Service UI to have finally have time travel available for testing in dev environment.

@jpellizzari I'd be curious what your ideas are how we could reconcile the two different state trees that actions are getting depending on the context they're called from (see my `TODO` comment).

### DIFF
--- a/client/app/scripts/actions/app-actions.js
+++ b/client/app/scripts/actions/app-actions.js
@@ -635,13 +635,15 @@ function updateFromNodesDeltaBuffer(dispatch, state) {
 }
 
 export function clickResumeUpdate() {
-  return (dispatch, getServiceState) => {
+  return (dispatch, getState) => {
     dispatch({
       type: ActionTypes.CLICK_RESUME_UPDATE
     });
+    // TODO: Find a better way to do this (see the comment above).
+    const state = getState().scope || getState();
     // Periodically merge buffered nodes deltas until the buffer is emptied.
     nodesDeltaBufferUpdateTimer = setInterval(
-      () => updateFromNodesDeltaBuffer(dispatch, getServiceState().scope),
+      () => updateFromNodesDeltaBuffer(dispatch, state),
       NODES_DELTA_BUFFER_FEED_INTERVAL,
     );
   };
--- a/client/app/scripts/components/footer.js
+++ b/client/app/scripts/components/footer.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 
 import Plugins from './plugins';
+import PauseButton from './pause-button';
 import { trackMixpanelEvent } from '../utils/tracking-utils';
 import {
   clickDownloadGraph,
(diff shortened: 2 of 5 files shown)


client/app/scripts/actions/app-actions.js
```diff
@@ -635,13 +635,15 @@ function updateFromNodesDeltaBuffer(dispatch, state) {
 }
 
 export function clickResumeUpdate() {
-  return (dispatch, getServiceState) => {
+  return (dispatch, getState) => {
     dispatch({
       type: ActionTypes.CLICK_RESUME_UPDATE
     });
+    // TODO: Find a better way to do this (see the comment above).
+    const state = getState().scope || getState();
```
jpellizzari: This should not be necessary if the action is being called from inside the `<Namespace />` by a component, and the entirety of Scope should be in there. I will pull this down and take a look.

fbarl: Yeah, the problem is that there are some actions that are called from two different context :)

jpellizzari: @fbarl It looks like the `clickResumeUpdate` action is called by the `PauseButton` as well. Should that not also be wrapped in a `<CloudFeature />`? That would solve this issue.

fbarl: I'd say the whole point of this PR was to move the pause button back to the footer so that it's not a part of `<CloudFeature />`, so that it would stay visible in Scope standalone as was the case so far. :)

fbarl:
@jpellizzari I'll merge it as it is for now since it needs to get into `release-1.5` but maybe we'll eventually have to come up with a better strategy for split cases like above.

jpellizzari:
@fbarl I guess I misunderstood what the pause button is doing. My understanding is that the pause button stops the time-travel feature from incrementing. How can this be used anywhere but Weave Cloud?

rade:
Time moves forward even outside Weave Cloud :)

In the absence of time travel, Scope shows what the system looks "now"; it's continuously updating information, as "now" moves forward. 'pause' pauses the updating.

There *is* a problem here: if you bring up the detail panel, or perform any other action that requires fetching info the client doesn't know already, then the info will be fetched for "now", rather than the time at which the user pressed "pause". @fbarl does the time-travel feature make 'pause' behave correctly here?

jpellizzari:
> Time moves forward even outside Weave Cloud :)

I refuse to believe this.

> 'pause' pauses the updating.

So it seems we are re-using the pause button for both pausing updates and for pausing time travel. Works for me!

PR #2587, opened 2017-06-12
Fix the table sorting order bug for numerical values

### LINKED ISSUE #2573: unable to change sort orders of numeric columns
The sort order of numeric columns in the details panel and table modes is fixed as "descending". By contrast, the sort order of text and datetime columns can be switched between ascending and descending.

Examples:
- Deployments table mode - "# PODS"
- Pods table mode - "CPU"
- Pods detail panel - "PORT"

fbarl:
Fixes #2573.

### DIFF
--- a/client/app/scripts/components/node-details/node-details-table.js
+++ b/client/app/scripts/components/node-details/node-details-table.js
@@ -202,7 +202,7 @@ export default class NodeDetailsTable extends React.Component {
 
     const sortedBy = this.state.sortedBy || getDefaultSortedBy(columns, this.props.nodes);
     const sortedByHeader = this.getColumnHeaders().find(h => h.id === sortedBy);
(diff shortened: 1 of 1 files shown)


fbarl:
> ...the order is now ASC by default. It should be DESC, as the user wants to see the busiest items first.

@davkal I am getting *DESC* by default for numericals (at least for fields like `CPU` and `Memory` in `Pods` table from dev), for which columns did you get *ASC* first?

davkal:
It's ASC on first load (when opening node details for the first time).

fbarl:
> It's ASC on first load (when opening node details for the first time).

@davkal Yes, that's a good catch! I fixed it now by replacing `undefined` with `null` (that part of the code is still a bit fragile :/) and it seems to work fine now, so PTAL.

davkal:
LGTM

Issue #2565, opened 2017-06-05
Sub-topology selector overflows onto the table


jpellizzari:
The list of topologies is overflowing onto the table:
![screen shot 2017-06-05 at 11 56 11 am](https://cloud.githubusercontent.com/assets/2802257/26798759/aed0ae7c-49e6-11e7-97f6-bf28fda1ce26.png)

jpellizzari:
No longer an issue since we reduced the number of sub-topologies

PR #2588, opened 2017-06-12
Adjust header menu vertical position

fbarl:
To make the header menu aligned with the `CONFIGURE` button again after https://github.com/weaveworks/service-ui/pull/624 is merged, I have moved the `.header` div 17px up.

IMO this still looks good (if not better, given the current footer margins) and helps us a bit with issues like #2565, since the vertical space is becoming a problem.

### DIFF
--- a/client/app/styles/_base.scss
+++ b/client/app/styles/_base.scss
@@ -155,9 +155,8 @@
   pointer-events: none;
 
   position: absolute;
-  top: 32px;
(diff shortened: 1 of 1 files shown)


fbarl:
Configure button was moved differently in https://github.com/weaveworks/service-ui/pull/624, so we still want to keep the menu aligned with it.

PR #2600, opened 2017-06-13
Make API calls with time travel timestamp

### LINKED ISSUE #2609: Time Travel: Use timestamp in calls to /api/topology
From #2524:

> Clicking on node details makes a direct call to `api/topology/[topId]/[nodeId]`, disregarding the timestamp. So the API needs to be extended similarly to websockets to make the nodes details data consistent with time travel control. The same goes for `api/topology?` calls which give the info about all the available (sub)topologies as well as their stats (node/edge count, etc...). Here we need to give some thought to the UI, e.g. what happens when the topology we are currently at disappears if we go back in time and it's not currently there?

fbarl:
Fixes #2609. Now we (hopefully) only need to adress #2608 before removing the feature flag.

##### Changes

* Node details now display timestamps relative to the ttime we jumped to with the *Time Travel* control.
* List of topologies as well as topology stats also change when travelling in time.
* Modified the overlay displayed when jumping in time so that it affects the whole rest of the screen (with node details and topologies lists) and not just the rendered nodes.
* Improved the *empty state* messaging for the resource view, as it became more important with time travelling.

##### Pain points

* Would be nice to update the timestamps in *Node Details* every second, e.g. by wrapping them in some sort of `<AutoRefreshable />` component.
* A lot of selectors/actions logic is becoming unnecessarily complicated because *Resource View* is an exception to a lot of things, drawing its displayed nodes from `nodesByTopology` global state instead of `nodes` - would be nice to unify these two global states into one.


app/api_topologies.go
```diff
@@ -415,6 +415,16 @@ type topologyStats struct {
 	FilteredNodes      int `json:"filtered_nodes"`
 }
 
+// deserializedTimestamp converts the ISO8601 query param into a proper timestamp.
+func deserializedTimestamp(timestampStr string) time.Time {
```
2opremio: `s/deserialized/deserialize/`

2opremio: `s/timestampStr/timestamp/`

2opremio:
Backend code LGTM (modulo minor naming nits)

app/api_topologies.go
```diff
@@ -415,6 +415,16 @@ type topologyStats struct {
...
+	// If no timestamp is given, assume the current time.
+	timestamp := time.Now()
+	if timestampStr != "" {
+		timestamp, _ = time.Parse(time.RFC3339, timestampStr)
```
fbarl: Yes, I will try to handle it somehow :)

fbarl:
> I am assuming we are going to do another iteration on the UI design for time travel?

@jpellizzari Yes! #2608 is the design follow-up issue, thanks for reminding me to reference it here.

Issue #2577, opened 2017-06-09
Connection tracking doesn't work when IPv6 is disabled


2opremio:
See https://github.com/st3v/weave-scope-release/issues/3#issuecomment-307163016

st3v:
Thanks for looking into this!

alban:
@2opremio For testing, once https://github.com/kinvolk/stage1-builder/issues/23 is done, we could use `RKT_HYPERVISOR_EXTRA_KERNEL_PARAMS=ipv6.disable=1` in semaphore.sh when `rkt` is called.

rade:
@2opremio what's left to do here?

2opremio:
We are now tolerant towards ipV6 being disabled in the `/proc` walker but not in the eBPF connection tracker.

PR #2604, opened 2017-06-15
Do not read tcp6 files if TCP version 6 isn't supported

2opremio:
This will help with #2577 when using `/proc` parsing. Fixing the eBPF connection tracker requeres a separate PR.

### DIFF
--- a/probe/endpoint/procspy/proc_linux.go
+++ b/probe/endpoint/procspy/proc_linux.go
@@ -22,7 +22,18 @@ var (
 	procRoot               = "/proc"
 	namespaceKey           = []string{"procspy", "namespaces"}
 	netNamespacePathSuffix = ""
+	ipv6IsSupported        = tcp6FileExists()
 )
+
+func tcp6FileExists() bool {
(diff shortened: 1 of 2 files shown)


probe/endpoint/procspy/proc_linux.go
```diff
@@ -22,8 +22,19 @@ var (
 	procRoot               = "/proc"
 	namespaceKey           = []string{"procspy", "namespaces"}
 	netNamespacePathSuffix = ""
+	ipv6IsSupported        = tcp6FileExists()
 )
 
+func tcp6FileExists() bool {
+	filename := filepath.Join(procRoot, "self/net/tcp6")
+	f, err := fs.Open(filename)
+	if err != nil {
```
ekimekim: Ideally we should be checking this is ENOENT, and log an error otherwise. I can imagine it being super confusing if some other crazy circumstance meant a different error and we had no clue why ipv6 conns weren't showing up.

2opremio: Being a proc file (living in the kernel's memory), I don't think we gain much by adding this refinement.

PR #2613, opened 2017-06-18
faster matching of known services

rade:
We hit this code *a lot* during rendering.

I have tested rendering performance against a recent Weave Cloud prod report. Run scope like this:
```
$ prog/scope --mode=app --weave=false --app.collector=file:/tmp/prod-on-dev-report.json
```
and then time the fetching of a topology with
```
$ time curl -so /dev/null http://localhost:4040/api/topology/<topology>
```
and then repeat for other topologies (NB: scope must be restarted in-between in order to defeat memoisation between requests).

Results (all times in seconds):

| topology   | master | branch |
| ------------ | -------- | -------- |
| processes  |   0.14 |   0.11 |
| containers |   0.92 |   0.67 |
| pods       |   0.74 |   0.48 |
| services   |   0.72 |   0.45 |
| hosts      |   0.76 |   0.47 |

So that's a nearly 40% improvement in some cases.

### DIFF
--- a/render/theinternet.go
+++ b/render/theinternet.go
@@ -3,6 +3,7 @@ package render
 import (
 	"net"
 	"regexp"
+	"strings"
 
 	"github.com/weaveworks/scope/probe/host"
 	"github.com/weaveworks/scope/report"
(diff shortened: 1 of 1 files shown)


2opremio:
Great catch @rade

PR #2616, opened 2017-06-19
Time Travel: remove the feature flag and make the availability depend on historic reports capability

### LINKED ISSUE #2607: Time Travel: make use of /api endpoint report persistence capability flag
Since #2575, we are able to tell in the UI whether Scope reports are stored persistently. Instead of treating *Time Travel* as a Cloud-only feature, we should be showing it whenever historical reports are available. Apart from Weave Cloud, that would enable the use of time travel in Scope standalone whenever the reports are stored persistently.

**Note:** This should only be addressed after the feature-flag for *Time Travel* has been removed.

fbarl:
Resolves #2607 and removes the feature flag - this will finally make Time Travel appear on production.


client/app/scripts/reducers/root.js
```diff
@@ -652,10 +653,12 @@ export function rootReducer(state = initialState, action) {
       state = state.set('errorUrl', null);
 
       return state.merge({
+        capabilities: action.capabilities,
         hostname: action.hostname,
-        version: action.version,
         plugins: action.plugins,
-        versionUpdate: action.newVersion
+        version: action.version,
+        // TODO: Why is that one never forwarded here?
```
rndstr: this was fixed in #2720

fbarl: Cool, thanks! I'll remove the comment when I get back to this PR.

client/app/scripts/selectors/time-travel.js
```diff
@@ -1,5 +1,15 @@
...
+import { isResourceViewModeSelector } from '../selectors/topology';
+
+
+export const showingTimeTravelSelector = createSelector(
+  [
+    state => state.getIn(['capabilities', 'report_persistence'], false),
+    isResourceViewModeSelector,
+  ],
+  (hasReportPersistence, isResourceViewMode) => hasReportPersistence && !isResourceViewMode
+);
```
rade: This needs updating in light of #2774

PR #2618, opened 2017-06-19
Pass build tags to unit tests

bboreham:
On master, just after `make`:
```
$ time make tests
real	1m18.575s
```

On this branch:
```
$ time make tests
real	1m1.074s
```

Note this relies on https://github.com/weaveworks/build-tools/pull/103

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ shell: $(SCOPE_BACKEND_BUILD_UPTODATE)
 	/bin/bash
 
 tests: $(SCOPE_BACKEND_BUILD_UPTODATE) $(CODECGEN_TARGETS) prog/staticui/staticui.go prog/externalui/externalui.go
-	./tools/test -no-go-get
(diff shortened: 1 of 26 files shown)


bboreham:
I realised those timings were skewed by `TestPipeRouter` taking 10 seconds longer in the second run.

This turns out to be because the mock Consul inside that test has a 10-second timeout that it may or may not hit at the end of the test.  I ran out of energy to fix that, alongside the 1-second timeout that it pretty much always hits retrying a closed connection.

Anyway, a fair comparison is 78 to 51 seconds.

rade:
I had a go at re-basing this but ran away.

rade:
@bboreham looks like your re-base didn't work :(

bboreham:
@rade I am unclear what you are referring to.  CI tests passed when I clicked 'rebuild'.

rade:
> CI tests passed when I clicked 'rebuild'.

It wasn't obvious that was going to happen.

LGTM

PR #2622, opened 2017-06-20
Ignore ipv6

rade:
There is no point in paying attention to them since scope connection tracking only deals in IPv4.

In Weave Cloud Prod, 138 of the 218 local networks in the combined report are IPv6 addresses. That slows down endpoint projection during rendering since that needs to detect whether an address is local. This PR doesn't solve that completely - matching against 80 networks is still expensive - but I have plans for that :)

### DIFF
--- a/probe/host/reporter.go
+++ b/probe/host/reporter.go
@@ -93,9 +93,8 @@ var GetLocalNetworks = func() ([]*net.IPNet, error) {
 	}
 	localNets := report.Networks{}
 	for _, addr := range addrs {
-		// Not all addrs are IPNets.
-		if ipNet, ok := addr.(*net.IPNet); ok {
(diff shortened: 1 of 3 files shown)


2opremio:
Uhm, my only reservation about this is that we should probably be moving closer to support ipv6 instead of going way from it.

Granted, it doesn't make sense to track ipv6 networks if we don't support ipv6 but maybe we should be making easy to reenable instead of removing all the code?

2opremio:
Although, thinking about it, it's just a matter of removing `ipv4Nets` so I guess that's fine

rade:
> (re-adding support for IPV5 here is) just a matter of removing `ipv4Nets`

Exactly.

PR #2623, opened 2017-06-20
Improved rendering order of nodes/edges in Graph View

fbarl:
As a first step towards making #2431 work, we need to *fix* the rendering order of elements in `<NodesChart />`. The current order *all edges -> all nodes* doesn't look that bad as long as the background nodes are kept relatively transparent, but even so, on a closer look it can be seen how background nodes are rendered on top of foreground edges. This PR introduces a new rendering order:

* blurred edges
* blurred nodes
* normal edges
* normal nodes
* highlighted edges
* highlighted nodes
* hovered edges
* hovered nodes

Since SVG 1.1 doesn't support anything like CSS `z-index`, and SVG 2.0 standard still seems far away, the SVG rendering order necessarily corresponds to the DOM order, so I had to collapse both components `<NodesChartNodes />` and `<NodesChartEdges />` into `<NodesChartElements />` to make the lists of nodes and edges intertwined.

No measurable performance loses were observed.

Now #2431 could be fixed e.g. by not making background nodes transparent, but changing their colors to similar effect.


client/app/scripts/charts/nodes-chart-elements.js
```diff
@@ -1,30 +1,236 @@
...
+      .groupBy(this.nodeDisplayLayer);
+
+    const edges = this.props.layoutEdges.toIndexedSeq()
+      .map(this.edgeHighlightedDecorator)
+      .map(this.edgeFocusedDecorator)
+      .map(this.edgeBlurredDecorator)
+      .map(this.edgeScaleDecorator)
+      .groupBy(this.edgeDisplayLayer);
+
+    // NOTE: The elements need to be arranged into a single array outside
```
jpellizzari: Curious about this. Any reference material you can send?

fbarl: I was trying to find some reference to it but couldn't :/

I figured it was true by first having the lists in the DOM and noticing that the `render` method was being invoked for too many nodes. Then I just tried making one flat list outside and it worked..

fbarl:
@jpellizzari I noticed later how the focused edges were blinking due to the CSS `fading` animation on them that was now being triggered on every hover with the changes I did. I added one more commit that fixes that by calculating edge thickness from `<EdgeContainer />` and animating it through `<Motion />` instead so PTAL at the behaviour before I merge it to make sure no new errors sneaked in.

jpellizzari:
@fbarl Looks reasonably smooth with no problems. LGTM.

PR #2625, opened 2017-06-21
fast network membership check

rade:
The rendering code checks whether endpoint IPs are part of cluster-local networks. Due to the prevalence of endpoints - medium sized reports can contain many thousands of endpoints - this is performance critical. Alas the existing code performs the check via a linear scan of a list of networks. That is slow when there are more than a few, which will be the case in the context of k8s, since there the probes register service IPs as local /32 networks.

Here we change representation of the set of networks to a prefix tree (aka trie), which is well-suited for IP network membership checks since networks are in fact a bitstring prefixes.

The specific representation is a crit-bit tree, but that choice was purely based on implementation convenience - the chosen library is the only one I could find that directly supports IP networks.

In my benchmarks this change improves topology rendering performance of Weave Cloud production reports by 3-7%.

### DIFF
--- a/render/container_test.go
+++ b/render/container_test.go
@@ -2,7 +2,6 @@ package render_test
 
 import (
 	"fmt"
-	"net"
 	"testing"
 
 	"github.com/weaveworks/common/test"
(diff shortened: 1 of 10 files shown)


2opremio:
Related: https://github.com/kubernetes/kubernetes/issues/25533

PR #2631, opened 2017-06-21
correct `nodeSummaryGroupSpec`

### LINKED ISSUE #2630: `nodeSummaryGroupSpec` looks wrong
The [`nodeSummaryGroupSpec`](https://github.com/weaveworks/scope/blob/v1.5.0/render/detailed/node.go#L130) looks wrong. AFAICT the data is (only) used to display children in the details panel. But several of the included items cannot be children of anything: Hosts, K8S Services, ECS Services.

For a moment I thought that perhaps it's a "list of all the things", but Deployments and DaemonSets aren't in it.

@2opremio @ekimekim Am I misunderstanding the purpose of this list?

rade:
make `nodeSummaryGroupSpecs` only include what's needed

Fixes #2630.

### DIFF
--- a/render/detailed/node.go
+++ b/render/detailed/node.go
@@ -126,103 +126,101 @@ func controls(r report.Report, n report.Node) []ControlInstance {
 	return []ControlInstance{}
 }
 
-var (
-	nodeSummaryGroupSpecs = []struct {
-		topologyID string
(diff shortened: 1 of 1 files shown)


rade:
best reviewed by looking at the two commits individually.

PR #2639, opened 2017-06-24
ensure connections from /proc/net/tcp{,6} get the right pid

### LINKED ISSUE #2638: connection to dead process associated with different process
Running
```
$ scope launch --weave=false --probe.ebpf.connections=false
$ tcpserver -v 0 1122 echo foo
```
and then
```
$ nc localhost 1122
```
produces
screenshot from 2017-06-24 08-36-45
i.e. the connection is erroneously associated with the chrome process.

rade:
`ProcNet.Next` does not allocate `Connection` structs, for efficiency. Instead it always returns a `*Connection` pointing to the same instance. As a result, any mutations by the caller to struct elements that aren't actually set by `ProcNet.Next`, in particular `Connection.Proc`, are carried across to subsequent calls.

This had hilarious consequences: connections referencing an inode which we hadn't come across during proc walking would be associated with the process corresponding to the last successfully looked up inode.

The fix is to clear out the garbage left over from previous calls.

Fixes #2638.

### DIFF
--- a/probe/endpoint/procspy/spy_linux.go
+++ b/probe/endpoint/procspy/spy_linux.go
@@ -28,6 +28,11 @@ func (c *pnConnIter) Next() *Connection {
 	}
 	if proc, ok := c.procs[n.Inode]; ok {
 		n.Proc = *proc
+	} else {
+		// ProcNet.Next() always returns a pointer to the same
+		// struct. We therefore must clear any garbage left over from
+		// the previous call.
+		n.Proc = Proc{}
 	}
 	return n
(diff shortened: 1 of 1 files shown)


rade:
Note that with this change, the troublesome connection from #2638 doesn't show up at all, which arguably is the correct outcome, given that one of the connection ends belongs to a short-lived process that has since vanished.

rade:
tests are failing because of #2640.

2opremio:
Really good catch

PR #2645, opened 2017-06-25
correct polarity of initial connections

### LINKED ISSUE #2644: initial connections have wrong polarity
With eBPF connection tracking, any connections that exist when scope is launched do get the wrong polarity:
```
$ scope launch
$ nc -l 1122
```
```
$ nc localhost 1122
```
screenshot from 2017-06-25 10-29-16
Notice that the highlighted process is the client, as per the command shown in the details panel, yet the connection edge points towards it, and the details panel lists the connection as inbound.

rade:
Fixes #2644.


rade:
It's best to review the individual commits rather than the combined diff.

PR #2648, opened 2017-06-26
Show multiple relatives in the nodes-grid view

### LINKED ISSUE #2627: containers-by-image table mode only shows one host
The same container image can reside (and by used) on multiple hosts. This is reflected in the containers-by-image report structure, which contains multiple parents for the host topology:
```
    "memcached;<container_image>": {
     ...
      "parents": [
        {
          "id": "ip-172-20-1-136;<host>",
          "label": "ip-172-20-1-136",
          "topologyId": "hosts"
        },
        {
          "id": "ip-172-20-2-118;<host>",
          "label": "ip-172-20-2-118",
          "topologyId": "hosts"
        },
        {
          "id": "ip-172-20-2-90;<host>",
          "label": "ip-172-20-2-90",
          "topologyId": "hosts"
        },
        {
          "id": "ip-172-20-3-251;<host>",
          "label": "ip-172-20-3-251",
          "topologyId": "hosts"
        },
        {
          "id": "ip-172-20-3-52;<host>",
          "label": "ip-172-20-3-52",
          "topologyId": "hosts"
        },
        {
          "id": "ip-172-20-3-76;<host>",
          "label": "ip-172-20-3-76",
          "topologyId": "hosts"
        }
      ],
...
```
And indeed that is what gets shown in parents area at the top of the details panel. But not in the 'hosts' column of the table view, which only shows a single host:

screenshot from 2017-06-21 15-42-58

foot:
Fixes #2627

### DIFF
--- a/client/app/scripts/components/node-details/node-details-table-row.js
+++ b/client/app/scripts/components/node-details/node-details-table-row.js
@@ -1,8 +1,6 @@
 import React from 'react';
 import classNames from 'classnames';
-import groupBy from 'lodash/groupBy';
-import forEach from 'lodash/forEach';
-import mapValues from 'lodash/mapValues';
+import { groupBy, mapValues } from 'lodash';
 import { intersperse } from '../../utils/array-utils';
 
 
@@ -11,7 +9,7 @@ import NodeDetailsTableNodeMetric from './node-details-table-node-metric';
 import { formatDataType } from '../../utils/string-utils';
 
 function getValuesForNode(node) {
-  const values = {};
+  let values = {};
   ['metrics', 'metadata'].forEach((collection) => {
     if (node[collection]) {
       node[collection].forEach((field) => {
@@ -32,9 +30,10 @@ function getValuesForNode(node) {
       relatives,
     }));
 
-    forEach(relativesByTopologyId, (columnData, topologyId) => {
-      values[topologyId] = columnData;
-    });
+    values = {
+      ...values,
+      ...relativesByTopologyId,
+    };
   }
 
   return values;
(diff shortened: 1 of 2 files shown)


rade:
@fbarl ping

client/app/scripts/utils/array-utils.js
```diff
@@ -26,3 +26,7 @@ export function moveElement(array, from, to) {
   }
   return insertElement(removeElement(array, from), to, array[from]);
 }
+
+export function intersperse(items, value) {
```
fbarl: Would be nice to create a unit test for this function and/or add a comment with an example transform above the function.

client/app/scripts/components/node-details/node-details-table-row.js
```diff
@@ -1,5 +1,10 @@
 import React from 'react';
 import classNames from 'classnames';
+import groupBy from 'lodash/groupBy';
+import forEach from 'lodash/forEach';
+import mapValues from 'lodash/mapValues';
```
fbarl: Now that we are using Webpack 2 in Scope, tree shaking should already do this for you if you go with `import { groupBy, forEach, mapValues } from 'lodash';`.

client/app/scripts/components/node-details/node-details-table-row.js
```diff
@@ -17,22 +22,28 @@ function getValuesForNode(node) {
...
+      value: relatives.map(relative => relative.label).join(', '),
       valueType: 'relatives',
-    };
-  });
+      relatives,
+    }));
+
+    forEach(relativesByTopologyId, (columnData, topologyId) => {
+      values[topologyId] = columnData;
+    });
```
fbarl: Doesn't this basically just basically merge `relativesByTopologyId` over values? Would it be any different from simply doing something like `values = {...values, ...relativesByTopologyId}`?

foot: Good catch, yep!

PR #2649, opened 2017-06-26
Honor DOCKER_* env variables in probe and app

### LINKED ISSUE #1975: Honor DOCKER_HOST in the probes
Right now we hardcode the docker endpoint to `unix:///var/run/docker.sock`

rndstr:
Changed default for flag `-app.docker` to use the DOCKER_* env variables
instead of hardcoded /var/run/docker.sock; uses docker's default if
no DOCKER_HOST defined, for both probe and app.

Fixes #1975

---
This is technically a breaking change (if someone had DOCKER_HOST set before and not provided an alternate location in `-app.docker`, it will now take a different location); albeit one I would assume impact is very low. Still, how do we handle this?

### DIFF
--- a/prog/main.go
+++ b/prog/main.go
@@ -337,7 +337,7 @@ func setupFlags(flags *flags) {
 	flag.StringVar(&flags.app.weaveAddr, "app.weave.addr", app.DefaultWeaveURL, "Address on which to contact WeaveDNS")
 	flag.StringVar(&flags.app.weaveHostname, "app.weave.hostname", "", "Hostname to advertise in WeaveDNS")
 	flag.StringVar(&flags.app.containerName, "app.container.name", app.DefaultContainerName, "Name of this container (to lookup container ID)")
-	flag.StringVar(&flags.app.dockerEndpoint, "app.docker", app.DefaultDockerEndpoint, "Location of docker endpoint (to lookup container ID)")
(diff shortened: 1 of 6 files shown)


prog/probe.go
```diff
@@ -236,7 +234,7 @@ func probeMain(flags probeFlags, targets []appclient.Target) {
 
 	if flags.weaveEnabled {
 		client := weave.NewClient(sanitize.URL("http://", 6784, "")(flags.weaveAddr))
-		weave, err := overlay.NewWeave(hostID, client, dockerEndpoint)
+		weave, err := overlay.NewWeave(hostID, client, "")
```
2opremio: Do we need this parameter at all? Can't we just remove it?

prog/main.go
```diff
@@ -394,7 +394,7 @@ func main() {
 	if flags.probe.httpListen != "" {
 		_, _, err := net.SplitHostPort(flags.probe.httpListen)
 		if err != nil {
-			log.Fatalf("Invalid value for -app.http.address: %v", err)
+			log.Fatalf("Invalid value for -probe.http.address: %v", err)
```
2opremio: Good catch :)

prog/probe.go
rndstr: Good point, I was thinking to keep `overlay.NewWeave` backwards compatible but I presume there is no need.. will remove it!

PR #2652, opened 2017-06-27
do not filter endpoints by procspied/ebpf in renderers

### LINKED ISSUE #2551: no container connections when `probe.proc.spy=false`

### LINKED ISSUE #2558: connections between local containers via host IP not shown
Scope fails to show connections between local containers when these connections go via the host IP.

e.g. launch scope and then two containers as follows
1. `docker run --rm -p 1122:1122 -ti alpine nc -lkp 1122`
2. `docker run --rm -ti alpine nc <host-ip> 1122`

and then look at the container topology in scope, which shows two containers with no connection between them.

By contrast, when the 2nd container is run on a different machine, with a scope probe running on that too, then the connection *does* show up.

rade:
The filtering of endpoints causes some connections to get missed for non-eBPF-tracked connections. Furthermore, the filtering of endpoints is entirely pointless when the probes run eBPF since the filters just pass through eBPF-tracked endpoints (for good reason too; because otherwise some connections would be missed). So in that case it is just costing CPU and removing it actually improves performance.

Note that removing the filtering does not result in over-counting connections since that is done by source ip:port pairs.

Fixes #2551.
Fixes #2558.

### DIFF
--- a/render/process.go
+++ b/render/process.go
@@ -22,7 +22,7 @@ func renderProcesses(rpt report.Report) bool {
 }
 
 // EndpointRenderer is a Renderer which produces a renderable endpoint graph.
-var EndpointRenderer = FilterProcspiedOrEBPF(SelectEndpoint)
+var EndpointRenderer = SelectEndpoint
 
 // ProcessRenderer is a Renderer which produces a renderable process
 // graph by merging the endpoint graph and the process topology.
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -227,7 +227,7 @@ func MakeRegistry() *Registry {
 		},
 		APITopologyDesc{
 			id:       containersID,
-			renderer: render.ContainerWithImageNameRenderer,
+			renderer: render.FilterUnconnectedPseudo(render.ContainerWithImageNameRenderer),
 			Name:     "Containers",
 			Rank:     2,
 			Options:  containerFilters,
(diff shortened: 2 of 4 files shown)


rade:
Performance comparison using a recent eBPF-based Weave Cloud prod report, timing the http api calls to retrieve topologies:

| topology                 | master | branch |
| ---------------- | ------ | ------ |
| processes        | 0.59 | 0.45 |
| containers       |  1.6 |  1.3 |
| pods             |  1.5 |  1.2 |
| kube-controllers |  1.5 |  1.3 |
| services         |  1.5 |  1.3 |
| hosts            |  2.0 |  1.6 |

(all times in seconds)

2opremio:
More accurate, cleaner code and it even improves performance: fantastic

render/process.go
```diff
@@ -22,7 +22,7 @@ func renderProcesses(rpt report.Report) bool {
 }
 
 // EndpointRenderer is a Renderer which produces a renderable endpoint graph.
-var EndpointRenderer = FilterProcspiedOrEBPF(SelectEndpoint)
+var EndpointRenderer = SelectEndpoint
```
2opremio: Can't we simply replace `EndpointRenderer` by `SelectEndpoint` wherever it's used?

rade: That's what I had originally, but there is a general theme in the code of having a renderer for every topology. So, for example, in
```golang
var HostRenderer = MakeReduce(
	MakeMap(
		MapEndpoint2Host,
		EndpointRenderer,
	),
	MakeMap(
		MapX2Host,
		ColorConnectedProcessRenderer,
	),
	MakeMap(
		MapX2Host,
		ContainerRenderer,
	),
	MakeMap(
		MapX2Host,
		ContainerImageRenderer,
	),
	MakeMap(
		MapX2Host,
		PodRenderer,
	),
	SelectHost,
)
```
changing `EndpointRenderer` to `SelectEndpoint` does make the latter stick out as different.

rade:
> and it even improves performance

For the eBPF case. It hurts performance for the non-eBPF case.

PR #2659, opened 2017-06-28
Use new k8s go client

2opremio:
This namely involved importing new libraries and using the new Clientset.

Changes worth mentioning:

* The new kubernetes library doesn't provide `StoreToListe`r wrappers, so now I am doing the casting directly.
* Deleting the pods and getting their logs is done in a cleaner way (using the
  Clientset instead of the lower-level RESTclient).
* `HostNetwork` is now inlined in the pod spec

After merging this PR we should probably create a new one to replace the raw [`cache.Store`](https://godoc.org/k8s.io/client-go/tools/cache#Store)s by the newer [`SharedIndexInformer`](https://godoc.org/k8s.io/client-go/tools/cache#SharedIndexInformer). This will allow us to:

* Get rid of `EventStore` (we can set event handlers directly with the Informer interface)
* Stop using raw, low-level `cache.Store`s and castings (we can create listers like [PodLister](https://godoc.org/k8s.io/client-go/listers/core/v1#PodLister) using the indexer).

TODO:
- [x] Update and successfully run unit tests
- [x] Run in dev to check nothing broke
- [x] Remove old vendored libraries


2opremio:
Rebased due to conflict in `vendor/manifest`

2opremio:
Tests started failing after I rebased:
```
>>> Test ./330_process_edge_test.sh finished after 178.0 secs with error: exit status 1
Error response from daemon: failed deleting network: legacy plugin: Post http://%2Frun%2Fdocker%2Fplugins%2Fweavemesh.sock/Plugin.Activate: dial unix /run/docker/plugins/weavemesh.sock: connect: connection refused
```

ekimekim:
It's failed with 3 different error messages across 3 runs. And I can't run the integration tests locally, apparently they use GCE and I need creds.

rade:
> And I can't run the integration tests locally, apparently they use GCE and I need creds.

You should be able to run the tests against local VMs. If that is not working, please raise an issue and fix it.

2opremio:
Rebased again to see if the error was transient (no other PR seems to be suffering from this)

Issue #2620, opened 2017-06-19
combined, comprehensive kubernetes views


rade:
This is an umbrella issue:

- [x] #1444 add DaemonSets (released in 1.3)
- [x] #2239 add StatefulSets
- [x] #2571 add CronJobs
- [x] #2552 combined view of k8s controllers
- [x] ~enable searching/filtering by controller type when in combined view~~
- [x] #2670 get rid of type selector
- [x] #2661 get rid of ReplicaSets (at least from the UI)

rade:
![](https://github.trello.services/images/mini-trello-icon.png) [Scope Kubernetes View](https://trello.com/c/iJzYERG7/10-scope-kubernetes-view)

jpellizzari:
> enable searching/filtering by controller type when in combined view, and get rid of type selector

The current convention dictates that we use the search term `controller_type:daemonset` or something similar. Maybe `node_type:`? `type:`?

jpellizzari:
Whoops accidentally closed

rade:
> The current convention dictates that we use the search term `controller_type:daemonset` or something similar. Maybe `node_type:`? `type:`?

Let @ekimekim decide.

jpellizzari:
I can already search by node type since its part of the metadata fields:
![screen shot 2017-06-29 at 3 31 34 pm](https://user-images.githubusercontent.com/2802257/27713371-3dbb979e-5ce0-11e7-9987-46336df41ce1.png)

In fact, that is one of the examples. @ekimekim Maybe I am missing the requirement here?

ekimekim:
No, it was me and @rade missing that this was already possible.

ekimekim:
Referencing here for visibility: https://github.com/weaveworks/scope/issues/2693
Feature request to have a Volumes topology that connects to the pods/containers which access that volume. Shouldn't be too hard, and seems useful.

rade:
> Shouldn't be too hard,

It's harder than you might think. See my comment in the issue.

PR #2661, opened 2017-06-29
Remove replica sets

ekimekim:
Use a special kind of selector renderer to elide replica sets from pod nodes
and directly reference deployment parents instead.
Do the inverse (replace replica sets with pods) during the mapping from pod to deployment.

Note we can no longer use renderParents since we're using a non-standard Selector


ekimekim:
This approach is a little hacky, but is the best we can do without changing the probe, which won't work retroactively. One other option would be changing the probe, and then doing this re-mapping to remove replica sets at `report.Upgrade()`.

rade:
> One other option would be changing the probe, and then doing this re-mapping to remove replica sets at `report.Upgrade()`.

Ultimately that is the correct way to go, but only once we are sure we won't change our minds about not representing replicasets. I have raised #2662 for this.

rade:
FTR, this PR is part of #2620.

rade:
PS: please make the linter happy

ekimekim:
note: though github claims '1 of 2 test passed', it was netlify that was failing, which doesn't even test anything afaict.

PR #2664, opened 2017-06-29
Link scope-ui graphs clickable to prometheus queries

rndstr:
scope-app:
- Adds `-app.metrics-graph` cli flag for configuring the base url to
  use for graph links; supports `:orgID` and `:query` placeholders
- Assigns query URLs to existing metrics and creates an empty metric if it's missing 

scope-ui:
- Extends `<CloudFeature />` with option `alwaysShow`
- Adds `<CloudLink />` to simplify routing when in cloud vs not in cloud
- Links metric graphs in the ui's node details view for all k8s
  toplogies and containers so far

---
**TODO**
- [x] receive design suggestions
- [x] deal with overflow items (remove dead code in case not in use)
- [x] verify naming of pods in k8s for queries
- [x] ~integration tests~ (postponed)
- [x] lint errors

---

As long as there is no `--app.metrics-graph` configured, this features is invisible.

**Screenshots**
scope-metric-links-hover
scope-metric-links-overflow
scope-metric-links-tableview


prog/main.go
```diff
@@ -353,6 +354,7 @@ func setupFlags(flags *flags) {
 	flag.IntVar(&flags.app.memcachedCompressionLevel, "app.memcached.compression", gzip.DefaultCompression, "How much to compress reports stored in memcached.")
 	flag.StringVar(&flags.app.userIDHeader, "app.userid.header", "", "HTTP header to use as userid")
 	flag.BoolVar(&flags.app.externalUI, "app.externalUI", false, "Point to externally hosted static UI assets")
+	flag.StringVar(&flags.app.metricsGraphURL, "app.metrics-graph", "", "Enable extended metrics graph by providing a templated URL (supports :orgID and :query)")
```
2opremio: I would include an example URL

render/detailed/links.go
```diff
@@ -0,0 +1,167 @@
...
+func queryParamsAsJSON(query string) (string, error) {
+	type cell struct {
+		Queries []string `json:"queries"`
+	}
+	type queryParams struct {
+		Cells []cell `json:"cells"`
+	}
+	params := &queryParams{[]cell{{[]string{query}}}}
+
+	bs, err := json.Marshal(params)
```
2opremio: We use the ugorji library for marshaling in most places (if not all) since it's faster.

2opremio:
Please fix the linting errors.

bboreham:
I note that #2620 anticipates some further additions, e.g. StatefulSet/PetSet.  Should we also anticipate them here?

render/detailed/links.go
```diff
@@ -0,0 +1,169 @@
...
+	// Available metric links
+	linkTemplates = []MetricLink{
+		{ID: docker.CPUTotalUsage, Label: "CPU", Priority: 1},
+		{ID: docker.MemoryUsage, Label: "Memory", Priority: 2},
+	}
+
+	// Prometheus queries for topologies
+	topologyQueries = map[string]map[string]*template.Template{
+		report.Pod: {
+			docker.MemoryUsage:   prepareTemplate(`sum(container_memory_usage_bytes{pod_name="{{.Label}}"})`),
```
bboreham: I think there should be some explanation here about where these rules come from: are they assuming some particular exporter is in place?  Are they assuming anything in forwarding or recording rules?

render/detailed/links.go
```diff
@@ -0,0 +1,169 @@
...
+	"strings"
+	"text/template"
+
+	"github.com/weaveworks/scope/probe/docker"
+	"github.com/weaveworks/scope/report"
+
+	"github.com/ugorji/go/codec"
+)
+
+// MetricLink describes a link referencing a metric.
```
bboreham: This comment didn't help me much.  Maybe it could say that "link" is something like "URL"?

render/detailed/links.go
```diff
@@ -0,0 +1,169 @@
...
+	URL      string `json:"url"`
+	Priority int    `json:"priority"`
+}
+
+// Variable name for the query within the metrics graph url
+const urlQueryVarName = ":query"
+
+var (
+	// As configured by the user
+	metricsGraphURL = ""
```
bboreham: Please, no singletons.

rndstr: I talked about this with @2opremio; I first went the route of _passing it down_ which lead to exponential increase of function signature adjustments (I think I stopped when I crossed 10 and saw there is no end in sight).

I'll take another look at it and am happy to receive alternative suggestions in the meantime.

render/detailed/links.go
rndstr: Since "link" seems to be easily confused, do you think it should be renamed to something like `MetricURL` or `MetricHref`?

render/detailed/links.go
rndstr: What do you mean by «forwarding rule»?

render/detailed/links.go
rndstr: @bboreham I took the plunge in https://github.com/weaveworks/scope/pull/2664/commits/b43dfd274f97eb57cb4d8cf8a0def3dca51af730 feedback welcome

render/detailed/links.go
bboreham: I guess I used the wrong term - there is a [relabel_config](https://prometheus.io/docs/operating/configuration/#<relabel_config>) applied within [remote_write](https://prometheus.io/docs/operating/configuration/#<remote_write>)

client/app/scripts/components/node-details/node-details-health-link-item.js
```diff
@@ -0,0 +1,59 @@
...
+    this.setState({hovered: true});
+  }
+
+  onMouseOut() {
+    this.setState({hovered: false});
+  }
+
+  onClick() {
+    trackMixpanelEvent('scope.node.metric.click', {
+      layout: GRAPH_VIEW_MODE,
```
davkal: The name seems off, maybe `scope.node.details.metric.click`?

Also, the layout could be either table, graph, or resource view, no?

client/app/scripts/components/node-details/node-details-health.js
```diff
@@ -20,7 +21,12 @@ export default class NodeDetailsHealth extends React.Component {
   }
 
   render() {
-    const metrics = this.props.metrics || [];
+    const {
+      metrics = makeList(),
```
davkal: I'd be worried that you're mixing immutablejs and JS array functions further down. E.g., immutablejs prefers `.size` to check list length.

client/app/scripts/components/node-details/node-details-table-node-metric-link.js
```diff
@@ -0,0 +1,50 @@
...
+class NodeDetailsTableNodeMetricLink extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.onClick = this.onClick.bind(this);
+  }
+
+  onClick() {
+    trackMixpanelEvent('scope.node.metric.click', {
+      layout: TABLE_VIEW_MODE,
```
davkal: `layout: TABLE_VIEW_MODE` refers to the overall layout for the scope canvas, as opposed to the table inside the details panel

client/app/scripts/components/node-details/node-details-table-node-metric-link.js
```diff
@@ -0,0 +1,50 @@
...
+      topologyId: this.props.topologyId,
+    });
+  }
+
+  static dismissEvent(ev) {
```
davkal: This could just be a module level function.

client/app/scripts/components/node-details/node-details-table.js
```diff
@@ -109,7 +109,7 @@ function getSortedNodes(nodes, sortedByHeader, sortedDesc) {
   const getValue = getValueForSortedBy(sortedByHeader);
   const withAndWithoutValues = groupBy(nodes, (n) => {
     const v = getValue(n);
-    return v !== null && v !== undefined ? 'withValues' : 'withoutValues';
+    return !n.valueEmpty && v !== null && v !== undefined ? 'withValues' : 'withoutValues';
```
davkal: Living dangerously checking props of `n` here first, before checking `null`-ness.

client/app/scripts/utils/color-utils.js
```diff
@@ -82,3 +82,13 @@ export function brightenColor(c) {
   }
   return color.toString();
 }
+
+export function darkenColor(c) {
+  let color = hsl(c);
+  if (hsl.l < 0.5) {
```
davkal: Did you play around to find that threshold?

client/package.json
```diff
@@ -12,17 +12,18 @@
...
     "d3-selection": "1.0.5",
     "d3-shape": "1.0.6",
     "d3-time-format": "2.0.5",
     "d3-transition": "1.0.4",
-    "d3-drag": "1.0.4",
     "d3-zoom": "1.1.4",
     "dagre": "0.7.4",
     "debug": "2.6.6",
     "filesize": "3.5.9",
+    "filter-invalid-dom-props": "^2.0.0",
```
davkal: Should use `--save-exact`.

client/app/scripts/components/node-details.js
```diff
@@ -192,7 +192,11 @@ class NodeDetails extends React.Component {
         <div className="node-details-content">
           {details.metrics && <div className="node-details-content-section">
             <div className="node-details-content-section-header">Status</div>
-            <NodeDetailsHealth metrics={details.metrics} />
+            <NodeDetailsHealth
+              metrics={details.metrics}
+              topologyId={topologyId}
+              nodeColor={nodeColor}
```
davkal: I would not use the node color to draw attention here. The node color distinguishes a node from other nodes. Here we merely want to highlight availability of more details. That should be distinct in its decoration.

render/detailed/links.go
2opremio: That took a bit of work, it looks good.

render/detailed/links.go
```diff
@@ -0,0 +1,230 @@
...
+			continue
+		}
+
+		tpl := queries[metadata.ID]
+		if tpl == nil {
+			continue
+		}
+
+		bs.Reset()
+		if err := tpl.Execute(&bs, summary); err != nil {
```
2opremio: Isn't using templating a bit overkill for a simple query/replace? It's probably fine as long as it doesn't have a noticeable performance impact.

render/detailed/links.go
```diff
@@ -0,0 +1,230 @@
...
+	if strings.Contains(metricsGraphURL, urlQueryVarName) {
+		return strings.Replace(metricsGraphURL, urlQueryVarName, url.QueryEscape(query), -1)
+	}
+
+	params, err := queryParamsAsJSON(query)
+	if err != nil {
+		return ""
+	}
+
+	if metricsGraphURL[len(metricsGraphURL)-1] != '/' {
```
2opremio: Is `len(metricsGraphURL)` > 0 in all cases?

render/detailed/links.go
rndstr: @2opremio I reverted this and created #2740, Bryan mentioned he had a few thoughts on improving it, maybe we can continue the discussion there

client/app/scripts/components/node-details/node-details-health-link-item.js
rndstr: I was deriving the name from the `scope.node.relative.click` that's in use for a similar purpose. 

The other table mixpanel event `scope.node.metric.click`  is sometimes in the table view and sometimes in the details view. So, technically, we would need to switch there to `scope.node.details.metric.click` if it's within the details. Maybe we can add a property that tells the immediate context/container we are in instead? That would also help to answer the query «do people click on metric links anywhere» w/o the need to combine multiple events.

`'scope.node.metric.click', { context: 'node-details' }`
`'scope.node.metric.click', { context: 'main' }`

In case we want to differentiate further ("it was a table", "it was the status metric section") we can just add more properties. I presume these are filterable in mixpanel. Wdyt?

> Also, the layout could be either table, graph, or resource view, no?

Indeed, will remove.

client/app/scripts/components/node-details/node-details-table-node-metric-link.js
rndstr: Right, I was first under the impression this render path is only in use for the table view but it's not, good DRY, also removing it.

client/app/scripts/components/node-details/node-details-table.js
rndstr: To be fair, the `getValue` above would've failed before it reached this line. I modified the whole block to abort early if that's the case.

client/app/scripts/utils/color-utils.js
rndstr: Nope, blatantly copied this from the `brightenColor()` version. It made sense to me, if it's already on the darker/brighter side, go a bit slower towrads black/white.

client/package.json
rndstr: Good catch!

client/app/scripts/components/node-details.js
rndstr: I went back to the `btn-opacity` which is not as visible but the cursor pointer should be enough once we are hovering.

render/detailed/links.go
rndstr: Good point, I started off thinking there might be other parts in there that will be used but it now looks like we only care about the `label` and nothing else. Will take a look

render/detailed/links.go
rndstr: Yes, we bail early in https://github.com/weaveworks/scope/pull/2664/files#diff-7a21e593fb345568b7d0bc0774375c97R112

client/app/scripts/components/node-details/node-details-health.js
rndstr: Good catch!

client/app/styles/_base.scss
```diff
@@ -12,6 +12,10 @@
 // TODO: Remove this line once Service UI CONFIGURE button stops being added to Scope.
 .scope-wrapper .setup-nav-button { display: none; }
 
+a {
+  text-decoration: none;
```
rndstr: Note that this is scope-wide and brings it on par with the defaults of service-ui.

rndstr:
@2opremio @davkal @bboreham PTAL?

davkal:
Latest JS changes LGTM. Cant wait to see it on dev!

PR #2666, opened 2017-06-29
Prevents 6 digit pids from being truncated in details panel/table mode

### LINKED ISSUE #2660: 6-digit pids truncated in details panel
I think we assumed that pids would be at most 5 digits. Wrong.
screenshot from 2017-06-28 22-09-00

(The pids in question are 108063 and 107575)

foot:
Fixes #2660

### DIFF
--- a/client/app/scripts/constants/styles.js
+++ b/client/app/scripts/constants/styles.js
@@ -46,7 +46,8 @@ export const CANVAS_MARGINS = {
 // Node details table constants
 export const NODE_DETAILS_TABLE_CW = {
   XS: '32px',
-  S: '50px',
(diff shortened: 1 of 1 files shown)


foot:
Thanks @fbarl! I think it might be linux too ~_~.

56 it is.

rndstr:
FWIW removing the `truncate` would also make sure it never cuts it off but just mis-align it in case it doesn't fit. Haven't looked closely at the code, might be more work than anticipated to selectively remove `truncate` from columns. (my test was on Linux btw)

foot:
@rndstr Thats an interesting thought actually, will have a think.

The `...` is a nice "fail fast" soln which kinda highlights something's wrong with our width expectations too.

Issue #2665, opened 2017-06-29
connections attributed to wrong process when ephemeral ports are re-used quickly


rade:
Let's say the probes on two hosts detect the following sequence of events

1. on host1 we see p1 connect from e1 to e2
2. on host2 we see p2 accept a connection on e2 from e1
3. on host1 we see p3 connect from e1 to e3

Now consider what edges we show in the process topology for possible combined reports:

a) 1+2 - edge from p1 to p2 - good
b) 1+3 - no edge - ok (though see notes below)
c) 2+3 - edge from p1 to p3 - bad
d) 1+2+3 - edge from p1 to p3 - bad

In case b we end up with e1 having a pid of p3 and edges to both e2 and e3. This is wrong, but causes no harm unless e2 is mapped to a display node, which isn't happening here.

rade:
This is a fundamental problem with the representation of connections in reports. A connection from addr1:port1 to addr2:port2 is represented as two endpoints, with the first having an adjacency record referencing the latter. Each endpoint may have at most one pid, with timestamp-based merge semantics applied to resolve conflicts. And adjacency sets are simply merged.

Such a representation is incapable of correctly representing two connections from the same source port but different processes, or the same destination port but different processes.

rade:
It would be ok, though not ideal for us to
- not show an edge in cases c+d
- only show one (correct) edge in case d (in the example that's the best that can happen anyway, since i didn't include a 4th event detecting the acceptance of the 2nd connection, but even if there was such an event, it would be ok to only show the first or 2nd edge)

on the basis that in general we don't expect scope to detect all connections.

rade:
#2668 ameliorates this by not showing edges when their source endpoint has more than one edge.

PR #2668, opened 2017-06-29
ignore endpoints with >1 adjacency in process rendering

rade:
This eliminates the worst effects of #2665.

It is imperfect because it
- simply drops all connections that originate from the same endpoint, if there is more than one. Not only will this eliminate the problematic connections (i.e. those where it is indeed the case that the same source ip+port was used by different processes), but also those where the source ip+port was used multiple times by the same process. That is unlikely for ephemeral ports (the odds of the *same* process getting given a particular ephemeral port more than once in a short period of time are much lower than the odds of *any two* processes getting given a particular ephemeral port in a short period of time, which is the problematic case we are wanting to eliminate here), but a dead certainty when the process is specifically choosing a source port, which is unusual but possible.
- does not eliminate incorrect process attribution when *destination* ip+ports are used by different processes. The most common case for that is pre-forking in the likes of apache. In that scenario, all the processes are "siblings". Incorrect attribution is not that bad here; yes, things will be wrong at the process level, but not in a "talking to some completely random process" sort of way, and higher level topologies are unaffected.

I have checked the effects of this change on recent Weave Cloud dev topologies, and it makes them look sane, while not obviously removing any edges that should be there.

### DIFF
--- a/render/process.go
+++ b/render/process.go
@@ -123,6 +123,13 @@ func MapEndpoint2Process(n report.Node, local report.Networks) report.Nodes {
 		return report.Nodes{}
 	}
 
+	if len(n.Adjacency) > 1 {
+		// We cannot be sure that the pid is associated with all the
+		// connections. It is better to drop such an endpoint than
+		// risk rendering bogus connections.
+		return report.Nodes{}
+	}
+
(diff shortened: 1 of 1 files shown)


2opremio:
I don't like this solution, but I also don't have a better suggestion, so merge away.

PR #2674, opened 2017-07-03
Add non-eBPF variants of tests 330 and 340

2opremio:
Leftover from #2535 

Also, print debug information since they are flakey.

### DIFF
--- a/integration/330_process_edge_test.sh
+++ b/integration/330_process_edge_test.sh
@@ -19,5 +19,7 @@ wait_for processes "$HOST1" 60 "nginx: worker process" nc
 has processes "$HOST1" "nginx: worker process"
 has processes "$HOST1" nc
 has_connection processes "$HOST1" nc "nginx: worker process"
+# Print connections in case of test failure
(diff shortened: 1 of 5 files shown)


rade:
Perhaps we should have two variants - ebpf and non-ebpf - of these tests, as we have for some others (300/301, 310/311).

rade:
Could we grab the full report in these tests and make it available as a test artefact? That way we can investigate whether it is missing individual endpoints, processes, etc.

2opremio:
> Perhaps we should have two variants - ebpf and non-ebpf - of these tests, as we have for some 

Sure.

> Could we grab the full report in these tests and make it available as a test artefact?

Yep, I just reworked this PR to do that.

rade:
> Yep, I just reworked this PR to do that.

Dumping the report on stdout isn't quite what I had in mind. I was suggesting to add the report to the test artifacts, just as we do for coverage data in the Weave Net CI. See https://circleci.com/gh/weaveworks/weave/9335#artifacts/containers/0 for an example.

PR #2678, opened 2017-07-04
Make graph layout zoom limits constant

### LINKED ISSUE #2642: max zoom level varies by window height
The maximum zoom level depends on the window height, i.e. one can zoom further when the window is taller.

max zoom in short window:
screenshot from 2017-06-25 10-16-23

max zoom in tall window:
screenshot from 2017-06-25 10-17-29

fbarl:
Fixes #2642.


foot:
You can zoom a tiny bit too far out perhaps.

But otherwise LGTM!

fbarl:
> You can zoom a tiny bit too far out perhaps.

Yeah, I increased minimal nodes size from `1px` to `3px`, should be better now :)

p.s. I didn't wanna increase it even more because of the very wide layout we sometimes get for `Processes` topology.

PR #2679, opened 2017-07-04
Make Resource view nodes clickable

### LINKED ISSUE #2663: Resource view - make nodes clickable
When clicking on a node in the resource view, *Node Details* panel should open up, like it does in the other two views. Noticed by @2opremio (from Slack conversation):

> I think it’s a very important feature
the view is crippled without being able to see details
(once I identify something odd I want to be able to drill down on the cause)

fbarl:
Resolves #2663 by making resource view nodes *selectable*. Ideally, we would also zoom into the selected node like we do in the graph view, but the animation there is a bit more involved, so it shouldn't be a blocker for this action that feels very basic.

### DIFF
--- a/client/app/scripts/actions/app-actions.js
+++ b/client/app/scripts/actions/app-actions.js
@@ -323,13 +323,14 @@ export function setResourceView() {
   };
 }
 
-export function clickNode(nodeId, label, origin) {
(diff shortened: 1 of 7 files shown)


fbarl:
@davkal ping

rndstr:
I played with it and it works well!

PR #2680, opened 2017-07-05
correct determination of a host's container images

### LINKED ISSUE #2629: container image list in host details panel is incomplete
The list of container images in the host details panel is incomplete, e.g. notice how in
screenshot from 2017-06-21 16-29-36
only billing images are listed yet there are plenty of non-billing containers running.

Furthermore, the list keeps changing.

I've looked at the API response from a statically served captured report, and it is incomplete, so this is a b/e problem.

rade:
The same container image may be in use on multiple hosts. The latest-map `HostNodeID` of a node in the `ContainerImage` topology is therefore meaningless - it gets set to whatever host reported that image last in the time window covered by the report. Hence using it as a basis for mapping to hosts, as we did, a) fails to associate images with all the right hosts (hence they are missing from the host details panel), and b) causes that association to change semi-randomly (hence the list in the details panel is unstable).

By contrast, the host topology parents of container image nodes is the complete set of all hosts the image is used on. So making that the basis for mapping fixes the problem.

The same technique - mapping based on host parents rather than latest-map `HostNodeID` - also works for the other nodes we are mapping: processes, containers and pods. So we might as well use it there
too. That's also what `Map2Parent` does in other topology renderers.

Fixes #2629.

### DIFF
--- a/render/host.go
+++ b/render/host.go
@@ -35,25 +35,31 @@ var HostRenderer = MakeReduce(
 // If this function is given a node without a hostname
 // (including other pseudo nodes), it will drop the node.
 //
-// Otherwise, this function will produce a node with the correct ID
+// Otherwise, this function will produce nodes with the correct ID
 // format for a host, but without any Major or Minor labels.  It does
 // not have enough info to do that, and the resulting graph must be
 // merged with a host graph to get that info.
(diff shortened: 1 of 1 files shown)


rade:
> I'm sorely tempted to refactor this into Map2Parent

Believe me, I tried.

PR #2683, opened 2017-07-05
Show table overflow only if limit exceeded by 2+

rndstr:
Having to toggle a +1 feels weird while that +1 takes up
the same vertical space as displaying that row.

---
Just something small I came across.

Behavior on default (or configured) row limit equals 5:

Items | Display,Overflow | (Alternative)
-|-|-
4|4,0|4,0
5|5,0|5,0
6|6,0|4,+2
7|5,+2|4,+3
8|5,+3|4,+4

Current implementation shows generally $limit rows, alternative would be if you want a max row (including toggle) to not exceed $limit.

### DIFF
--- a/client/app/scripts/components/node-details/node-details-table.js
+++ b/client/app/scripts/components/node-details/node-details-table.js
@@ -132,7 +132,7 @@ class NodeDetailsTable extends React.Component {
     super(props, context);
 
     this.state = {
-      limit: props.limit || NODE_DETAILS_DATA_ROWS_DEFAULT_LIMIT,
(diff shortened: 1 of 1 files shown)


rndstr:
@foot thanks! I missed that `notShown` should've been adjusted as well.

(I now also stumbled over a bug that if you pass a non-default props value for `limit`, after you toggle twice it would return back to default instead of the value you passed in initially.)

foot:
Yeah good catch w/ the default prop.

Looks good now and works really nicely! LGTM

PR #2688, opened 2017-07-06
Last line of defense against overlapping nodes in graph layout

### LINKED ISSUE #2261: External connections can lay on top of other nodes
It is possible for external connections to lay on top of other nodes:
screenshot from 2017-02-19 19-00-57

This appears to happen when a new connection to an external server (a cloud-shape node) pops into the graph. Forcing a re-layout corrects the problem.

fbarl:
Fixes #2261 by introducing an additional test at the end of the layout function to see if any the nodes would overlap - if they would, the whole layout is refreshed by calling the render engine with a clean cache.

*Note*: The `layoutSingleNodes` has just been moved to the top, nothing was changed there.

### DIFF
--- a/client/app/scripts/charts/__tests__/node-layout-test.js
+++ b/client/app/scripts/charts/__tests__/node-layout-test.js
@@ -407,4 +407,35 @@ describe('NodesLayout', () => {
     expect(nodes.n6.x).toBeGreaterThan(nodes.n3.x);
     expect(nodes.n6.y).toEqual(nodes.n3.y);
   });
+
+  it('rerenders the nodes completely after the coordinates have been messed up', () => {
+    // Take an initial setting
+    let result = NodesLayout.doLayout(
+      nodeSets.rank4.nodes,
+      nodeSets.rank4.edges,
+    );
+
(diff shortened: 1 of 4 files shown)


fbarl:
> If this introduces too many re-layouts, you could try another approach: keep track of nodes that are too close, and only modify their position, and then check again recursively.

In the ideal case, this *last line of defense* will never be executed - additional re-layout will be triggered **only** if the bug is imminent, otherwise nothing changes (minus the check for overlaps which is rather lightweight and could be made even faster).

So the point of this PR is to prevent overlap bugs above all and give us more time to come up with a smarter solution. Any attempts to introduce another smart heuristics vs. brute force failsafe approach would decrease the chance of the final layout being bug-free :)

davkal:
> In the ideal case, this last line of defense will never be executed - additional re-layout will be triggered only if the bug is imminent

If it helped you, you could track those cases with mixpanel.

fbarl:
> If it helped you, you could track those cases with mixpanel.

I have a good idea *why* they happen, but tracking should help us identify *how often* they happen. Thanks for the tip!

PR #2692, opened 2017-07-06
Set package.json version to 0.0.0

rndstr:
The package version is irrelevant for the build process
and is not read anywhere.

The package is not published and causes confusion if the
bump is forgotten.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "weave-scope",
-  "version": "1.5.1",
+  "version": "0.0.0",
   "description": "SPA JS app for Weave Scope visualising the application network.",
   "repository": "weaveworks/scope",
   "license": "Apache-2.0",


jpellizzari:
We do install Scope as an npm module in service-ui. This change shouldn't affect, but a test wouldn't hurt.

rndstr:
Thanks Jordan (it did work when I tried it locally)

fbarl:
I'm afraid this PR have screwed something with `yarn.lock` in Service UI. It was only updated in https://github.com/weaveworks/service-ui/commit/89dd0d3ee08d6c1991262b5178fe2f599a35d062, but that commit didn't really update the Scope version in the UI (even though the right version was shown in the footer). Still not sure it has to do with this PR though, but maybe we should see if reverting fixes it. :/

PR #2694, opened 2017-07-07
Fix `yarn pack` ignoring directory cli flag

rndstr:


rade:
So what's currently broken that this fixes?

rade:
> So what's currently broken that this fixes?

Ignore me; just figured it out.

PR #2699, opened 2017-07-10
EbpfTracker refactoring / cleanup

rade:
a bunch of refactors, and race condition fixes.

Best reviewed commit-by-commit.

### DIFF
--- a/probe/endpoint/connection_tracker.go
+++ b/probe/endpoint/connection_tracker.go
@@ -114,6 +114,21 @@ func (t *connectionTracker) performFlowWalk(rpt *report.Report) map[string]fourT
 	return seenTuples
 }
 
+func (t *connectionTracker) existingFlows() map[string]fourTuple {
+	seenTuples := map[string]fourTuple{}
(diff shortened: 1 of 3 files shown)


probe/endpoint/connection_tracker.go
```diff
@@ -109,6 +111,24 @@ func (t *connectionTracker) performFlowWalk(rpt *report.Report, seenTuples map[s
 		seenTuples[tuple.key()] = tuple
 		t.addConnection(rpt, tuple, "", extraNodeInfo, extraNodeInfo)
 	})
+	return seenTuples
+}
+
+func (t *connectionTracker) existingFlows() map[string]fourTuple {
+	seenTuples := map[string]fourTuple{}
+	if !t.conf.UseConntrack {
+		log.Warnf("Not using conntrack: disabled")
```
2opremio: I don't think this warning is really useful (particularly without further context on the impact)

rade: Fair enough. Will remove.

It turns out this change actually fixed a logic bug too: Previously, it would run conntrack here even when the flag wasn't set - all the flag did was suppress the "not supported by the kernel" warning.

Issue #2677, opened 2017-07-04
Time Travel: Make slider into timeline


fbarl:
In order to display different historic metadata in the *Time Travel* component (which would give better clues as to where we want to navigate), current slider will be made into a timeline which would be holding the metadata. The idea is to make that timeline both *zoomable* and *pannable*.

fbarl:
@bia is currently working on the sketches for it - the very initial ones start from [here](https://www.dropbox.com/sh/45hq01047u7zfir/AABQugkx5FjhOOpIGilBB0pBa?dl=0&preview=170628timelineSketch9.png).

rade:
zoom and pan are important features in their own right, and quite independent of representing historic meta data on the timeline. So let's tackle that *first*. I've raised #2686 for that.

This issue is therefore only about representing events on the timeline.

fbarl:
Some comments @foot made in #2651 (and still need to be considered):

> * I got stressed by the "busy" cursor that appeared as soon as I started dragging. Maybe show that onMouseUp when stuff starts happening.
> * Not a biggie but I'd prefer a "Go!" button next to the textfield, if I'm using it I'm probably making a couple of changes and it feels a bit hectic things changing around as you're trying to enter a value.
timestamp is truncated a tiny bit for me :).
> * Not sure about "live", but can't think of anything else right now.
> * Show the date next to the mouse pointer when dragging? Bit of eye flicking there.

fbarl:
Addressed by #2703.

Issue #2686, opened 2017-07-06
Time Travel: zoom and pan for easy navigation in time


rade:
Zooming and panning are well understood control operations for exploring/navigating a space, e.g. on online maps. The same controls work for exploring/navigating in the time dimension.

So imagine a timeline spanning the width of the screen. The mid-point represents the point in time for which scope shows data. The user can
- click-and-drag the timeline to move backward and forward
- hover over any point and use the scroll wheel to expand/contract the window represented by the timeline, with the zoom operation centered around the point under the mouse
- alternatively, use a slider to control zoom around the center point - important for mouse-less operation
- (maybe) click to re-center the timeline of the point under the mouse

fbarl:
This is pretty much the direction where we are already going with the [mockups](https://www.dropbox.com/sh/45hq01047u7zfir/AABQugkx5FjhOOpIGilBB0pBa?dl=0&preview=170705timelineSketch11.png) @bia is working on.

Actually, all the points you mentioned above will be addressed with the new design except maybe the 3rd one (slider for zoom control). The timeline would be zoomable when scrolling over it but ideally we'd want to have an additional way to control the zoom which doesn't look like another control. I'd say this is the main point we'd have to figure for the new design, how to make zooming obvious without ruining the simplicity of the *Time Travel* component.

rade:
> The timeline would be zoomable when scrolling over it but ideally we'd want to have an additional way to control the zoom which doesn't look like another control. 

Exactly. Because we know from experience with the Scope main zoom that a good proportion of users do not realise that the scroll wheel zooms. Which is why we added the little slider.

Issue #1632, opened 2016-07-04
Mode, or overlay view filter navigation in bottom LHS is becoming cluttered.


pidster:
As we add more ways to overlay data and filter the contents of a view, the nav in the bottom LHS is becoming cluttered.

pidster:
See also: https://github.com/weaveworks/scope/issues/1369

Issue #1369, opened 2016-04-21
Bottom right of UI is getting a little crowded


tomwilkie:
With #1366, #1365 and various other PRs, the bottom right of the UI is starting to get a little crowded:

![screen shot 2016-04-21 at 19 13 00](https://cloud.githubusercontent.com/assets/859729/14718205/8dcf1314-07f5-11e6-8a61-ef860f9cc954.png)

Somewhat related is I'd like somewhere to report all the probe versions (#1322).

Perhaps this whole area is worth a redesign?

pidster:
Perhaps the "update available" notice could go top-right; this is a location frequently used for notifications.

PR #2702, opened 2017-07-11
[WIP] Minor repositioning of controls

fbarl:
Moved around some controls as preparation for #2677 (see [latest mockups](https://www.dropbox.com/sh/45hq01047u7zfir/AABQugkx5FjhOOpIGilBB0pBa?dl=0&preview=170710timelineSketch2.png)) and addressed https://github.com/weaveworks/service-ui/issues/746 in the context of Scope.

##### Semantic grouping

* *Topologies* - top left
* *View selectors* (view mode, resource metrics) - top center
* *Time control* - top right
* *Filters* (topology options, search) - bottom left
* *Footer* (version, help, misc controls, zooming) - bottom right

##### Specific changes

* Moved the *Search bar* to the bottom left corner as it belongs to filters semantically.
* Moved the *Scope logo* down to the footer (next to version) and made it much smaller.
* Moved *Topologies* selector to the freed up space on the left and split it from the *View mode* selector.
* Centered *View mode* selector horizontally.
* Hidding the *CONFIGURE* button for Scope.

##### Notes

* We should consider moving the *Networks* button together with the view selectors (as is the case with resource metrics).

### DIFF
--- a/client/app/scripts/components/app.js
+++ b/client/app/scripts/components/app.js
@@ -4,7 +4,6 @@ import classNames from 'classnames';
 import { connect } from 'react-redux';
 import { debounce } from 'lodash';
 
-import Logo from './logo';
 import Footer from './footer';
 import Sidebar from './sidebar';
 import HelpPanel from './help-panel';
@@ -172,7 +171,6 @@ class App extends React.Component {
       timeTravelTransitioning, showingTimeTravel } = this.props;
 
     const className = classNames('scope-app', { 'time-travel-open': showingTimeTravel });
-    const isIframe = window !== window.top;
 
     return (
       <div className={className} ref={this.saveAppRef}>
@@ -187,12 +185,6 @@ class App extends React.Component {
         <div className="header">
           <TimeTravel />
           <div className="selectors">
-            <div className="logo">
-              {!isIframe && <svg width="100%" height="100%" viewBox="0 0 1089 217">
-                <Logo />
-              </svg>}
-            </div>
-            <Search />
             <Topologies />
             <ViewModeSelector />
             <TimeControl />
@@ -205,6 +197,7 @@ class App extends React.Component {
           {showingNetworkSelector && isGraphViewMode && <NetworkSelector />}
           {!isResourceViewMode && <Status />}
           {!isResourceViewMode && <TopologyOptions />}
+          {!isResourceViewMode && <Search />}
         </Sidebar>
 
         <Footer />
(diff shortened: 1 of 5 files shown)


rade:
screenshot?

fbarl:
@rade Taken from my 1920 x 1080 pixels screen:

context-scope
context-cloud

pidster:
I think I understand what you're trying to achieve, but I'm not at all keen on the weavescope logo moving down to the bottom.

fbarl:
@pidster Are we putting this story on hold before we organize around the general Scope interface redesign project or shall we proceed with the code review?

I do see this particular PR as a prerequisite for #2677 (and #2686) though, unless we want to hack the mockups a bit in the meantime.

pidster:
I don't think this is a pre-requisite, given our intent to fix the general problem. 
We should proceed with getting the timeline in and live with a bit of clutter until we can deal with that separately.

fbarl:
> I don't think this is a pre-requisite, given our intent to fix the general problem.

The reason I'm seeing it as a prerequisite rests on the assumption we made in the mockups that the timestamp (together with the red vertical line over the timeline) is always going to be horizontally centered. 

If we're not doing that yet, we'll have the red line shift more and more to the right on the lower screen resolutions (or more weird, be much to the left on wide screen, so most of the timeline would be showing the future by default, when at *now*). That's maybe not that terrible as a temporary solution, but IMO it does change the experience of using time travel.

rade:
So the main issue really is the logo, since everything else easily shifts down. Yes?

rade:
but the logo isn't present in Weave Cloud, which is the only thing that has timetravel for now, so that's not an issue.

fbarl:
> So the main issue really is the logo, since everything else easily shifts down. Yes?

I'd say it's the logo + the search bar.

rade:
You can shift down the search bar together with everything else.

fbarl:
> You can shift down the search bar together with everything else.

You mean that the timestamp input field is put above search/topologies/etc., occupying its own vertical space and then we could center it horizontally? That could be another way to split the issues, but I'm afraid it will look very hacky - I'll start in that direction and see how it goes.

pidster:
@fbarl a hack that minimises changes is better than making two bigger changes in succession.

rade:
We have several issues open regarding the UI clutter: #1369 #1632

fbarl:
Closing as this will be addressed as a part of #2800.

PR #2703, opened 2017-07-11
Time Travel 3.0

### LINKED ISSUE #2686: Time Travel: zoom and pan for easy navigation in time

fbarl:
Resolves #2686 and moves towards #2677.

**Notes**

* Zooming discoverability might be an issue, but we won't know that for sure until we get some tracking data. The only way now to control the zooming is via scrolling (as the tooltip suggests), so we have to test if it works on touchpads.
* There are no mockups for the timestamps that I implemented here - they are a result of brainstorming between @bia and me:
image_uploaded_from_ios_1024


foot:
Looks and feels really great! Nice one 👍 👍 

- browser window moves a bit onScroll under osx (maybe needs prevDefault?)
- "go back a bit further" controls are the edge are nice, no "go forward" a bit equiv?
- clicking on a visible label on the timeline feels a tiny bit disorienting: where did it go? what state am in now. Feels like a place where a little animation might actually help.
- I like being able to toggle by clicking time travel a lot!
- small glitch? timetravel > live > (while its still going into "live") quickly click pause.
  - stuck in limbo. can recover okay though.

fbarl:
@foot 

> * browser window moves a bit onScroll under osx (maybe needs prevDefault?)

Could you try fixing that on your laptop? I can't really test it without getting live feedback.

> * "go back a bit further" controls are the edge are nice, no "go forward" a bit equiv?

The idea of these markers on the left side was to signal the overall context in the timeline and the fact that they are clickable is just a nice bonus. I was thinking of also adding them to the right side, but then I thought it would make it more cluttered and the context wouldn't be that obvious from a glance. I'm still not sure about it, but for the start I'd just leave it as it is.

> * clicking on a visible label on the timeline feels a tiny bit disorienting: where did it go? what state am in now. Feels like a place where a little animation might actually help.

I added the animation and it looks much better now!

> * I like being able to toggle by clicking time travel a lot!

Me too :)

> * small glitch? timetravel > live > (while its still going into "live") quickly click pause.
stuck in limbo. can recover okay though.

I fixed it, i.e. made it consistent with what happens when you click on LIVE -> TIME TRAVEL. Not sure that fixes all the glitches, but it does seem to behave better.

Issue #2628, opened 2017-06-21
No such container: weaveproxy / weaveplugin


andrey01:
Dear Devs,

we have noticed on our machines the following messages occurring every 10 seconds:

journalctl -f
```
Jun 21 16:26:19 srv-master01 docker[12463]: time="2017-06-21T16:26:19.709531017+02:00" level=error msg="Error setting up exec command in container weaveproxy: No such container: weaveproxy"
Jun 21 16:26:19 srv-master01 docker[12463]: time="2017-06-21T16:26:19.709960879+02:00" level=error msg="Handler for POST /containers/weaveproxy/exec returned error: No such container: weaveproxy"
Jun 21 16:26:19 srv-master01 docker[12463]: time="2017-06-21T16:26:19.710237838+02:00" level=error msg="Handler for GET /containers/weaveplugin/json returned error: No such container: weaveplugin"

Jun 21 16:26:29 srv-master01 docker[12463]: time="2017-06-21T16:26:29.710678227+02:00" level=error msg="Error setting up exec command in container weaveproxy: No such container: weaveproxy"
Jun 21 16:26:29 srv-master01 docker[12463]: time="2017-06-21T16:26:29.710714574+02:00" level=error msg="Handler for POST /containers/weaveproxy/exec returned error: No such container: weaveproxy"
Jun 21 16:26:29 srv-master01 docker[12463]: time="2017-06-21T16:26:29.710867804+02:00" level=error msg="Handler for GET /containers/weaveplugin/json returned error: No such container: weaveplugin"
```

This is coming when weaveworks/scope is deployed.

Any pointers what could be the cause of it?
We are using our custom docker registry, so instead of ``image: weaveworks/weave...`` we are using ``image: our-registry.com/weaveworks/weave``

Our versions are:
kubernetes 1.6.5
weaveworks/scope 1.5.0

To deploy weave-net we are applying the https://github.com/weaveworks/weave/blob/master/prog/weave-kube/weave-daemonset-k8s-1.6.yaml

To deploy the weaveworks/scope we are using the manifest file generated by the ``https://cloud.weave.works/k8s/v1.6/scope.yaml?k8s-service-type=NodePort``

I have found the weaveplugin & weaveproxy mentioned in:

https://github.com/weaveworks/scope/blob/fa56c030e217de82c06f37cf08beb46e410417d7/probe/overlay/weave.go#L315

[...]

PR #2719, opened 2017-07-17
Gather Weave Net plugin and proxy info from report

### LINKED ISSUE #2634: Scope needs updating for Weave Net 2.0
This code: https://github.com/weaveworks/scope/blob/master/probe/overlay/weave.go  
It's attempting to talk to a weaveproxy container, which no longer exists.

2opremio:
Instead of using Docker, because after Weave Net 2.0 there are no proxy nor
plugin containers.

This has the drawback of not detecting the plugin/proxy in systems running
Weave Net < 2.0 , but I think we can live with it.

Fixes #2634

### DIFF
--- a/common/weave/client.go
+++ b/common/weave/client.go
@@ -34,6 +34,8 @@ type Status struct {
 	Router  Router
 	DNS     *DNS
 	IPAM    *IPAM
+	Proxy   *Proxy
+	Plugin  *Plugin
 }
 
 // Router describes the status of the Weave Router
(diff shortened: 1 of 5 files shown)


rade:
@2opremio did you check that this gets rid of the docker daemon log entries shown in #2628? Both for Weave Net 2.0 and pre-2.0 (not essential, but would be lovely).

2opremio:
> did you check that this gets rid of the docker daemon log entries shown

Yes it does.

PR #2723, opened 2017-07-18
Back off when writing to Dynamo and S3

### LINKED ISSUE #2345: Retry dynamoDB writes with backoff in collector
We had an incident today that was caused by insufficient provisioned capacity: weaveworks/service#1157

The collector should implement retries e.g. https://github.com/weaveworks/cortex/blob/611bf7c1561a0318a9303e7b0683ac433b03c870/chunk/dynamodb_client.go#L98

2opremio:
Fixes #2345 

I am intentionally not backing off when writing to Memcached nor NATS. The first one is an optimization and the second one is just used for shortcut reports (not essential).

I tried to use https://github.com/weaveworks/common/blob/master/backoff/backoff.go but it's not suitable for this usecase (it doesn't forward errors and doesn't accept retry limits).

### DIFF
--- a/app/multitenant/aws_collector.go
+++ b/app/multitenant/aws_collector.go
@@ -10,6 +10,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/bluele/gcache"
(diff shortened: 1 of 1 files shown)


app/multitenant/aws_collector.go
```diff
@@ -355,6 +355,19 @@ func calculateReportKey(rowKey, colKey string) (string, error) {
...
 
+func withBackoff(f func() error) error {
+	retries := 0
+	backoff := 50 * time.Millisecond
+	err := f()
+	for err != nil && retries <= 5 {
+		time.Sleep(backoff)
+		err = f()
+		retries += 1
+		backoff *= 2
```
ekimekim: if it's easy, maybe add some small random variance here for thundering herd protection?

app/multitenant/aws_collector.go
```diff
@@ -391,24 +409,27 @@ func (c *awsCollector) Add(ctx context.Context, rep report.Report, buf []byte) e
...
+					},
+					reportField: {
+						S: aws.String(reportKey),
+					},
 				},
-			},
-			ReturnConsumedCapacity: aws.String(dynamodb.ReturnConsumedCapacityTotal),
+				ReturnConsumedCapacity: aws.String(dynamodb.ReturnConsumedCapacityTotal),
+			})
+			return err
```
rade: Surely we should only back off on `provisionedThroughputExceededException`. That's what the [cortex code](https://github.com/weaveworks/cortex/blob/611bf7c1561a0318a9303e7b0683ac433b03c870/chunk/dynamodb_client.go#L98) does.

app/multitenant/aws_collector.go
```diff
@@ -368,7 +381,12 @@ func (c *awsCollector) Add(ctx context.Context, rep report.Report, buf []byte) e
 		return err
 	}
 
-	reportSize, err := c.s3.StoreReportBytes(ctx, reportKey, buf)
+	var reportSize int
+	err = withBackoff(func() error {
+		var err error
+		reportSize, err = c.s3.StoreReportBytes(ctx, reportKey, buf)
```
rade: I don't see why we would want to retry S3. If there are specific retry-able errors (are there?) then maybe, but not in general.

PR #2726, opened 2017-07-18
render: In minor labels, display '0 things' instead of blank if zero things present

### LINKED ISSUE #2722: Controller minor labels don't display properly when controller has no pods
In this example, `configs-db` has 1 pod whereas `configs-db-exporter` has 0. This results in the Pods counter not being present at all, and the string formatting breaking.

Related, the '# Pods' metadata item also doesn't display in the metadata tab, but this is more arguable.

We can fix the minor label issue by changing the minor label formatting code directly, or we can fix the root cause by making all Counter lookups default to 0 for items that aren't there.

ekimekim:
This also fixes a bug where k8s controller nodes would show up as 'Deployment of' without any number

Fixes #2722
screenshot

### DIFF
--- a/render/detailed/summary.go
+++ b/render/detailed/summary.go
@@ -344,13 +344,14 @@ func groupNodeSummary(base NodeSummary, r report.Report, n report.Node) (NodeSum
 }
 
 func pluralize(counters report.Counters, key, singular, plural string) string {
-	if c, ok := counters.Lookup(key); ok {
-		if c == 1 {
-			return fmt.Sprintf("%d %s", c, singular)
-		}
-		return fmt.Sprintf("%d %s", c, plural)
+	c, ok := counters.Lookup(key)
+	if !ok {
+		c = 0
 	}
-	return ""
+	if c == 1 {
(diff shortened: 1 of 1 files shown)


render/detailed/summary.go
```diff
@@ -344,13 +344,14 @@ func groupNodeSummary(base NodeSummary, r report.Report, n report.Node) (NodeSum
...
 
 func pluralize(counters report.Counters, key, singular, plural string) string {
-	if c, ok := counters.Lookup(key); ok {
-		if c == 1 {
-			return fmt.Sprintf("%d %s", c, singular)
-		}
-		return fmt.Sprintf("%d %s", c, plural)
+	c, ok := counters.Lookup(key)
+	if !ok {
+		c = 0
```
rade: surely `c=0` when `!ok`; so could we simply drop that check?

ekimekim: Force of habit, I tend to assume the value is 'some undefined value' when err is not nil. Of course, this is ok, not err, so slightly different conventions, but I'd still rather keep the explicit zero value to make it obvious what's going on.

PR #2728, opened 2017-07-19
Fix Pods number in graph not updating (minor label)

### LINKED ISSUE #2401: Pods number doesn't update on scale up or scale down
When I press the + or - (scale up or scale down) buttons, the number in the topology doesn't update until I press refresh.

See picture below: on the right it says "# PODS: 2" while on the left it says "3 pods".

image

rndstr:
Fixes #2401.

### DIFF
--- a/client/app/scripts/charts/nodes-layout.js
+++ b/client/app/scripts/charts/nodes-layout.js
@@ -487,7 +487,7 @@ export function doLayout(immNodes, immEdges, opts) {
 
     // cache results
     cache.cachedLayout = layout;
-    cache.nodeCache = cache.nodeCache.merge(layout.nodes);
+    cache.nodeCache = cache.nodeCache.merge(layout.nodes.map(n => fromJS({ x: n.get('x'), y: n.get('y') })));
     cache.edgeCache = cache.edgeCache.merge(layout.edges);
   }
 
--- a/client/app/scripts/charts/nodes-layout.js
+++ b/client/app/scripts/charts/nodes-layout.js
@@ -1,6 +1,7 @@
 import dagre from 'dagre';
 import debug from 'debug';
 import { fromJS, Map as makeMap, Set as ImmSet } from 'immutable';
+import pick from 'lodash/pick';
 
 import { NODE_BASE_SIZE, EDGE_WAYPOINTS_CAP } from '../constants/styles';
 import { EDGE_ID_SEPARATOR } from '../constants/naming';
@@ -487,7 +488,10 @@ export function doLayout(immNodes, immEdges, opts) {
 
     // cache results
     cache.cachedLayout = layout;
-    cache.nodeCache = cache.nodeCache.merge(layout.nodes.map(n => fromJS({ x: n.get('x'), y: n.get('y') })));
+    // only cache layout-related properties
+    // NB: These properties must be immutable wrt a given node because properties of updated nodes
+    // will be overwritten with the cached values, see copyLayoutProperties()
+    cache.nodeCache = cache.nodeCache.merge(layout.nodes.map(n => fromJS(pick(n.toJS(), ['x', 'y', 'rank']))));
     cache.edgeCache = cache.edgeCache.merge(layout.edges);
   }
 


client/app/scripts/charts/nodes-layout.js
```diff
@@ -402,7 +402,7 @@ function cloneLayout(layout, nodes, edges) {
 function copyLayoutProperties(layout, nodeCache, edgeCache) {
   const result = Object.assign({}, layout);
   result.nodes = layout.nodes.map(node => (nodeCache.has(node.get('id'))
-    ? node.merge(nodeCache.get(node.get('id'))) : node));
+    ? nodeCache.get(node.get('id')).merge(node) : node));
```
fbarl: I'd like to play with this a bit to see if it really works as it should. I think the idea here was that the cached `x`, `y` values should override the current ones. If `node` object doesn't have them set, then this fix should work just fine. Otherwise, the old direction should be kept for `x`, `y` field, while this direction should be used for all the other node data.

Maybe a more robust solution would be to cache *only* `x`, `y` fields for the nodes (since they are the only fields relevant for the layouts + maybe adjacencies), i.e. to replace

```
cache.nodeCache = cache.nodeCache.merge(layout.nodes)
```

with something like

```
cache.nodeCache = cache.nodeCache.merge(layout.nodes.map(n => fromJS({ x: n.get('x'), y: n.get('y') }))
```

rndstr: I see, that also aligns much better with the method name and its jsdoc. Updated!

fbarl:
Actually, `nodeCache` seems to be used in a lot of places in `nodes-layout.js`, namely in functions:

* `hasUnseenNodes`
* `hasNewNodesOfExistingRank`
* `doLayoutNewNodesOfExistingRank`
* (`copyLayoutProperties`)

The problem is that except for the `x` and `y` fields, `rank` field seems to be used in some of them as well for some of the heuristic algorithms we use to bypass full layout recalculation. As far as I can see, `rank` is the only such field, but it would be good to *double check* that.

So, to make sure nothing is broken I think we could do one of the following:

1. Cache `rank` for nodes alongside [`x`, `y`] if it's a fixed property of a node (not sure about that, but maybe @2opremio knows)
2. See if we could replace `nodeCache` with `immNodes` in the functions listed above

I think a quick fix would be to go with *1* but it's important to note that we'd still be assuming that ranks are immutable, which deserves a comment in the code even if it's always true.

client/app/scripts/charts/nodes-layout.js
```diff
@@ -487,7 +487,7 @@ export function doLayout(immNodes, immEdges, opts) {
 
     // cache results
     cache.cachedLayout = layout;
-    cache.nodeCache = cache.nodeCache.merge(layout.nodes);
+    cache.nodeCache = cache.nodeCache.merge(layout.nodes.map(n => fromJS({ x: n.get('x'), y: n.get('y') })));
```
fbarl: `pick(n.toJS(), ['x', 'y'])` feels a bit more robust and readable (`lodash/pick`), sorry for not mentioning it before!

fbarl: `fromJS` conversion part should be automatic since the parent structure is already an Immutable/Map.

rndstr: I was looking for `pick` in immutablejs but it doesn't exist. I'm fine either way but in this case here it would be

```js
cache.nodeCache = cache.nodeCache.merge(layout.nodes.map(n => fromJS(pick(n.toJS(), ['x', 'y']))));
```
which is less readable IMHO than your initial suggestion. I also don't see the robustness advantage since `get() `deals with missing properties as well.

fbarl: It's completely up to you - I just wanted to make this suggestion alongside your current solution :)

client/app/scripts/charts/nodes-layout.js
```diff
@@ -487,7 +488,10 @@ export function doLayout(immNodes, immEdges, opts) {
 
     // cache results
     cache.cachedLayout = layout;
-    cache.nodeCache = cache.nodeCache.merge(layout.nodes);
+    // only cache layout-related properties
+    // NB: These properties must be immutable wrt a given node because properties of updated nodes
+    // will be overwritten with the cached values, see copyLayoutProperties()
+    cache.nodeCache = cache.nodeCache.merge(layout.nodes.map(n => pick(n.toJS(), ['x', 'y', 'rank'])));
```
fbarl: This doesn't work for me after all, but it does when I add `fromJS` here.

PR #2732, opened 2017-07-20
Allows to skip client build when doing make prog/scope

rndstr:
When you run the client locally with `yarn start` and modify the
app/probe code you don't want to compile the client on `make prog/scope`.

This allows you to do `SCOPE_SKIP_UI_ASSETS=true make prog/scope` to
only build the scope binary.

---
Is there a place I should document this?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,19 +144,19 @@ ifeq ($(BUILD_IN_CONTAINER),true)
 
 client/build/index.html: $(shell find client/app -type f) $(SCOPE_UI_BUILD_UPTODATE)
 	mkdir -p client/build
-	test "true" = "$(SCOPE_SKIP_UI_ASSETS)" || { \
+	if test "true" != "$(SCOPE_SKIP_UI_ASSETS)"; then \
 		$(SUDO) docker run $(RM) $(RUN_FLAGS) -v $(shell pwd)/client/app:/home/weave/app \
 			-v $(shell pwd)/client/build:/home/weave/build \
 			$(SCOPE_UI_BUILD_IMAGE) yarn run build; \
-	}
+	fi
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -144,15 +144,19 @@ ifeq ($(BUILD_IN_CONTAINER),true)
...
+			-v $(shell pwd)/client/build:/home/weave/build \
+			$(SCOPE_UI_BUILD_IMAGE) yarn run build; \
+	}
 
 client/build-external/index.html: $(shell find client/app -type f) $(SCOPE_UI_BUILD_UPTODATE)
-	mkdir -p client/build
-	$(SUDO) docker run $(RM) $(RUN_FLAGS) -v $(shell pwd)/client/app:/home/weave/app \
-		-v $(shell pwd)/client/build-external:/home/weave/build-external \
-		$(SCOPE_UI_BUILD_IMAGE) yarn run build-external
+	mkdir -p client/build-external
```
fbarl: Good catch!

Makefile
```diff
@@ -144,15 +144,19 @@ ifeq ($(BUILD_IN_CONTAINER),true)
 
 client/build/index.html: $(shell find client/app -type f) $(SCOPE_UI_BUILD_UPTODATE)
 	mkdir -p client/build
-	$(SUDO) docker run $(RM) $(RUN_FLAGS) -v $(shell pwd)/client/app:/home/weave/app \
-		-v $(shell pwd)/client/build:/home/weave/build \
-		$(SCOPE_UI_BUILD_IMAGE) yarn run build
+	test "true" = "$(SCOPE_SKIP_UI_ASSETS)" || { \
```
fbarl: I'd personally find `test "true" != "$(SCOPE_SKIP_UI_ASSETS)" && { \` (or even `if test "true" != "$(SCOPE_SKIP_UI_ASSETS)"; then`) more readable, but maybe that's just me :)

Issue #2650, opened 2017-06-27
probes falling back to proc parsing due to late eBPF events


rade:
We see this:
```
<probe> ERRO: 2017/06/27 05:02:25.336993 tcp tracer received event with timestamp 615263068049191 even though the last timestamp was 615263068049504. Stopping the eBPF tracker. 
```
after running probes for hours, even days.

The message is produced by the logic introduced in #2334, even though we are running Ubuntu with a 4.4.0-81 kernel, which should not exhibit the problem described in that issue.

@alban any idea how we might go about tracking this down?

Also, as a workaround, is there any downside to simply *restarting* all the eBPF tracking when getting this error, i.e. going through the same initialisation as when initially starting the probes, including the initial proc walk? We'd need to guard against "flapping", obviously.

2opremio:
@alban has suggested offline:

> if it is reproducible, I would add logs to print the cpu number for both events..
>
> I would try to reproduce an issue with the tcptracer-bpf test program (without scope) in the dev environment with your kernel, using a script like https://github.com/weaveworks/scope/pull/2507#issuecomment-300735719 with different cpu# in different terminals

2opremio:
Could this be due to https://github.com/iovisor/gobpf/issues/42 ?

alban:
@rade How many cpus do you have on the machine where the bug was produced?

I am currently trying to reproduce this on GCE with this config:

```
$ uname -r
4.4.0-81-generic
$ grep PRETTY /etc/os-release
PRETTY_NAME="Ubuntu 14.04.5 LTS"
$ cat /sys/devices/system/cpu/{possible,present,online}
0
0
0
```

rade:
```
# uname -r
4.4.0-83-generic
# grep PRETTY /etc/os-release
PRETTY_NAME="Ubuntu 16.04.1 LTS"
# cat /sys/devices/system/cpu/{possible,present,online}
0-127
0-15
0-15
```

rade:
FYI, on our dev cluster of seven machines, all the probes were started on 2017/06/29 21:51. Four of them are still running in ebpf mode now, the other three encountered a timestamp discrepancy at
- 2017/06/30 07:50:49.966603
- 2017/07/01 22:42:24.916713
- 2017/07/04 02:03:31.517207

respectively.

So this error does appear to be quite rare.

alban:
I can reproduce the bug using `tcptracer-bpf/tests/tracer` in about 1 minute execution with the following:
```
[...]

PR #2735, opened 2017-07-20
restart eBPF tracking on error

alban:
This is a workaround for https://github.com/weaveworks/scope/issues/2650

Tested manually the following way:

Start Scope with:
```
    sudo WEAVESCOPE_DOCKER_ARGS="-e SCOPE_DEBUG_BPF=1" ./scope launch
```    
Simulate EbpfTracker failure with:
```
    echo foo | sudo tee /proc/$(pidof scope-probe)/root/var/run/scope/debug-bpf
```

TODO:
- [x] implement guard against "flapping"
- [x] tests
- [x] ~pending on https://github.com/weaveworks/tcptracer-bpf/pull/49~
- [x] pending on https://github.com/weaveworks/tcptracer-bpf/pull/50

### DIFF
--- a/probe/endpoint/connection_tracker.go
+++ b/probe/endpoint/connection_tracker.go
@@ -2,6 +2,7 @@ package endpoint
 
 import (
 	"strconv"
+	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/weaveworks/scope/probe/endpoint/procspy"
@@ -29,6 +30,9 @@ type connectionTracker struct {
 	flowWalker      flowWalker // Interface
 	ebpfTracker     *EbpfTracker
 	reverseResolver *reverseResolver
+
+	// time of the previous ebpf failure, or zero if it didn't fail
+	ebpfLastFailureTime time.Time
 }
 
 func newConnectionTracker(conf connectionTrackerConfig) connectionTracker {
@@ -87,8 +91,29 @@ func (t *connectionTracker) ReportConnections(rpt *report.Report) {
 			t.performEbpfTrack(rpt, hostNodeID)
 			return
 		}
-		log.Warnf("ebpf tracker died, gently falling back to proc scanning")
-		t.useProcfs()
+
+		// We only restart the EbpfTracker if the failures are not too frequent to
+		// avoid repeatitive restarts.
+
+		ebpfLastFailureTime := t.ebpfLastFailureTime
+		t.ebpfLastFailureTime = time.Now()
+
+		if ebpfLastFailureTime.After(time.Now().Add(-5 * time.Minute)) {
+			// Multiple failures in the last 5 minutes, fall back to proc parsing
+			log.Warnf("ebpf tracker died again, gently falling back to proc scanning")
+			t.useProcfs()
+		} else {
+			// Tolerable failure rate, restart the tracker
+			log.Warnf("ebpf tracker died, restarting it")
+			err := t.ebpfTracker.restart()
+			if err == nil {
+				go t.getInitialState()
+				t.performEbpfTrack(rpt, hostNodeID)
+				return
+			}
+			log.Warnf("could not restart ebpf tracker, falling back to proc scanning: %v", err)
+			t.useProcfs()
+		}
 	}
 
 	// consult the flowWalker for short-lived (conntracked) connections
--- a/probe/endpoint/ebpf.go
+++ b/probe/endpoint/ebpf.go
@@ -3,8 +3,11 @@ package endpoint
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"regexp"
 	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 
@@ -30,8 +33,20 @@ type EbpfTracker struct {
 	sync.Mutex
 	tracer          *tracer.Tracer
 	ready           bool
+	stopping        bool
 	dead            bool
 	lastTimestampV4 uint64
+
+	// debugBPF specifies if EbpfTracker must be started in debug mode. This
+	// allows to easily debug issues like:
+	// https://github.com/weaveworks/scope/issues/2650
+	//
+	// Scope could be started this way:
+	//   $ sudo WEAVESCOPE_DOCKER_ARGS="-e SCOPE_DEBUG_BPF=1" ./scope launch
+	//
+	// Then, EbpfTracker could be tricked into restarting with:
+	//   $ echo stop | sudo tee /proc/$(pidof scope-probe)/root/var/run/scope/debug-bpf
+	debugBPF bool
 
 	openConnections   map[fourTuple]ebpfConnection
 	closedConnections []ebpfConnection
@@ -77,24 +92,35 @@ func newEbpfTracker() (*EbpfTracker, error) {
 		return nil, fmt.Errorf("kernel not supported: %v", err)
 	}
 
-	tracker := &EbpfTracker{
-		openConnections:  map[fourTuple]ebpfConnection{},
-		closedDuringInit: map[fourTuple]struct{}{},
+	var debugBPF bool
+	if os.Getenv("SCOPE_DEBUG_BPF") != "" {
+		log.Infof("ebpf tracker started in debug mode")
+		debugBPF = true
 	}
 
-	tracer, err := tracer.NewTracer(tracker)
-	if err != nil {
+	tracker := &EbpfTracker{
+		debugBPF: debugBPF,
+	}
+	if err := tracker.restart(); err != nil {
 		return nil, err
 	}
-
-	tracker.tracer = tracer
-	tracer.Start()
 
 	return tracker, nil
 }
 
 // TCPEventV4 handles IPv4 TCP events from the eBPF tracer
 func (t *EbpfTracker) TCPEventV4(e tracer.TcpV4) {
+	if t.debugBPF {
+		debugBPFFile := "/var/run/scope/debug-bpf"
+		b, err := ioutil.ReadFile("/var/run/scope/debug-bpf")
+		if err == nil && strings.TrimSpace(string(b[:])) == "stop" {
+			os.Remove(debugBPFFile)
+			log.Warnf("ebpf tracker stopped as requested by user")
+			t.stop()
+			return
+		}
+	}
+
 	if t.lastTimestampV4 > e.Timestamp {
 		// A kernel bug can cause the timestamps to be wrong (e.g. on Ubuntu with Linux 4.4.0-47.68)
 		// Upgrading the kernel will fix the problem. For further info see:
@@ -102,6 +128,7 @@ func (t *EbpfTracker) TCPEventV4(e tracer.TcpV4) {
 		// https://github.com/weaveworks/scope/issues/2334
 		log.Errorf("tcp tracer received event with timestamp %v even though the last timestamp was %v. Stopping the eBPF tracker.", e.Timestamp, t.lastTimestampV4)
 		t.stop()
+		return
 	}
 
 	t.lastTimestampV4 = e.Timestamp
@@ -287,10 +314,45 @@ func (t *EbpfTracker) isDead() bool {
 
 func (t *EbpfTracker) stop() {
 	t.Lock()
-	alreadyDead := t.dead
-	t.dead = true
+	alreadyDead := t.dead || t.stopping
+	t.stopping = true
 	t.Unlock()
-	if !alreadyDead && t.tracer != nil {
-		t.tracer.Stop()
+
+	// Do not call tracer.Stop() in this thread, otherwise tracer.Stop() will
+	// deadlock waiting for this thread to pick up the next event.
+	go func() {
+		if !alreadyDead && t.tracer != nil {
+			t.tracer.Stop()
+			t.tracer = nil
+		}
+
+		// Only advertise the tracer as dead after the tracer is fully stopped so that
+		// restart() is not called in parallel in another thread.
+		t.Lock()
+		t.stopping = false
+		t.dead = true
+		t.Unlock()
+	}()
+}
+
+func (t *EbpfTracker) restart() error {
+	t.Lock()
+	defer t.Unlock()
+
+	t.dead = false
+	t.ready = false
+
+	t.openConnections = map[fourTuple]ebpfConnection{}
+	t.closedDuringInit = map[fourTuple]struct{}{}
+	t.closedConnections = []ebpfConnection{}
+
+	tracer, err := tracer.NewTracer(t)
+	if err != nil {
+		return err
 	}
+
+	t.tracer = tracer
+	tracer.Start()
+
+	return nil
 }
--- a/probe/endpoint/ebpf_test.go
+++ b/probe/endpoint/ebpf_test.go
@@ -5,6 +5,7 @@ import (
 	"reflect"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/weaveworks/tcptracer-bpf/pkg/tracer"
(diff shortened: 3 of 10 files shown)


probe/endpoint/ebpf.go
```diff
@@ -287,10 +310,33 @@ func (t *EbpfTracker) isDead() bool {
 
 func (t *EbpfTracker) stop() {
 	t.Lock()
+	defer t.Unlock()
+
```
rade: I deliberately didn't do this, because it requires a lot of assumptions about the behaviour of tracer, in order to avoid deadlocks. It's fine right now.

Was there a specific reason why you changed this?

probe/endpoint/ebpf.go
```diff
@@ -287,10 +310,33 @@ func (t *EbpfTracker) isDead() bool {
...
+	tracer, err := tracer.NewTracer(t)
+	if err != nil {
+		return err
+	}
+
+	t.tracer = tracer
+	tracer.Start()
+
+	return nil
+}
```
rade: I'd be wary of holding the lock while calling `tracer.Start()`.

Also, `newEbpfTracker()` should be refactored to call this.

probe/endpoint/ebpf.go
alban: Actually, it fails if I let the Unlock as it was before:

```
<probe> WARN: 2017/07/24 13:45:09.816261 ebpf tracker stopped as requested by user
<probe> WARN: 2017/07/24 13:45:10.304710 ebpf tracker died, trying to restart it
<probe> WARN: 2017/07/24 13:45:10.312468 could not restart ebpf tracker, falling back to proc scanning: cannot write "p:ptcp_v4_connect tcp_v4_connect\n" to kprobe_events: write /sys/kernel/debug/tracing/kprobe_events: device or resource busy
```

The stop is done in the event handler thread, when I receive an event with a wrong timestamp. And the restart is done in the publisher thread when it notices that `isDead == true`. I think that the publisher thread starts the restart before the tracer is fully stopped, so the kprobe is not completely removed yet...

alban:
I updated the code. It seems to restart fine. But I had one instance of my test where Scope was talking 100% of the cpu in the gobpf code when I tested the fallback to proc parsing.

<details>

I got the stack of the go routine taking all the cpu with the following command. The stack changes all the time since it is running.

```
sudo WEAVESCOPE_DOCKER_ARGS="-e SCOPE_DEBUG_BPF=1" ./scope launch --probe.http.listen :4041
http://localhost:4041/debug/pprof/goroutine?debug=2
```

```
goroutine 4442 [runnable, locked to thread]:
github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf._C2func_poll(0xc42385aea0, 0x4, 0x1f4, 0x4, 0x0, 0x0)
	github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf/_obj/_cgo_gotypes.go:313 +0x68
github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf.perfEventPoll(0xc4218f0e40, 0x4, 0x4, 0xc42385ae00, 0xc423857510)
	/go/src/github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf/perf.go:258 +0xf6
github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf.(*PerfMap).PollStart.func1(0xc42026c910, 0xc4245df1c0, 0xc4245db180)
	/go/src/github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf/perf.go:171 +0x695
created by github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf.(*PerfMap).PollStart
	/go/src/github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf/perf.go:238 +0xf1
```

</details>

` `

Also, the integration test needs fixing too: https://circleci.com/gh/kinvolk/scope/990
integration/313_container_to_container_edge_with_ebpf_proc_fallback_test.sh

alban:
I fixed the regression in integration test 313 and added a new integration test 315 that restarts EbpfTracker.

probe/endpoint/ebpf.go
```diff
@@ -77,31 +92,43 @@ func newEbpfTracker() (*EbpfTracker, error) {
 		return nil, fmt.Errorf("kernel not supported: %v", err)
 	}
 
-	tracker := &EbpfTracker{
-		openConnections:  map[fourTuple]ebpfConnection{},
-		closedDuringInit: map[fourTuple]struct{}{},
+	var debugBPF bool
+	if os.Getenv("SCOPE_DEBUG_BPF") != "" {
+		log.Warnf("ebpf tracker started in debug mode")
```
2opremio: I would make this an Info

probe/endpoint/ebpf_test.go
```diff
@@ -220,6 +221,8 @@ func TestInvalidTimeStampDead(t *testing.T) {
 	if cnt != 2 {
 		t.Errorf("walkConnections found %v instead of 2 connections", cnt)
 	}
+	// EbpfTracker is marked as dead asynchronously.
+	time.Sleep(100 * time.Millisecond)
```
2opremio: Can't you poll with a timeout? magic sleep times lead to flaky tests.

probe/endpoint/connection_tracker.go
```diff
@@ -87,7 +91,27 @@ func (t *connectionTracker) ReportConnections(rpt *report.Report) {
...
+		// avoid repeatitive restarts.
+		if t.ebpfFailureCount > 0 && t.ebpfFailureTime.Before(time.Now().Add(-5*time.Minute)) {
+			t.ebpfFailureCount = 0
+		}
+		t.ebpfFailureTime = time.Now()
+
+		if t.ebpfFailureCount == 0 {
+			log.Warnf("ebpf tracker died, restarting it")
+			t.ebpfFailureCount++
+			err := t.ebpfTracker.restart()
```
2opremio: please check for errors first

```
if err := t.ebpfTracker.restart(); err != nil {
  log.Warnf("could not restart ebpf tracker, falling back to proc scanning: %v", err)
}
go t.getInitialState()
t.performEbpfTrack(rpt, hostNodeID)
return
```

probe/endpoint/connection_tracker.go
```diff
@@ -87,7 +91,27 @@ func (t *connectionTracker) ReportConnections(rpt *report.Report) {
...
-		log.Warnf("ebpf tracker died, gently falling back to proc scanning")
+
+		// We only restart the EbpfTracker if the failures are not too frequent to
+		// avoid repeatitive restarts.
+		if t.ebpfFailureCount > 0 && t.ebpfFailureTime.Before(time.Now().Add(-5*time.Minute)) {
+			t.ebpfFailureCount = 0
+		}
+		t.ebpfFailureTime = time.Now()
+
+		if t.ebpfFailureCount == 0 {
```
2opremio: This logic is hard to read. I don't think there's a need to split this in multiple `if`s. I think it's preferable to repeat the `t.useProcfs()`. Also, it's more idiomatic in Go to check for errors in `if`s and continue with the least erroneous path.

```
t.ebpfFailureTime = time.Now()
// Multiple failures in the last 5 minutes, fall back to proc parsing
if t.ebpfFailureCount > 0 && t.ebpfFailureTime.After(time.Now().Add(-5*time.Minute))  {
   log.Warnf("ebpf tracker died again, gently falling back to proc scanning")
   t.useProcfs()
  return
}

// Tolerable failure rate, let's just reset the failure count and restart the traker
t.ebpfFailureCount = 0
log.Warnf("ebpf tracker died, restarting it")
if err := t.ebpfTracker.restart(); err != nil {
  log.Warnf("could not restart ebpf tracker, falling back to proc scanning: %v", err)
  t.useProcfs()
  return
}

go t.getInitialState()
t.performEbpfTrack(rpt, hostNodeID)
return
```

probe/endpoint/connection_tracker.go
```diff
@@ -29,6 +30,9 @@ type connectionTracker struct {
 	flowWalker      flowWalker // Interface
 	ebpfTracker     *EbpfTracker
 	reverseResolver *reverseResolver
+
+	ebpfFailureCount int
+	ebpfFailureTime  time.Time
```
2opremio: Wouldn't it be more accurate to call this `ebpfLastFailureTime`?

probe/endpoint/connection_tracker.go
```diff
@@ -29,6 +30,9 @@ type connectionTracker struct {
 	flowWalker      flowWalker // Interface
 	ebpfTracker     *EbpfTracker
 	reverseResolver *reverseResolver
+
+	ebpfFailureCount int
```
2opremio: Do we need counts here? We only check it to be higher than 0, right? then we can simply call it `didEBPFFail`?

2opremio: In fact, I think we can remove this altogether and use the zero value of time (https://golang.org/pkg/time/#Time.IsZero ) to reset the failures.

probe/endpoint/ebpf.go
```diff
@@ -287,10 +314,45 @@ func (t *EbpfTracker) isDead() bool {
...
-	alreadyDead := t.dead
-	t.dead = true
+	alreadyDead := t.dead || t.stopping
+	t.stopping = true
 	t.Unlock()
-	if !alreadyDead && t.tracer != nil {
-		t.tracer.Stop()
+
+	// Do not call tracer.Stop() in this thread, otherwise tracer.Stop() will
+	// deadlock waiting for this thread to pick up the next event.
```
2opremio: How would it deadlock? I am probably missing something, but  [Stop()](https://github.com/weaveworks/tcptracer-bpf/blob/master/pkg/tracer/tracer.go#L130-L135 ) just seems to close the channel, which should be fine.

probe/endpoint/connection_tracker.go
```diff
@@ -87,7 +91,27 @@ func (t *connectionTracker) ReportConnections(rpt *report.Report) {
...
+			t.ebpfFailureCount = 0
+		}
+		t.ebpfFailureTime = time.Now()
+
+		if t.ebpfFailureCount == 0 {
+			log.Warnf("ebpf tracker died, restarting it")
+			t.ebpfFailureCount++
+			err := t.ebpfTracker.restart()
+			if err == nil {
+				go t.getInitialState()
```
2opremio: Shouldn't the invocation of `getInitialState()` be part of the tracker initialization in `restart()`?

2opremio: We seem to always call it immediately after it.

probe/endpoint/ebpf.go
2opremio: In other words why can't `stop()` simply be:
```
t.Lock()
defer t.Unlock()
if !t.dead {
  return
}
t.tracer.Stop()
t.tracer = nil
t.dead = true
```

2opremio: BTW, the `stop` channel should have a higher priority than the other channels here: https://github.com/weaveworks/tcptracer-bpf/blob/master/pkg/tracer/tracer.go#L80-L99

probe/endpoint/connection_tracker.go
alban: yes, but `getInitialState()` uses other objects besides just ebpfTracker, such as consulting conntrack. And `restart()` in ebpf.go is independent from conntrack. So better to keep it separate.

probe/endpoint/ebpf.go
alban: The deadlock was not with tcptracer-bpf's channel `t.stopChan` but with gobpf's channels:
- `t.perfMapIPV4.PollStop()` / `pm.pollStop <- true` ([perf.go#L244](https://github.com/iovisor/gobpf/blob/51b4a5bc15ab504f9efb7702591ae04bc6c6f068/elf/perf.go#L244))
- `pm.receiverChan <- incoming.bytesArray[0]` ([perf.go#L229](https://github.com/iovisor/gobpf/blob/51b4a5bc15ab504f9efb7702591ae04bc6c6f068/elf/perf.go#L229))

I could reliably reproduce the deadlock before using pprof to print the go routine stacks (following this [doc](https://stackoverflow.com/questions/19094099/how-to-dump-goroutine-stacktraces/19145992#19145992)).

Maybe this should be fixed at the gobpf level but I don't immediately see how, because the caller of gobpf's `PollStop()` needs to know when it's done...

probe/endpoint/ebpf_test.go
alban: ok, will do

probe/endpoint/connection_tracker.go
alban: ok, I will remove ebpfFailureCount

probe/endpoint/connection_tracker.go
alban: :+1:

alban:
Branch updated.

alban:
Tests:
- CircleCI passes
- tested manually on my laptop (simulating EbpfTracker failure)
- tested on the GCE instance with the [test script](https://github.com/weaveworks/scope/issues/2650#issuecomment-314788229) to reproduce it the bug

It seemed to work fine.

But then, I had a look at the go routine stacks with pprof after Scope fell back on proc parsing:
```
goroutine 26 [chan send, 4 minutes]:
github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf.(*PerfMap).PollStart.func1(0xc4202c0370, 0xc422dbbf00, 0xc423f508a0)
	/go/src/github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf/perf.go:229 +0x4f7
created by github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf.(*PerfMap).PollStart
	/go/src/github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf/perf.go:238 +0xf1
```

- the go routine in gobpf should not be there anymore but it is. It does not take CPU, but that's still a leak. Probably related to @2opremio 's comment https://github.com/weaveworks/scope/pull/2735#discussion_r130365172
- the go import path shows a double vendoring. Is it expected?

alban:
Fix in gobpf, tested on GCE: https://gist.github.com/alban/cb4b899e9558e231d080ec7e7b3abbc5
Explanation: https://github.com/iovisor/gobpf/blob/d0a3e1b/elf/perf.go#L241-L242

It seems to resolve the leaking go routine. I'll do the PR tomorrow.

alban:
I updated the vendoring with https://github.com/weaveworks/tcptracer-bpf/pull/49  and marked this PR as WIP because the dependency is not merged yet.

alban:
I updated the PR (pending on https://github.com/weaveworks/tcptracer-bpf/pull/50) and it works for me now. I don't see leaking go routines nor panics anymore. I tested several times both with waiting 5 minutes between the ebpf tracker crashes and without waiting.

alban:
@2opremio new review welcome :)

PR #2736, opened 2017-07-21
optimisation: don't copy report stream unnecessarily

rade:
We don't need to copy from the reader when publishing to just one destination.

### DIFF
--- a/probe/appclient/multi_client.go
+++ b/probe/appclient/multi_client.go
@@ -166,13 +166,21 @@ func (c *multiClient) Stop() {
 // reader, and recreate new readers for each publisher. Note that it will
 // publish to one endpoint for each unique ID. Failed publishes don't count.
 func (c *multiClient) Publish(r io.Reader) error {
+	c.mtx.Lock()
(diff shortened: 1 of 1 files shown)


probe/appclient/multi_client.go
```diff
@@ -166,13 +166,21 @@ func (c *multiClient) Stop() {
...
+		return nil
+	}
+
 	buf, err := ioutil.ReadAll(r)
 	if err != nil {
 		return err
 	}
 
-	c.mtx.Lock()
-	defer c.mtx.Unlock()
```
rade: Minor downside here is that the lock now covers the `ioutil.ReadAll(r)`. That's in the >1 case only though, so not the common path, and eliminating it would require unlocking and relocking which may well be more expensive.

ekimekim: can't you avoid it just by keeping this code and moving the lock/unlock above inside the if statement?

rade: no, because evaluating `len(c.clients)` requires the lock.

PR #2741, opened 2017-07-24
Add test for pods number not updating

rndstr:
Follow-up to #2728

### DIFF
--- a/client/app/scripts/charts/__tests__/nodes-layout-test.js
+++ b/client/app/scripts/charts/__tests__/nodes-layout-test.js
@@ -131,6 +131,18 @@ describe('NodesLayout', () => {
       edges: fromJS({
         [edge('n1', 'n4')]: {id: edge('n1', 'n4'), source: 'n1', target: 'n4'}
       })
+    },
+    layoutProps: {
+      nodes: fromJS({
+        n1: {id: 'n1', label: 'lold', labelMinor: 'lmold', rank: 'rold'},
+      }),
+      edges: fromJS({})
+    },
+    layoutProps2: {
+      nodes: fromJS({
+        n1: {id: 'n1', label: 'lnew', labelMinor: 'lmnew', rank: 'rnew', x: 111, y: 109},
+      }),
+      edges: fromJS({})
     }
   };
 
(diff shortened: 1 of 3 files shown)


client/app/scripts/charts/nodes-layout.js
```diff
@@ -481,7 +481,8 @@ export function doLayout(immNodes, immEdges, opts) {
 
   if (layout) {
     // Last line of defense - re-render everything if two nodes are too close to one another.
-    if (minEuclideanDistanceBetweenPoints(layout.nodes) < NODE_CENTERS_SEPARATION_FACTOR) {
+    if (layout.nodes.count() > 1 &&
```
fbarl: I just discovered that Javascript has `Infinity`! :) Maybe it would be easier to simplify `minEuclideanDistanceBetweenPoints` by setting `let minDistance = Infinity;` there (and then removing `foundPair` and extra logic that it introduces). That would also make this check unnecessary.

client/app/scripts/charts/__tests__/nodes-layout-test.js
```diff
@@ -438,4 +450,26 @@ describe('NodesLayout', () => {
...
+      nodeSets.layoutProps2.nodes,
+      nodeSets.layoutProps2.edges,
+      {}
+    );
+
+    // new labels should not be overwritten by cache
+    nodes = second.nodes.toJS();
+    expect(nodes.n1.label).toEqual('lnew');
+    expect(nodes.n1.labelMinor).toEqual('lmnew');
+    expect(nodes.n1.rank).toEqual('rold');
```
fbarl: Maybe it would also be good to test for `x` and `y` values here.

rndstr:
@fbarl good suggestions, mind taking another look?

client/app/scripts/charts/__tests__/nodes-layout-test.js
```diff
@@ -131,6 +131,18 @@ describe('NodesLayout', () => {
       edges: fromJS({
         [edge('n1', 'n4')]: {id: edge('n1', 'n4'), source: 'n1', target: 'n4'}
       })
+    },
+    layoutProps: {
+      nodes: fromJS({
+        n1: {id: 'n1', label: 'lold', labelMinor: 'lmold', rank: 'rold'},
```
fbarl: Shouldn't we also set `x`, `y` values here, like we do with `rank`, to make sure they are not overrwritten?

rndstr: First run ignores the cache and will always have a fresh `x` & `y` due to the `runLayout()`. That's why I store the result of the first run in `first` and then make sure in the second run they are not overwritten by `layoutProps2`.

fbarl: Fair enough! :)

PR #2743, opened 2017-07-24
new full reports are more important than old and shortcut reports

### LINKED ISSUE #2738: nothing displayed when there is a high rate of container/pod churn
A high volume of shortcut reports - produced by container/pod births and deaths - can drown out full reports, causing the latter to be dropped, especially when reports take a while to be processed (see #2464). This results in data loss and in extreme cases - when full reports get dropped for longer than the 15s report aggregation interval in the app - nothing being displayed at all.

To reproduce, apply
```diff
diff --git a/app/router.go b/app/router.go
index 414f45f9..7e57bb3d 100644
--- a/app/router.go
+++ b/app/router.go
@@ -159,6 +159,9 @@ func RegisterReportPostHandler(a Adder, router *mux.Router) {
                        respondWith(w, http.StatusInternalServerError, err)
                        return
                }
+
+               time.Sleep(3 * time.Second)
+
                w.WriteHeader(http.StatusOK)
        }))
 }
```
and then start scope normally with `scope launch` and point a browser at it. Then introduce some container churn by running
```
$ while true; do docker run --rm alpine true; sleep 1; done
```
Every now and then, the display in the browser will say that there is "Nothing to show".

The underlying reason is that `appClient.Publish` drops reports when the internal buffer (of just two reports, plus one in flight) is full, making no distinction between full and shortcut reports. So when there is a high volume of shortcut reports, there is a good chance that the buffer will be filled with them, causing full reports to be dropped.

rade:
so when there is backpressure in publishing reports, drop shortcut reports in preference to full reports, and drop old full reports in preference to new full reports.

Fixes #2738.

### DIFF
--- a/probe/appclient/app_client.go
+++ b/probe/appclient/app_client.go
@@ -30,7 +30,7 @@ type AppClient interface {
 	ControlConnection()
 	PipeConnection(string, xfer.Pipe)
 	PipeClose(string) error
-	Publish(r io.Reader) error
+	Publish(io.Reader, bool) error
 	Target() url.URL
 	ReTarget(url.URL)
 	Stop()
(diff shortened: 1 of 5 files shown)


probe/appclient/app_client.go
```diff
@@ -311,13 +311,25 @@ func (c *appClient) startPublishing() {
...
+			return nil
+		}
+		// drop an old report to make way for new one
+		c.mtx.Lock()
+		defer c.mtx.Unlock()
+		select {
+		case <-c.readers:
+		default:
+		}
+		c.readers <- r
```
2opremio: Would it make sense to increase the buffer size of readers?

rade: Only for shortcut reports. For full reports we'd actually would rather have a single-place buffers.

I originally planned to do something along those lines, but it's a bigger change.

2opremio: Alright

PR #2746, opened 2017-07-25
do not back off on timeouts when sending reports

### LINKED ISSUE #2745: nothing displayed when app takes >4s to process reports
Apply
```
diff --git a/app/router.go b/app/router.go
index 414f45f9..40075048 100644
--- a/app/router.go
+++ b/app/router.go
@@ -159,6 +159,9 @@ func RegisterReportPostHandler(a Adder, router *mux.Router) {
                        respondWith(w, http.StatusInternalServerError, err)
                        return
                }
+
+               time.Sleep(5 * time.Second)
+
                w.WriteHeader(http.StatusOK)
        }))
 }
```
and then run scope normally with `scope launch` and point a browser at it.

Initially this will show containers as expected, but after ~1m the "nothing to show" message appears. This then remains there but for a ~15s periods approximately every minute, when containers are shown.

The problem is as follows: The client has an http timeout set to 4s, and exponentially backs off on all errors, including timeouts, up to a max backoff of 1m. Since the app displays data from aggregating reports over a 15s window, any time the client backs off for longer than that window, results in no data being available.

Solution: we shouldn't be backing off on timeouts.

rade:
...since doing so unnecessarily throttles report sending, to the point where the app is receiving reports so infrequently that often it has no data to show.

The timeout period itself is sufficient to prevent thrashing.

Fixes #2745.

### DIFF
--- a/probe/appclient/app_client.go
+++ b/probe/appclient/app_client.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/rpc"
 	"net/url"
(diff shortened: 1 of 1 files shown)


probe/appclient/app_client.go
```diff
@@ -215,7 +216,11 @@ func (c *appClient) doWithBackoff(msg string, f func() (bool, error)) {
 			backoff = initialBackoff
 			continue
 		}
-
+		if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
```
2opremio: I would add a comment explaining why the 4s timeout is enough

rade: done

PR #2764, opened 2017-07-28
Update build tools to 35679ee

leth:


leth:
Hmm the linter failed. I would have expected this to be linted to the same standard by the other repo...

leth:
I've raised https://github.com/weaveworks/build-tools/issues/111

PR #2771, opened 2017-07-29
add 'realclean' make target to clear out container images

rade:
...and use that in bin/release, so that we don't build releases with ancient images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all deps static clean client-lint client-test client-sync backend frontend shell lint ui-upload
+.PHONY: all deps static clean realclean client-lint client-test client-sync backend frontend shell lint ui-upload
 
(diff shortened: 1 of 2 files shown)


rade:
I've tested this by running
```shell
docker rmi -f $(docker images -a)
docker images -a # should show nothing
make
make realclean
docker images -a # should show nothing
```

Makefile
```diff
@@ -212,16 +212,20 @@ ui-pkg-upload: tmp/weave-scope.tgz
 
 clean:
 	$(GO) clean ./...
-# Don't actually rmi the build images - rm'ing the .uptodate files is enough to ensure
```
ekimekim: keep this comment but add 'in most cases'. below, add a comment on why you sometimes do need realclean.

rade: Done. PTAL

PR #2772, opened 2017-07-30
get rid of endpoint type indicators

rade:
The app stopped paying attention to these some time ago.

Removing them shrinks reports by 3-10%.


2opremio:
I had forgotten that we don't check procspied to match connections by container IP. LGTM

PR #2774, opened 2017-07-31
rename 'report_persistence' capability to 'historic_reports'

rade:
since that better captures the intent - the UI doesn't care about reports get stored, but what reports it can retrieve.

Part of #2607.


2opremio:
Nicer without the explicit S3 check.

Issue #2302, opened 2017-03-05
CPU intensive render in FF (not Chrome)


rade:
When viewing [Weave Cloud Prod in prod, services, all namespaces, graph mode](https://cloud.weave.works/app/loud-breeze-77#!/state/{%22controlPipe%22:null,%22topologyViewMode%22:%22topo%22,%22nodeDetails%22:[],%22pinnedMetricType%22:null,%22pinnedSearches%22:[],%22searchQuery%22:%22%22,%22selectedNodeId%22:null,%22gridSortedBy%22:null,%22gridSortedDesc%22:null,%22topologyId%22:%22services%22,%22topologyOptions%22:{%22processes%22:{%22unconnected%22:%22hide%22},%22containers%22:{%22system%22:%22application%22,%22stopped%22:%22running%22,%22pseudo%22:%22hide%22},%22pods%22:{%22namespace%22:%22all%22,%22pseudo%22:%22hide%22}},%22contrastMode%22:false}), firefox uses an entire core on my machine. By contrast, Chrome uses less than 5%.

rade:
I wonder whether #2291 would help here. Profiling tells me that most of the time is going on lots of "Recalculate Style" with "Restyle Hint: CSSAnimations".

rade:
> I wonder whether #2291 would help here.

Apparently not. It's running on dev now but I am still seeing the same behaviour on [prod-in-dev](https://frontend.dev.weave.works/app/nameless-night-79#!/state/{%22controlPipe%22:null,%22topologyViewMode%22:%22topo%22,%22nodeDetails%22:[],%22pinnedMetricType%22:null,%22pinnedSearches%22:[],%22searchQuery%22:%22%22,%22selectedNodeId%22:null,%22gridSortedBy%22:%22label%22,%22gridSortedDesc%22:null,%22topologyId%22:%22services%22,%22topologyOptions%22:{%22processes%22:{%22unconnected%22:%22hide%22},%22containers%22:{%22system%22:%22all%22,%22stopped%22:%22running%22,%22pseudo%22:%22hide%22},%22pods%22:{%22namespace%22:%22all%22,%22pseudo%22:%22hide%22}},%22contrastMode%22:false}).

fbarl:
[...]

PR #2777, opened 2017-07-31
Adjust timeline zoom sensitivity on Firefox

### LINKED ISSUE #2768: Time Travel: zoom sensitivity wildly different between Chrome and Firefox
In Firefox, to zoom out from a range of 6 minutes to a range of 1 hour, takes 55(!) full drags of my scroll wheel (one drag is about 8 clicks, so about 440 clicks), which is way too insensitive.

On Chrome, it takes just 3 full drags.

For comparison, zooming by the same 10x factor on Google Maps takes just 13 clicks.

fbarl:
Resolves #2768 by taking the `defaultWheelDelta` logic from https://github.com/d3/d3-zoom.

Zooming transitions were made also animated to make them smoother (especially on Firefox which sometimes does discrete jumps).

**Edit**: Tracking of zoom events was wrapped into a debounce method to prevent excessive Mixpanel reporting.


rade:
The FF performance is probably #2302.

fbarl:
> The zoom track still seems a bit excessive. What do you hope to learn? If it's binary use vs non-use, maybe use some `once` logic, or otherwise massively turn up the debounce interval (request handling eats into drawing performance).

I hope to see at which *depth* the users are using the timeline, i.e. whether they are more interested in moving through minutes/hours/days/etc... With that in mind, I will increase `ZOOM_TRACK_DEBOUNCE_INTERVAL` from `1s` to `10s` and also replace the `durationPerPixelInMilliseconds` float with a rounded humanized time unit (`minutes`, `hours`, etc..) to make it easier to interpret the Mixpanel data.

fbarl:
@davkal I've updated the PR.

@rade Maybe you'd want to test the FF behaviour (modulo lagging) before I merge it.

rade:
> Maybe you'd want to test the FF behaviour (modulo lagging) before I merge it.

Nah, it can hardly have gotten worse ;)

PR #2779, opened 2017-07-31
synthesise k8s service network from service IPs

### LINKED ISSUE #2748: "Local Networks" is confusing
I don't really know what it is supposed to be, but maybe strip out the `/32` entries because they are not "networks"?

image

rade:
This prevents cluttering host.LocalNetworks with lots of /32 addresses. These were unsightly and rather distracting in the UI. They also bloated the report and slowed down server-side rendering.

Fixes #2748.

### DIFF
--- a/probe/kubernetes/reporter.go
+++ b/probe/kubernetes/reporter.go
@@ -2,6 +2,7 @@ package kubernetes
 
 import (
 	"fmt"
+	"net"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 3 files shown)


probe/kubernetes/reporter.go
```diff
@@ -294,23 +295,32 @@ func (r *Reporter) serviceTopology() (report.Topology, []Service, error) {
...
-//        The right way of fixing this is performing DNAT mapping on persistent
-//        connections for which we don't have a robust solution
-//        (see https://github.com/weaveworks/scope/issues/1491)
+// FIXME: Hideous hack to remove persistent-connection edges to
+// virtual service IPs attributed to the internet. The global
+// service-cluster-ip-range is not exposed by the API server (see
+// https://github.com/kubernetes/kubernetes/issues/25533), so instead
+// we synthesise it by computing the smallest network that contains
+// all service IPs. That network may be smaller than the actual range
+// but that is ok, since in the end all we care about is that it
```
2opremio: Since we show the networks in the UI, this will likely cause flapping when services are created and destroyed. This is arguably better than having a gazillion networks though.

rade: Not sure what you mean by flapping. Yes, the synthesized network may change as services come and go, but `host.LocalNetworks` will actually change less often than now, since atm *every* service coming/going changes that value.

report/networks.go
```diff
@@ -105,3 +105,70 @@ func ipv4Nets(addrs []net.Addr) []*net.IPNet {
...
+	}
+	if len(a) != len(b) {
+		return 0
+	}
+	// If IPv6, only up to the prefix (first 64 bits)
+	if len(a) > 8 {
+		a = a[:8]
+		b = b[:8]
+	}
+	for len(a) > 0 {
```
2opremio: This seems really convoluted. I would simply transform it to an 8 byte int and progressively shift a logical `&` mask. I am not sure if distinguishing full bytes and bits is worth the extra performance (If I were to use tricks I would probably use binary search starting with the middle and using precomputed masks).

rade: I copied this code from the golang stdlib so I didn't have to think about it.

rade: I've changed my mind. Simpler now.

report/networks.go
```diff
@@ -105,3 +106,31 @@ func ipv4Nets(addrs []net.Addr) []*net.IPNet {
...
+		Mask: net.CIDRMask(net.IPv4len*8, net.IPv4len*8),
+	}
+	for _, ip := range ips[1:] {
+		network.Mask = net.CIDRMask(commonIPv4PrefixLen(network.IP, ip), net.IPv4len*8)
+		network.IP = network.IP.Mask(network.Mask)
+	}
+	return &network
+}
+
+func commonIPv4PrefixLen(a, b net.IP) (cpl int) {
```
2opremio: I would make the return parameter explicit.

rade: At the cost of an extra line. Oh well.

Issue #2783, opened 2017-08-01
segfault in cronjob probing


rade:
This was reported by user on slack; they are running 1.6.1
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x169f35e]

goroutine 446 [running]:
github.com/weaveworks/scope/probe/kubernetes.(*cronJob).GetNode(0xc4262b7bc0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/weaveworks/scope/probe/kubernetes/cronjob.go:70 +0x19e
github.com/weaveworks/scope/probe/kubernetes.(*Reporter).cronJobTopology.func1(0x2f194a0, 0xc4262b7bc0, 0x2f194a0, 0xc4262b7bc0)
	/go/src/github.com/weaveworks/scope/probe/kubernetes/reporter.go:368 +0x64
github.com/weaveworks/scope/probe/kubernetes.(*client).WalkCronJobs(0xc4202b82a0, 0xc423bcdd40, 0x0, 0x0)
	/go/src/github.com/weaveworks/scope/probe/kubernetes/client.go:313 +0x1b5
github.com/weaveworks/scope/probe/kubernetes.(*Reporter).cronJobTopology(0xc42026f0a0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/weaveworks/scope/probe/kubernetes/reporter.go:371 +0x3f1
github.com/weaveworks/scope/probe/kubernetes.(*Reporter).Report(0xc42026f0a0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/weaveworks/scope/probe/kubernetes/reporter.go:253 +0x26d
github.com/weaveworks/scope/probe.(*Probe).report.func1(0xc42040a510, 0xc4211f57a0, 0x2f01dc0, 0xc42026f0a0)
	/go/src/github.com/weaveworks/scope/probe/probe.go:155 +0x118
created by github.com/weaveworks/scope/probe.(*Probe).report
/go/src/github.com/weaveworks/scope/probe/probe.go:165 +0xec
```

PR #2785, opened 2017-08-01
k8s probe: Fix a panic (nil pointer deref) when a cronjob has never been scheduled

ekimekim:
in which case cj.Status.LastScheduled is nil.
New behaviour is to omit it from the map (and therefore the display) if it has never been scheduled.

Fixes https://github.com/weaveworks/scope/issues/2783


ekimekim:
Tested with and without a value for LastScheduled.

Issue #2787, opened 2017-08-02
Nodes layout zoom sensitivity differs between Firefox and Chrome


fbarl:
See #2768. As long as we are using `d3-zoom`, this will be a problem. However, we could make `ZoomableCanvas` use the same triggers and constants as `TimeTravelTimeline` does for zooming, which would make the sensitivity consistent between the browsers and the zooming behaviour consistent between different controls in our app (and would also give us more control over zooming).

PR #2788, opened 2017-08-02
Balance timeline zooming sensitivity between Firefox and Chrome

### LINKED ISSUE #2768: Time Travel: zoom sensitivity wildly different between Chrome and Firefox

fbarl:
Resolves #2768.

If this will behave nicely, consider extending the same logic to #2787.

### DIFF
--- a/client/app/scripts/utils/zoom-utils.js
+++ b/client/app/scripts/utils/zoom-utils.js
@@ -1,5 +1,15 @@
 
+const ZOOM_SENSITIVITY = 0.0025;
+const DOM_DELTA_LINE = 1;
+
 // See https://github.com/d3/d3-zoom/blob/807f02c7a5fe496fbd08cc3417b62905a8ce95fa/src/zoom.js
-export function defaultWheelDelta(ev) {
-  return ev.deltaY * (ev.deltaMode ? 15 : 1) * 0.0035;
+function wheelDelta(ev) {
+  // Only Firefox seems to use the line unit (which we assume to
+  // be 25px), otherwise the delta is already measured in pixels.
+  const unitInPixels = (ev.deltaMode === DOM_DELTA_LINE ? 25 : 1);
+  return ev.deltaY * unitInPixels * ZOOM_SENSITIVITY;
+}
+
+export function zoomFactor(ev) {
+  return Math.exp(wheelDelta(ev));
 }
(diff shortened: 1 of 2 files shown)


client/app/scripts/utils/zoom-utils.js
```diff
@@ -1,5 +1,5 @@
 
 // See https://github.com/d3/d3-zoom/blob/807f02c7a5fe496fbd08cc3417b62905a8ce95fa/src/zoom.js
 export function defaultWheelDelta(ev) {
-  return ev.deltaY * (ev.deltaMode ? 120 : 1) * 0.002;
+  return ev.deltaY * (ev.deltaMode ? 15 : 1) * 0.0035;
```
rndstr: Maybe add a comment to elaborate that we are checking `deltaMode` to see whether `deltaY` is in pixel or lines. It seems like only FF uses `1` (`DOM_DELTA_LINE`) while other browsers use `0` (`DOM_DELTA_PIXEL`).

rndstr:
This still has a large discrepancy in sensitivity between FF (lower) and Chromium (higher) for me.

A line-to-pixel ratio in the area of 35 puts it at about the same sensitivity. This would be roughly double what you suggest so maybe there is some fast-scrolling setting in effect for your env (https://stackoverflow.com/a/37474225 suggests that Windows uses 2x).

client/app/scripts/utils/zoom-utils.js
fbarl: Yup, makes sense!

fbarl:
@rndstr After reading about this more, I came to conclusion that any attempt to balance the cross-browser sensitivity perfectly is futile because it will always depend on the platform and some special settings.

Although we seem to be using very similar browsers/environments, line-to-pixel ratio for you seems to be `35px` while for me it's `15px`. I changed it to `25px` in the code and I'd say that if the real sensitivity on browsers will be +/- 50% from that value, that will still result in a good UX.

PR #2793, opened 2017-08-03
Fix timeline label vertical displacement on some Chromes

### LINKED ISSUE #2789: Time Travel: timeline labels are cut off at bottom on Chrome
time-travel-line-height-chromium

I played around with `line-height` which would fix the issue for Chrome (but possibly break it for FF) 

OS | Linux
Chromium | 59.0.3071.115 (Developer Build) (64-bit)
&&
Google Chrome | 60.0.3112.66 (Official Build) beta (64-bit)

(Looks fine in FF)

rndstr:
Fixes #2789

### DIFF
--- a/client/app/scripts/components/time-travel-timeline.js
+++ b/client/app/scripts/components/time-travel-timeline.js
@@ -289,7 +289,7 @@ class TimeTravelTimeline extends React.Component {
       <g transform={`translate(${position}, 0)`} key={timestamp.format()}>
         {!isBehind && <line y2="75" stroke="#ddd" strokeWidth="1" />}
         {!disabled && <title>Jump to {timestamp.utc().format()}</title>}
-        <foreignObject width="100" height="20">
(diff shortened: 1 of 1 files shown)


client/app/scripts/components/time-travel-timeline.js
```diff
@@ -284,12 +284,14 @@ class TimeTravelTimeline extends React.Component {
     // Ticks are disabled if they are in the future or if they are too transparent.
     const disabled = timestamp.isAfter(this.state.timestampNow) || opacity < 0.4;
     const handleClick = () => this.jumpTo(timestamp);
+    // Fixes Chrome effective line-height miscalculation in certain environments
+    const foreignObjectStyle = { lineHeight: '150%' };
```
fbarl: I think a bit more robust solution would be to set line height equal to the height of the foreign object, i.e. `{ lineHeight: '20px' }`. That would be a bit more self-explanatory and it would still work well if the font size ever changed.

PR #2531, opened 2017-05-22
Release 1.4.0

2opremio:
@alban @fbarl @davkal @foot @ekimekim @bowenli @jpellizzari @abuehrle @pidster @errordeveloper @rade Please chip in by testing this release branch in the scenarios below and tick off what you tested afterwards (including your name at the end). Have a look at the CHANGELOG to prioritize your testing.

To make things easier to test I have pushed the image `weaveworks/scope:release-1.4-5af0a6d`
Thanks!

- [ ] Start a 2 node cluster and leave a UI connected, watch for performance regression
- [x] Run a Scope cluster on ECS [ML]
- [x] Run Scope on https://github.com/microservices-demo/microservices-demo/ [SW]
- [x] Run Scope on Docker For Mac [SW]
- [x] Start and stop Scope in standalone mode [AA]
  - [x] Linux [AA]
  - [ ] docker-machine
- [x] Connect probes from release candidate to Weave Cloud [SW]
  - [x] Local scope in probe-only mode connects to Weave Cloud and renders correctly
    - [x] dev
    - [x] prod
  - [x] Controls appear for containers and function correctly
    - [x] dev
    - [x] prod
  - [x] No new errors in logging or telemetry
    - [x] dev
    - [x] prod
- [ ] UI integration
  - [ ] Hot-reloading `cd client && npm i && npm start` and open http://localhost:4042
  - [ ] UI under prefix path  `cd client && npm run build && npm start-production` and open http://localhost:4043/scoped/
  - [ ] URL parameters: Run previous release and click around, keep UI open, start new release, reload page
- [x] UI features
  (To test: clone [microservices demo](https://github.com/microservices-demo/microservices-demo) and run `cd deploy/docker-compose && docker-compose up`)
  - [x] General: switch between topologies, see nodes and edges, details panel [AA]
  - [x] Zoom and pan, switch between topologies [AA]
  - [x] Topology filters: namespaces, system containers, etc. [SW]
  - [x] Metrics on canvas, try pinning/unpinning [SW]
  - [x] Search [AA]
    - [x] by container name
    - [x] by usage `cpu < 2%`
  - [x] Terminal
    - [x] Attach to container to see logs [SW]
    - [x] Exec to container and type `ls` [SW]
    - [x] Pop out terminal [SW]
  - [x] Contrast mode [AA] (*NOT PRESENT ANYMORE?* [SW])
  - [x] Download report [SW]
  - [x] Pause button [SW]
[...]

PR #2794, opened 2017-08-04
Fix rendering of exported SVG

### LINKED ISSUE #2534: SVG exporting is sometimes broken
From https://github.com/weaveworks/scope/pull/2531#issuecomment-303087988

svg-error

Sample report, with resulting broken SVG: [report-svg.zip](https://github.com/weaveworks/scope/files/1021535/report-svg.zip)

rndstr:
Replace invalid characters in attribute value; fixes the error

        Unescaped '<' not allowed in attributes values

---
Fixes #2534

There are a bunch of other errors when [validating](https://validator.w3.org/#validate_by_upload) the SVG but renderers might be okay with these

    element "div" undefined
    element "span" undefined

@pidster would you mind opening the following SVG file in the browser where you get the other error and see whether it works now? [weave-cloud-1501839338756.svg.zip](https://github.com/weaveworks/scope/files/1199912/weave-cloud-1501839338756.svg.zip)

### DIFF
--- a/client/app/scripts/charts/edge.js
+++ b/client/app/scripts/charts/edge.js
@@ -3,6 +3,7 @@ import { connect } from 'react-redux';
 import classNames from 'classnames';
 
 import { enterEdge, leaveEdge } from '../actions/app-actions';
+import { encodeIdAttribute, decodeIdAttribute } from '../utils/dom-utils';
 
 class Edge extends React.Component {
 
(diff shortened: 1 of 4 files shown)


rade:
It was @2opremio that reported the problem, not @pidster.

The original report gives me exactly the shown error when opening in Chrome. Opening the above report works fine, so I think this is fixed.

rndstr:
@rade thanks for confirming (the comment that reported it was by @pidster https://github.com/weaveworks/scope/pull/2531#issuecomment-303087988)

pidster:
LGTM

PR #2795, opened 2017-08-04
Improve Firefox performance

### LINKED ISSUE #2431: Graph edges are visible through the nodes in the background

fbarl:
Improves on #2302.

##### Changes

* Added a non-reactive semi-transparent layer between background and foreground nodes and edges in the graph view, removing excessive opacity transitions for individual elements (which also fixes #2431).
* The semi-transparent layer mentioned above is now applied instantly (without a transition) to avoid the glitches and is made a bit lighter (I thought it looked better that way, but that can easily be reverted).
* Made default *spring* animation a bit faster so that it appears smoother and more reactive on Firefox.
* Decreased the graph animation treshold again to make it bearable on Firefox.

### DIFF
--- a/client/app/scripts/charts/node-container.js
+++ b/client/app/scripts/charts/node-container.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import { omit } from 'lodash';
 import { Motion, spring } from 'react-motion';
 
 import { NODES_SPRING_ANIMATION_CONFIG } from '../constants/animation';
@@ -16,8 +15,7 @@ const transformedNode = (otherProps, { x, y, k }) => (
 
 export default class NodeContainer extends React.PureComponent {
   render() {
-    const { dx, dy, isAnimated, scale } = this.props;
-    const forwardedProps = omit(this.props, 'dx', 'dy', 'isAnimated', 'scale');
+    const { dx, dy, isAnimated, scale, ...forwardedProps } = this.props;
 
     if (!isAnimated) {
       // Show static node for optimized rendering
--- a/client/app/scripts/charts/node-container.js
+++ b/client/app/scripts/charts/node-container.js
@@ -6,24 +6,22 @@ import { NODES_SPRING_ANIMATION_CONFIG } from '../constants/animation';
 import Node from './node';
 
 
-const transformedNode = (otherProps, { x, y, k, opacity }) => (
+const transformedNode = (otherProps, { x, y, k }) => (
   // NOTE: Controlling blurring and transform from here seems to re-render
   // faster than adding a CSS class and controlling it from there.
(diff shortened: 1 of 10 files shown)


rade:
This supersedes #2794, right?

rade:
> This supersedes #2794, right?

Err, I meant #2432.

rade:
Some of the commits should probably be squashed.

fbarl:
> This supersedes #2432, right?

Yes, good catch @rade!

> Some of the commits should probably be squashed.

I actually always squash all of my commits into one when merging a PR - it would make it way more readable when I was working on big PRs of 20+ commits, so I just adopted that practice.

client/app/scripts/charts/node-container.js
```diff
@@ -6,24 +6,22 @@ import { NODES_SPRING_ANIMATION_CONFIG } from '../constants/animation';
...
 );
 
 export default class NodeContainer extends React.PureComponent {
   render() {
-    const { dx, dy, isAnimated, scale, blurred, contrastMode } = this.props;
-    const nodeBlurOpacity = contrastMode ? 0.6 : 0.25;
-    const forwardedProps = omit(this.props, 'dx', 'dy', 'isAnimated', 'scale', 'blurred');
-    const opacity = blurred ? nodeBlurOpacity : 1;
+    const { dx, dy, isAnimated, scale } = this.props;
+    const forwardedProps = omit(this.props, 'dx', 'dy', 'isAnimated', 'scale');
```
fbarl: Sweet! :)

PR #2799, opened 2017-08-08
Time Travel Docs

abuehrle:
@fbarl I did a very rudimentary pass at the description for this feature.  What more should we add? 
@bia  is that screen capture ok? 

Note also, that I want rearrange these docs so that they are more like Net docs and will add in a tasks section where we can expand on this feature and the resources.


bia:
Hi Anita, it would be best to always refer to Time Travel with consistent capitalisation throughout the description. Sometimes you write 'Time travel' or 'time travel,' but Filip and I both prefer 'Time Travel.'

The screenshot that you have taken is not the most current version - the marker of the time being displayed should be red. Will you style the screen capture as you have [here](https://www.weave.works/docs/scope/latest/features/)?

abuehrle:
@bia thanks. good feedback. I will change the capitalization.  Can you please send me an updated screen capture or invite me to an instance with the current version. I took this using the dev server with a feature flag.

abuehrle:
moved this to weave cloud docs

PR #2801, opened 2017-08-08
Time Travel: unmount in the shutdown() action

fbarl:
Resolves https://github.com/weaveworks/service-ui/issues/814 by making timeline deconstruction part of the `shutdown()` action, avoiding the attempt to establish new connections.

Also, triggering `resumeTime` action only if previously paused.

### DIFF
--- a/client/app/scripts/actions/app-actions.js
+++ b/client/app/scripts/actions/app-actions.js
@@ -798,10 +798,16 @@ export function changeInstance() {
 }
 
 export function shutdown() {
-  stopPolling();
-  teardownWebsockets();
-  return {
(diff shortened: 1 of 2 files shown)


jpellizzari:
LGTM, without any context, but still have the [WIP] tag

fbarl:
@jpellizzari PTAL

> LGTM, without any context, but still has the [WIP] tag.

I think the first commit accidentaly solved the case described in the referenced issue, but you'd still get the same bug if you tried clicking on Time Travel before switching to *Monitor* - now that part has been fixed as well.

p.s. the branch name is still a bit of a misnomer :)

PR #2803, opened 2017-08-09
Make nodes graph animations a bit faster

fbarl:
From Slack conversation (referring to animation speed changes in #2795):

```
matthias [11:52 PM] 
TBH, it's a bit too snappy for me now. 
...
matthias [11:54 PM] 
does it not defeat the purpose of the animations?
...
jordan
[11:55 PM] 
yeah might be a little fast
...
matthias [11:57 PM] 
I don't think animations need to be quite as lethargic as before. somewhere in between.
```


rndstr:
@rade asked me to take a look and for me when comparing to the previous speed it *is* an improvement. The animations I couldn't follow previously are now a bit easier on the eye.

PR #2810, opened 2017-08-11
Make Scope's URL message more precise

### LINKED ISSUE #1678: scope tells me it's available at URLs that it's not available at
Running scope 0.16.1 on my Mac using docker-machine, I see the following:

```
$ ./scope launch
c46ebb9bc663573b9ca39000352cba4da5184c5ad61c9450f39ad37757298527
  * http://10.32.0.1:4040/
  * http://192.168.99.100:4040/
  * http://10.0.2.15:4040/
```

However, when I click on these URLs, I get "address unreachable" errors for the first and third. Only the second URL is accessible.

Ideally, Scope would only tell me about URLs that I can actually reach. Failing that, the wording should be amended to be more truthful.

2opremio:
Doesn't work just yet since the curl commands get a connection refused since the App is not ready.

Fixes #1678

### DIFF
--- a/scope
+++ b/scope
@@ -201,8 +201,8 @@ launch() {
 print_app_endpoints() {
     HOST_SUFFIX=""
     if [ -n "${DOCKER_HOST+x}" ]; then
-	DOCKER_HOSTNAME=$(docker run --rm $USERNS_HOST --net=host --entrypoint /bin/sh "$SCOPE_IMAGE" -c hostname)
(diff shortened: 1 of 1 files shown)


rade:
This might be heading in the same direction as [`wait_for_status` in Weave Net](https://github.com/weaveworks/weave/blob/v2.0.1/weave#L663), the main difference being that we want to try several addresses.

PR #2811, opened 2017-08-14
Remove whitespace from empty connection lists

rndstr:
Removes empty container that introduce unnecessary margin
scope-details-whitespace


rndstr:
Good point, maybe something like

> INBOUND
> No connections
> 
> OUTBOUND
> No connections

PR #2812, opened 2017-08-15
Unified zooming sensitivity

### LINKED ISSUE #2787: Nodes layout zoom sensitivity differs between Firefox and Chrome

### LINKED ISSUE #2809: Do not allow for node graph to completely leave the viewport
I have "lost" the graph more than once and had to reload.

fbarl:
Replaces all `d3-zoom` code with our own handlers for zooming/panning events in order to have unified behaviour between all components and across all browsers. Having our own logic also gives us better control over the behaviour.

##### Changes

  * Use the same zooming logic in `ZoomableCanvas` as in `TimeTravelTimeline` (fixes #2787).
  * Use two modes of translation extent for the nodes canvas:
    * `CONTENT_COVERING` - the nodes content bounding box must alwas cover the viewport. This is the case with the resource view, where the content is _sticky_ so that there is never empty space to its left or right.
    * `CONTENT_INCLUDED` - this is introduced as a new mode for the nodes graph to fix #2809. Translation is bounded in such a way that a part of the nodes graph bounding box must always be in the viewport.
  * `ZoomableCanvas` now always forwards the transform instead of applying it on itself for consistency.

##### TODOs

  * Ideally, we should split `ZoomableCanvas` into the caching component and the zooming component, so that the _pure_ zooming component could also be used in `TimeTravelTimeline` and that the whole zooming logic is indeed in one place.

### DIFF
--- a/client/app/scripts/components/zoomable-canvas.js
+++ b/client/app/scripts/components/zoomable-canvas.js
@@ -1,14 +1,15 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import { debounce, pick } from 'lodash';
+import { clamp, debounce, pick } from 'lodash';
 import { fromJS } from 'immutable';
 
+import { drag } from 'd3-drag';
 import { event as d3Event, select } from 'd3-selection';
-import { zoom, zoomIdentity } from 'd3-zoom';
 
 import Logo from '../components/logo';
 import ZoomControl from '../components/zoom-control';
 import { cacheZoomState } from '../actions/app-actions';
+import { zoomFactor } from '../utils/zoom-utils';
 import { transformToString } from '../utils/transform-utils';
 import { activeTopologyZoomCacheKeyPathSelector } from '../selectors/zooming';
 import {
@@ -19,6 +20,15 @@ import {
 
 import { ZOOM_CACHE_DEBOUNCE_INTERVAL } from '../constants/timer';
 
+const transformF = ({ x, y }, t) => ({
+  x: t.translateX + (t.scaleX * x),
+  y: t.translateY + (t.scaleY * y),
+});
+
+const inverseTransform = ({ x, y }, t) => ({
+  x: (x - t.translateX) / t.scaleX,
+  y: (y - t.translateY) / t.scaleY,
+});
 
 class ZoomableCanvas extends React.Component {
   constructor(props, context) {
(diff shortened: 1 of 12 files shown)


fbarl:
[Unify zooming sensitivity in Scope](https://trello.com/c/ajhhHXvx/29-unify-zooming-sensitivity-in-scope)

client/app/scripts/components/zoomable-canvas.js
```diff
@@ -144,64 +133,122 @@ class ZoomableCanvas extends React.Component {
...
+  handlePanStart() {
+    this.setState({ isPanning: true });
+  }
+
+  handlePanEnd() {
+    this.setState({ isPanning: false });
+  }
+
+  handlePan() {
+    let state = { ...this.state };
```
fbarl: Good catch! Although note that this has trivial complexity compared to all the rerendering stuff that follows each panning event :)

fbarl:
> * Chrome+OSX zooming on canvas now elastically bounces the viewport a little. (Maybe needs a prevDefault somewhere.)

I added it at the end of `handleZoom` like it is in `time-travel-timeline`, hopefully that solves the problem!

> * Timeline zooming is very jerky now on Firefox.
>   * master/what-is-currently-on-fdww is quite smooth under Firefox.
> * Resource zooming no longer silky smooth in chrome.

Unfortunately I wasn't able to reproduce neither of the two on my machine - the performance on both Chrome and Firefox hasn't changed for me compared to _master_. That makes me think those two might also be OS X specific. Maybe we can look into them together tomorrow?

I hope that the resource one might already be resolved by the newly added `ev.preventDefault()`, otherwise I'm not so sure what's causing it. As for timeline zooming, that one's a complete mystery as the only thing that has changed there is `x` -> `-x` and `x` -> `1 / x`. Maybe inverse is an expensive operation on OS X :thinking:

foot:
Ah! Interesting, might be dev vs prod builds, will test in a bit.

fbarl:
> ...might be dev vs prod builds, will test in a bit.

Yeah, I was also comparing dev-to-dev & prod-to-prod... found no diff :)

foot:
Yes. w/ `build-prod` + `start-prod` you get the speed back. 👌 

LGTM

PR #2815, opened 2017-08-15
scope/cortex: fix typo in query filter

rndstr:
& slight refactor

### DIFF
--- a/render/detailed/links.go
+++ b/render/detailed/links.go
@@ -58,8 +58,8 @@ var (
 			[]string{docker.MemoryUsage, docker.CPUTotalUsage},
 		),
 		// Pod naming: // https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pod-template-hash-label
-		"__k8s_controllers": formatMetricQueries(`pod_name=~"^{{label}}-[^-]+-[^-]+$"}`, []string{docker.MemoryUsage, docker.CPUTotalUsage}),
-		report.DaemonSet:    formatMetricQueries(`pod_name=~"^{{label}}-[^-]+$"}`, []string{docker.MemoryUsage, docker.CPUTotalUsage}),
+		"__k8s_controllers": formatMetricQueries(`pod_name=~"^{{label}}-[^-]+-[^-]+$"`, []string{docker.MemoryUsage, docker.CPUTotalUsage}),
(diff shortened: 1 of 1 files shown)


render/detailed/links.go
```diff
@@ -58,8 +58,8 @@ var (
 			[]string{docker.MemoryUsage, docker.CPUTotalUsage},
 		),
 		// Pod naming: // https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pod-template-hash-label
-		"__k8s_controllers": formatMetricQueries(`pod_name=~"^{{label}}-[^-]+-[^-]+$"}`, []string{docker.MemoryUsage, docker.CPUTotalUsage}),
-		report.DaemonSet:    formatMetricQueries(`pod_name=~"^{{label}}-[^-]+$"}`, []string{docker.MemoryUsage, docker.CPUTotalUsage}),
+		"__k8s_controllers": formatMetricQueries(`pod_name=~"^{{label}}-[^-]+-[^-]+$"`, []string{docker.MemoryUsage, docker.CPUTotalUsage}),
+		report.DaemonSet:    formatMetricQueries(`pod_name=~"^{{label}}-[^-]+$"`, []string{docker.MemoryUsage, docker.CPUTotalUsage}),
```
rade: daemonsets are part of controllers, so why do they show up separately here?

Related, surely the list of `k8sControllers` should be refactored and made the same as in `KubeControllerRenderer`.

rndstr: Since `DaemonSet` does not share the same pod name filter as the other controller-types, I got rid of the term `controllers` here altogether.

render/detailed/links.go
```diff
@@ -155,11 +155,7 @@ func RenderMetricURLs(summary NodeSummary, n report.Node, metricsGraphURL string
 
 // metricQuery returns the query for the given node and metric.
 func metricQuery(summary NodeSummary, n report.Node, metricID string) string {
-	t := n.Topology
-	if _, ok := k8sControllers[n.Topology]; ok {
-		t = "__k8s_controllers"
-	}
-	queries := topologyQueries[t]
+	queries := topologyQueries[n.Topology]
```
rndstr: This method to return "" and the eventual metric URL to be empty as well

PR #2819, opened 2017-08-16
Fix query filters by adding namespaces and using docker container name

rndstr:


render/detailed/links.go
```diff
@@ -82,17 +83,17 @@ func formatMetricQueries(filter string, ids []string) map[string]string {
 		// All  `container_*`metrics  are provided by cAdvisor in Kubelets
 		switch id {
 		case docker.MemoryUsage:
-			queries[id] = fmt.Sprintf("sum(container_memory_usage_bytes{%s})/1024/1024", filter)
+			queries[id] = fmt.Sprintf("sum(container_memory_usage_bytes{%s})", filter)
```
2opremio: Why the change of units?

render/detailed/links.go
```diff
@@ -160,7 +161,19 @@ func metricQuery(summary NodeSummary, n report.Node, metricID string) string {
...
+	namespace, _ := n.Latest.Lookup(kubernetes.Namespace)
+	name, _ := n.Latest.Lookup(docker.ContainerName)
+	replacements := map[string]string{
+		"{{label}}":         summary.Label,
+		"{{namespace}}":     namespace,
+		"{{containerName}}": name,
+	}
+
+	q := queries[metricID]
+	for old, new := range replacements {
```
2opremio: What about using a `Replacer` for this? https://golang.org/pkg/strings/#Replacer

2opremio:
LGTM, only an optional tweak and a question.

render/detailed/links.go
rndstr: My initial idea was to keep it the same unit as the metric you are clicking on (which is MB) since the hover of the graph displays always bytes which are hard to decipher.

But it turns out to be weird in the graph once you have GBs and the axis labels show `16K` (but it's actually `16GB`). The proper place to improve this would be the graph itself an a way to attach units.

PR #2824, opened 2017-08-17
Fix processes/hosts table not appearing

### LINKED ISSUE #2822: Table mode: unable to see processes
It was working before, then I refreshed the window and I just get a blank page:

image

Switching to topology view I see lots of processes.

URL is https://cloud.weave.works/app/loud-breeze-77#!/state/{"controlPipe":null,"nodeDetails":[],"topologyViewMode":"grid","pinnedSearches":[],"searchQuery":"","gridSortedBy":"docker_cpu_total_usage","gridSortedDesc":true,"topologyId":"processes","topologyOptions":{"processes":{"unconnected":["hide"]},"containers":{"system":["application"],"stopped":["running"],"pseudo":["hide"],"namespace":[""]},"pods":{"pseudo":["hide"],"namespace":[""]}},"contrastMode":false}

rndstr:
Closes #2822

### DIFF
--- a/client/app/scripts/components/cloud-link.js
+++ b/client/app/scripts/components/cloud-link.js
@@ -57,7 +57,7 @@ class LinkWrapper extends React.Component {
   render() {
     const { url, children, ...props } = this.props;
     if (!url) {
-      return children;
+      return React.isValidElement(children) ? children : (<span>{children}</span>);
     }
 
     const href = this.buildHref(url);


client/app/scripts/components/cloud-link.js
```diff
@@ -57,7 +57,7 @@ class LinkWrapper extends React.Component {
   render() {
     const { url, children, ...props } = this.props;
     if (!url) {
-      return children;
+      return React.isValidElement(children) ? children : (<span>{children}</span>);
```
jpellizzari: Under what circumstances would this not be valid? This still wraps an invalid element in span tags.

rndstr: That would be a number and by being wrapped it becomes valid.

jpellizzari: 👍

PR #2828, opened 2017-08-18
Update topologies and namespaces when time travelling

### LINKED ISSUE #2802: Time Travel: namespace selection does not respect time
This is a picture of Weave Cloud dev from about a year ago...
screenshot from 2017-08-08 23-44-20
Notice that I've selected the 'cortex' namespace. But at that time no cortex namespace existed, so it shouldn't really be shown. And even if it was shown, selecting it certainly shouldn't display any nodes. The graph that is displayed actually shows *all* nodes.

fbarl:
Resolves #2802.

##### Changes

* Unblock the calls to topology when time travelling, triggering updates of both *topology selector* and *topology options*. Disappearance of the selected option is handled a bit differently between the two:
  * The topology selection never changes over time, meaning that e.g. if we select *Processes* and go back in time to the point when that topology wasn't there, we'd still keep it selected but show it as empty. This is to avoid confusion with context switching and also because not sure which one we should default to.
  * The namespace selection is taken as an intersection of the full selection and the currently available options. If the intersection is empty, we default to the default option (in case of namespaces, that's the _All namespaces_ option). Here we use the fact that we have somewhere to default to, but in a long run, this logic should probably be consistent with how topologies selection is handled.

### DIFF
--- a/client/app/scripts/components/topology-options.js
+++ b/client/app/scripts/components/topology-options.js
@@ -83,6 +83,7 @@ class TopologyOptions extends React.Component {
     // TODO: This logic should probably be made consistent with how topology
     // selection is handled when time travelling, especially when the name-
     // spaces are brought under category selection.
+    // TODO: Consider extracting this into a global selector.
     let activeValue = option.get('defaultValue');
     if (activeOptions && activeOptions.has(optionId)) {
       const activeSelection = makeSet(activeOptions.get(optionId));
--- a/client/app/scripts/components/topology-options.js
+++ b/client/app/scripts/components/topology-options.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import { Map as makeMap } from 'immutable';
+import { Set as makeSet, Map as makeMap } from 'immutable';
 import includes from 'lodash/includes';
 
 import { trackMixpanelEvent } from '../utils/tracking-utils';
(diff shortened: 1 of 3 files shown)


client/app/scripts/utils/web-api-utils.js
```diff
@@ -235,20 +241,19 @@ export function getResourceViewNodesSnapshot(state, dispatch) {
   getNodesForTopologies(state, dispatch, topologyIds);
 }
 
-// NOTE: getState is called every time to make sure the up-to-date state is used.
-export function getTopologies(getState, dispatch, initialPoll = false) {
+function pollTopologies(getState, dispatch, initialPoll = false) {
```
foot: Would it be tricky to write `pollTopologies` in terms of `getTopologiesOnce`?

I'm trying to get over my DRYOCD so maybe this is not important. `reqwest` seems to support Promises but we're not using them very much.

client/app/scripts/components/topology-options.js
```diff
@@ -73,9 +73,26 @@ class TopologyOptions extends React.Component {
...
+    // spaces are brought under category selection.
+    let activeValue = option.get('defaultValue');
+    if (activeOptions && activeOptions.has(optionId)) {
+      const activeSelection = makeSet(activeOptions.get(optionId));
+      const availableOptions = makeSet(option.get('options').map(o => o.get('value')));
+      const intersection = activeSelection.intersect(availableOptions);
+      if (!intersection.isEmpty()) {
+        activeValue = intersection.toJS();
+      }
+    }
```
foot: This stuff feels ripe to be pulled out into a selector!

client/app/scripts/utils/web-api-utils.js
fbarl: Probably it wouldn't be that hard to put it all in one method especially with help of Promises, but my thinking was the following:

* There is very little redundancy right now in `getTopologiesOnce` as all the complex logic is behind `topologiesUrl`, `receiveTopologies` and `receiveError`.
* I wanted to follow the example of `getNodes` and to make an early split between frozen and polling state consistent across our API calls (maybe this pattern could be generalized somehow).
* I think the logical split between polling/websockets and one-time calls is good as it enables us to manage them independently - e.g. what if we wanted to make `pollTopologies` use websockets in the future?

Also, doing it like this was simpler for now, I am aware that big parts of this code should probably be redone from scratch :)

client/app/scripts/components/topology-options.js
fbarl: Yeah, it certainly feels ripe for a selector - I'll try pulling it out!

fbarl: Actually I don't think it's worth the effort at this point (it would have to look a bit different as a global selector, working with all the options and not per option like here), but I'll add a _TODO_ for the future!

PR #2830, opened 2017-08-21
Use Node 8.4 for builds

jpellizzari:
Related to https://github.com/weaveworks/service-ui/issues/930

Builds should be a little faster. It may also be possible to turn off some transpiling features, since Node supports new ES syntax.

### DIFF
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -1,5 +1,4 @@
-FROM node:6.9.0
-RUN npm install -g yarn
+FROM node:8.4.0
 WORKDIR /home/weave
 COPY package.json yarn.lock /home/weave/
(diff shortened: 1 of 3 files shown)


client/Dockerfile
```diff
@@ -1,5 +1,4 @@
-FROM node:6.9.0
-RUN npm install -g yarn
```
fbarl: Nice that it's built in now! :)

client/package.json
```diff
@@ -128,6 +128,6 @@
     ]
   },
   "engines": {
-    "node": "^6.9.0"
+    "node": "^8.4.0"
```
jpellizzari: Scope is being installed as a dependency, `service-ui` is not (at least that was my thinking at the time). `service-ui` has the engines key added back in now.

PR #2836, opened 2017-08-30
Fix phrasing in readme

johananl:
The sentence "Easily to find the container using the most CPU or memory for a given host or service" should probably be "*Easily find* the container using the most CPU or memory for a given host or service".


rade:
Thanks!

PR #2837, opened 2017-08-31
make circleci ui-upload unconditional

rade:
It doesn't depend on `$DOCKER_USER`. Also, because it was tagged on without a `&&`, it was masking errors in the dockerhub image uploads.


aaron7:
Updated DOCKER_USER and DOCKER_PASS to weaveworksci.

PR #2843, opened 2017-09-01
More precise compose instructions

rade:


site/installing.md
```diff
@@ -204,7 +206,7 @@ Where,
 
 * `SCOPE_SERVICE_TOKEN=<token>` is the token you obtained after you signed up for Weave Cloud.
 
-Note that you will need to launch Scope onto every node that you want to monitor.
+Note that you will need to launch Scope on every node that you want to monitor.
```
abuehrle: you can use the new note format here: 

`> note:`  and it will add a shaded box around this. 

I

rade: I'll leave the formatting changes to somebody more qualified, i.e. you :)

abuehrle: Ok I'll fix that up later.  We need to add V3 here as well...

PR #2851, opened 2017-09-11
Fix edges disappearing in graph mode

### LINKED ISSUE #2850: Graph view edges sometimes disappear
@rade noticed in #2848 that _occasionally all the edges disappear_, but probably the issue has been there for a while.

edges-missing

fbarl:
Fixes #2850.

It seems that under some circumstances `isAnimated` prop would turn from _false_ to _true_ after the `waypoints` would be fixed, so `prepareWaypointsForMotion` would never be called given how the conditionals were set, giving an empty `waypointsMap` and consequently an ill-defined spline path for animated edges.

##### Changes

* Simplified and broadened the conditions for `waypointsMap` recalculation to make sure they are called when necessary (but not making excessive calls).
* Made helper functions for `spring` to avoid repeated code.

### DIFF
--- a/client/app/scripts/charts/edge-container.js
+++ b/client/app/scripts/charts/edge-container.js
@@ -1,8 +1,8 @@
 import React from 'react';
 import { Motion } from 'react-motion';
-import { fromJS, Map as makeMap } from 'immutable';
+import { Repeat, fromJS, Map as makeMap } from 'immutable';
 import { line, curveBasis } from 'd3-shape';
-import { times, constant } from 'lodash';
+import { times } from 'lodash';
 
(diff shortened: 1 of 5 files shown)


foot:
Is there an easy way to repro this one?

rade:
I see this all the time in Chrome on Linux when I

1. go to https://cloud.weave.works/app/cold-sky-72
2. click 'bug' -> 'Reset your local view state' (this should bring up the table view of 'Controllers')
3. click graph view

The edges show up for a few seconds and then disappear.

foot:
Perfect, ta.

PR #2854, opened 2017-09-18
Adds ECS Cluster Region option

galindro:
### Propourse

As talked with @ekimekim [on slack](https://weave-community.slack.com/archives/C2ND76PAA/p1505756045000138), this PR adds an ECS Cluster Region option.

### Usage
``` bash
./scope -probe.ecs.cluster.region us-east-1
```

The default behaviour is to get the region from the instance metadata.

### DIFF
--- a/probe/awsecs/client.go
+++ b/probe/awsecs/client.go
@@ -70,16 +70,19 @@ type EcsInfo struct {
 	TaskServiceMap map[string]string
 }
 
-func newClient(cluster string, cacheSize int, cacheExpiry time.Duration) (EcsClient, error) {
+func newClient(cluster string, cacheSize int, cacheExpiry time.Duration, clusterRegion string) (EcsClient, error) {
 	sess := session.New()
+	var err error
 
-	region, err := ec2metadata.New(sess).Region()
-	if err != nil {
-		return nil, err
+	if clusterRegion == "" {
+		clusterRegion, err = ec2metadata.New(sess).Region()
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	return &ecsClientImpl{
-		client:       ecs.New(sess, &aws.Config{Region: aws.String(region)}),
+		client:       ecs.New(sess, &aws.Config{Region: aws.String(clusterRegion)}),
 		cluster:      cluster,
 		taskCache:    gcache.New(cacheSize).LRU().Expiration(cacheExpiry).Build(),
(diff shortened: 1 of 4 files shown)


prog/main.go
```diff
@@ -29,11 +29,12 @@ var (
 	// set at build time
 	version = "dev"
 	// tokens to be elided when logging
-	serviceTokenFlag       = "service-token"
-	probeTokenFlag         = "probe.token"
-	kubernetesPasswordFlag = "probe.kubernetes.password"
-	kubernetesTokenFlag    = "probe.kubernetes.token"
-	sensitiveFlags         = []string{
+	serviceTokenFlag              = "service-token"
+	probeEcsClusterRegionFlag     = "probe.ecs.cluster.region"
```
ekimekim: I wouldn't consider the ECS region sensitive enough to be hidden in logging output by default (for example, IPs aren't considered sensitive either). The hidden things are all passwords or authentication tokens.

probe/awsecs/client.go
```diff
@@ -70,16 +70,19 @@ type EcsInfo struct {
...
 }
 
-func newClient(cluster string, cacheSize int, cacheExpiry time.Duration) (EcsClient, error) {
+func newClient(cluster string, cacheSize int, cacheExpiry time.Duration, clusterRegion string) (EcsClient, error) {
 	sess := session.New()
-
-	region, err := ec2metadata.New(sess).Region()
-	if err != nil {
-		return nil, err
+  var err error
```
ekimekim: style: this line isn't correctly indented. Make sure all code has passed through `gofmt -s`

prog/main.go
galindro: Done!

probe/awsecs/client.go
galindro: Done for all files

galindro:
@ekimekim When my commit will be tagged? I've noticed that it was mentioned in [1.6.5 changelog](https://github.com/weaveworks/scope/blob/v1.6.5/CHANGELOG.md) but, I can't see it in [commit list](https://github.com/weaveworks/scope/compare/v1.6.5...master) nor in files changed

ekimekim:
It's there. Check https://github.com/weaveworks/scope/commits/v1.6.5
The commit list you linked compares v1.6.5 to master. Since your commit is on both, it doesn't appear.

PR #2857, opened 2017-09-21
tracking: Add Mixpanel event on Control clicks

dlespiau:
We'd like to gain some insight on how controls are used.

### DIFF
--- a/client/app/scripts/components/node-details/node-details-control-button.js
+++ b/client/app/scripts/components/node-details/node-details-control-button.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { connect } from 'react-redux';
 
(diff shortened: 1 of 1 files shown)


client/app/scripts/components/node-details/node-details-control-button.js
```diff
@@ -21,6 +22,7 @@ class NodeDetailsControlButton extends React.Component {
 
   handleClick(ev) {
     ev.preventDefault();
+    trackMixpanelEvent('scope.node.control.click', { id: this.props.control.id });
```
fbarl: I'd pass in `title: this.props.control.human` as well - maybe it would give us easier filtering in Mixpanel.

dlespiau: Added title!

PR #2861, opened 2017-09-22
tracking: Use segment for tracking

dlespiau:
Align scope to the rest of the platform.


dlespiau:
Renamed the trackMixepanelEvent() to trackAnalyticsEvent() to have the function name reflect the change as discussed with @aaron7 offline.

PR #2863, opened 2017-09-23
Update ugorji/go/codec

### LINKED ISSUE #2862: upgrade ugorji codec to latest version for major performance improvements
[codec: major update with MAJOR PERFORMANCE improvements](https://github.com/ugorji/go/commit/54210f4e076c57f351166f0ed60e67d3fca57a36)

bboreham:
The commit log talks about all the speed optimisations, but when I ran some tests of msgpack encoding and decoding it was about the same speed as before.

Fixes #2862 

Nice that our use of the library doesn't seem to break.


rade:
Have you done any backward compatibility testing, i.e. decoding a newly encoded report with the old decoder and decoding an old encoded report with the new decoder?

bboreham:
Yes, I have checked this with the ugly code in branch `test-marshall-backwards`, which was originally proposed in #2399.  The new encoder generates exactly the same bytes as the old one, and can read a selection of older reports identically as before.

Issue #2662, opened 2017-06-29
don't report on replicasets from probe


rade:
In #2661 we removed replicasets from the display, instead adding deployments as direct parents of pods. Once we are sure that is the right way to go we should in fact remove replicasets from the probes, performing the mapping to deployments there, and convert old reports to the new format in `report.Upgrade()` for backward compatibility.

That should improve performance in several ways:
- reduced report size, and hence probe report writing, app report reading, and app report storage performance - replicasets constitute a monotonically growing data set, since during normal operation the replicasets of old deployment generations are never removed
- improved rendering performance since we no longer need to do the mapping there

rade:
I've looked at a recent set of reports from one of our clusters. These contain nearly 2500 replicasets. And they do get reported _by every node_. That means that overall they are the most numerous report entity.

rade:
>  remove replicasets from the probes, performing the mapping to deployments there

It would be lovely if the probes were able to do that w/o retrieving all replicasets, which is what they do know.

> and convert old reports to the new format in `report.Upgrade()` for backward compatibility.

We can probably leave that for later. By my reading of `selectPodsWithDeployments.Render()`, it will Just Work. OTOH, moving that code into `Report.Upgrade()` shouldn't be hard either.

bboreham:
btw `report.Upgrade()` currently does a copy on every call, which is expensive in the common case that it doesn't need upgrading.  Can we have a version number or something like that so we can easily tell which is which?

rade:
> Can we have a version number or something like that so we can easily tell which is which?

File an issue.

rade:
> `report.Upgrade()` currently does a copy on every call, which is expensive in the common case that it doesn't need upgrading

Filed #2959.

PR #2865, opened 2017-09-25
"Intern" map keys

bboreham:
Somewhat speculative approach; it works well in micro-benchmarks.

This is based after #2863 because it needs the `DecodeStringAsBytes` (re-)introduced there.

I created a static map so we don't have to lock  access from multiple threads and don't have to worry about it  getting clogged with values that are only used once or twice.

Relates to #1010


report/map_helpers.go
```diff
@@ -110,7 +110,13 @@ func mapRead(decoder *codec.Decoder, decodeValue func(isNil bool) interface{}) p
 		var key string
 		z.DecSendContainerState(containerMapKey)
 		if !r.TryDecodeAsNil() {
-			key = r.DecodeString()
+			b := r.DecodeStringAsBytes()
+			// Try to avoid an allocation by looking the key up
+			var ok bool
+			key, ok = commonKeys[string(b)]
+			if !ok {
```
rade: `if key, ok = ...; !ok {`

report/map_keys.go
```diff
@@ -0,0 +1,69 @@
...
+	"name":              "name",
+	"pid":               "pid",
+	"pod":               "pod",
+	"port":              "port",
+	"ppid":              "ppid",
+	"procspied":         "procspied",
+	"reverse_dns_names": "reverse_dns_names",
+	"snooped_dns_names": "snooped_dns_names",
+	"threads":           "threads",
+}
```
rade: I suggest you use the defined constants here. That will also identify entries in the above which are no longer present in current reports (`conntracked`, `procspied` comes to mind).

bboreham: That seems to end up with a rather ugly mish-mash of imports from all over Scope plus a few keys that are runtime-generated.

Also I don't think we want to reduce the performance on older reports every time we remove keys from newer ones.  It costs basically nothing to have one extra string in this list, for a given order of magnitude.

bboreham:
Ideas:

Maybe the lookup should be done via a perfect hash function.

I also experimented with putting all the values into `case` statements on the grounds that the compiler ought to know how to do it; so far I don't understand what I am seeing.

report/map_keys.go
rade: I am struggling to work out how this list is maintained. Or indeed how I would go about reviewing its contents as part of reviewing this PR. How do we make sure it contains all keys we want? How do we make sure it contains no spurious keys (e.g. because of typos etc)? My suggestion to use the constants was aimed at (imperfectly) addressing that.

bboreham: Perhaps if I checked in the code that creates the list?

Currently that doesn't exist as one thing - it went via some Go code, some awk code, some regexp-replace, etc.

rade:
I have re-based this.

rade:
In my benchmarks I am only seeing a ~3.2% drop in allocations during report reading.
```
go test -tags 'netgo unsafe' -run=x -cpu=2 -bench=ReportUnmarshal --bench-report-path=/tmp/prod-20171129/
```
master
```
BenchmarkReportUnmarshal-2   	       1	12567724701 ns/op	2679082008 B/op	19529959 allocs/op
```
branch
```
BenchmarkReportUnmarshal-2   	       1	11961347915 ns/op	2666462296 B/op	18906998 allocs/op
```
Moreover, nearly all of that comes from just 8 keys: `deployment`, `host`, `container`, `container_image`, `pod`, `deployment`, `reverse_dns_names`, `snooped_dns_names`.  None of these are latest-map keys. The first four are entries in parent sets (and `deployment` shows up frequently there because of #2662). The latter two are common "sets" entries for endpoints.

Looking into this further, it appears that the change in this PR doesn't affect latest-map decoding because we have a custom decoder for that.

rade:
>  it appears that the change in this PR doesn't affect latest-map decoding because we have a custom decoder for that.

Fixed. With that change I am seeing an allocation reduction of ~24% compared to master
```
BenchmarkReportUnmarshal-2   	       1	11665006452 ns/op	2532008296 B/op	14407494 allocs/op
```

rade:
I applied a quick hack to ascertain the possible gain from interning all keys:
```diff
diff --git a/extras/generate_latest_map b/extras/generate_latest_map
index 48d44863..2bc78d12 100755
--- a/extras/generate_latest_map
+++ b/extras/generate_latest_map
@@ -244,6 +244,7 @@ function generate_latest_map() {
                var ok bool
                if key, ok = commonKeys[string(b)]; !ok {
                        key = string(b)
+                       commonKeys[key] = key
                }
             }
             i := m.locate(key)
diff --git a/report/latest_map_generated.go b/report/latest_map_generated.go
index 2956447d..ddc136ca 100644
--- a/report/latest_map_generated.go
+++ b/report/latest_map_generated.go
@@ -212,6 +212,7 @@ func (m *StringLatestMap) CodecDecodeSelf(decoder *codec.Decoder) {
                        var ok bool
                        if key, ok = commonKeys[string(b)]; !ok {
                                key = string(b)
+                               commonKeys[key] = key
                        }
                }
                i := m.locate(key)
@@ -434,6 +435,7 @@ func (m *NodeControlDataLatestMap) CodecDecodeSelf(decoder *codec.Decoder) {
                        var ok bool
                        if key, ok = commonKeys[string(b)]; !ok {
                                key = string(b)
+                               commonKeys[key] = key
                        }
                }
                i := m.locate(key)
diff --git a/report/map_helpers.go b/report/map_helpers.go
index 79dd278e..0d2b0311 100644
--- a/report/map_helpers.go
+++ b/report/map_helpers.go
@@ -115,6 +115,7 @@ func mapRead(decoder *codec.Decoder, decodeValue func(isNil bool) interface{}) p
                        var ok bool
                        if key, ok = commonKeys[string(b)]; !ok {
                                key = string(b)
+                               commonKeys[key] = key
                        }
                }
 ```

This actually only led to a modest improvement to ~28% compared to master
```
BenchmarkReportUnmarshal-2   	       1	11494403535 ns/op	2518734696 B/op	14039184 allocs/op
```

**Comment by rade** [comment]
> it appears that the change in this PR doesn't affect latest-map decoding because we have a custom decoder for that.

In fact since #2870 `LatestMap` is based on a sorted slice, rather than on `ps.Map`. Allocations for keys are a major cost their too though.

**Comment by rade** [comment]
Here's how I get a count of the top keys from a scope report
```
cat report.json | jq '.[].nodes? | .[]? | (.sets,.parents,.latest,.latestControls) | select(. != null) | keys | .[]' | sort | uniq -c | sort -rn | head -n 30
```

ATM this is massively skewed for weaveworks clusters by #2662
```

rade:
@bboreham please take a look at my last commit and let me know what you think. I am constructing the map from constants declared in our code. That is less brittle but means we won't catch common labels or env vars, or deprecated keys in old reports. We could add these manually, but given the benchmark results below I don't think it's worth it.

I minimised the diff in that commit. There are some extra changes I want to make:
- remove the constant aliases in the `probe` packages, and instead use the report constants directly
- remove the imports of `probe` packages from `render`, and instead use the report constants

I reckon that's a good rearrangements of the dependencies: `report` defines the protocol between probes and the app/renderer.

Here's a comparison of running
```
go test -tags 'netgo unsafe' -run=x -cpu=2 -benchmem -bench='ReportUnmarshal' --bench-report-path=/tmp/prod-20171204-15s
```
against master, this branch prior to my commit, and this branch after my commit
```
BenchmarkReportUnmarshal-2   	       1	2542708115 ns/op	458429904 B/op	 3477396 allocs/op
BenchmarkReportUnmarshal-2   	       1	2419631234 ns/op	447161120 B/op	 2947783 allocs/op
BenchmarkReportUnmarshal-2   	       1	2445833014 ns/op	447220928 B/op	 2941192 allocs/op
```

report/map_keys.go
```diff
@@ -0,0 +1,192 @@
...
+	ECSTaskFamily:          ECSTaskFamily,
+	ECSServiceDesiredCount: ECSServiceDesiredCount,
+	ECSServiceRunningCount: ECSServiceRunningCount,
+	ECSScaleUp:             ECSScaleUp,
+	ECSScaleDown:           ECSScaleDown,
+}
+
+// lookupCommonKey tris to avoid an allocation in populating the key
+// by looking it up
+func lookupCommonKey(key *string, b []byte) {
```
bboreham: I would just have this return the string rather than taking a pointer.

report/map_keys.go
```diff
@@ -0,0 +1,192 @@
...
+	ECSCluster:             ECSCluster,
+	ECSCreatedAt:           ECSCreatedAt,
+	ECSTaskFamily:          ECSTaskFamily,
+	ECSServiceDesiredCount: ECSServiceDesiredCount,
+	ECSServiceRunningCount: ECSServiceRunningCount,
+	ECSScaleUp:             ECSScaleUp,
+	ECSScaleDown:           ECSScaleDown,
+}
+
+// lookupCommonKey tris to avoid an allocation in populating the key
```
bboreham: s/tris/tries/

probe/docker/tagger.go
```diff
@@ -10,9 +10,8 @@ import (
 
 // Node metadata keys.
 const (
-	ContainerID = "docker_container_id"
-	Domain      = "domain" // TODO this is ambiguous, be more specific
-	Name        = "name"   // TODO this is ambiguous, be more specific
+	ContainerID = report.DockerContainerID
+	Name        = report.Name // TODO this is ambiguous, be more specific
```
bboreham: I don't understand this comment

rade: I didn't write the original comment. I suspect it stems from the fact that key is "name" rather than, say, "container_name". So it's prone to clashing, and indeed does with `ImageName` and `ProcessName` that are also defined as "name". However, none of them are used in the same context. I am happy to get rid of the warning.

report/map_keys.go
rade: I am pretty sure I tried that and got more allocations.

rade: Just tried again and it's fine. Will change.

rade:
@bboreham PTAL

bboreham:
LGTM.  Since this was my PR originally I can't Approve it.

PR #2881, opened 2017-10-08
[WIP] Upgrade weaveworks/common and lower Sirupsen/logrus

marccarre:
As discussed [on Slack](https://weaveworks.slack.com/archives/C6NSJ9942/p1507299349000715), this PR:
- upgrades to latest `weaveworks/common`, and
- uses `sirupsen/logrus` instead of `Sirupsen/logrus` -- see: https://github.com/sirupsen/logrus/issues/570.

I am admittedly not happy with this PR, as there are few hacks involved (mainly because Go's vendoring/dependencies management currently s*cks):

1. mixing `gvt` and `dep` for dependencies management in various projects seems to lead to the needs for hacks like f5b6937, see also [this explanation](https://github.com/peterbourgon/wtf) -- proper resolution *might* be to manage `scope`'s dependencies with `dep`,
2. some other dependencies still had references to `Sirupsen/logrus`, which is addressed in a hacky way in 76807eb -- proper resolution would be to lowercase `logrus` for these dependencies upstream and, then update to releases having the change,
3. upgrading to latest to `weaveworks/common` led to the need to upgrade to latest `context` (for `Context`) and `golang.org/x/net/context` (for `ctxhttp`),

but at least:
- we can build successfully with latest dependencies and `sirupsen/logrus`,
- details of all steps followed and commands ran are in the commit messages, so the upgrade is reproducible.

**Reviewers**: given how large the `diff` is for this PR, you may want to pay closer attention to the commit messages, which explain what has been done, how, and for most commits, why.


rade:
What is happening with this? It's the kind of PR that surely is prone to bitrot, so we ought to get it reviewed and merged quickly.

bboreham:
I really don't like editing the vendor'd dependencies as in 76807eb54ffaa14b3524b4c8fc71469c1fe0e365

The rest of it looks pretty good.

marccarre:
Agreed, I really dislike this too. Having another look at this, it seems the latest versions of:

- [github.com/Azure/go-ansiterm](https://github.com/Azure/go-ansiterm)
- [github.com/moby/moby](https://github.com/moby/moby)
- [github.com/opencontainers/runc](https://github.com/opencontainers/runc)

use `sirupsen` instead of `Sirupsen`.
I'll see if we can upgrade easily.

bboreham:
Of the ones edited:
* one moved to lowercase in June https://github.com/Azure/go-ansiterm/pull/21 (and subsequently removed)
* one in July https://github.com/moby/moby/pull/34272
* two changes are in test files vendor/github.com/docker/libcontainer/cgroups/fs/stats_util_test.go vendor/github.com/opencontainers/runc/libcontainer/cgroups/fs/stats_util_test.go which we can ignore
* two are Weaveworks libs

2opremio:
> I really don't like editing the vendor'd dependencies as in 76807eb

There's also f5b6937 which removes a vendored package in `promrus` ... but I don't have a better solution. It's basically a manual implementation of what's proposed at https://github.com/Masterminds/glide/issues/303

marccarre:
I'm trying to set up `dep` so that dependencies management becomes (hopefully) easier to manage and more standardised across Weaveworks projects.
I'll see how that goes, but `dep init` still had not returned after ~4 hours yesterday :sweat_smile:.

marccarre:
Still no success with `dep` yesterday. I'll try a cleaner approach with `gvt`, i.e. use the latest versions of:

- [github.com/Azure/go-ansiterm](https://github.com/Azure/go-ansiterm)
- [github.com/moby/moby](https://github.com/moby/moby)
- [github.com/opencontainers/runc](https://github.com/opencontainers/runc)

marccarre:
Next blocker: [github.com/fsouza/go-dockerclient](https://github.com/fsouza/go-dockerclient) still uses `docker` and not `moby`.
This causes issues when trying a cleaner approach with `gvt`.

PR #2882, opened 2017-10-09
Use a context sensitive route for the popped out terminal.html

foot:
- Service-ui has learned a new route which corresponds to the existing
  scope terminal.html route. Both standalone and service-ui scope can now
  respond to terminal.html:
  - If using standalone scope path/to/scope/ -> path/to/scope/terminal.html will be used.
  - If using service-ui app/instanceid -> app/instanceId/terminal.html

### DIFF
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -9,7 +9,7 @@ import Term from 'xterm';
 import { clickCloseTerminal } from '../actions/app-actions';
 import { getNeutralColor } from '../utils/color-utils';
 import { setDocumentTitle } from '../utils/title-utils';
-import { getPipeStatus, doResizeTty, getWebsocketUrl, getApiPath } from '../utils/web-api-utils';
+import { getPipeStatus, doResizeTty, getWebsocketUrl, basePath } from '../utils/web-api-utils';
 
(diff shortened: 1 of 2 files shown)


client/app/scripts/components/terminal.js
```diff
@@ -243,7 +243,7 @@ class Terminal extends React.Component {
 
     const bcr = this.node.getBoundingClientRect();
     const minWidth = (this.state.characterWidth * 80) + (8 * 2);
-    openNewWindow(`${getApiPath()}/terminal.html#!/state/${paramString}`, bcr, minWidth);
+    openNewWindow(`${basePath(window.location.pathname)}/terminal.html#!/state/${paramString}`, bcr, minWidth);
```
jpellizzari: `getApiPath` was already using `basePath` internally.

jpellizzari:
@foot What is the context for this? I am confused as to what this is fixing.

client/app/scripts/components/terminal.js
foot: This will now try and open:

`/app/:instanceId/terminal.html` instead of `/api/app/:instanceId/terminal.html` e.g. the service-ui will now handle terminal.html rather than query service

PR #2886, opened 2017-10-11
Flat background instead of gradient

bia:
As decided in last week's design sync.
@fbarl was there another reason why the overlay-wrapper would need a background colour? I remember talking about it...

### DIFF
--- a/client/app/styles/_base.scss
+++ b/client/app/styles/_base.scss
@@ -102,7 +102,7 @@ a {
 
 .overlay-wrapper {
   align-items: center;
-  background-color: fade-out($background-average-color, 0.1);
+  background-color: transparent;
   border-radius: 4px;
   color: $text-tertiary-color;
   display: flex;
--- a/client/app/styles/_base.scss
+++ b/client/app/styles/_base.scss
@@ -102,7 +102,7 @@ a {
 
 .overlay-wrapper {
   align-items: center;
-  background-color: transparent;
+  background-color: fade-out($background-average-color, 0.1);
   border-radius: 4px;
   color: $text-tertiary-color;
   display: flex;
--- a/client/app/styles/_variables.scss
+++ b/client/app/styles/_variables.scss
@@ -58,5 +58,5 @@ $search-border-width: 1px;
 $timeline-height: 55px;
 
 /* specific elements */
-$body-background-color: linear-gradient(30deg, $background-color 0%, $background-lighter-color 100%);
+$body-background-color: $background-average-color;
 $label-background-color: fade-out($background-average-color, .3);


client/app/styles/_variables.scss
```diff
@@ -58,5 +58,5 @@ $search-border-width: 1px;
 $timeline-height: 55px;
 
 /* specific elements */
-$body-background-color: linear-gradient(30deg, $background-color 0%, $background-lighter-color 100%);
+$body-background-color: $background-average-color;
```
fbarl: Finally! :)

btw. there's this whole mix of finely graded shades of background color defined and they all seem to be used somewhere - hopefully we'll be able to remove some as we keep unifying our color language:

```
$background-color: lighten($primary-color, 66%);
$background-lighter-color: lighten($background-color, 8%);
$background-average-color: mix($background-color, $background-lighter-color);
$background-darker-color: darken($background-color, 8%);
$background-darker-secondary-color: darken($background-color, 4%);
$background-dark-color: $primary-color;
```

client/app/styles/_base.scss
```diff
@@ -102,7 +102,7 @@ a {
 
 .overlay-wrapper {
   align-items: center;
-  background-color: fade-out($background-average-color, 0.1);
+  background-color: transparent;
```
fbarl: > @fbarl was there another reason why the overlay-wrapper would need a background colour? I remember talking about it...

Yeah, I made it non-transparent about a year ago to make overlays more readable when there is content in the background. Compare the footer text:

**now**
image

**before**
image

So yeah, I'd rather keep it mostly opaque, unless we figure a better solution (which we could also do in a separate PR).

client/app/styles/_variables.scss
bia: Yes, we'll get rid of a lot of those as time goes on. :)

bia:
Thanks for the comments! The overlay background colour makes sense and I reverted the transparent colour change.

How do I remove _Merge branch 'master' into background-colors_?
Roli suggested that I make a one-time `git config --global pull.rebase true` for the future.

fbarl:
> How do I remove Merge branch 'master' into background-colors?
Roli suggested that I make a one-time `git config --global pull.rebase true` for the future.

I'm not sure how @rndstr's command works, but `git rebase` is definitely your friend here and by running it in interactive mode you can manipulate (and delete) individual commits from your PR (e.g. try `git rebase -i HEAD~3` in this branch).

If you don't manage to fix your git tree for this PR, maybe you could just reset your branch at master and apply that one line change in a new clean commit (normally I wouldn't recommend that but this PR is really just a one-liner).

Hope that helps! :)

bia:
Thanks @fbarl ! That's fixed now.

rndstr:
> I'm not sure how @rndstr's command works

That appends `--rebase` to any `git pull` automatically so it doesn't merge in remote commits.

PR #2916, opened 2017-10-31
MAINTAINER is deprecated, now using LABEL

cglewis:
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated


dlespiau:
Indeed, from the referenced page:

```
MAINTAINER in Dockerfile
Deprecated In Release: v1.13.0

MAINTAINER was an early very limited form of LABEL which should be used instead.
```

I note there's no removal date (compare that to other deprecated features) and this introduces a dependency on docker 1.13 (Jan 2017) for build, at least. I don't see any problem with this at the moment, but we may encounter one down the line.

Still, this seems fine.

PR #2917, opened 2017-11-01
Simplify Utsname string conversion

tklauser:
Use Utsname from golang.org/x/sys/unix instead of syscall to simplify the conversion of the Utsname members to strings. This requires re-vendoring golang/x/sys/unix.

It is then possible to get rid of the marshal.FromUtsname conversion functions and replace them by a simple string conversion (e.g. string(uname.Sysname[:])), because unix.Utsname has byte array instead of the int8/uint8 members that syscall had. Also see golang/go#20753 for details.


dlespiau:
Kicked a test run. Something tells me we need to update the tests: https://github.com/weaveworks/scope/blob/c0adf40ad7ff8a6638b07ad693b000418b604046/probe/host/system_linux_test.go#L19
and
https://github.com/weaveworks/scope/blob/c0adf40ad7ff8a6638b07ad693b000418b604046/probe/host/system_linux_test.go#L35

dlespiau:
And indeed, from the test run:

```
# github.com/weaveworks/scope/probe/host_test

probe/host/system_linux_test.go:23: cannot use func literal (type func(*syscall.Utsname) error) as type func(*unix.Utsname) error in assignment

FAIL	github.com/weaveworks/scope/probe/host [build failed]
```

tklauser:
Sorry, forgot about the test. Fixed now and pushed out, PTAL.

PR #2919, opened 2017-11-01
Use timestamp in URL

### LINKED ISSUE #2823: Time-travel mode does not change the URL
This means I can't ask someone else to look at the same screen I'm seeing.
Also I cannot use the 'back' button to go to a time-travel page; it reverts to live mode.

fbarl:
Fixes #2823.

##### Changes

* Store the `pausedAt` value in the URL (as `timestamp` field), effectively storing the Time Travel state.
* Use a fallback mechanism to go into the _paused_ state at the current timestamp if Time Travel functionality is not provided (which is the case for the typical standalone _Scope_ setup).
* Bump the Time Travel component version and forward it timestamps as formatted strings.
* Don't go out of Time Travel mode when unmounting the UI.
* Empty `version` field state is now `null` instead of `...` for state tree consistency.

### DIFF
--- a/client/app/scripts/actions/app-actions.js
+++ b/client/app/scripts/actions/app-actions.js
@@ -345,6 +345,7 @@ export function pauseTimeAtNow() {
     dispatch({
       type: ActionTypes.PAUSE_TIME_AT_NOW
     });
+    updateRoute(getState);
     if (!getState().get('nodesLoaded')) {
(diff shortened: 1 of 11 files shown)


fbarl:
[Use Scope timestamp in URL](https://trello.com/c/xkvdKv2h/174-use-scope-timestamp-in-url)

PR #2920, opened 2017-11-04
Optimisation: replace three map-reduces with Renderers

bboreham:
In three cases: `ConnectionJoin`, `ProcessRenderer` and `HostRenderer`, the code was running a `Map` to speculatively create `Node`s, then a `Reduce` to match them all up with something else.

If we replace those multiple operations with a single routine which does the matching directly, we save a lot of garbage.

I also added benchmark to time a single topology rendering, equivalent to hitting an endpoint like`/api/topology/hosts`. Best time of five runs; input is a 33MB report from production on Oct 15:
Before
```
$ go test -tags 'unsafe netgo' -run=x -bench=Topology -bench-report-file=prod-report1.json
BenchmarkTopologyList-2         	       1	1400508627 ns/op	352382920 B/op	 2913537 allocs/op
BenchmarkTopologyHosts-2        	       1	1124918587 ns/op	302509672 B/op	 2627937 allocs/op
BenchmarkTopologyContainers-2   	       2	 691307933 ns/op	194143244 B/op	 1723873 allocs/op
```

After
```
$ go test -tags 'unsafe netgo' -run=x -bench=Topology -bench-report-file=prod-report1.json
BenchmarkTopologyList-2         	       2	 692289010 ns/op	146370204 B/op	 1529315 allocs/op
BenchmarkTopologyHosts-2        	       3	 485628292 ns/op	95328261 B/op	 1245029 allocs/op
BenchmarkTopologyContainers-2   	      10	 208411667 ns/op	36021493 B/op	  515582 allocs/op
```

That is a 50-70% reduction in time and objects created.

### DIFF
--- a/app/benchmark_internal_test.go
+++ b/app/benchmark_internal_test.go
@@ -32,19 +32,43 @@ func loadReport() (report.Report, error) {
 }
 
 func BenchmarkTopologyList(b *testing.B) {
+	benchmarkRender(b, func(report report.Report) {
+		request := &http.Request{
+			Form: url.Values{},
+		}
+		topologyRegistry.renderTopologies(report, request)
+	})
+}
+
+func benchmarkRender(b *testing.B, f func(report.Report)) {
 	report, err := loadReport()
 	if err != nil {
 		b.Fatal(err)
 	}
 	b.ReportAllocs()
 	b.ResetTimer()
-	request := &http.Request{
-		Form: url.Values{},
-	}
 	for i := 0; i < b.N; i++ {
 		b.StopTimer()
 		render.ResetCache()
 		b.StartTimer()
-		topologyRegistry.renderTopologies(report, request)
+		f(report)
 	}
+}
+
+func BenchmarkTopologyHosts(b *testing.B) {
+	benchmarkOneTopology(b, "hosts")
+}
+
+func BenchmarkTopologyContainers(b *testing.B) {
+	benchmarkOneTopology(b, "containers")
+}
+
+func benchmarkOneTopology(b *testing.B, topologyID string) {
+	benchmarkRender(b, func(report report.Report) {
+		renderer, decorator, err := topologyRegistry.RendererForTopology(topologyID, url.Values{}, report)
+		if err != nil {
+			b.Fatal(err)
+		}
+		renderer.Render(report, decorator)
+	})
 }
--- a/render/container.go
+++ b/render/container.go
@@ -71,8 +71,7 @@ func (c connectionJoin) Render(rpt report.Report, dct Decorator) report.Nodes {
 			}
 		}
 	}
-	var ret = make(report.Nodes)
-	var mapped = map[string]string{} // input node ID -> output node ID
+	ret := newJoinResults()
 
 	// Now look at all the endpoints and see which map to IP nodes
 	for _, m := range endpoints {
@@ -83,7 +82,7 @@ func (c connectionJoin) Render(rpt report.Report, dct Decorator) report.Nodes {
 		// Nodes without a hostid may be pseudo nodes - if so, pass through to result
 		if _, ok := m.Latest.Lookup(report.HostNodeID); !ok {
 			if id, ok := externalNodeID(m, addr, local); ok {
-				addToResults(m, id, ret, mapped, newPseudoNode)
+				ret.addToResults(m, id, newPseudoNode)
 				continue
 			}
 		}
@@ -95,20 +94,15 @@ func (c connectionJoin) Render(rpt report.Report, dct Decorator) report.Nodes {
 			id, found = ipNodes[report.MakeScopedEndpointNodeID(scope, addr, port)]
 		}
 		if found && id != "" { // not one we blanked out earlier
-			addToResults(m, id, ret, mapped, func(id string) report.Node {
+			ret.addToResults(m, id, func(id string) report.Node {
 				return inputNodes[id]
 			})
 		}
 	}
-	// Copy through any unmatched input nodes
-	for _, n := range inputNodes {
-		if _, found := ret[n.ID]; !found {
-			ret[n.ID] = n
-		}
-	}
-	fixupAdjancencies(inputNodes, ret, mapped)
-	fixupAdjancencies(endpoints, ret, mapped)
-	return ret
+	ret.copyUnmatched(inputNodes)
+	ret.fixupAdjacencies(inputNodes)
+	ret.fixupAdjacencies(endpoints)
+	return ret.nodes
 }
 
 func (c connectionJoin) Stats(rpt report.Report, _ Decorator) Stats {
--- a/render/container.go
+++ b/render/container.go
@@ -39,85 +39,82 @@ var ContainerRenderer = Memoise(MakeFilter(
 	),
 ))
 
-var mapEndpoint2IP = Memoise(MakeMap(endpoint2IP, SelectEndpoint))
-
 const originalNodeID = "original_node_id"
 
 // ConnectionJoin joins the given renderer with connections from the
 // endpoints topology, using the toIPs function to extract IPs from
 // the nodes.
 func ConnectionJoin(toIPs func(report.Node) []string, r Renderer) Renderer {
-	nodeToIP := func(n report.Node, _ report.Networks) report.Nodes {
-		result := report.Nodes{}
-		for _, ip := range toIPs(n) {
-			result[ip] = report.MakeNode(ip).
-				WithTopology(IP).
-				WithLatests(map[string]string{
-					originalNodeID: n.ID,
-				}).
-				WithCounters(map[string]int{IP: 1})
-		}
-		return result
-	}
-
-	return MakeReduce(
-		MakeMap(
-			ipToNode,
-			MakeReduce(
-				MakeMap(nodeToIP, r),
-				mapEndpoint2IP,
-			),
-		),
-		r,
-	)
+	return connectionJoin{toIPs: toIPs, r: r}
 }
 
-func ipToNode(n report.Node, _ report.Networks) report.Nodes {
-	// propagate non-IP nodes
-	if n.Topology != IP {
-		return report.Nodes{n.ID: n}
-	}
-	// If an IP is shared between multiple nodes, we can't reliably
-	// attribute an connection based on its IP
-	if count, _ := n.Counters.Lookup(IP); count > 1 {
-		return report.Nodes{}
-	}
-	// If this node is not of the original type, exclude it.  This
-	// excludes all the nodes we've dragged in from endpoint that we
-	// failed to join to a node.
-	id, ok := n.Latest.Lookup(originalNodeID)
-	if !ok {
-		return report.Nodes{}
-	}
-
-	return report.Nodes{id: NewDerivedNode(id, n)}
+type connectionJoin struct {
+	toIPs func(report.Node) []string
+	r     Renderer
 }
 
-// endpoint2IP maps endpoint nodes to their IP address, for joining
-// with container nodes.
-func endpoint2IP(m report.Node, local report.Networks) report.Nodes {
-	scope, addr, port, ok := report.ParseEndpointNodeID(m.ID)
-	if !ok {
-		return report.Nodes{}
-	}
+func (c connectionJoin) Render(rpt report.Report, dct Decorator) report.Nodes {
+	local := LocalNetworks(rpt)
+	inputNodes := c.r.Render(rpt, dct)
+	endpoints := SelectEndpoint.Render(rpt, dct)
 
-	// Nodes without a hostid may be pseudo nodes
-	if _, ok := m.Latest.Lookup(report.HostNodeID); !ok {
-		if externalNode, ok := NewDerivedExternalNode(m, addr, local); ok {
-			return report.Nodes{externalNode.ID: externalNode}
+	// Collect all the IPs we are trying to map to, and which ID they map from
+	var ipNodes = map[string]string{}
+	for _, n := range inputNodes {
+		for _, ip := range c.toIPs(n) {
+			if _, exists := ipNodes[ip]; exists {
+				// If an IP is shared between multiple nodes, we can't reliably
+				// attribute an connection based on its IP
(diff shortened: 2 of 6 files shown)


render/host.go
```diff
@@ -7,10 +7,7 @@ import (
 // HostRenderer is a Renderer which produces a renderable host
 // graph from the host topology.
 var HostRenderer = MakeReduce(
-	MakeMap(
-		MapEndpoint2Host,
-		EndpointRenderer,
-	),
+	endpoints2Hosts{},
```
rade: Unlike the `MakeMap` this replaces, this is not `Memoise`d. Which got me thinking about memoisation (again) and I have filed #2923 as a result.

render/host.go
```diff
@@ -62,18 +59,73 @@ func MapX2Host(n report.Node, _ report.Networks) report.Nodes {
...
+				out.Adjacency = out.Adjacency.Add(mappedDest)
+			}
+		}
+		ret[outID] = out
+	}
+}
+
+// Dummy struct for Renderers that do not return useful stats
+type blankStats struct {
+}
```
rade: Is `blankStats` really worthwhile, compared to implementing blank `Stats` functions in each of the renderers?

If you really do think so then it should be used in a few other places too, namely `Map`, `ConstantRenderer`, `TopologySelector`, `selectPodsWithDeployment`, `mockRenderer`.

render/container.go
```diff
@@ -39,85 +39,93 @@ var ContainerRenderer = MakeFilter(
...
+// node if not already there, and updating the old-id to new-id mapping
+func addToResults(m report.Node, id string, ret report.Nodes, mapped map[string]string, create func() report.Node) {
+	result, exists := ret[id]
+	if !exists {
+		result = create()
 	}
+	result.Children = result.Children.Add(m)
+	result.Children = result.Children.Merge(m.Children)
+	ret[result.ID] = result
+	mapped[m.ID] = result.ID
```
rade: It is most unfortunate that this function differs from the analogous code in `endpoint2Host.Render`. Do we know these differences are there for good reasons, rather than accidents or even bugs?

rade: I've dug into this a bit...

Both here and in `endpoint2Host.Render`, the node we are mapping (`m` in the above) is an endpoint.

I am quite sure endpoints don't have children, so the `result.Children.Merge(m.Children)` here and in `endpoint2Host.Render` are no-ops and superfluous.

I am also quite sure that we do not need to count endpoints. And I certainly cannot think of a reason why we would count them in `endpoint2Host.Render` but not here.

render/host.go
bboreham: Yes, I couldn't see when memoisation would be useful for this particular case.
I did unify two calls to the `r` Renderer in `ConnectionJoin`, but that's trivial.

render/host.go
bboreham: To me it makes the intent plainer.  I would spread it around more, yes.

bboreham:
While working on this I had to write out the following analysis of the starting state; pasting it here for preservation:

```
TestShortLivedInternetNodeConnections calls
render.ContainerWithImageNameRenderer.Render(rpt, FilterNoop))

ContainerWithImageNameRenderer is a variable initialised with ContainerRenderer

First thing ContainerWithImageNameRenderer.Render() does is call
ContainerRenderer to get a list of "containers"

  ContainerRenderer is
    A merge of
      Map ColorConnectedProcessRenderer into MapProcess2Container
        ColorConnectedProcessRenderer calls ProcessRenderer then a function
          ProcessRenderer, if the report has any process nodes, is a merge of
            Map EndpointRenderer into MapEndpoint2Process
            SelectProcess
          the ColorConnected function adds a flag IsConnected to any nodes with edges
        MapProcess2Container creates container nodes if there is an ID, otherwise puts in 'uncontained'
      ConnectionJoin, which is a merge of
        Map a merge of
            Map r (SelectContainer) into nodeToIP via MapContainer2IP
            mapEndpoint2IP which is Map SelectEndpoint into endpoint2IP
          into ipToNode which removes any Endpoint2IP nodes that didn't match a container
        r (SelectContainer)
    Filtered to just nodes that have a ContainerState that is non-deleted (or pseudo-nodes)

Next Render() calls SelectContainerImage.Render(rpt, dct) to get a list of images

Then Render does:
  For each input node, if it has an image
    if the image is in the list of images
      if the image has a name
        updates the image details in the container object, and adds it to the output
  otherwise pass the input node straight through to the output
```

render/container.go
bboreham: I have made this change; dropping the counter doesn't change the final output.

The superfluous `Merge(m.Children)` shouldn't cost much (it detects that case and just returns the input) so I left it; future cases may have children.

rade: Now the function is shared, it should be moved near `fixupAdjacencies`.

The function itself is not specific to endpoints, even though the call sites are. I therefore reckon we should add back the counter, or, if that turns out to be really expensive, add a comment. I've spent ages in the past chasing missing counters :(

render/host.go
rade: Perhaps leave it out, and add it here and elsewhere in a subsequent PR.

render/id.go
```diff
@@ -39,12 +39,21 @@ func NewDerivedPseudoNode(id string, node report.Node) report.Node {
 
 // NewDerivedExternalNode figures out if a node should be considered external and creates the corresponding pseudo node
 func NewDerivedExternalNode(n report.Node, addr string, local report.Networks) (report.Node, bool) {
+	id, ok := externalNodeID(n, addr, local)
+	if !ok {
+		return report.Node{}, false
+	}
+	return NewDerivedPseudoNode(id, n), true
+}
```
rade: Inverting the condition would allow you to inline the first line.

render/process.go
```diff
@@ -87,18 +87,27 @@ var ProcessNameRenderer = ConditionalRenderer(renderProcesses,
...
 // MapEndpoint2Pseudo makes internet of host pesudo nodes from a endpoint node.
 func MapEndpoint2Pseudo(n report.Node, local report.Networks) report.Nodes {
-	_, addr, _, ok := report.ParseEndpointNodeID(n.ID)
+	id, ok := pseudoNodeID(n, local)
 	if !ok {
 		return report.Nodes{}
 	}
+	externalNode := NewDerivedPseudoNode(id, n)
+	return report.Nodes{externalNode.ID: externalNode}
+}
```
rade: Inverting the condition would allow you to inline the first line.

app/benchmark_internal_test.go
```diff
@@ -48,3 +48,30 @@ func BenchmarkTopologyList(b *testing.B) {
...
+	}
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		render.ResetCache()
+		b.StartTimer()
+		renderer.Render(report, decorator)
+	}
+}
```
rade: a `benchmarkRender(render func(report.Report))` helper would extract all the report loading and benchmarking machinery that is shared between this and `BenchmarkTopologyList`

bboreham:
How I tested it produces the same output as before:

First, obtain a test report, e.g. from your production system, and save it as a JSON file.

```
git checkout master
make prog/scope
prog/scope --mode=app --weave=false --app.collector=file:some-report.json &
curl -so /tmp/before http://localhost:4040/api/topology/containers
jq -S . < /tmp/before > /tmp/before-jq
```
(`-S` is used to sort keys so they are consistent)

then repeat on the branch for the 'after' case, and diff the `jq` output before and after.

Repeat for other topologies - `hosts`, etc.

rade:
> Repeat for other topologies - `hosts`, etc.

I think `/api/topology` does that.

render/process.go
bboreham: function NewDerivedExternalNode has been removed now

render/id.go
bboreham: function NewDerivedExternalNode has been removed now

render/process.go
```diff
@@ -85,56 +77,82 @@ var ProcessNameRenderer = ConditionalRenderer(renderProcesses,
 	),
 )
 
-// MapEndpoint2Pseudo makes internet of host pesudo nodes from a endpoint node.
-func MapEndpoint2Pseudo(n report.Node, local report.Networks) report.Nodes {
+func pseudoNodeID(n report.Node, local report.Networks) (string, bool) {
```
rade: I reckon this should move to id.go.

render/host.go
```diff
@@ -62,18 +59,73 @@ func MapX2Host(n report.Node, _ report.Networks) report.Nodes {
...
+			})
+		}
 	}
+	fixupAdjancencies(ns, ret, mapped)
+	return ret
+}
+
+// Add Node M to the result set ret under id, creating a new result
+// node if not already there, and updating the old-id to new-id mapping
+// Note we do not update any counters for child topologies here
```
rade: That last bit explains what it does but should really explain why.

Is it because of performance? Is that actually noticeable? And why is it safe? Because in current usage `m` is always an endpoint, and we don't depend on endpoint counters anywhere?

render/container.go
```diff
@@ -39,85 +39,82 @@ var ContainerRenderer = MakeFilter(
...
+				// If an IP is shared between multiple nodes, we can't reliably
+				// attribute an connection based on its IP
+				ipNodes[ip] = "" // blank out the mapping so we don't use it
+			} else {
+				ipNodes[ip] = n.ID
+			}
+		}
 	}
+	var ret = make(report.Nodes)
+	var mapped = map[string]string{} // input node ID -> output node ID
```
rade: suggestion: wrap these two in a struct. Then `addToResults` and `fixupAdjacencies` can become methods on that.

render/host.go
```diff
@@ -62,18 +59,73 @@ func MapX2Host(n report.Node, _ report.Networks) report.Nodes {
...
-	// Nodes without a hostid are treated as pseudo nodes
-	hostNodeID, timestamp, ok := n.Latest.LookupEntry(report.HostNodeID)
-	if !ok {
-		return MapEndpoint2Pseudo(n, local)
+type endpoints2Hosts struct {
+}
+
+func (e endpoints2Hosts) Render(rpt report.Report, dct Decorator) report.Nodes {
+	ns := SelectEndpoint.Render(rpt, dct)
+	local := LocalNetworks(rpt)
```
rade: The order of these two statements here and in `connectionJoin.Render` should be the same. Also, I'd name the vars the same, i.e. `endpoints` instead of `ns`.

render/host.go
```diff
@@ -62,18 +59,73 @@ func MapX2Host(n report.Node, _ report.Networks) report.Nodes {
...
+func addToResults(m report.Node, id string, ret report.Nodes, mapped map[string]string, create func() report.Node) {
+	result, exists := ret[id]
+	if !exists {
+		result = create()
+	}
+	result.Children = result.Children.Add(m)
+	result.Children = result.Children.Merge(m.Children)
+	ret[result.ID] = result
+	mapped[m.ID] = result.ID
+}
```
rade: There are a few too many ids here for my liking. AFAICT we expect `result.ID == id` here, right? So perhaps
1. always use `id`
2. pass `id` to `create()`
3. assert that the `result.ID` returned by `create()` is `id`

The second rather clutters call sites though, since they all use closures which already have access to the id.

render/host.go
```diff
@@ -62,18 +59,73 @@ func MapX2Host(n report.Node, _ report.Networks) report.Nodes {
...
+
+// Rewrite Adjacency for new nodes in ret, original nodes in input, and mapping old->new IDs in mapped
+func fixupAdjancencies(input, ret report.Nodes, mapped map[string]string) {
```
rade: `fixupAdjacencies`

render/host.go
rade: Actually, doing the 2nd would allow us to introduce a `NewPseudoNode(id string)` function, thus eliminating the closure in three of the `addToResults` call sites.

render/host.go
bboreham: It added 4-5% to the number of allocations, and makes no difference to the output.
I did not get the bottom of why it makes no difference.
`m` is a `process` in one case; `endpoint` in two cases.

rade: `m` is always an endpoint.

rade: which is why it makes no difference.

bboreham: Sorry, you're right, it's always an endpoint

render/process.go
```diff
@@ -89,56 +81,67 @@ var ProcessNameRenderer = ConditionalRenderer(renderProcesses,
...
-func MapEndpoint2Process(n report.Node, local report.Networks) report.Nodes {
-	// Nodes without a hostid are treated as pseudo nodes
-	if _, ok := n.Latest.Lookup(report.HostNodeID); !ok {
-		return MapEndpoint2Pseudo(n, local)
-	}
-
-	pid, timestamp, ok := n.Latest.LookupEntry(process.PID)
-	if !ok {
+func (e endpoints2Processes) Render(rpt report.Report, dct Decorator) report.Nodes {
+	if len(rpt.Process.Nodes) == 0 {
```
rade: This is rather at odds with the "we have a pid, but no matching process node; create a new one rather than dropping the data" behaviour below, and differs from the existing behaviour.

bboreham: I believe it does match the behaviour of `ProcessRenderer = Memoise(ConditionalRenderer(renderProcesses, ...` - that condition is `len(rpt.Process.Nodes) >= 1`

bboreham:
All comments addressed

render/process.go
rade: Hmm. Good point. But odd nonetheless.

rade:
Benchmarking this with a recent prod report by running
```
$ prog/scope --mode=app --weave=false --app.collector=file:prod-on-prod-report.json
```
and then timing topology rendering with
```
time curl -so /dev/null http://localhost:4040/api/topology/<topology>
```
yields

| topology   | before | after |
| ------------  | ------- | ------- |
| processes  | 0.307 | 0.252 |
| containers | 1.277 | 0.796 |
| pods       | 0.936 | 0.506 |
| hosts      | 1.227 | 0.711 |

So this is a significant improvement indeed.

LGTM. Merge it!

PR #2921, opened 2017-11-04
Avoid object creation when scanning DNS names

bboreham:
Since we do this a lot, scanning the lists in-place saves work.

Also we don't need to sort them since StringSet is implemented as a sorted set of strings

It's actually benchmarking a little slower, when the best of five runs is taken, but with reduced memory allocation.  The timing of this benchmark is highly variable on my VM; the new benchmarks in #2920 are more reliably faster with this change.

Before
```
BenchmarkTopologyList-2   	       1	1479321809 ns/op	351308800 B/op	 2913864 allocs/op
```

After
```
BenchmarkTopologyList-2   	       1	1511797615 ns/op	347651256 B/op	 2753226 allocs/op
```

### DIFF
--- a/render/id.go
+++ b/render/id.go
@@ -2,7 +2,6 @@ package render
 
 import (
 	"net"
-	"sort"
 	"strings"
 
 	"github.com/weaveworks/scope/probe/endpoint"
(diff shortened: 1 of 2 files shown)


render/id.go
```diff
@@ -62,15 +59,24 @@ func NewDerivedExternalNode(n report.Node, addr string, local report.Networks) (
...
-// DNSNames returns a prioritized list of snooped and reverse-resolved
-// DNS names associated with node n.
-func DNSNames(n report.Node) []string {
-	snoopedNames, _ := n.Sets.Lookup(endpoint.SnoopedDNSNames)
-	reverseNames, _ := n.Sets.Lookup(endpoint.ReverseDNSNames)
-	// sort the names, to make selection for display more
+// DNSFirstMatch returns the first DNS name where match() returns
+// true, from a prioritized list of snooped and reverse-resolved DNS
+// names associated with node n.
+func DNSFirstMatch(n report.Node, match func(name string) bool) (string, bool) {
```
rade: D'oh.

rade: Making this higher-order seems overkill. We could call the function `FirstKnownService`.

bboreham: You see it's called differently in two places?

PR #2922, opened 2017-11-04
Replace knownService gcache with simpler map

bboreham:
The function `isKnownService` uses `gcache` as a speed-up; however because `gcache` is `interface{}`->`interface{}` every lookup causes a heap allocation as the string parameter escapes.  A `map[string]bool` avoids this problem, at the cost of a harsher cache trimming strategy.

Before
```
$ go test -tags 'unsafe netgo' -run=x -bench=Topology -bench-report-file=prod-report1.json
BenchmarkTopologyList-2   	       1	1479321809 ns/op	351308800 B/op	 2913864 allocs/op
```

After
```
$ go test -tags 'unsafe netgo' -run=x -bench=Topology -bench-report-file=prod-report1.json
BenchmarkTopologyList-2   	       1	1392126833 ns/op	351000392 B/op	 2859230 allocs/op
```

### DIFF
--- a/render/theinternet.go
+++ b/render/theinternet.go
@@ -3,8 +3,7 @@ package render
 import (
 	"regexp"
 	"strings"
-
-	"github.com/bluele/gcache"
+	"sync"
 
 	"github.com/weaveworks/scope/probe/host"
 	"github.com/weaveworks/scope/report"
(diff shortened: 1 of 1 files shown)


render/theinternet.go
```diff
@@ -46,18 +45,26 @@ var (
 	//
 	// Since names are generally <50 bytes, this shouldn't weight in
 	// at more than a few MB of memory.
-	knownServiceCache = gcache.New(10000).ARC().Build()
+	knownServiceCache        = map[string]bool{}
```
rade: Why not `map[string]struct{}{}`?

bboreham: It caches both positive and negative lookups, just as the previous code did.

bboreham:
replaced by #2939

PR #2926, opened 2017-11-07
produce stats as part of rendering

rade:
Stats are easily (and efficiently) produced as part of `Render`ing so there is no need for a separate `Stats` method.

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -504,19 +504,19 @@ func decorateWithStats(rpt report.Report, renderer render.Renderer, decorator re
 		realNodes int
 		edges     int
 	)
-	for _, n := range renderer.Render(rpt, decorator) {
+	r := renderer.Render(rpt, decorator)
+	for _, n := range r.Nodes {
 		nodes++
 		if n.Topology != render.Pseudo {
 			realNodes++
 		}
 		edges += len(n.Adjacency)
 	}
-	renderStats := renderer.Stats(rpt, decorator)
 	return topologyStats{
 		NodeCount:          nodes,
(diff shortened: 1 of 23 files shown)


rade:
I benchmarked this by feeding a recent production report to
```
prog/scope --mode=app --weave=false --app.collector=file:report.json
```
and then timing the fetching of various topologies with
```
time curl -so /dev/null http://localhost:4040/api/topology/<topology>
```
taking care to restart the above after every test.

| topology | master | branch |
| -- | -- | -- |
| processes  | 0.244 | 0.242 |
| containers | 0.765 | 0.786 |
| pods       | 0.522 | 0.521 |
| hosts      | 0.689 | 0.666 |
| (all)      | 0.831 | 0.736 |

Results are best run of 5.

"(all)" represents fetching `/api/topology`. There is a decent improvement. That api endpoint returns stats, and previously the stats calculation executed some renderers twice.

I also checked the validity of the result by comparing the output of master and branch of
```
curl -s "http://localhost:4040/api/topology" | jq -S '.' > topology.json
curl -s "http://localhost:4040/api/topology?namespace=kube-system&pseudo=show&system=system&stopped=stopped" | jq -S '.' > topology-filtered.json
```

bboreham:
Not something you changed in this PR, but is the "merge" of filtering stats correct?
Seems to me if two overlapping sets are filtered we will double-count.

rade:
>  Seems to me if two overlapping sets are filtered we will double-count.
Correct. I noticed that too. I think it's fine in most cases because overlapping filtered sets are rare.

PR #2929, opened 2017-11-10
Upgrade to React 16

aaron7:
- Upgrade to React 16
- Upgrade `weaveworks-ui-components` to 0.2.0 with React 16 support
- Create `nowInSecondsPrecision` as it was removed in https://github.com/weaveworks/ui-components/pull/59/files
- Remove `react-addons-perf` in debug toolbar as it is no longer supported in React 16
- Replace redux devtools with https://github.com/zalmoxisus/redux-devtools-extension which does not require any packages and redux devtools includes dependencies locked to versions which do not support React 16


client/package.json
```diff
@@ -43,8 +42,8 @@
     "reqwest": "2.0.5",
     "reselect": "3.0.1",
     "reselect-map": "1.0.3",
-    "styled-components": "^2.2.1",
-    "weaveworks-ui-components": "git+https://github.com/weaveworks/ui-components.git#v0.1.51",
+    "styled-components": "2.2.1",
+    "weaveworks-ui-components": "git+https://github.com/weaveworks/ui-components.git#v0.2.0",
```
fbarl: We need to merge https://github.com/weaveworks/scope/pull/2919 first because of Time Travel API changes in `v0.1.52`. Will test all the behaviour more thoroughly once that's done.

fbarl: Alternatively, just copying the changes from the TT wrapper https://github.com/weaveworks/scope/pull/2919/files#diff-d79edc7959396ddc637201c514c8638b would probably do the trick.

fbarl:
I pushed a commit that fixes two warnings I was getting:

err0
err1

Other thing is that part of debug toolbar panel seems to be broken, but that wasn't introduced in this branch.

@aaron7 please take a quick look at my last commit, the rest LGTM! :tada:

PR #2930, opened 2017-11-10
Remove optional dependencies

jpellizzari:
Optional dependencies were incorrectly declared, causing warnings when using Scope as a component.

`optionalDependencies` are not actually optional for the user. They are optional in the sense that if they fail to install or are not found, they can be skipped.

https://docs.npmjs.com/files/package.json#optionaldependencies


jpellizzari:
Service-ui was complaining that things listed in the `optionalDependencies` had unmet peer deps. This will remove them from the service-ui package, since it won't install devDeps.

PR #2939, opened 2017-11-14
Change isKnownService() cache to reduce memory allocation

bboreham:
`gcache` took an `interface{}`, so every time it is called a string header object gets allocated on the heap.
`go-cache` takes a `string` so avoids this allocation

Cache eviction is time-based so not as nice as before.

Benchmarks before (c5bdebdffebfa7cc0839770a3bf70429260f1bc9):
```
BenchmarkTopologyList-2         	       2	 603197223 ns/op	112161732 B/op	 1253503 allocs/op
BenchmarkTopologyHosts-2        	       3	 385415232 ns/op	82957173 B/op	 1015204 allocs/op
BenchmarkTopologyContainers-2   	      10	 192766565 ns/op	33145659 B/op	  407698 allocs/op
```

after this PR:
```
BenchmarkTopologyList-2         	       2	 528033687 ns/op	110843492 B/op	 1179523 allocs/op
BenchmarkTopologyHosts-2        	       3	 379481628 ns/op	81654378 B/op	  941519 allocs/op
BenchmarkTopologyContainers-2   	      10	 164464821 ns/op	32557828 B/op	  371446 allocs/op
```

### DIFF
--- a/render/theinternet.go
+++ b/render/theinternet.go
@@ -3,8 +3,9 @@ package render
 import (
 	"regexp"
 	"strings"
+	"time"
 
-	"github.com/bluele/gcache"
+	gocache "github.com/patrickmn/go-cache"
 
 	"github.com/weaveworks/scope/probe/host"
 	"github.com/weaveworks/scope/report"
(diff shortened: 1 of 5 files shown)


rade:
> Perhaps give https://github.com/gocql/gocql/blob/master/internal/lru/lru.go a try

That didn't work out because the package has "internal" in the path, which go refuses to import. I found a better implementation though, that includes locking. See #2952 for a PR replacing this one.

bboreham:
Replaced by #2952

Issue #2945, opened 2017-11-18
namespace filter calculation is horrendously inefficient


rade:
The namespace filter calculation in `updateKubeFilters` trawls through (nearly) all k8s topologies. Instead we should just get the probes to tell us what namespaces exist. That does mean we'd be including empty namespaces in filters in the UI, but that surely is a price worth paying. And arguably less surprising for users.

PR #2946, opened 2017-11-18
Expand the app's k8s namespace calculation

rade:
to include recently added k8s types.

This is all rather inefficient. See #2945.


bboreham:
How often is this called?

rade:
> How often is this called?

It's called for every topology request.

PR #2947, opened 2017-11-19
Decorators, begone!

rade:
Decorators turn out to be an unnecessary abstraction that rather obscures what's going on. So let's get rid of them!

Decorators were used to dynamically construct filters from UI selections - e.g. namespaces, system vs app containers, etc; the selectors in the bottom left corner of the current scope UI - and apply them to the output of topology renderers - pods, containers-by-image, etc; the selectors at the top centre of the current scope UI. Decorators were accomplishing that in a rather complicated way, by composing renderers that wrap the base topology renderers.

This PR unravels all that.

The filter selection from the UI is now turned into a `FilterFunc`, which we `Apply()` to the output of topology renderers. The `render.Render()` function does exactly that, and very clearly so. The one place where we need to do something special is when rendering a single node, e.g. for the details panel in the UI. See comments in the `handleNode` code.

The PR looks larger than it actually is. There are a bunch of mechanical changes that touch several files, especially the removal of Decorators from the `Renderer.Render()` signature, and use of the new `render.Render()` function. The individual commits are self-contained, but the first one is very much an intermediate step which is largely superseded by later commits. Also note that GH does a poor job of showing the diff for the change that moves the filter logic to `FilterFunc.Apply()`.

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -485,17 +485,17 @@ func (r *Registry) renderTopologies(rpt report.Report, req *http.Request) []APIT
 	req.ParseForm()
 	r.walk(func(desc APITopologyDesc) {
 		renderer, filter, _ := r.RendererForTopology(desc.id, req.Form, rpt)
-		desc.Stats = decorateWithStats(rpt, renderer, filter)
+		desc.Stats = computeStats(rpt, renderer, filter)
 		for i, sub := range desc.SubTopologies {
 			renderer, filter, _ := r.RendererForTopology(sub.id, req.Form, rpt)
-			desc.SubTopologies[i].Stats = decorateWithStats(rpt, renderer, filter)
+			desc.SubTopologies[i].Stats = computeStats(rpt, renderer, filter)
 		}
 		topologies = append(topologies, desc)
 	})
 	return updateFilters(rpt, topologies)
 }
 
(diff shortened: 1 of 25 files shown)


rade:
I've tested the correctness of this by running
```
echo > report.json
for r in processes containers pods hosts \
         "processes?unconnected=hide" \
         "containers?pseudo=hide&namespace=default,scope&system=application&stopped=running" \
         "pods?pseudo=hide&namespace=default,scope" ; do \
    echo $r
    curl -s "http://localhost:4040/api/topology/$r" | jq -S '.' >> report.json; \
done
```
against master and this branch and diffing the resulting reports.

app/api_topologies.go
```diff
@@ -487,24 +484,24 @@ func (r *Registry) renderTopologies(rpt report.Report, req *http.Request) []APIT
...
+			renderer, filter, _ := r.RendererForTopology(sub.id, req.Form, rpt)
+			desc.SubTopologies[i].Stats = decorateWithStats(rpt, renderer, filter)
 		}
 		topologies = append(topologies, desc)
 	})
 	return updateFilters(rpt, topologies)
 }
 
-func decorateWithStats(rpt report.Report, renderer render.Renderer, decorator render.Decorator) topologyStats {
+func decorateWithStats(rpt report.Report, renderer render.Renderer, filter render.FilterFunc) topologyStats {
```
bboreham: feels like this name should change - `computeFilterStats` ?

PR #2952, opened 2017-11-25
optimisation: faster knownServiceCache

rade:
We use an LRU cache implementation that is keyed on strings, which avoids conversion to/from interface and the memory allocation that entails.

This supersedes #2939.

### DIFF
--- a/render/render.go
+++ b/render/render.go
@@ -230,5 +230,5 @@ func (ret *joinResults) result() Nodes {
 // cache.
 func ResetCache() {
 	renderCache.Purge()
-	knownServiceCache.Purge()
+	purgeKnownServiceCache()
 }
--- a/render/theinternet.go
+++ b/render/theinternet.go
@@ -4,7 +4,7 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/bluele/gcache"
+	"github.com/camlistore/camlistore/pkg/lru"
 
 	"github.com/weaveworks/scope/probe/host"
 	"github.com/weaveworks/scope/report"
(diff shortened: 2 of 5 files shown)


rade:
benchmarks...

Running
```
go test -tags 'netgo unsafe' -run=x -cpu=2 -bench=Topology -bench-report-file=/tmp/prod-on-prod-report.json
```

master
```
BenchmarkTopologyList-2         	       3	 445479318 ns/op	97718400 B/op	  917789 allocs/op
BenchmarkTopologyHosts-2        	       3	 392563280 ns/op	71954213 B/op	  732940 allocs/op
BenchmarkTopologyContainers-2   	      10	 177969226 ns/op	30143139 B/op	  303605 allocs/op
```
branch
```
BenchmarkTopologyList-2         	       5	 348859962 ns/op	95724460 B/op	  799855 allocs/op
BenchmarkTopologyHosts-2        	      10	 270307551 ns/op	69933334 B/op	  614790 allocs/op
BenchmarkTopologyContainers-2   	      20	 125542963 ns/op	29164020 B/op	  244730 allocs/op
```
So that's a significant reduction in allocations and runtime. The latter isn't quite as strong in
```
time curl -o /dev/null -s http://localhost:4040/api/topology/<topology>
```
but still noticable:

|            |   master |   branch |
| ------------ | ------- | ------- |
| processes  | 0.144 | 0.139 |
| containers | 0.389 | 0.335 |
| pods       | 0.326 | 0.190 |
| services   | 0.253 | 0.176 |
| hosts      | 0.328 | 0.235 |

PR #2956, opened 2017-12-02
optimisation: allocate less memory in LatestMap merging

rade:
Most maps we merge have the same keys, or at least one set of keys is the subset of the other. Therefore, allocate a result slice capable of holding only the max of number keys, rather than the sum.


rade:
benchmarks, using 10s worth of reports, best results of three runs of
```
go test -tags 'netgo unsafe' -run=x -cpu=2 -bench='ReportMerge|Topology' --bench-report-path=/tmp/prod-20171129-10s/
```

master
```
BenchmarkReportMerge-2          	       1	1607241469 ns/op	493586816 B/op	 2617770 allocs/op
BenchmarkTopologyList-2         	       3	 475206422 ns/op	85075605 B/op	  761465 allocs/op
BenchmarkTopologyHosts-2        	       5	 328691841 ns/op	61066236 B/op	  572573 allocs/op
BenchmarkTopologyContainers-2   	      10	 170538039 ns/op	27480500 B/op	  247852 allocs/op
```
branch
```
BenchmarkReportMerge-2          	       1	1576221340 ns/op	410041360 B/op	 2635871 allocs/op
BenchmarkTopologyList-2         	       3	 463841864 ns/op	84564133 B/op	  759849 allocs/op
BenchmarkTopologyHosts-2        	       5	 343044301 ns/op	60801827 B/op	  572695 allocs/op
BenchmarkTopologyContainers-2   	      10	 165845244 ns/op	27234359 B/op	  247834 allocs/op
```
So we allocate ~17% less memory during merging.

I included the rendering benchmarks because rendering also merges nodes, so I wanted to check that we didn't negatively impact performance, in particular increase allocation count.

bboreham:
Did you run this with memory profiling?  That would show how many times the `append()` causes an alloc after this change.

rade:
Surely the increase in the figure should be 263083-2609989.

rade:
I don't see any allocations in the `append`s when running `BenchmarkReportMerge`:
```
$ go test -tags 'netgo unsafe' -run=x -cpu=2 -memprofile=/tmp/mem.out -memprofilerate=111 -bench='ReportMerge' --bench-report-path=/tmp/prod-20171129-10s/
BenchmarkReportMerge-2   	       1	9429282808 ns/op	398308736 B/op	 2595429 allocs/op
PASS
$ go tool pprof -alloc_objects -list=StringLatestMap.Merge /tmp/mem.out
Total: 8517392
ROUTINE ======================== github.com/weaveworks/scope/report.StringLatestMap.Merge in /home/matthias/go/src/github.com/weaveworks/scope/report/latest_map_generated.go
    132977     132977 (flat, cum)  1.56% of Total
         .          .     53:	}
         .          .     54:	l := len(m.entries)
         .          .     55:	if len(n.entries) > l {
         .          .     56:		l = len(n.entries)
         .          .     57:	}
    132977     132977     58:	out := make([]stringLatestEntry, 0, l)
         .          .     59:
         .          .     60:	i, j := 0, 0
         .          .     61:	for i < len(m.entries) {
         .          .     62:		switch {
         .          .     63:		case j >= len(n.entries) || m.entries[i].key < n.entries[j].key:

```
Same for `NodeControlDataLatestMap.Merge`

To check that I wasn't doing anything wrong in the profiling, I changed the code to allocate `l-1` instead of `l`. That _does_ result in allocations in `append`.

So now I mystified why my the benchmark results show an increase in the `allocs/op`.

PR #2957, opened 2017-12-04
Stop reporting ReplicaSets

### LINKED ISSUE #2662: don't report on replicasets from probe

rbruggem:
Also, add Deployment as Pod parent.
Fixes #2662

### DIFF
--- a/probe/kubernetes/reporter.go
+++ b/probe/kubernetes/reporter.go
@@ -260,11 +260,7 @@ func (r *Reporter) Report() (report.Report, error) {
 	if err != nil {
 		return result, err
 	}
-	replicaSetTopology, replicaSets, err := r.replicaSetTopology(r.probeID, deployments)
-	if err != nil {
-		return result, err
-	}
-	podTopology, err := r.podTopology(services, replicaSets, daemonSets, statefulSets, cronJobs)
+	podTopology, err := r.podTopology(services, deployments, daemonSets, statefulSets, cronJobs)
 	if err != nil {
 		return result, err
 	}
@@ -275,7 +271,6 @@ func (r *Reporter) Report() (report.Report, error) {
 	result.StatefulSet = result.StatefulSet.Merge(statefulSetTopology)
 	result.CronJob = result.CronJob.Merge(cronJobTopology)
 	result.Deployment = result.Deployment.Merge(deploymentTopology)
-	result.ReplicaSet = result.ReplicaSet.Merge(replicaSetTopology)
 	return result, nil
 }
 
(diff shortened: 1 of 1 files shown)


probe/kubernetes/reporter.go
```diff
@@ -446,7 +393,7 @@ func match(namespace string, selector labels.Selector, topology, id string) func
 	}
 }
 
-func (r *Reporter) podTopology(services []Service, replicaSets []ReplicaSet, daemonSets []DaemonSet, statefulSets []StatefulSet, cronJobs []CronJob) (report.Topology, error) {
+func (r *Reporter) podTopology(services []Service, daemonSets []DaemonSet, statefulSets []StatefulSet, cronJobs []CronJob, deployments []Deployment) (report.Topology, error) {
```
rade: pls put `deployments` in the same place where `replicaSets` used to be.

probe/kubernetes/reporter.go
```diff
@@ -474,16 +421,16 @@ func (r *Reporter) podTopology(services []Service, replicaSets []ReplicaSet, dae
...
 			return pods, err
 		}
 		selectors = append(selectors, match(
-			replicaSet.Namespace(),
+			deployment.Namespace(),
 			selector,
-			report.ReplicaSet,
-			report.MakeReplicaSetNodeID(replicaSet.UID()),
+			report.Deployment,
+			report.MakeDeploymentNodeID(deployment.UID()),
```
rade: How does this work?

rbruggem: `match` will add the deployment as the pod's parent if the following conditions are true:
* the pod and the deployment belong to the same namespace
* the pod and the deployment match the same labels.

rbruggem: https://github.com/weaveworks/scope/blob/master/probe/kubernetes/reporter.go#L403

rbruggem: Deployments, ReplicaSets and Pods will be assigned the same label in the manifest.
The label is required: https://github.com/kubernetes/kubernetes/issues/28475

PR #2961, opened 2017-12-06
Remove struct wrapping LatestMap

bboreham:
It isn't necessary, and it prevents the codec from seeing maps as empty

Partial fix for https://github.com/weaveworks/scope/issues/1201#issuecomment-235205270 - this makes `omitempty` work for `controls` and `latest`.

### DIFF
--- a/extras/generate_latest_map
+++ b/extras/generate_latest_map
@@ -62,7 +62,7 @@ function generate_latest_map() {
     }
 
     // ${latest_map_type} holds latest ${data_type} instances, as a slice sorted by key.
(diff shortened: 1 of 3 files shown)


rade:
Interesting. Would similar considerations apply to `report.Sets`?

rade:
Have you got any report size comparisons? And codec benchmarks?

bboreham:
The size impact is small on one report I checked - 3,173,719 - >3,151,135 bytes - it had mostly non-empty `latest`.  Benchmark impact will be similar.

`Sets` gets the same benefit in #2962

Issue #2959, opened 2017-12-05
more efficient report upgrade


rade:
ATM `report.Upgrade()`
- does a lot of work even when no upgrading is necessary
- gets called on merged reports, which means upgraded reports aren't stored, cached, etc, so the upgrading work is repeated over and over

Instead we should make `report.Upgrade()`
- do very little work when no upgrading is necessary. Report schema versioning (#1061) would help here, but I don't think is essential for now
- operate on the "ingest" path, so we store upgraded reports (we still need it in the query path too, because there will be stored, non-upgraded reports)

rade:
>  - do very little work when no upgrading is necessary.

Done in #2965.

>   - operate on the "ingest" path

I have created #2967 for that.

Issue #2960, opened 2017-12-05
Move selectPodsWithDeployments.Render() logic to report.Upgrade()


rbruggem:
`selectPodsWithDeployments.Render()` ensures pods reference deployments directly as parents.
This logic should be moved to `report.Upgrade()`.

PR #2963, opened 2017-12-07
`report` upgrade based on probe version

rbruggem:
* `report.node.LatestControls` and `NodeControlData` was introduced before version 0.16.2, therefore only upgrade the report if the probe's version is below.
* Do not upgrade the report if both the probe and the app have git revisions as version and they are equal.
* The probe's version is stored in the header and in the report itself. Only the version in the report is used to avoid issues when merging reports.
* `Upgrade()` is now also called on the ingest path

See #2959

### DIFF
--- a/common/version/version.go
+++ b/common/version/version.go
@@ -1,0 +1,75 @@
+package version
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+)
+
+const (
+	versionRegexp = "^v?(\\d+)\\.(\\d+)\\.(\\d+)$"
+)
+
+// Version represents a release version decomposed in (major,minor,patch)
+type Version struct {
+	major int
+	minor int
+	patch int
+}
+
+// New returns a new Version
+func New(major, minor, patch int) Version {
+	return Version{major, minor, patch}
+}
+
+// ParseVersion parses a version string and, if successful, returns a Version
+func ParseVersion(version string) (*Version, error) {
+	// the regexp catches this, short circuiting does not hurt though
+	if version == "" {
+		return nil, fmt.Errorf("Invalid version: %v", version)
+	}
+
+	re := regexp.MustCompile(versionRegexp)
+	match := re.FindStringSubmatch(version)
+	if len(match) != 4 {
+		return nil, fmt.Errorf("Invalid version: %v", version)
+	}
+
+	major, err := strconv.Atoi(match[1])
+	if err != nil {
+		return nil, fmt.Errorf("Invalid version: %v", version)
+	}
+	minor, err := strconv.Atoi(match[2])
+	if err != nil {
+		return nil, fmt.Errorf("Invalid version: %v", version)
+	}
+	patch, err := strconv.Atoi(match[3])
+	if err != nil {
+		return nil, fmt.Errorf("Invalid version: %v", version)
+	}
+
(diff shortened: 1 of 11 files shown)


common/version/version.go
```diff
@@ -0,0 +1,75 @@
+package version
```
rade: We are already vendoring https://github.com/blang/semver. There's also https://github.com/Masterminds/semver, https://github.com/coreos/go-semver and https://github.com/hashicorp/go-version. Do none of these do what we need here?

report/report.go
```diff
@@ -422,4 +486,7 @@ const (
 	HostNodeID = "host_node_id"
 	// ControlProbeID is the random ID of the probe which controls the specific node.
 	ControlProbeID = "control_probe_id"
+
+	// ScopeVersion is the probe's version
+	ScopeVersion = "host_scope_version"
```
rade: I reckon this should be called `HostScopeVersion`.

probe/host/reporter.go
```diff
@@ -23,7 +23,6 @@ const (
 	Load1         = "load1"
 	CPUUsage      = "host_cpu_usage_percent"
 	MemoryUsage   = "host_mem_usage_bytes"
-	ScopeVersion  = "host_scope_version"
```
rade: I'd alias that as `ScopeVersion = report.HostScopeVersion`, which requires fewer code changes and keeps the call sites shorter.

report/report.go
```diff
@@ -340,7 +345,11 @@ func (r Report) Validate() error {
 // Upgrade returns a new report based on a report received from the old probe.
 //
 // This for now creates node's LatestControls from Controls.
-func (r Report) Upgrade() Report {
+func (r Report) Upgrade(appVersion string) Report {
+	if !r.RequiresUpgrade(appVersion) {
```
rade: I reckon `RequiresUpgrade` is too high-level an API. We will eventually have multiple different kinds of upgrade to perform, depending on the version. E.g. consider how you'd expand this function to deal with #2960.

I reckon a `err, semver := Report.ScopeVersion()` function that returns the lowest scope semver or error (if none can be found) would work better.

Then this function can dispatch different upgrade tasks depending on comparing that version to specific known versions like `latestControlsRelease`.

rade: There is another problem with the approach here. In #2959 I suggested that upgrades should take place in the "ingest" path, so we store upgraded reports. But the check here would not notice that, so upgrades would still be performed when an already-upgraded stored report is subsequently read.

report/report.go
```diff
@@ -360,6 +369,61 @@ func (r Report) Upgrade() Report {
 	return cp
 }
 
+// RequiresUpgrade returns whether the report needs upgrading
+func (r Report) RequiresUpgrade(appVersion string) bool {
```
rade: Feeding the app version through here, and the whole `requiresUpgradeRevisionCheck()` logic, is overkill. If we get a report from a probe with a git revision then we should always upgrade. That will only affect us, for probes we are running in dev and during local development.

rade:
Given the results in #2965, I don't think this change is worthwhile.

rade:
Given the results in #2965, I don't think this change is worthwhile. Let's revisit when we encounter a case where there is no cheap no-op path in an upgrade.

PR #2965, opened 2017-12-07
make report upgrading fast when it's a no-op

rade:
The vast majority of the cost is memory allocation, so doing a first pass to see whether any upgrading is necessary at all, and thus avoiding allocation when it isn't, is a massive saving.

Part of resolving #2959.

### DIFF
--- a/report/report.go
+++ b/report/report.go
@@ -341,6 +341,18 @@ func (r Report) Validate() error {
 //
 // This for now creates node's LatestControls from Controls.
 func (r Report) Upgrade() Report {
+	needUpgrade := false
+	r.WalkTopologies(func(topology *Topology) {
+		for _, node := range topology.Nodes {
+			if node.LatestControls.Size() == 0 && len(node.Controls.Controls) > 0 {
+				needUpgrade = true
+			}
+		}
+	})
+	if !needUpgrade {
+		return r
+	}
+
 	cp := r.Copy()
 	ncd := NodeControlData{
 		Dead: false,
--- a/app/benchmark_internal_test.go
+++ b/app/benchmark_internal_test.go
@@ -67,6 +67,20 @@ func BenchmarkReportMerge(b *testing.B) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


rade:
benchmarks from running
```
go test -tags 'netgo unsafe' -run=x -cpu=2 -benchmem -bench='Report.*' --bench-report-path=/tmp/prod-20171204-15s
```
with some sizable reports that need to upgrading.

master
```
BenchmarkReportUnmarshal-2   	       1	2438382196 ns/op	458421984 B/op	 3477383 allocs/op
BenchmarkReportMerge-2       	       1	1549196525 ns/op	365303288 B/op	 1994498 allocs/op
BenchmarkReportUpgrade-2     	      20	 112642220 ns/op	26312183 B/op	   85406 allocs/op
```
branch
```
BenchmarkReportUnmarshal-2   	       1	2471503100 ns/op	458430336 B/op	 3477405 allocs/op
BenchmarkReportMerge-2       	       1	1537666375 ns/op	357025768 B/op	 1974748 allocs/op
BenchmarkReportUpgrade-2     	     200	   7071493 ns/op	    3072 B/op	       2 allocs/op
PASS
```

So upgrading was already only costing about 7% of merging. With this change, it costs 0.46%.

report/report.go
```diff
@@ -341,6 +341,18 @@ func (r Report) Validate() error {
 //
 // This for now creates node's LatestControls from Controls.
 func (r Report) Upgrade() Report {
+	needUpgrade := false
+	r.WalkTopologies(func(topology *Topology) {
+		for _, node := range topology.Nodes {
+			if node.LatestControls.Size() == 0 && len(node.Controls.Controls) > 0 {
+				needUpgrade = true
```
rade: An early exit would be nice here, but that would require introducing an alternative to `WalkTopologies`, which just isn't worth the effort, given that we hit this branch when an upgrade _is_ required, and upgrading is orders of magnitudes more expensive than doing this walk to the end, so we'd be saving relatively little.

Issue #2967, opened 2017-12-08
upgrade reports on ingest


rade:
`report.Upgrade` gets called on merged reports, which means upgraded reports aren't stored, cached, etc, so the upgrading work is repeated over and over.

Instead we should perform the upgrade on the "ingest" path, so we store upgraded reports.

We still need to upgrade in the query path too, because there will be stored, non-upgraded reports, but that will only be needed for historic reports.

This issue is a follow-up to #2959.

rade:
This is a definite win in query latency but at possible higher base line CPU utilisation. Performing the upgrade on ingest means it is done for every report. Whereas on query it is done only for reports which are visualised.

We may want to gather some stats here before proceeding.

PR #2968, opened 2017-12-08
Upgrade reports on the app's ingest path

rbruggem:
See #2967


rade:
See comments in #2967 why this may not necessarily be a good idea.

PR #2971, opened 2017-12-08
probe: Use an absolute FQDN for cloud.weave.works by default

dlespiau:
This can save a few DNS queries here and there!


dlespiau:
Even tested it now! reports are correctly send when using this:

```
time="2017-12-08T16:58:31Z" level=info msg="publishing to: https://cloud.weave.works.:443" 
```

Issue #2974, opened 2017-12-11
don't embed docker binary


rade:
The scope container currently embeds a docker binary, for use by the Weave Net API. This should no longer be necessary.

PR #2975, opened 2017-12-11
bump embedded Weave Net version to 2.1.3

rade:
I've tested this while running
- no Weave Net
- Weave Net 1.9.0
- Weave Net 2.1.3

Precursor to addressing #2974.


rade:
> Weave Net 1.9.0

There's actually a problem here:
```
<app> WARN: 2017/12/11 14:24:00.073314 Error updating weaveDNS, backing off 20s: Error running weave expose: exit status 4: ""
```
which occurs because in 2.1, `expose` is done via a new endpoint on the http API.

That breaks registering of the app with weaveDNS and hence discovery of the app via weaveDNS by probes.

I think it's ok to say to users who want that functionality in Scope 1.7 that they need to run a 2.1 Weave Net. @bboreham what do you reckon?

bboreham:
Why is it calling `expose` ?  I haven't dug all the way through the chain of dependencies.

rade:
> Why is it calling `expose`

I was wondering the same. My theory is that it's done in order to get a Weave IP (and register that in WeaveDNS) even though the app is running in the host netns.

bboreham:
Seems to come from #302, and the code seems to have used DNS before that.

Certainly the code now registers all available IP addresses with WeaveDNS, so it isn't specifically trying to get one on the Weave network.

Looks unnecessary to me.

rade:
> Looks unnecessary to me.

It is necessary for probes to able to connect to apps over the weave network.

Note that it checks first whether there's already an exposed address, So this would still work for users of Weave Net prior to version 2.1 provided they run `weave expose` themselves.

bboreham:
> It is necessary for probes to able to connect to apps over the weave network.

Do you mean it is aimed at the case where the probe cannot connect to the app except over the Weave network?  

AFAICS it registers every address the app is listening on, so which address the probe chooses is unpredictable, so in those cases it would fail much of the time.

rade:
> Do you mean it is aimed at the case where the probe cannot connect to the app except over the Weave network?

Yes.

> AFAICS it registers every address the app is listening on,

Correct.

>  so which address the probe chooses is unpredictable, so in those cases it would fail much of the time.

It sends to all of them.

PR #2976, opened 2017-12-11
Lock styled-components version and upgrade ui-components

aaron7:
Lock `styled-components` version. This was causing problems when the version did not match in service-ui (https://github.com/styled-components/styled-components/issues/1305).

Also:
- upgrade ui-components version to latest (now via npm)
- upgrade `d3-selection` to 1.2.0


client/yarn.lock
```diff
@@ -1920,10 +1920,26 @@ d3-scale@1.0.6:
...
+  dependencies:
+    d3-array "^1.2.0"
+    d3-collection "1"
+    d3-color "1"
+    d3-format "1"
+    d3-interpolate "1"
+    d3-time "1"
+    d3-time-format "2"
+
 d3-selection@1, d3-selection@1.1.0:
```
fbarl: Damn, we should get rid of this.

client/package.json
```diff
@@ -42,8 +42,8 @@
     "reqwest": "2.0.5",
     "reselect": "3.0.1",
     "reselect-map": "1.0.3",
-    "styled-components": "^2.2.1",
-    "weaveworks-ui-components": "git+https://github.com/weaveworks/ui-components.git#v0.2.1",
+    "styled-components": "2.2.4",
+    "weaveworks-ui-components": "0.3.6",
```
fbarl: Hope this will not be causing `service-ui` CI to fail once it's pulled in.

PR #2978, opened 2017-12-12
don't exclude NATed connections in mapping to processes

### LINKED ISSUE #2943: no connections between containers on GKE
We've got a cluster on GKE, and scope shows pod connections fine, e.g.
screenshot from 2017-11-17 10 05 03
but it does not show container connections
screenshot from 2017-11-17 10 05 47
except when they are in the host netns. Neither does it show process connections unless they are on the host netns.

rade:
We used to ignore source endpoints that are associated with multiple destination endpoints, which is a partial workaround for our inability to correctly represent two connections from the same source ip/port
but different processes, or the same destination ip/port but different processes. See #2665.

However, that condition is too coarse. In particular, we end up ignoring endpoints that are connected to NATed destinations, since the latter are represented by two (or more) endpoints.

The change here corrects that.

Fixes #2943 (or, rather, it fixes the bit that isn't covered by other issues).

### DIFF
--- a/render/process.go
+++ b/render/process.go
@@ -2,6 +2,7 @@ package render
 
 import (
 	"github.com/weaveworks/scope/probe/docker"
+	"github.com/weaveworks/scope/probe/endpoint"
 	"github.com/weaveworks/scope/probe/process"
 	"github.com/weaveworks/scope/report"
 )
@@ -98,11 +99,7 @@ func (e endpoints2Processes) Render(rpt report.Report) Nodes {
 			if !ok {
 				continue
 			}
-
-			if len(n.Adjacency) > 1 {
-				// We cannot be sure that the pid is associated with all the
-				// connections. It is better to drop such an endpoint than
-				// risk rendering bogus connections.
+			if hasMoreThanOneConnection(n, endpoints.Nodes) {
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


render/process.go
```diff
@@ -124,6 +121,32 @@ func (e endpoints2Processes) Render(rpt report.Report) Nodes {
...
+// them, since the source endpoint may have been re-used by a
+// different process. See #2665. It is better to drop such an endpoint
+// than risk rendering bogus connections.  Aliased connections - when
+// all the remote endpoints represent the same logical endpoint, due
+// to NATing - are fine though.
+func hasMoreThanOneConnection(n report.Node, endpoints report.Nodes) bool {
+	if len(n.Adjacency) < 2 {
+		return false
+	}
+	realEndpointID := ""
```
bboreham: What is this?  Its value changes every time round the loop; how can it be "real" ?

render/process.go
```diff
@@ -124,6 +121,32 @@ func (e endpoints2Processes) Render(rpt report.Report) Nodes {
...
+	for _, endpointID := range n.Adjacency {
+		if ep, ok := endpoints[endpointID]; ok {
+			if copyID, _, ok := ep.Latest.LookupEntry(endpoint.CopyOf); ok {
+				endpointID = copyID
+			}
+		}
+		if realEndpointID != "" && realEndpointID != endpointID {
```
bboreham: convince me that this is the correct test for 3 or 4 items in `n.Adjacency`

render/process.go
rade: It's the first real endpoint id we encounter. And also the last, since the loop exits when we encounter another one.

render/process.go
rade: See above. The loop exists (by the `return false`) when encountering a 2nd real endpoint.

rade:
@bboreham PTAL

PR #2980, opened 2017-12-12
Upgrade weaveworks-ui-components to 0.3.10

aaron7:


dlespiau:
I think those ones could be done without review, the original PR in ui-components has been reviewed after all.

dlespiau:
Note: this will need to be bumped a bit more once the npm package isn't missing exports.

PR #2981, opened 2017-12-12
Update the ugorji json/msgpack library

bboreham:
I wouldn't ordinarily pick a beta version, but it seems the library is about to make a breaking change to the msgpack encoding.

Benchmarks are about the same, after some work:
Master:
```
BenchmarkReportUnmarshal-2   	       1	1770790307 ns/op	458396816 B/op	 3477329 allocs/op
BenchmarkDecode-2            	     100	  14350934 ns/op	  3005481 B/op	   51906 allocs/op
BenchmarkEncode-2            	     100	  16523432 ns/op	  6290948 B/op	   29857 allocs/op
```

This PR:
```
BenchmarkReportUnmarshal-2   	       1	1608796336 ns/op	458262800 B/op	 3477360 allocs/op
BenchmarkDecode-2            	     100	  15738725 ns/op	  3003764 B/op	   51905 allocs/op
BenchmarkEncode-2            	     100	  16321384 ns/op	  6318272 B/op	   29857 allocs/op
```


rade:
> I wouldn't ordinarily pick a beta version, but it seems the library is about to make a breaking change to the msgpack encoding.

Why update at all?

bboreham:
> Why update at all?

I started down this rabbit-hole to try to get rid of the memory allocations from `time.Time.MarshalBinary` encoding; this turns out to be inordinately hard and I wondered if the newer codec library would help.

It didn't, but it did trigger a discussion which fixed the `omitempty` for `NodeSet` etc.  Sadly this went in after the breaking change to `time.Time` encoding. I should be able to work something out with that.

So, this PR is a step along the way to somewhere better; not particularly worth merging today but useful to have on file.

2opremio:
@bboreham Are you planning to finish up this work anytime soon? Otherwise I would like to close it (I am cleaning up the lingering PRs)

bboreham:
Yeah, close it: the next ugorji update will have to deal with the change in `Time` encoding. 

Bryan

PR #2983, opened 2017-12-14
cheap probe connectedness api endpoint

### LINKED ISSUE #2982: low-cost app api endpoint to determine probe connectedness
In Weave Cloud we'd benefit from having an API endpoint that tells is whether there are any probes connected. ATM the UI is invoking `/api/probes` for that, which entails fetching and merging the reports for the current app window and hence is expensive.

rade:
This introduce a cheap `/api/probes?sparse` variant.

In many cases we only need to know whether there are _any_ connected probes, and not the probe details. Obtaining that info is cheaper since it requires no reading or merging or reports.

Fixes #2982.

### DIFF
--- a/app/api_report.go
+++ b/app/api_report.go
@@ -32,6 +32,16 @@ type probeDesc struct {
 // Probe handler
 func makeProbeHandler(rep Reporter) CtxHandlerFunc {
 	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+		r.ParseForm()
+		if _, sparse := r.Form["sparse"]; sparse {
+			// if we have reports, we must have connected probes
+			hasProbes, err := rep.HasReports(ctx, time.Now())
+			if err != nil {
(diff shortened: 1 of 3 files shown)


app/api_report.go
```diff
@@ -32,6 +32,16 @@ type probeDesc struct {
...
 func makeProbeHandler(rep Reporter) CtxHandlerFunc {
 	return func(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+		r.ParseForm()
+		if _, sparse := r.Form["sparse"]; sparse {
+			// if we have reports, we must have connected probes
+			hasProbes, err := rep.HasReports(ctx, time.Now())
+			if err != nil {
+				respondWith(w, http.StatusInternalServerError, err)
+			}
+			respondWith(w, http.StatusOK, hasProbes)
```
leth: Does this just respond with 'true' or 'false' as JSON?

rade: Yes. Is that a problem?

PR #2985, opened 2017-12-14
Probe reports namespaces

### LINKED ISSUE #2945: namespace filter calculation is horrendously inefficient

rbruggem:
If the probe reports namespace topologies, the app, in particular `updateKubeFilters()`, is able to avoid unnecessary loops.

Fixes #2945.

This PR also fixes incorrect comments and removes `WalkNodes()` which was unused.

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -95,26 +95,13 @@ func updateSwarmFilters(rpt report.Report, topologies []APITopologyDesc) []APITo
 }
 
 func updateKubeFilters(rpt report.Report, topologies []APITopologyDesc) []APITopologyDesc {
-	namespaces := map[string]struct{}{}
-	// We exclude ReplicaSets since we don't show them anywhere.
-	for _, t := range []report.Topology{rpt.Pod, rpt.Service, rpt.Deployment, rpt.DaemonSet, rpt.StatefulSet, rpt.CronJob} {
-		for _, n := range t.Nodes {
-			if state, ok := n.Latest.Lookup(kubernetes.State); ok && state == kubernetes.StateDeleted {
-				continue
-			}
-			if namespace, ok := n.Latest.Lookup(kubernetes.Namespace); ok {
-				namespaces[namespace] = struct{}{}
-			}
-		}
-	}
-	if len(namespaces) == 0 {
-		// We only want to apply k8s filters when we have k8s-related nodes,
-		// so if we don't then return early
-		return topologies
-	}
 	ns := []string{}
-	for namespace := range namespaces {
-		ns = append(ns, namespace)
+	for _, n := range rpt.Namespace.Nodes {
+		name, ok := n.Latest.Lookup(kubernetes.Name)
+		if !ok {
+			continue
+		}
+		ns = append(ns, name)
 	}
 	sort.Strings(ns)
 	topologies = append([]APITopologyDesc{}, topologies...) // Make a copy so we can make changes safely
--- a/probe/kubernetes/namespace.go
+++ b/probe/kubernetes/namespace.go
@@ -1,0 +1,28 @@
+package kubernetes
+
+import (
+	"github.com/weaveworks/scope/report"
+
+	apiv1 "k8s.io/client-go/pkg/api/v1"
+)
+
+// NamespaceResource represents a Kubernetes namespace
+// `Namespace` is already taken in meta.go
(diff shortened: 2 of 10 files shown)


probe/kubernetes/namespace.go
```diff
@@ -0,0 +1,24 @@
...
+// The GetNode function does not follow the conventions of other kubernetes api resources
+// (see cronjob.go, daemonsets.go, deployments.go, pod.go, replica_set.go, service.go, statefulset.go).
+// The above modules declare an interface nameed after the kubernetes resource and an implementation which includes
+// the kubernetes api resource and the meta struct.
+//
+// Namespace does not follow the convention because of the following:
+// 1. Declaring a Namespace interface clashes with the already existent kubernetes.Namespace constant in meta.go.
+//    This can be solved by naming the interface differently, but there's second issue:
+// 2. Defining a Namespace implementation and composing it with *apiv1.Namespace and Meta also produces name clashes
+//    between *apiv1.Namespace and meta.Namespace()
```
rade: so give the `*apiv1.Namespace` field a name.

report/report.go
```diff
@@ -409,6 +426,36 @@ func (r Report) upgradePodNodes() Report {
...
+		}
+	}
+
+	nodes := Nodes{}
+	for ns := range namespaces {
+		// Namespace ID
+		// Probes did not use to report namespace ids, but since creating a report node requires an id,
+		// the namespace name, which is unique, is passed to `MakeNamespaceNodeID
+		namespaceID := MakeNamespaceNodeID(ns)
+		nodes[namespaceID] = MakeNodeWith(namespaceID, map[string]string{KubernetesName: ns})
```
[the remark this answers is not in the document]
rade: Oh, is that because "proper" namespace nodes do get field courtesy of `meta.MetaNode()`? If so, should we be adding a `KubernetesNamespace` key too, since `meta.MetaNode()` adds that too?

It's a bit silly to have the name in three places, but probably not worth breaking the pattern for that.

report/report.go
```diff
@@ -428,6 +439,36 @@ func (r Report) upgradePodNodes() Report {
...
+			if namespace, ok := n.Latest.Lookup(KubernetesNamespace); ok {
+				namespaces[namespace] = struct{}{}
+			}
+		}
+	}
+
+	nodes := Nodes{}
+	for ns := range namespaces {
+		// Namespace ID:
+		// Probes did not use to report namespace ids, but since creating a report node requires an id,
```
rade: typo: use/used

report/report.go
```diff
@@ -428,6 +439,36 @@ func (r Report) upgradePodNodes() Report {
...
+			if state, ok := n.Latest.Lookup(KubernetesState); ok && state == KubernetesStateDeleted {
+				continue
+			}
+			if namespace, ok := n.Latest.Lookup(KubernetesNamespace); ok {
+				namespaces[namespace] = struct{}{}
+			}
+		}
+	}
+
+	nodes := Nodes{}
```
rade: `nodes := make(Nodes, len(namespaces))`

report/report_test.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/weaveworks/common/mtime"
 	"github.com/weaveworks/common/test"
+	"github.com/weaveworks/scope/probe/kubernetes"
```
rade: There should be suitable constants in `report` to avoid this import.

app/api_topologies.go
```diff
@@ -95,26 +95,14 @@ func updateSwarmFilters(rpt report.Report, topologies []APITopologyDesc) []APITo
...
-	}
 	ns := []string{}
-	for namespace := range namespaces {
-		ns = append(ns, namespace)
+	for _, n := range rpt.Namespace.Nodes {
+		name, ok := n.Latest.Lookup(kubernetes.Name)
+		if !ok {
+			log.Errorf("Error while fetching the namespace's name: %v", n)
+			continue
+		}
```
rade: I'd just drop this silently. The rendering code generally doesn't complain about absent data.

PR #2990, opened 2017-12-15
Suggest how to disable weave errors and warnings

### LINKED ISSUE #1726: weaveDNS warnings in deployments without Weave Net still too alarming
It causes questions from users, see https://weaveworks.slack.com/archives/weave-users/p1469675356000173

We had an issue covering this ( #1008 ) but the solution (downgrading the errors to warnings) doesn't make the user think that the messages are harmless.

rbruggem:
This applies when running scope deployments without Weave Net.

Fixes #1726.

### DIFF
--- a/common/weave/client.go
+++ b/common/weave/client.go
@@ -136,7 +136,7 @@ func (c *client) Status() (Status, error) {
 	req.Header.Add("Accept", "application/json")
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		return Status{}, err
+		return Status{}, errorf("%v", err)
 	}
 	defer resp.Body.Close()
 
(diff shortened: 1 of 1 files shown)


common/weave/client.go
```diff
@@ -165,7 +169,7 @@ func (c *client) AddDNSEntry(fqdn, containerID string, ip net.IP) error {
 	req.Header.Add("Content-Length", strconv.Itoa(len(data.Encode())))
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		return err
+		return fmt.Errorf("%v. %v", err, noWeaveTip)
```
rade: Given how often we do this, I suggest creating an `errorf` function with the same signature as `fmt.Errorf` that appends the message.

common/weave/client.go
```diff
@@ -251,3 +251,7 @@ func weaveCommand(arg ...string) exec.Cmd {
 	cmd.SetEnv(append(os.Environ(), "DOCKER_API_VERSION="+dockerAPIVersion))
 	return cmd
 }
+
+func errorf(format string, a...interface{}) error {
+	return fmt.Errorf(format + ". If you are not running Weave Net, you may wish to suppress this warning by launching scope with the `--weave=false` option.", a...)
```
rade: This package is the wrong place for that message. We have no idea that we are being called in the context of the scope app or probe here.

A possible fix here is to define a special error type, so that call sites can then distinguish between errors that may be due to Weave Net not being there, and others.

PR #2992, opened 2017-12-17
remove Node.Edges

rade:
It is unused and none of the adjacency mapping code in the renderer takes any notice of it. Removing this shrinks the report size.

Edges were introduced in #838. At the time we had an experimental packet sniffer under experimental/sniff/sniffer.go. That got removed in #1646.

We can resurrect this if we ever decide to add meta data to edges.


bboreham:
There are a number of comments in `report/report.go` saying "Edges are present" or "Edges are not present"; should they be updated ?

rade:
> There are a number of comments in report/report.go saying "Edges are present" or "Edges are not present"; should they be updated ?

Adjacencies *are* edges; we use the two interchangeably. I've updated the comments to clarify that relationship, and removed vestiges of edge metadata from comments too, and fixed an incorrect "are not present" assertion (edges on Overlay _are_ present).

Issue #2901, opened 2017-10-19
Display the OS name in the host details panel


dlespiau:
A variety of things can be used to get a user friendly OS name. These days `os-release`'s `PRETTY_NAME` ([spec](https://www.freedesktop.org/software/systemd/man/os-release.html)) should cover most of the OSes of what we care about.

dlespiau:
Hum, we may have a problem. Clearly I hadn't thought it through entirely, sorry.

The `/etc/os-release` file you can open from scope is the one from the scope Docker image. We need access to the underlying OS file. The simplest way I can think of is to add another volume/bind mount for the host `/etc/os-release` into scope's mount namespace. We already do something similar for a few host files.

In the `/scope` control script:

```
docker_args() {
    echo --privileged $USERNS_HOST --net=host --pid=host \
        -v /var/run/docker.sock:/var/run/docker.sock \
        -v /var/run/scope/plugins:/var/run/scope/plugins \
        -v /sys/kernel/debug:/sys/kernel/debug \
        -e CHECKPOINT_DISABLE
}
```
We need to choose where to put it, I wouldn't be shocked to have it mounted as `/var/run/scope/host-os-release`.

We can definitely have a series of fallback options for the string. I think I'd always use the same `OS` field, trying to get the most user friendly string as possible. This could look like:

1. `$PRETTY_NAME`
2. `$NAME $VERSION`
3. `$ID $VERSION_ID`
4. runtime.GOOS

JordanJLopez:
So we have mounted the host's /etc/os-release and that seems to work great. Now seeing the correct OS for the host.

![image](https://user-images.githubusercontent.com/8593393/34085749-81e3143c-e35a-11e7-95c8-02739808f4c8.png)

It strikes me that we have all this other info in addition to $PRETTY_NAME, I think it would be cool if we added it in the hovertext over OS in the details panel, but I notice none of the other host details have hovertext that differs or adds more info.

Another thing that concerns me, I don't think /etc/os-release is a given on OSX, so will have to explore some other way to get more detailed OS info.

PR #2995, opened 2017-12-18
Linux pretty name Issue #2901

JordanJLopez:
Display Linux OS's PRETTY_NAME rather than Go's runtime.GOOS in the Host details panel.

### DIFF
--- a/probe/host/reporter.go
+++ b/probe/host/reporter.go
@@ -2,11 +2,9 @@ package host
 
 import (
 	"fmt"
-	"runtime"
 	"strconv"
 	"sync"
 	"time"
-
 	"github.com/weaveworks/common/mtime"
 	"github.com/weaveworks/scope/probe/controls"
 	"github.com/weaveworks/scope/report"
--- a/probe/host/reporter.go
+++ b/probe/host/reporter.go
@@ -124,10 +124,6 @@ func (r *Reporter) Report() (report.Report, error) {
 		return rep, err
 	}
 
-	if os == "unknown" {
-		os = runtime.GOOS
-	}
-
 	now := mtime.Now()
 	metrics := GetLoad(now)
 	cpuUsage, max := GetCPUUsagePercent()
--- a/probe/host/reporter.go
+++ b/probe/host/reporter.go
@@ -117,6 +117,17 @@ func (r *Reporter) Report() (report.Report, error) {
 	rep.Host = rep.Host.WithMetadataTemplates(MetadataTemplates)
 	rep.Host = rep.Host.WithMetricTemplates(MetricTemplates)
 
+
+	os, err := GetPrettyName()
+
+	if err != nil {
+		return rep, err
+	}
+
+	if os == "unknown" {
+		os = runtime.GOOS
+	}
+
 	now := mtime.Now()
 	metrics := GetLoad(now)
 	cpuUsage, max := GetCPUUsagePercent()
@@ -129,7 +140,7 @@ func (r *Reporter) Report() (report.Report, error) {
 			report.ControlProbeID: r.probeID,
 			Timestamp:             mtime.Now().UTC().Format(time.RFC3339Nano),
 			HostName:              r.hostName,
-			OS:                    runtime.GOOS,
+			OS:                    os,
 			KernelVersion:         kernel,
 			Uptime:                strconv.Itoa(int(uptime / time.Second)), // uptime in seconds
 			ScopeVersion:          r.version,
--- a/probe/host/system_darwin.go
+++ b/probe/host/system_darwin.go
@@ -30,6 +30,12 @@ var GetKernelReleaseAndVersion = func() (string, string, error) {
 	return string(release), string(version), nil
 }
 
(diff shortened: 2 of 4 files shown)


probe/host/reporter.go
```diff
@@ -117,6 +117,17 @@ func (r *Reporter) Report() (report.Report, error) {
 	rep.Host = rep.Host.WithMetadataTemplates(MetadataTemplates)
 	rep.Host = rep.Host.WithMetricTemplates(MetricTemplates)
 
+
+	os, err := GetPrettyName()
```
dlespiau: - This function could use a more explicit name, how about `GetOSName`?

- I would simplify the logic here: don't return a special `"unknown"` value but fallback to `runtime.GOOS` on any error:

```
os, err := GetOSName()
if err != nil {
        os = runtime.GOOS
}
```

probe/host/system_darwin.go
```diff
@@ -30,6 +30,12 @@ var GetKernelReleaseAndVersion = func() (string, string, error) {
 	return string(release), string(version), nil
 }
 
+// GetPrettyName extracts the PRETTY_NAME from /etc/os-release
+var GetPrettyName = func () (string, error) {
+
+	return "unknown", nil
```
dlespiau: The suggestion above implies returning en error here.

probe/host/system_linux.go
```diff
@@ -31,6 +33,67 @@ var GetKernelReleaseAndVersion = func() (string, string, error) {
...
 }
 
+// GetPrettyName extracts the PRETTY_NAME from host's /etc/os-release
+// SPEC: https://www.freedesktop.org/software/systemd/man/os-release.html
+var GetPrettyName = func () (string, error) {
+
+	var buffer bytes.Buffer
+
+	buf, err := ioutil.ReadFile("/var/run/scope/host-os-release")
+	if err != nil {
```
dlespiau: we already fallback to runtime.GOOS in the common code (common to all GetPrettyName() functions). No need to do it here as well.

probe/host/system_linux.go
```diff
@@ -31,6 +33,67 @@ var GetKernelReleaseAndVersion = func() (string, string, error) {
...
+		return runtime.GOOS, nil
+	}
+
+	// If can't find host-os-release, get container's os-release
```
dlespiau: Nope. We really don't want to surface the scope container base OS as the node OS :)

probe/host/system_linux.go
```diff
@@ -31,6 +33,67 @@ var GetKernelReleaseAndVersion = func() (string, string, error) {
...
+		} else {
+			buffer.WriteString(string(name[1]))
+			buffer.WriteString(" ")
+			buffer.WriteString(string(version[1]))
+		}
+	} else {
+		buffer.WriteString(string(prettyName[1]))
+	}
+
+	// Remove quotation marks from return value
```
dlespiau: The explanation of the *why* you need to do this would be most welcomed.

probe/host/system_linux.go
```diff
@@ -31,6 +33,67 @@ var GetKernelReleaseAndVersion = func() (string, string, error) {
...
+	IDParse := regexp.MustCompile("ID=(.+?)\n")
+	versionIDParse := regexp.MustCompile("VERSION_ID=(.+?)\n")
+
+	prettyName := prettyNameParse.FindStringSubmatch(string(buf))
+	name := nameParse.FindStringSubmatch(string(buf))
+	version := versionParse.FindStringSubmatch(string(buf))
+	prettyID := IDParse.FindStringSubmatch(string(buf))
+	versionID := versionIDParse.FindStringSubmatch(string(buf))
+
+	if prettyName == nil {
```
dlespiau: This `if/else` ladder looks a bit ugly:

- `runtime.GOOS` is returned instead of letting the caller deal with that fallback
- For the `if / else` ladder, I'd invert the conditions so we don't end up shifting the logic to the right

```
if (prettyName != nil) {
        ...
}  else if (.... ) {

} else if (...) {

}
```

probe/host/system_linux.go
```diff
@@ -31,6 +33,67 @@ var GetKernelReleaseAndVersion = func() (string, string, error) {
...
+	// 3. $ID $VERSION_ID
+	// 4. runtime.GOOS
+
+	prettyNameParse := regexp.MustCompile("PRETTY_NAME=(.+?)\n")
+	nameParse := regexp.MustCompile("NAME=(.+?)\n")
+	versionParse := regexp.MustCompile("VERSION=(.+?)\n")
+	IDParse := regexp.MustCompile("ID=(.+?)\n")
+	versionIDParse := regexp.MustCompile("VERSION_ID=(.+?)\n")
+
+	prettyName := prettyNameParse.FindStringSubmatch(string(buf))
```
dlespiau: Instead of using a lot of regex to parse this file (several times!) I'd cycle though the lines and use simple string functions to extract the interesting values.

2opremio:
@JordanJLopez any chance of implementing the suggested changes anytime soon?

PR #2999, opened 2017-12-24
refactor: inline summarisation of metadata, metrics, tables

rade:
This removes a bunch of duplication and scattering of little pieces of code.


rade:
I've merged this since it's a rather straightforward PR.

PR #3000, opened 2017-12-24
benchmark report summarization

rade:
This is the last phase before result marshalling in the /api/topology/{name} handler.

There's bunch of tidying up of the benchmark code too.


rade:
I've merged this since it's a rather straightforward PR and only affects tests.

PR #3003, opened 2017-12-25
refactor: banish TheInternet

rade:
TheInternet hasn't existed as a single node for a long time.

We also move & export the IsInternetNode predicate so it can be used in more places.


rade:
I've merged this since it's a pretty straightforward refactor.

PR #3005, opened 2017-12-27
refactor: move RenderContext where it belongs

rade:
It shouldn't be in 'report' because that is shared between the probe and the app, and RenderContext, as the name suggests, is only needed in the app.


bboreham:
Feels like `RenderContext` should be an interface so other parts don't need to know what it is being used for. 

But I don't see evidence that this causes a problem so perhaps yagni.

PR #3007, opened 2017-12-28
remove unnecessary metadata propagation

rade:
It turns out that nothing is actually looking at the metadata fields we were propagating in the Map/join functions. Instead, all the metadata being rendered comes from nodes we are joining with. Furthermore, summarisation does a reasonable job of rendering nodes knowing just their ID and topology.

This change makes the Map/join functions much more regular. With the exception of `PropagateSingleMetric` (which really shouldn't be a Map - see  #3008), all these functions now produce nodes with just an id & topology, and no metadata. This allows us to simply `joinResults.add*`. It also brings some performance benefits - a few % saved allocations in the rendering benchmarks. And it's a step on the way to unifying Map and join.

### DIFF
--- a/render/container.go
+++ b/render/container.go
@@ -268,9 +268,7 @@ func MapProcess2Container(n report.Node) report.Nodes {
 	}
 
 	// Otherwise, if the process is not in a container, group it into
-	// an per-host "Uncontained" node.  If for whatever reason this
-	// node doesn't have a host id in their node metadata, it'll all
-	// get grouped into a single uncontained node.
+	// a per-host "Uncontained" node.
 	var (
 		id   string
 		node report.Node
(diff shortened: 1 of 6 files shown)


bboreham:
I think the last commit message could be improved. 

> It turns out that our usage of joinResults.add* only ever ends
> creating minimal nodes, from just an id and topology

I would phrase it more like "after dropping extra metadata in the rest of this PR, our usage of ..."

render/container.go
```diff
@@ -93,9 +93,10 @@ func (c connectionJoin) Render(rpt report.Report) Nodes {
 		if !found {
 			id, found = ipNodes[report.MakeScopedEndpointNodeID(scope, addr, port)]
 		}
-		if found && id != "" { // not one we blanked out earlier
-			// We are guaranteed to find the id, so no need to pass a node constructor.
-			ret.addChild(m, id, nil)
+		if found && id != "" { // not one we blanked out earlier We
```
bboreham: Comment splice

render/container.go
```diff
@@ -268,9 +269,7 @@ func MapProcess2Container(n report.Node) report.Nodes {
 	}
 
 	// Otherwise, if the process is not in a container, group it into
-	// an per-host "Uncontained" node.  If for whatever reason this
-	// node doesn't have a host id in their node metadata, it'll all
-	// get grouped into a single uncontained node.
+	// an per-host "Uncontained" node.
```
bboreham: an->a

render/container.go
rade: fixed

render/container.go
rade: fixed

rade:
> I would phrase it more like "after dropping extra metadata in the rest of this PR, our usage of ..."

Fixed.

PR #3011, opened 2018-01-02
Simplifying backgrounds to match lightgray in service-ui and ui-compo…

bia:
…nents
Breaks unless service-ui is also updated to https://github.com/weaveworks/service-ui/pull/1672

The corresponding gray shift in Explore to addresses the issues in https://github.com/weaveworks/ui-components/pull/93
Trying to consolidate and simplify some colouring and settled on this after quite a few more more drastic attempts..
Is this enough contrast?

screen shot 2018-01-02 at 20 48 16

### DIFF
--- a/client/app/styles/_base.scss
+++ b/client/app/styles/_base.scss
@@ -96,7 +96,7 @@ a {
 
 .overlay-wrapper {
   align-items: center;
-  background-color: fade-out($background-average-color, 0.1);
+  background-color: fade-out($background-color, 0.1);
   border-radius: 4px;
   color: $text-tertiary-color;
   display: flex;
--- a/client/app/styles/_base.scss
+++ b/client/app/styles/_base.scss
@@ -2044,7 +2044,7 @@ a {
     }
 
     tr:nth-child(even) {
-      background: $background-color;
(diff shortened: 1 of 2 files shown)


rade:
The highlighting of the controls at the top is rather faint.

bia:
What if I shift up the darkness of both 'secondary-color's?
It kind of bothers me that all of these nuanced shades don't appear anywhere else in the app.
screen shot 2018-01-03 at 09 15 17

rade:
That looks fine.

bia:
Thanks! That's fixed now.

screen shot 2018-01-05 at 17 46 10

rade:
...which has just reminded me to ask... why is there such a huge gap between the top controls and the table?

Vertical space is precious here; the table should use as much as it can.

Separate issue, clearly.

rade:
> why is there such a huge gap between the top controls and the table?

Ah, I think it's there to accommodate the expansion of the top when timetravelling.

bia:
@rade, but it shouldn't be that way when in Live mode?

rade:
> but it shouldn't be that way when in Live mode?

Ideally not, no. Have filed #3015.

Issue #2234, opened 2017-02-11
show pod logs for multi-container pods


rade:
Right now a user can only view the logs of containers, not of whole pods. We have a hack in place so that when a pod only has one non-pause container then in the details panel of the pod, we let the user view the logs of that container. This doesn't work for the likes of Weave Net pods, which comprise several containers.

Essentially we want the equivalent of `kubectl logs -f <pod>`.

rade:
> Right now a user can only view the logs of containers, not of whole pods. We have a hack in place so that when a pod only has one non-pause container then in the details panel of the pod, we let the user view the logs of that container.

I completely misinterpreted what I was seeing...

The 'Get logs' button on the pod detail panel *does* fetch the pod logs using the k8s API. It even shows history.

HOWEVER, it appears that said k8s API requires a container to be specified when the pod contains more than one (plus plause). Otherwise a "a container name must be specified for pod blah, choose one of: [blah, blah]" error is returned.

ekimekim:
What are you suggesting design-wise? Interleaving the logs in the output window? Should we add a prefix saying where they're from? eg.
```
[container_a] foo
[container_a] bar
[container_b] foo
[container_a] baz
[container_b] bar
```

rade:
see also #1493.

rade:
@awh has just found https://github.com/wercker/stern - perhaps we could embed that?

rade:
This got resolved by #3013

PR #3013, opened 2018-01-03
Reading pod logs returns all container logs

rbruggem:
This is achieved by issuing an http request for each container to kubernetes' API, which yields one Reader for the corresponding container.
`logReadCloser' then reads from the above readers in parallel as data is available, buffering when necessary, forwarding it to clients by implementing the io.ReadCloser interface.

Part of #2234.

### DIFF
--- a/probe/kubernetes/logreadcloser.go
+++ b/probe/kubernetes/logreadcloser.go
@@ -6,30 +6,30 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"sync"
 )
 
 type logReadCloser struct {
-	labels       []string
-	labelLength  int
-	readClosers  []io.ReadCloser
-	eof          []bool
-	buffer       bytes.Buffer
-	dataChannel  chan []byte
-	stopChannels []chan struct{}
-	eofChannel   chan int
+	labels      []string
+	labelLength int
+	readClosers []io.ReadCloser
+	eof         []bool
+	buffer      bytes.Buffer
+	dataChannel chan []byte
+	eofChannel  chan int
+	wg          sync.WaitGroup
 }
 
 // NewLogReadCloser reads from multiple io.ReadCloser, where data is available,
 // and annotates each line with the reader's label
 func NewLogReadCloser(readClosersWithLabel map[io.ReadCloser]string) io.ReadCloser {
-	stopChannels := make([]chan struct{}, len(readClosersWithLabel))
-	labels := make([]string, len(readClosersWithLabel))
-	readClosers := make([]io.ReadCloser, len(readClosersWithLabel))
+	n := len(readClosersWithLabel)
+	labels := make([]string, n)
+	readClosers := make([]io.ReadCloser, n)
 
 	i := 0
 	labelLength := 0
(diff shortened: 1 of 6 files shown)


probe/kubernetes/logreadcloser.go
```diff
@@ -0,0 +1,148 @@
...
+		n, err := l.readClosers[idx].Read(tmpBuffer)
+		if err == io.EOF {
+			if n > 0 {
+				l.dataChannel <- tmpBuffer[:n]
+			}
+			l.eofChannel <- idx
+			break
+		}
+		if err != nil {
+			log.Errorf("Failed to read: %v", err)
```
rade: I'd probably ditch this. It's too low-level to be of much use.

rade:
How does this do the interleaving? https://github.com/wercker/stern interleaves *by line", which IMO is quite sensible. It also prefixes every line with the namespace, pod and container, which is rather nice (though in our case just prefixing with the container name would suffice).

probe/kubernetes/logreadcloser.go
```diff
@@ -0,0 +1,166 @@
...
+		if err != nil {
+			return err
+		}
+
+		// synchronous stop:
+		// the routines write to dataChannel which will be closed by this thread
+		select {
+		case <-l.stopChannels[i]:
+			break
+		}
```
rndstr: this select/case can be replaced by just `<-l.stopChannels[i]`

probe/kubernetes/logreadcloser.go
```diff
@@ -0,0 +1,166 @@
...
+	buffer       bytes.Buffer
+	dataChannel  chan []byte
+	stopChannels []chan struct{}
+	eofChannel   chan int
+}
+
+// NewLogReadCloser reads from multiple io.ReadCloser, where data is available,
+// and annotates each line with the reader's label
+func NewLogReadCloser(readClosersWithLabel map[io.ReadCloser]string) io.ReadCloser {
+	stopChannels := make([]chan struct{}, len(readClosersWithLabel))
```
rndstr: could also use a `sync.WaitGroup` for this (has the advantage of shutting down the readClosers in parallel)

probe/kubernetes/logreadcloser.go
rbruggem: so true!

probe/kubernetes/logreadcloser.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (l *logReadCloser) Close() error {
+	for i, rc := range l.readClosers {
+		err := rc.Close()
+		if err != nil {
+			return err
+		}
+
+		// synchronous stop:
+		// the routines write to dataChannel which will be closed by this thread
```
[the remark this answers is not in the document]
rbruggem: Yes, your comment is more appropriate. thanks

probe/kubernetes/logreadcloser.go
rbruggem: Thanks, as an extram it seems like I also won't need the `stopChannels` anymore.

PR #3014, opened 2018-01-03
Stop fetching ReplicaSets and ReplicationControllers

### LINKED ISSUE #3010: probes are fetching ReplicaSets for no reason
Since #2957 the probes no longer report on ReplicaSets. However, they still maintain a replicaSetStore that is kept in sync with k8s. There is no reason to do that.

rbruggem:
They are not reported back to the scope app.

Fixes #3010


probe/kubernetes/client.go
```diff
@@ -157,18 +153,16 @@ func NewClient(config ClientConfig) (Client, error) {
...
-	result.replicationControllerStore = result.setupStore(c.CoreV1Client.RESTClient(), "replicationcontrollers", &apiv1.ReplicationController{}, nil)
 	result.nodeStore = result.setupStore(c.CoreV1Client.RESTClient(), "nodes", &apiv1.Node{}, nil)
 	result.namespaceStore = result.setupStore(c.CoreV1Client.RESTClient(), "namespaces", &apiv1.Namespace{}, nil)
 
 	// We list deployments here to check if this version of kubernetes is >= 1.2.
 	// We would use NegotiateVersion, but Kubernetes 1.1 "supports"
-	// extensions/v1beta1, but not deployments, replicasets or daemonsets.
+	// extensions/v1beta1, but not deployments or daemonsets.
 	if _, err := c.Extensions().Deployments(metav1.NamespaceAll).List(metav1.ListOptions{}); err != nil {
 		log.Infof("Deployments, ReplicaSets and DaemonSets are not supported by this Kubernetes version: %v", err)
```
rade: The log message needs updating.

probe/kubernetes/controls.go
```diff
@@ -83,7 +83,6 @@ func (r *Reporter) CaptureResource(f func(xfer.Request, string, string, string)
 			f   func(string) (string, bool)
 		}{
 			{report.Deployment, report.ParseDeploymentNodeID},
-			{report.ReplicaSet, report.ParseReplicaSetNodeID},
```
rade: Now that there is just one item, we could radically simplify this entire function. Worth doing, _unless_ we believe we will be adding a new item here soon.

probe/kubernetes/controls.go
```diff
@@ -78,33 +78,21 @@ func (r *Reporter) CapturePod(f func(xfer.Request, string, string) xfer.Response
 func (r *Reporter) CaptureResource(f func(xfer.Request, string, string, string) xfer.Response) func(xfer.Request) xfer.Response {
```
rade: This function should be called `CaptureDeployment` now, by analogy with `CapturePod`.

probe/kubernetes/controls.go
```diff
@@ -78,33 +78,21 @@ func (r *Reporter) CapturePod(f func(xfer.Request, string, string) xfer.Response
...
-			f   func(string) (string, bool)
-		}{
-			{report.Deployment, report.ParseDeploymentNodeID},
-		} {
-			if u, ok := parser.f(req.NodeID); ok {
-				resource, uid = parser.res, u
-				break
-			}
+		if u, ok := report.ParseDeploymentNodeID(req.NodeID); ok {
+			resource, uid = report.Deployment, u
```
rade: `resource` isn't really used anywhere - the one is in an error message, where it can simply be replaced with "Deployment".

probe/kubernetes/controls.go
```diff
@@ -78,33 +78,21 @@ func (r *Reporter) CapturePod(f func(xfer.Request, string, string) xfer.Response
...
-				return f(req, "deployment", deployment.Namespace(), deployment.Name())
+		var deployment Deployment
+		r.client.WalkDeployments(func(d Deployment) error {
+			if d.UID() == uid {
+				deployment = d
 			}
+			return nil
+		})
+		if deployment != nil {
+			return f(req, "deployment", deployment.Namespace(), deployment.Name())
```
rade: I suggest making the structure the same here as in `CapturePod`, which a) has the condition reversed, and b) has `f` take just three arguments. The latter will require removing `resource` from the signature of `Reporter.Scale{Up,Down}`, and instead calling the client library with "deployment" as a constant. Incidentally, this will actually make `Reporter.Scale{Up,Down}` do what their doc comment says: "control to scale up a _deployment_".

rade: The change I proposed to `Reporter.Scale{Up,Down}` is still outstanding.

PR #3016, opened 2018-01-07
refactor: extract common code in endpoint mapping

rade:
This isn't quite as neat as we'd want to make it, because two of the three call sites still require a closure, but it's still an improvement.

Note that every instance of MapEndpoints only ever maps to one topology, which, furthermore, is a core report topology. Hence we can just parameterise MapEndpoints with that topology, and then the map functions can return just the node ids.

### DIFF
--- a/render/container.go
+++ b/render/container.go
@@ -82,12 +82,16 @@ func (c connectionJoin) Render(rpt report.Report) Nodes {
 			if !found {
 				id, found = ipNodes[report.MakeScopedEndpointNodeID(scope, addr, port)]
 			}
-			if found && id != "" { // not one we blanked out earlier
-				// We are guaranteed to find the id, so really this
-				// should never end up creating a node.
-				return id
+			if !found || id == "" {
+				return ""
 			}
-			return ""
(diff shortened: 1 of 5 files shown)


render/container.go
```diff
@@ -71,35 +68,28 @@ func (c connectionJoin) Render(rpt report.Report) Nodes {
...
+			if !found {
+				id, found = ipNodes[report.MakeScopedEndpointNodeID(scope, addr, port)]
+			}
+			if !found || id == "" {
+				return ""
+			}
+			// Not an IP we blanked out earlier. We are guaranteed to
+			// find the id, so really this should never end up
+			// creating a node.
+			return id
```
bboreham: It seems this comment has got a bit separated from the code that makes it make sense.
It would be `joinResult.addChild()` that created a node, right?
And the 'find' part is also under there.  How can we know here what will be in `joinResult.nodes` ?

rade: We know because the id is the result of a lookup in `ipNodes`, and all the ids in `ipNodes` are ids of nodes in `c.topology`.

Do you reckon I should say that in the comment?

PR #3030, opened 2018-01-18
Remove default values from URL state hash

### LINKED ISSUE #1727: Tighten up Scope url state.

fbarl:
Resolves #1727.

##### Changes

_major_
* Use only non-default values for `topologyOptions` in Scope state. These are calculated via new helper `hashDifferenceDeep` which basically omits all the deep non-hash values which correspond literally to default options.
* Filter out all (other) URL state fields which correspond to their initial Redux state.

_minor_
* Removed `label` field from `nodeDetails` Scope state hash array, as it wasn't used anywhere.
* Changed `searchQuery` Redux state default to `''` for better matching with the URL state.

Note: I'm aware that `topologyId` field could also have been omitted in the URL in case it has a default value, but that default topology is somewhat arbitrary and I didn't want to make a special case of it.

### DIFF
--- a/client/app/scripts/reducers/root.js
+++ b/client/app/scripts/reducers/root.js
@@ -71,7 +71,7 @@ export const initialState = makeMap({
   pinnedSearches: makeList(), // list of node filters
   routeSet: false,
   searchFocused: false,
-  searchQuery: null,
+  searchQuery: '',
   selectedNetwork: null,
   selectedNodeId: null,
   showingHelp: false,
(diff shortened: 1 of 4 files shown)


client/app/scripts/reducers/root.js
```diff
@@ -661,7 +658,7 @@ export function rootReducer(state = initialState, action) {
       state = setTopology(state, state.get('currentTopologyId'));
       // only set on first load, if options are not already set via route
       if (!state.get('topologiesLoaded') && state.get('topologyOptions').size === 0) {
-        state = setDefaultTopologyOptions(state, state.get('topologies'));
+        state = state.set('topologyOptions', getDefaultTopologyOptions(state));
```
foot: 👍

fbarl:
> if feasible would be cool to include topologyId too, which "defaults" to kube-controllers as its rank is 0.

This is not an ideal argument against it, but that change would make it possible to have an empty state hash (if we're on all defaults), which would be tricky to distinguish from no state at all which would default to localStorage. Something to consider in a future PR perhaps, but I think we can make a small exception with `topologyId` field for now and always display it.

PR #3036, opened 2018-01-19
Change pausedAt format from moment() back to ISO string

### LINKED ISSUE #3029: Change pausedAt state format back to ISO string
That's the format that `TimeTravel` component accepts and now that all the complex time logic happens there, it would be much easier (and more consistent) to treat the timestamp in Scope as a simple string.

fbarl:
Resolves #3029.

Everything should be working as before :)

### DIFF
--- a/client/app/scripts/components/node-details/node-details-health-link-item.js
+++ b/client/app/scripts/components/node-details/node-details-health-link-item.js
@@ -10,7 +10,7 @@ import { trackAnalyticsEvent } from '../../utils/tracking-utils';
 
 /**
  * @param {string} url
- * @param {Moment} time
+ * @param {string} time
  * @returns {string}
  */
 export function appendTime(url, time) {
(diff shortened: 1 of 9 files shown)


client/app/scripts/components/node-details/node-details-health-link-item.js
```diff
@@ -23,7 +24,7 @@ export function appendTime(url, time) {
     const json = decodeURIComponent(url.substr(pos + cloudLinkPathEnd.length));
     try {
       payload = JSON.parse(json);
-      payload.time = { queryEnd: time.unix() };
+      payload.time = { queryEnd: moment(time).unix() };
```
rndstr: jsdoc for `time` type needs correction

rndstr: tests need adjustment for `time` parameter

fbarl: Where is that? :) It's the same format being seint through (unix seconds), it's just needs to be wrapped in `moment()` here, because `time` in now a string, not a moment object internally anymore.

rndstr: the jsdoc for this function still says Moment https://github.com/weaveworks/scope/blob/77b691ca0b67d91812934442077d57219cd750f0/client/app/scripts/components/node-details/node-details-health-link-item.js#L13

fbarl: Good call, by fixing the tests, I realized `appendTime` function wasn't handling all the cases right so I fixed it as well :)

PR #3037, opened 2018-01-19
Check if k8s resources are supported in `runReflectorUntil`

### LINKED ISSUE #3027: failure to ever report on some k8s objects when apiserver is initially unreachable
In one of our systems we noticed that a bunch of pods were missing parent links to their deployments/daemonsets. And it turned out that this was the case for _all_ pods on a subset of nodes.

Looking at the probe log of one of these nodes reveals
```
2018-01-17T20:40:12.068781921Z <probe> INFO: 2018/01/17 20:40:12.068629 kubernetes: targeting api server https://10.0.0.1:443
2018-01-17T20:40:12.074645411Z <probe> INFO: 2018/01/17 20:40:12.074475 Deployments, ReplicaSets and DaemonSets are not supported by this Kubernetes version: Get https://10.0.0.1:443/apis/extensions/v1beta1/deployments: dial tcp 10.0.0.1:443: getsockopt: connection refused
2018-01-17T20:40:12.075199518Z <probe> INFO: 2018/01/17 20:40:12.075114 CronJobs are not supported by this Kubernetes version: Get https://10.0.0.1:443/apis/batch/v2alpha1/cronjobs: dial tcp 10.0.0.1:443: getsockopt: connection refused
2018-01-17T20:40:12.075696971Z <probe> INFO: 2018/01/17 20:40:12.075548 StatefulSets are not supported by this Kubernetesversion: Get https://10.0.0.1:443/apis/apps/v1beta1/statefulsets: dial tcp 10.0.0.1:443: getsockopt: connection refused
```
This node had just rebooted, and apparently the probe was unable to reach the k8s api server initially, even though it could connect successfully later on. Presumably that's just a startup sequencing/timing issue.

The problem here is that we make a final decision on whether certain controller types are supported based on a single connect attempt with a simple success/fail.

rbruggem:
`isResourceSupported` checks whether a kubernetes resource is supported by the api server.
This ensures that, if the probe is unable to communicate with the api server, the call is retried until a true/false response.

If `isResourceSupported` returns false, `ListAndWatch` is not called and `runReflectorUntil` just exits.

Fixes #3027.

Tested with kubernetes v1.3.0 and v1.8.0.

### DIFF
--- a/probe/kubernetes/client.go
+++ b/probe/kubernetes/client.go
@@ -9,8 +9,10 @@ import (
 	"github.com/weaveworks/common/backoff"
 
 	log "github.com/Sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	apiv1 "k8s.io/client-go/pkg/api/v1"
@@ -59,23 +61,6 @@ type client struct {
 
 	podWatchesMutex sync.Mutex
 	podWatches      []func(Event, Pod)
-}
-
-// runReflectorUntil runs cache.Reflector#ListAndWatch in an endless loop.
-// Errors are logged and retried with exponential backoff.
-func runReflectorUntil(r *cache.Reflector, stopCh <-chan struct{}, msg string) {
-	listAndWatch := func() (bool, error) {
-		select {
-		case <-stopCh:
-			return true, nil
-		default:
-			err := r.ListAndWatch(stopCh)
-			return false, err
-		}
-	}
-	bo := backoff.New(listAndWatch, fmt.Sprintf("Kubernetes reflector (%s)", msg))
-	bo.SetMaxBackoff(5 * time.Minute)
-	go bo.Start()
 }
 
 // ClientConfig establishes the configuration for the kubernetes client
@@ -134,7 +119,6 @@ func NewClient(config ClientConfig) (Client, error) {
 		if err != nil {
 			return nil, err
 		}
-
 	}
 	log.Infof("kubernetes: targeting api server %s", restConfig.Host)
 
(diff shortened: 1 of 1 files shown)


probe/kubernetes/client.go
```diff
@@ -151,34 +135,50 @@ func NewClient(config ClientConfig) (Client, error) {
...
+	case "cronjobs":
+		group = "batch/v2alpha1"
+	case "statefulsets":
+		group = "apps/v1beta1"
+	case "pods", "services", "nodes", "namespaces":
+		group = "v1"
+	case "jobs":
+		group = "batch/v1"
+	default:
+		return false, fmt.Errorf("Support check not available for resource %v", resource)
```
rade: We shouldn't need this mapping. The `RESTClient()` passed to `setupStore()` is a `rest.Interface`, which has an `APIVersion()` method that returns a `schema.GroupVersion`. So changing the type on `setupStore()` from `cache.Getter` to `rest.Interface` should give us access to the info we need.

probe/kubernetes/client.go
```diff
@@ -187,10 +187,35 @@ func (c *client) setupStore(kclient cache.Getter, resource string, itemType inte
 	if store == nil {
 		store = cache.NewStore(cache.MetaNamespaceKeyFunc)
 	}
-	runReflectorUntil(cache.NewReflector(lw, itemType, store, c.resyncPeriod), c.quit, resource)
+	c.runReflectorUntil(cache.NewReflector(lw, itemType, store, c.resyncPeriod), resource)
 	return store
 }
 
+// runReflectorUntil runs cache.Reflector#ListAndWatch in an endless loop.
+// Errors are logged and retried with exponential backoff.
```
rade: comment needs updating.

probe/kubernetes/client.go
```diff
@@ -24,6 +24,8 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
+const errResourceNotFound = "the server could not find the requested resource"
```
rade: Can we do better? I suspect the error that is coming back in this case is a [`StatusError`](https://godoc.org/k8s.io/apimachinery/pkg/api/errors#StatusError), implementing [`APIStatus`](https://godoc.org/k8s.io/apimachinery/pkg/api/errors#APIStatus), which ought to return [`StatusReasonNotFound`](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#StatusReason).

PR #3038, opened 2018-01-19
client: Fix uptime sort in table view

### LINKED ISSUE #3026: 'uptime' sorting broken
screenshot from 2018-01-17 20 33 23

rndstr:
Duration dataType for table columns is now sorted as numeric value.
Default sorting is kept as ascending and text is left-aligned, unlike
numbers.

Fixes #3026

### DIFF
--- a/client/app/scripts/components/node-details/__tests__/node-details-table-test.js
+++ b/client/app/scripts/components/node-details/__tests__/node-details-table-test.js
@@ -15,6 +15,7 @@ describe('NodeDetailsTable', () => {
     columns = [
       { id: 'kubernetes_ip', label: 'IP', dataType: 'ip' },
       { id: 'kubernetes_namespace', label: 'Namespace' },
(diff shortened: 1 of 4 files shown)


client/app/scripts/components/node-details/__tests__/node-details-table-test.js
```diff
@@ -125,4 +138,23 @@ describe('NodeDetailsTable', () => {
...
+            sortedBy="uptime"
+            nodeIdKey="id"
+            nodes={nodes}
+          />
+        </Provider>
+      ));
+
+      matchColumnValues('uptime', ['1 second', '2 seconds', '4 seconds', '30 seconds']);
+      clickColumn('Uptime');
+      matchColumnValues('uptime', ['30 seconds', '4 seconds', '2 seconds', '1 second']);
```
fbarl: Would be nice to test against longer durations as well, so that more units are involved than just _seconds_ :)

PR #3044, opened 2018-01-30
Fetch cronjobs from 'batch/v1beta1'

### LINKED ISSUE #3032: cronjobs not detected on k8s 1.8
I've noticed the following the probe logs in a GKE 1.8.6 cluster:
```
<probe> INFO: 2018/01/18 18:20:05.103074 CronJobs are not supported by this Kubernetes version: the server could not find the requested resource
```
I suspect the reason is the same as kubernetes/kube-state-metrics#295 - scope is looking for cronjobs under `batch/v2alpha1` when they actually are under `batch/v1beta1`.

According to the [API docs](https://v1-8.docs.kubernetes.io/docs/api-reference/v1.8/#cronjob-v1beta1-batch) the latter is indeed the correct version for k8s 1.8.

rbruggem:
This applies if kubernetes' version is >= 1.8.
Otherwise fetch cronjobs from 'batch/v2alpha1'.

Also, upgrade k8s.io/client-go to kubernetes-1.9.1.

Tested with kubernetes 1.7 and 1.9.

Fixes #3032.


probe/kubernetes/client.go
```diff
@@ -175,6 +180,20 @@ func (c *client) setupStore(kclient rest.Interface, resource string, itemType in
 	return store
 }
 
+func (c *client) setupCronjobsStore() (cache.Store, error) {
```
rade: should be `setupCronJobStore`

probe/kubernetes/cronjob.go
```diff
@@ -74,3 +86,29 @@ func (cj *cronJob) GetNode() report.Node {
 	}
 	return cj.MetaNode(report.MakeCronJobNodeID(cj.UID())).WithLatests(latest)
 }
+
+func upgrade(legacy *batchv2alpha1.CronJob) *batchv1beta1.CronJob {
```
rade: That's too generic a name for a function that lives in the `kubernetes` package. `upgradeCronJob`?

PR #3045, opened 2018-01-31
Close terminal pipe, when closing the pod panel

### LINKED ISSUE #3035: Closing Log terminal from the Pod panel does not close all resources in the probe.
Closing the pod log terminal from the pod panel instead of the terminal panel does not Close the Pipe, which also does not Close the underlying log reader.

rbruggem:
Moving `deletePipe()` in `componentWillUnmount()` ensures the pipe is deleted when:
1. the terminal is closed
2. the terminal is closed when closing the pod panel

The pipe still does not get deleted when popping out the terminal. Created https://github.com/weaveworks/scope/issues/3047.

Fixes #3035.

### DIFF
--- a/client/app/scripts/actions/app-actions.js
+++ b/client/app/scripts/actions/app-actions.js
@@ -234,15 +234,12 @@ export function clickCloseDetails(nodeId) {
   };
 }
 
-export function clickCloseTerminal(pipeId, closePipe) {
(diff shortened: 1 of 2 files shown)


rade:
> when the popped out terminal is closed the pipe is not deleted

Is that a leak?

rbruggem:
> Is that a leak?

yes

PR #3050, opened 2018-02-02
Create reflectors asynchronously

### LINKED ISSUE #3049: no k8s objects reported when apiserver is initially unreachable
#3044 introduced a more lethal version of #3027: Now when the apiserver isn't available at the start then _no_ k8s objects are reported.

rbruggem:
Reflectors are created and run within the same function, asynchronously from the main thread.
Creating reflectors may require calls to the kubernetes api, which can return errors.
API errors are not handled in the main thread, but are handled asynchronously by retries.

Fixes #3049.

### DIFF
--- a/probe/kubernetes/client.go
+++ b/probe/kubernetes/client.go
@@ -134,20 +134,17 @@ func NewClient(config ClientConfig) (Client, error) {
 		client:       c,
 	}
 
-	result.cronJobStore, err = result.setupCronjobStore()
-	if err != nil {
-		return nil, err
-	}
+	result.podStore = NewEventStore(result.triggerPodWatches, cache.MetaNamespaceKeyFunc)
+	result.runReflectorUntil("pods", result.podStore)
 
-	podStore := NewEventStore(result.triggerPodWatches, cache.MetaNamespaceKeyFunc)
-	result.podStore = result.setupStore(c.CoreV1().RESTClient(), "pods", &apiv1.Pod{}, podStore)
-	result.serviceStore = result.setupStore(c.CoreV1().RESTClient(), "services", &apiv1.Service{}, nil)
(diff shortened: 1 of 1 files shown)


probe/kubernetes/client.go
```diff
@@ -170,51 +166,81 @@ func (c *client) isResourceSupported(groupVersion schema.GroupVersion, resource
...
-		case <-c.quit:
-			return true, nil
-		default:
-			ok, err := c.isResourceSupported(groupVersion, resource)
-			if err != nil {
-				return false, err
+func (c *client) runReflectorUntil(resource string, store cache.Store) {
+	listAndWatch := func() func() (bool, error) {
+		var r *cache.Reflector
+		return func() (bool, error) {
```
rade: Why do we need that first closure above, given that we immediately evaluate it below? Can't we simply define `listAndWatch` as that 2nd function?

rbruggem: it's not necessarily needed, but the closure limits the scope or `r`

rade: AFAICT there is no semantic difference; an optimiser would be entitled to inline the `listAndWatch()` invocation, thus eliminating the outer closure.

probe/kubernetes/client.go
```diff
@@ -170,47 +166,75 @@ func (c *client) isResourceSupported(groupVersion schema.GroupVersion, resource
 	return false, nil
 }
 
-func (c *client) setupStore(kclient rest.Interface, resource string, itemType interface{}, nonDefaultStore cache.Store) cache.Store {
-	lw := cache.NewListWatchFromClient(kclient, resource, metav1.NamespaceAll, fields.Everything())
+func (c *client) setupStore(resource string, nonDefaultStore cache.Store) cache.Store {
```
rade: Now that this function does so little, we can ditch the last arg and change the one non-nil call site to directly call `c.runReflectorUntil` instead.

PR #3056, opened 2018-02-05
Save codegen output as CI artifact

bboreham:
Looks like the order of types in the generated output is unpredictable, so we need the actual file Scope was built with for debugging or profiling.

Also drive-by optimisation to stop saving the tarfile - we started in 441b71884b1d434221b8866e6719191eb975eff2, which offers no explanation as to why.


rade:
@bboreham what's happening with this?

bboreham:
Waiting for review?

I've restarted the CI build which failed for a reason that looks unrelated

PR #3057, opened 2018-02-05
Delete underlying pipe when popup terminal closes

### LINKED ISSUE #3047: When popping out the log terminal, the underlying Scope pipe will never get deleted

rbruggem:
Fixes #3047

### DIFF
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -56,7 +56,7 @@ function terminalCellSize(wrapperNode) {
 }
 
 
-function openNewWindow(url, bcr, minWidth = 200) {
+function openNewWindow(url, bcr, pipeId, minWidth = 200) {
   const screenLeft = window.screenX || window.screenLeft;
   const screenTop = window.screenY || window.screenTop;
   const popoutWindowToolbarHeight = 51;
@@ -164,6 +164,7 @@ class Terminal extends React.Component {
     if (this.props.connect) {
       this.mountTerminal();
     }
+    window.addEventListener('beforeunload', this.handleBeforeUnload);
   }
 
   mountTerminal() {
@@ -206,6 +207,7 @@ class Terminal extends React.Component {
     clearTimeout(this.resizeTimeout);
 
     window.removeEventListener('resize', this.handleResizeDebounced);
+    window.removeEventListener('beforeunload', this.handleBeforeUnload);
 
     if (this.term) {
       log('destroy terminal');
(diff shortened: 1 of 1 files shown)


foot:
- Looks good!
- I haven't tested it yet.
- We could do this more generally so that if you close scope it also cleans up pipes.
  - Terminal.cmpDidMount -> addEL('beforeunload', ...
  - Terminal.cmpWillUnmount -> removeEL('beforeunload', ...

rbruggem:
Doing Terminal.cmpDidMount -> addEL('beforeunload', ... is difficult since the window does not exist yet.

foot:
Ah, but once the window opens it will run cmpDidMount inside the window so we can hook it up to the global window object there:

```js

handleBeforeUnload = () => {
   deletePipe(this.getPipeId());
}

componentDidMount() {
    window.addEventListener('beforeunload', this.handleBeforeUnload);
}
```


But there are more cases here actually! Ooft.

1. No popout:
   - Open a terminal in weave cloud. 
   - cmpDidMount -> addEL('beforeunload'..). 
   - Close weave cloud window. 
   - Pipe deleted by beforeunload event.
2. Open popout and then close popout:
    - Open a terminal in weave cloud. 
    - cmpDidMount -> addEL('beforeunload'..)
    - popout terminal.
    - cmpWillUnmount is run inside of the Weave cloud window, which has a removeEL.
    - cmpDidMount -> addEL('beforeunload'..) runs inside new terminal window
    - close terminal window
    - Pipe deleted by beforeunload event inside terminal window
3. Open popout and then close weave cloud window:
    - same as previous scenario actually... when the terminal cmp is removed in weave cloud it should also remove the beforeunload handler so it won't delete the pipe.

I think that covers it? ...

foot:
Hey! Sorry I should have actually tested this earlier. I just did now and the DELETE is not being send onbeforeunload :(. AJAX + beforeunload doesn't seem to be supported by browsers https://stackoverflow.com/a/20322988

It seems like it has be done w/ a synchronous request. which neither `reqwest` or `window.fetch` support out of the box so that would mean writing our own xmlhttprequest sync helper.

You can test the request is being sent by clicking "preserve log" in the network panel of the dev tools, then typing in "google.com" `<enter>` into the url field. Filter by "pipe" to make this a bit more obvious.

I'm not sure how important it is to be so stringent w/ cleanup so your call if you wanna have a go at this!

rbruggem:
Closing this PR.
The scope app does close all underlying resources independently from the client.
The committed change does not achieve the desired cleanup.

Issue #3015, opened 2018-01-05
large gap between header and table


rade:
There's a rather large gap between the header and the table
![screenshot from 2018-01-05 17 13 41](https://user-images.githubusercontent.com/109109/34619984-e3d0bdee-f23b-11e7-8840-8199ad72843f.png)
which is unfortunate since tables really can do with all the vertical space they can get.

I believe the gap is there to accommodate timetravel:
![screenshot from 2018-01-05 17 13 54](https://user-images.githubusercontent.com/109109/34620036-0ae16000-f23c-11e7-80bd-9c0e9f2206a5.png)
which is fair enough, but it would be nice for the space not to go to waste when not timetravelling.

PR #3059, opened 2018-02-06
Time travel for local Scope

harshvkarn:
#3015
fixed the space between the header and table.

### DIFF
--- a/client/app/scripts/components/time-travel-wrapper.js
+++ b/client/app/scripts/components/time-travel-wrapper.js
@@ -12,7 +12,8 @@ const TimeTravelContainer = styled.div`
   transition: all .15s ease-in-out;
   position: relative;
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


fbarl:
Thanks for your PR @harshvkarn - showing Time Travel timeline in standalone Scope at all times is indeed what we've decided to do, but there was some more cleaning to be done, as well as making sure nothing breaks for our Scope in Weave Cloud use case, so I've put up PR #3065 that addresses all that.

fbarl:
> showing Time Travel timeline in standalone Scope at all times is indeed what we've decided to do

I need to correct myself here - in most cases Scope standalone won't even support historic reports, so Time Travel should be completely absent. What we've decided was to show Time Travel at all times _when available_ and one example of that is the Explore section in Weave Cloud.

PR #3060, opened 2018-02-06
Replace service cpu/mem link to 'Monitor'

### LINKED ISSUE #2925: Links to Monitoring make use of recording rules
We are using recording rules when deep-linking CPU/Memory graphs to Monitor, but we don't do any recording rule for our users at the moment.

To reproduce:

- Select Pods -> Service
- Click on the CPU graph
- This creates a new notebook with the query `namespace_label_name:container_cpu_usage_seconds_total:sum_rate{label_name="shipping",namespace="sock-shop"}`, which uses a recording rule
- No datapoints!

rbruggem:
The previous link to 'Monitor' contained a recording rule metric, and since not all Weave Clound instances have recording rules, the link would redirect to a 'Monitor' page without data points.
The recording rule metric has been substituted with the underlying query.

Fixes #2925

### DIFF
--- a/render/detailed/links.go
+++ b/render/detailed/links.go
@@ -71,8 +71,8 @@ var (
 			// NB: Pods need to be labeled and selected by their respective Service name, meaning:
 			// - The Service's `spec.selector` needs to select on `name`
 			// - The Service's `metadata.name` needs to be the same value as `spec.selector.name`
-			docker.CPUTotalUsage: `namespace_label_name:container_cpu_usage_seconds_total:sum_rate{label_name="{{label}}",namespace="{{namespace}}"}`,
-			docker.MemoryUsage:   `namespace_label_name:container_memory_usage_bytes:sum{label_name="{{label}}",namespace="{{namespace}}"}`,
+			docker.CPUTotalUsage: `sum(rate(container_cpu_usage_seconds_total{image!="",namespace="{{namespace}}",_weave_pod_name="{{label}}",job="cadvisor",container_name!="POD"}[5m]))`,
+			docker.MemoryUsage:   `sum(rate(container_memory_usage_bytes{image!="",namespace="{{namespace}}",_weave_pod_name="{{label}}",job="cadvisor",container_name!="POD"}[5m]))`,
 		},
 	}
 )
--- a/render/detailed/links.go
+++ b/render/detailed/links.go
@@ -67,10 +67,6 @@ var (
 		report.StatefulSet: podIDHashQueries,
 		report.CronJob:     podIDHashQueries,
 		report.Service: {
-			// These recording rules must be defined in the prometheus config.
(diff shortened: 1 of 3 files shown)


render/detailed/links.go
```diff
@@ -71,8 +71,8 @@ var (
 			// NB: Pods need to be labeled and selected by their respective Service name, meaning:
 			// - The Service's `spec.selector` needs to select on `name`
 			// - The Service's `metadata.name` needs to be the same value as `spec.selector.name`
-			docker.CPUTotalUsage: `namespace_label_name:container_cpu_usage_seconds_total:sum_rate{label_name="{{label}}",namespace="{{namespace}}"}`,
-			docker.MemoryUsage:   `namespace_label_name:container_memory_usage_bytes:sum{label_name="{{label}}",namespace="{{namespace}}"}`,
+			docker.CPUTotalUsage: `sum(label_replace(sum(rate(container_cpu_usage_seconds_total{image!="",namespace="{{namespace}}"}[5m])) by (pod_name, namespace), "pod", "$1", "pod_name", "(.*)") * on (pod) group_right(pod_name) kube_pod_labels{_weave_namespace="weave",_weave_service="kube-state-metrics",label_name="{{label}}"})`,
+			docker.MemoryUsage:   `sum(label_replace(sum(container_memory_usage_bytes{image!="",namespace="{{namespace}}"}) by (pod_name, namespace), "pod", "$1", "pod_name", "(.*)") * on (pod) group_right(pod_name) kube_pod_labels{_weave_namespace="weave",_weave_service="kube-state-metrics",label_name="{{label}}"})`,
```
dlespiau: Instead of joining on kube_pod_labels to grab the `name` label, we can use the special `_weave_pod_name` labels that is present on all cAdvisor metrics (the ones starting by `container_`). This is so we don't require our users to have a `name` label in their manifest, that's one we have because we enforce it with a linter not something that kubernetes requires.

I'm not sure where `{{label}}` comes from in this code, but the value you can give to `_weave_pod_name` should be the name of the top-level controller that has created the object.

rbruggem: `{{label}}` gets replaced with the service name as in `metadata.name`.

rade: Where does `metadata.name` come from? And why can we expect everything to have a `name` label that matches it?

rbruggem: `metadata.name` is the resource name which comes from the service manifest.
Every service manifest is required to specify a resource name.

The promql queries above apply to services only.

rbruggem: I think I can do better, so bare a moment

rbruggem: Above the promql queries is a comment. That comment still stands true.
Currently, the generated queries are only going to work if 'service name' == 'pod name'

rbruggem:
@dlespiau thanks!

I have added an additional commit which removes that assumption in the service metrics that the service name equals the deployment name.

PR #3061, opened 2018-02-07
 Move DNS name mapping from endpoint to report

### LINKED ISSUE #3058: Move dns names from endpoints to report level
In large reports we can see many repeats of the same name lookup:

```
      ";10.244.252.166;42230": {
        "id": ";10.244.252.166;42230",
        "topology": "endpoint",
        "counters": null,
        "sets": {
          "reverse_dns_names": [
            "ip-10-244-252-166.ec2.internal"
          ]
        },
...
      ";10.244.252.166;42170": {
        "id": ";10.244.252.166;42170",
        "topology": "endpoint",
        "counters": null,
        "sets": {
          "reverse_dns_names": [
            "ip-10-244-252-166.ec2.internal"
          ]
        },
...
      ";10.244.252.166;42200": {
        "id": ";10.244.252.166;42200",
        "topology": "endpoint",
        "counters": null,
        "sets": {
          "reverse_dns_names": [
            "ip-10-244-252-166.ec2.internal"
          ]
        },
...
```

so we could have a structure up at `Report` level that mapped them once, then just look this up when we need it.
For backwards-compatibility we could create this structure in `Report.Upgrade()`

Relates to #1201

bboreham:
Fixes #3058 

If the probe sees many TCP sockets on a host, it will repeat the same address information on each endpoint.  Moving it to report level saves all this repetition.  Also `Sets` is an expensive data structure; for the DNS mapping I used a straightforward `map`, similar to `Topology.Nodes`.

I see about a 6% reduction in JSON file size, on Weave Cloud production reports.

The mapping is used as a mapping from IP address to name, and the current implementation encodes it the same way, but to me it seems more logical to send it as a mapping from name to address.  Probably that doesn't matter.

~I considered moving `DNSFirstMatch` from `render` to `report`, to collect all the behaviour together.  However the decisions made are somewhat related to rendering, so I left it.~

### DIFF
--- a/render/detailed/connections.go
+++ b/render/detailed/connections.go
@@ -73,7 +73,7 @@ func newConnectionCounters() *connectionCounters {
 	return &connectionCounters{counted: map[string]struct{}{}, counts: map[connection]int{}}
 }
 
-func (c *connectionCounters) add(outgoing bool, localNode, remoteNode, localEndpoint, remoteEndpoint report.Node) {
+func (c *connectionCounters) add(dns report.DNSRecords, outgoing bool, localNode, remoteNode, localEndpoint, remoteEndpoint report.Node) {
 	// We identify connections by their source endpoint, pre-NAT, to
 	// ensure we only count them once.
 	srcEndpoint, dstEndpoint := remoteEndpoint, localEndpoint
@@ -94,10 +94,10 @@ func (c *connectionCounters) add(outgoing bool, localNode, remoteNode, localEndp
 		return
 	}
 	// For internet nodes we break out individual addresses
-	if conn.remoteAddr, ok = internetAddr(remoteNode, remoteEndpoint); !ok {
+	if conn.remoteAddr, ok = internetAddr(dns, remoteNode, remoteEndpoint); !ok {
 		return
 	}
-	if conn.localAddr, ok = internetAddr(localNode, localEndpoint); !ok {
+	if conn.localAddr, ok = internetAddr(dns, localNode, localEndpoint); !ok {
 		return
 	}
 
@@ -105,7 +105,7 @@ func (c *connectionCounters) add(outgoing bool, localNode, remoteNode, localEndp
 	c.counts[conn]++
 }
 
-func internetAddr(node report.Node, ep report.Node) (string, bool) {
+func internetAddr(dns report.DNSRecords, node report.Node, ep report.Node) (string, bool) {
 	if !render.IsInternetNode(node) {
 		return "", true
 	}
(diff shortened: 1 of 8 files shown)


rade:
This is supposed to fix #3058, right?

render/detailed/connections.go
```diff
@@ -73,7 +73,7 @@ func newConnectionCounters() *connectionCounters {
 	return &connectionCounters{counted: map[string]struct{}{}, counts: map[connection]int{}}
 }
 
-func (c *connectionCounters) add(outgoing bool, localNode, remoteNode, localEndpoint, remoteEndpoint report.Node) {
+func (c *connectionCounters) add(rpt report.Report, outgoing bool, localNode, remoteNode, localEndpoint, remoteEndpoint report.Node) {
```
rade: perhaps this should take just a `DNSRecords` instead of a whole `Report`.

bboreham:
@rade yes, I forgot I'd filed an issue.

report/report.go
```diff
@@ -469,6 +475,37 @@ func (r Report) upgradeNamespaces() Report {
 	return r
 }
 
+// Note in-place modification of Endpoint.Nodes
```
rade: This is way dodgy. Don't do it. If you must, I expect an essay here explaining why it is safe.

report/report.go
```diff
@@ -469,6 +475,37 @@ func (r Report) upgradeNamespaces() Report {
...
+		return r
+	}
+	dns := make(DNSRecords)
+	for endpointID, endpoint := range r.Endpoint.Nodes {
+		_, addr, _, ok := ParseEndpointNodeID(endpointID)
+		snoopedNames, foundS := endpoint.Sets.Lookup(SnoopedDNSNames)
+		reverseNames, foundR := endpoint.Sets.Lookup(ReverseDNSNames)
+		if ok && (foundS || foundR) {
+			// Add address and names to report-level data - we assume
+			// all endpoints with the same address have equally good data
```
rade: I don't see why that assumption would hold. And if it doesn't then the data here does not converge, which results in view instability.

You could _optimise_ for the case that the assumption holds, but not depend on it.

report/report.go
bboreham: I was under the impression the existing `upgrade` routines did it too, but on closer inspection they don't.  Will fix.

report/report.go
bboreham: The data comes from the caches in `resolver.go` and `dns_snooper_linux.go`; if that data is changing then it's a matter of timing what any particular endpoint has.

I can't imagine why the data would change in such a way that results didn't converge shortly.

rade: AFAICT you could process the same report twice and get different results. That's a recipe for view instability.

bboreham:
I made the requested changes and rebased.  Also moved `DNSFirstMatch` since it looks a lot more like a method on `DNSRecords` now.

PR #3065, opened 2018-02-09
Show Time Travel at all times in Weave Cloud

fbarl:
##### Changes

* Show Time Travel at all times if historic reports are supported (in particular in Weave Cloud) together with pause/live buttons, removing the Time Control _live/pause/time travel_ buttons in the top right corner.
* If historic reports are not supported, show only Time Control with _live/paused_ buttons as before.

**Scope in Weave Cloud**
image

**Scope standalone**
image

### DIFF
--- a/client/app/scripts/actions/app-actions.js
+++ b/client/app/scripts/actions/app-actions.js
@@ -339,14 +339,15 @@ export function clickNode(nodeId, label, origin, topologyId = null) {
 
 export function pauseTimeAtNow() {
   return (dispatch, getState) => {
+    const getScopeState = () => getState().scope || getState();
     dispatch({
       type: ActionTypes.PAUSE_TIME_AT_NOW
     });
-    updateRoute(getState);
-    if (!getState().get('nodesLoaded')) {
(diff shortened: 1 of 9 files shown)


rade:
screenshot?

foot:
Minor aesthetical points:

- The loading indicator looks a smidge awkward behind the top nav maybe bump it down a bit.
- on 1440x and hosts+graph the cpu/memory/load is overlapped by details panel by ~10px which looks a tiny bit awkward
- we've lost of a lot of vertical space! This will work better when top-nav moves to the side?

fbarl:
> The loading indicator looks a smidge awkward behind the top nav maybe bump it down a bit.

Good catch! I centered it vertically.

> On 1440px and hosts+graph the cpu/memory/load is overlapped by details panel by ~10px which looks a tiny bit awkward

Alright, I increased the top margin for the node details to avoid them overlapping.

> We've lost of a lot of vertical space! This will work better when top-nav moves to the side?

Yes, hopefully after this PR is merged, we'll be incentivized to push with layout changes faster :)

PR #3066, opened 2018-02-12
Relative positioning for Table view tables

### LINKED ISSUE #3015: large gap between header and table

fbarl:
Follow-up on #3065 (which needs to be merged first).

Fixes #3015.

Use relative positioning for `NodesGrid` and calculate its height from its _top_ coordinate + window height.


foot:
oops wrong pr. will review this properly in a second.

PR #3070, opened 2018-02-15
Disable detail panel link

rbruggem:
The link is disabled if the `monitor` flag is not true.
If Weave Cloud runs without Monitor, the link is dead. The flag allows to pass Weave Cloud's Monitor state down to Scope.


leth:
[Hide deep links to unobtainable product areas](https://trello.com/c/92JA9Lfq/271-hide-deep-links-to-unobtainable-product-areas)

PR #3072, opened 2018-02-16
Dispatch 'Monitor' flag `render` function

rbruggem:
This is instead of dispatching the flag from the constructor which was called only once, hence neutralizing updates.

### DIFF
--- a/client/app/scripts/components/app.js
+++ b/client/app/scripts/components/app.js
@@ -99,6 +99,12 @@ class App extends React.Component {
     this.router.stop();
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.monitor !== this.props.monitor) {
+      this.props.dispatch(setMonitorState(nextProps.monitor));
+    }
+  }
+
   onKeyUp(ev) {
     const { showingTerminal } = this.props;
     keyPressLog('onKeyUp', 'keyCode', ev.keyCode, ev);


client/app/scripts/components/app.js
```diff
@@ -184,6 +182,8 @@ class App extends React.Component {
       timeTravelTransitioning, timeTravelSupported
     } = this.props;
 
+    this.props.dispatch(setMonitorState(this.props.monitor));
```
rndstr: I think this should actually go into `componentWillReceiveProps()`

rndstr: I'm not that experienced with the reactjs lifecycle, should this be put in `constructor()` &&  `componentWillReceiveProps()` instead?

My assumption so far is if `this.props.monitor` changes it will do an extra `render()` in the next cycle? It shouldn't call render() twice if it doesn't change (which is the less likely occurrence of the two).

fbarl: > I'm not that experienced with the reactjs lifecycle, should this be put in `constructor()` && `componentWillReceiveProps()` instead?

Yes it should! :) In this case I don't think an extra call to render would happen, as nothing will change for this component's `props` & `state`, but it is generally a bad practice to update state from a `render` hook:

https://reactjs.org/docs/react-component.html#render

So going with @rndstr's suggestion and having this line both in `contructor` and in `componentWillReceiveProps` (with `nextProps.monitor` instead) should be the right way to go.

rndstr:
thanks @fbarl PTAL

PR #3076, opened 2018-02-19
Don't count page cache as part of the process memory

dlespiau:
The container_memory_usage_bytes metric contains the page cache allocations
done by the kernel when the process does some I/O to speed subsequent accesses
eg. when reading 32 bytes a file, the kernel I/O scheduler will optimistically
load more data into that page cache to speed subsequent read operations.

This page cache is an implementation detail of disk I/O and the process have
roughly no control over it (not exactly true, but a good enough approximation).

RSS is a truer measure of memory consumption ie. the actual number of system
memory pages used by the process eg. when allocating 10MB on the heap with
malloc() but not accessing it, that memory isn't actually pinned in RAM and not
accounted for in RSS.

### DIFF
--- a/render/detailed/links.go
+++ b/render/detailed/links.go
@@ -68,7 +68,7 @@ var (
 		report.CronJob:     podIDHashQueries,
 		report.Service: {
 			docker.CPUTotalUsage: `sum(rate(container_cpu_usage_seconds_total{image!="",namespace="{{namespace}}",_weave_pod_name="{{label}}",job="cadvisor",container_name!="POD"}[5m]))`,
-			docker.MemoryUsage:   `sum(rate(container_memory_usage_bytes{image!="",namespace="{{namespace}}",_weave_pod_name="{{label}}",job="cadvisor",container_name!="POD"}[5m]))`,
+			docker.MemoryUsage:   `sum(rate(container_memory_rss{image!="",namespace="{{namespace}}",_weave_pod_name="{{label}}",job="cadvisor",container_name!="POD"}[5m]))`,
 		},
 	}
 )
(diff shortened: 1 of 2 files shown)


bboreham:
Scope also has its own memory metrics - do they need changed similarly?

dlespiau:
yes, once we found the correct metric to use. That's definitely not the memory cgroup RSS.

Basically, we don't want to count the pages still in the memory  but in the inactive LRU lists as the OS is free to reclaim them and it's unlikely to cause a performance degradation. There's something still to understand though, [the claim that total_inactive_anon is where writes to tmpfs (assuming all shmem?) are accounted for when swap is off](https://github.com/google/cadvisor/commit/307d1b1cb320fef66fab02db749f07a459245451). I'm guessing this has to do with the tmpfs memory being really shmem, which pages are in the swap cache, the part of the page cache backed by the swap device.

Current status:
The metric we seem to be wanting with swap off: `usage - total_inactive_file`
The metric we seem to be wanting with swap on: `usage - total_inactive_anon - total_inactive_file`

rade:
@dlespiau what's happening with this?

dlespiau:
I now believe we should useto the `working_set` cAdvisor metric  (`usage - total_inactive_file`) in both the request and what scope reports. Just need to do it.

I still don't understand the claim that `usage - total_inactive_anon - total_inactive_file` undercount shmem pages when swap is off, but using what kubernetes uses for its eviction algorithm sounds like a good thing.

dlespiau:
Won't have the time to dig as much as I'd want to here. Spring cleaning it is!

PR #3080, opened 2018-02-20
Remove flag -probe.kubernetes.interval and stop re-syncing Kubernetes data

bboreham:
We set up a "watch" on everything we need so there is no reason to ~poll~ re-sync repeatedly.

### DIFF
--- a/probe/kubernetes/client.go
+++ b/probe/kubernetes/client.go
@@ -48,7 +48,6 @@ type Client interface {
 
 type client struct {
 	quit             chan struct{}
-	resyncPeriod     time.Duration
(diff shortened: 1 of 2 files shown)


rade:
Can we think of any reason why one would want to poll?

rade:
@bboreham ping

I am wondering whether we should get rid of the flag and associated machinery altogether.

bboreham:
To clarify, it doesn't actually poll the Kubernetes server on that interval; the resync is an in-process operation.

So yes, we can remove the flag and hard-code the zero parameter.

Issue #3082, opened 2018-02-22
Unconnected containers in sock-shop


dholbach:
Looking at the Explore → Containers view of sock-shop in WC, I can see quite a few containers which are not connected to anything. In a quick conversation, @rade pointed out that #1491, #1846, #2129 and #2586 could possibly be related.

![screenshot-unconnected-containers](https://user-images.githubusercontent.com/1346979/36541502-689fa0e0-17de-11e8-8c7e-59a3774b0702.png)

dholbach:
This seems to happen, both with "Show Uncontained" and "Hide Uncontained".

dholbach:
![screenshot-unconnected-containers-2](https://user-images.githubusercontent.com/1346979/36542239-901d7a1e-17e0-11e8-99a4-2d457dd6f541.png)

rade:
At first I thought it might be missing all short-lived connections, but I do occasionally see connections from front-end to user, so the problem must be something else.

The Pod topology looks fine, with all expected connections showing.

dholbach:
Toggling the filters did not help, FF 59.0b12 Ubuntu.

bboreham:
The POD containers are visible because Kubernetes changed the image of the ”pause container”. 
See IsPauseImageName() in the code.

rade:
Oh, that's yet _another_ bug.

rade:
To be clear, the issue here isn't all the unconnected POD containers - that is expected - but all the _other_ unconnected containers, especially since for a good few of them the corresponding pods _do_ show as connected.

rade:
> the POD containers are visible because Kubernetes changed the image of the ”pause container”.

Actually, they are visible because of the filtering bug #1999 - if you switch to show 'All' and then back to 'Application Containers' they go away.

bboreham:
Checking more closely I see POD containers with image `gcr.io/google_containers/pause-amd64:3.0` which would be matched by the current hard-coded string.  Sorry for confusion.

bboreham:
> for a good few of them the corresponding pods do show as connected.

I think the difference is that Scope knows the pod IPs and doesn't know the container IPs, and doesn't join those pieces of information together.

Some connections have a pid, and Scope does map processes to containers when rendering, so you see those connections.

rade:
> Some connections have a pid, and Scope does map processes to containers when rendering, so you see those connections.

Why don't _all_ the connections have a pid?

jrryjcksn:
[...]

PR #3089, opened 2018-02-23
[Paused] Find a container's IP address from its pod, if possible

### LINKED ISSUE #2586: no connections between containers when `probe.processes=false`
In a kubernetes cluster, containers do not have an associated IP address, as far as docker is concerned. We rely purely on connection tracking at the process level to determine connectivity for containers, so when such tracking is disabled with `probe.processes=false`, no container connections are show, which is unfortunate and surprising.

Note that we do track pod IPs and use those to figure out connections at the pod level.

NB: this is all based on my reading of the code and some scope reports from a k8s cluster. TODO: verify by running probes in such a cluster with `probe.processes=false`.

bboreham:
This adds some connections that are missing, such as reported at #3082 
Fixes #2586 

We still don't get all of them, because pods with more than one container are excluded from the mapping because we don't know which one to attach the edge to.

I don't like extending the `toIP()` function to take the whole report, but I can't think of a better way right now.

Also drive-by fix of a blank ID that was wasting some time.

### DIFF
--- a/render/container.go
+++ b/render/container.go
@@ -44,12 +44,12 @@ const originalNodeID = "original_node_id"
 // ConnectionJoin joins the given topology with connections from the
 // endpoints topology, using the toIPs function to extract IPs from
 // the nodes.
-func ConnectionJoin(toIPs func(report.Node) []string, topology string) Renderer {
+func ConnectionJoin(toIPs func(report.Report, report.Node) []string, topology string) Renderer {
 	return connectionJoin{toIPs: toIPs, topology: topology}
 }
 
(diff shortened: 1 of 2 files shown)


rade:
Overall, I am not keen on this approach. I'd rather a) figure out and fix why the endpoint->pid->process->container mapping isn't picking up these connections, and b) pick up container IPs directly. Both of these will improve accuracy more broadly than just in the context of k8s.

bboreham:
I started working on "b) pick up container IPs directly" - by entering the container's namespace.
I'll close this.

PR #3091, opened 2018-02-23
Exclude null entries for networks on container nodes in probe report

bboreham:
Also the "none" network. This just makes probe reports slightly smaller because they won't have this:

```
        "sets": {
          "docker_container_ips": null,
          "docker_container_ips_with_scopes": null,
          "docker_container_networks": [
            "none"
          ],
          "docker_container_ports": null
        },
```

This is quite typical for a container created by Kubernetes, which wires up networking via CNI so Docker doesn't know about it.

### DIFF
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -287,6 +287,9 @@ func (c *container) NetworkInfo(localAddrs []net.IP) report.Sets {
 	// here, and provide foreign key links from nodes to networks.
 	networks := make([]string, 0, len(c.container.NetworkSettings.Networks))
 	for name, settings := range c.container.NetworkSettings.Networks {
+		if name == "none" {
+			continue
+		}
 		networks = append(networks, name)
 		if settings.IPAddress != "" {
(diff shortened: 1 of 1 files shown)


probe/docker/container.go
```diff
@@ -303,11 +306,20 @@ func (c *container) NetworkInfo(localAddrs []net.IP) report.Sets {
...
+	if len(c.container.NetworkSettings.Ports) > 0 {
+		s = s.Add(ContainerPorts, c.ports(localAddrs))
+	}
+	if len(ipv4s) > 0 {
+		s = s.Add(ContainerIPs, report.MakeStringSet(ipv4s...))
+	}
+	if len(ipsWithScopes) > 0 {
+		s = s.Add(ContainerIPsWithScopes, report.MakeStringSet(ipsWithScopes...))
+	}
+	return s
```
rade: it's kinda tempting to introduce a `Sets.AddMulti` (or s.t. like that) which does the len check and conditional StringSet construction. There are probably a few other places where we could use that.

PR #3097, opened 2018-02-26
Pick up default topology options on initial load

fbarl:
Fixes https://github.com/weaveworks/service-ui/issues/1999.

As @foot discovered, the issue was actually easily reproducable - it would always show the _Uncontained_/_Unmanaged_ nodes on fresh page load, regardless of the options set in the UI.

The problem was that the initial `ROUTE_TOPOLOGY` handler would set `topologyOptions` to a certain value and then the condition in `RECEIVE_TOPOLOGIES` would not update with the default options once they'd be received with topologies data (with a certain delay).

The solution was to unconditionally update the options with defaults in `RECEIVE_TOPOLOGIES`  on first load, but using them only as 'fallback', still giving priority to the URL/local state options.


rade:
Is this issue something that affected standalone mode too?

Issue #3099, opened 2018-03-05
Enhanced support Metadata template data types [hyperlinks]


todkap:
In the Metadata template data types, the following is documented

https://www.weave.works/docs/scope/latest/plugins/#report-data-structures
**Metadata Templates section**
> dataType specifies the type of data, and determines how the value is displayed. Possible values for this attribute are: “number”, “ip”, “datetime” and “” for strings.

It would be great if there was support for URLs that support direct linking to resources.   The use case is that we have a Node that represents a Deployment of some service.   This service exposes a set of links for administration, logging, etc.   

**Food for thought**
If there was a more pluggable extension model, templates could define their own data types backed by the plugin to support other data types as well.

bboreham:
Hi, thanks for the suggestion!

How do you envisage this would look, both in the code and on the screen?

RcrsvSquid:
Hello,

I work with Todd.

We've implemented the feature by first running a plugin that posts metadata or mulit-column data including the `dataType: link` field. The frontend then conditionally renders either a link or the original component.

Our implementation is here: https://github.com/weaveworks/scope/pull/3105

rade:
Resolved by #3105.

PR #3105, opened 2018-03-08
Feat/link data type

RcrsvSquid:
Implements link data type proposal in #3099.

Adds the link dataType as a displayable field in both the NodeDetailsGenericTable and NodeDetailsInfo components via conditional render.


squaremo:
Thanks for this! I'm not qualified to review the code change; but, to get it in shape for someone who is, I recommend rebasing against master, and squashing the redundant commits out. `git rebase -i` is your friend (though sometimes a difficult-to-like friend).

RcrsvSquid:
Sorry about that @squaremo. I've squashed and rebased. Thanks for taking a look!

squaremo:
> Sorry about that @squaremo. I've squashed and rebased

No worries, thanks for that.

I like the idea in #3099, but we shall have to wait for someone more familiar with Scope to check the code. One thing that did occur to me: do you perhaps want to have both a link href and a link text?

RcrsvSquid:
@squaremo How might we get both values from a plugin into the view? We had tested this idea earlier by sending the value as `value: link-href:link-text` and parsing the string in the view but decided not to move forward with that solution. Is there a better way?

squaremo:
> How might we get both values from a plugin into the view?

Good question -- my understanding is too superficial to have grasped that you couldn't just supply a structure. I can certainly see that encoding the values in a string seems like a step too far. :-S

rndstr:
Appreciate the PR @RcrsvSquid! First feedback:

- extend `dataType` documentation in `site/plugins.md`
- links appear unstyled for me (should be same as others)
- I think there are some more places where these values are rendered such as 
https://github.com/weaveworks/scope/blob/da6995bd87d13a36608eff11d3ea155dc7a319c4/client/app/scripts/components/node-details/node-details-table-row.js#L43

RcrsvSquid:
@rndstr I've responded to you comments with changes

RcrsvSquid:
@foot @rndstr Is there anything else you need from me to be able to merge this pull request?

foot:
But @rndstr makes a good point against future proofing, if a future plugin proxies dodgy urls etc. So yes, if you could please add the noop/norel that would be grand! Ta!

RcrsvSquid:
@foot Done!

rndstr:
@RcrsvSquid awesome :sparkles: 

(did run tests locally, they passed)

Issue #3096, opened 2018-02-26
namespace selector doesn't handle lots of namespaces well


rade:
The namespace selector runs out of screen real estate when there are lots of namespaces.

ycao56:
Maybe by default show certain number of namespaces and add a button to expand to all

ycao56:
I created a PR to fix this issue. Please check https://github.com/weaveworks/scope/pull/3117

PR #3117, opened 2018-03-22
Hover to expand topology option if the list exceeds the max width

ycao56:
Fixes https://github.com/weaveworks/scope/issues/3096
1. Moved option `All` to the very beginning
2. Changed to use flexbox
3. Added hover to expand behavior to display more options if the list exceeds the max width
weave


rade:
Thanks for the PR. This looks like a pretty neat and low-cost (in terms of extra code) solution. Alas I am not a UI expert. @foot what do you reckon?

foot:
This looks really slick, thanks for the PR!

Its often dangerous to move elements around a lot on hover but I'm keen to take this for a spin. Adding an expansion button is not as cool but may be a bit easier to interact with. We'll give it a UX test on Monday.

foot:
While the hover expand is quite fun and playful you can quite easily get into that frustrating scenario where the item you intend to click jumps out from under the cursor at the last second.

A little expand/collapse caret would work fine I think. Especially if it didn't move around when clicked so you click, click, click to close / open / close.

ycao56:
@foot I thought about the expand/collapse caret, but I didn't figure out the position to place it. The current UI doesn't have this type of element so I am not sure the UX pattern to follow. Can you suggest where should be the caret placed?
Also once the caret is used, when to show/hide the caret itself becomes a difficulty due to the nature of a dynamic list. i.e. for filter list that doesn't go out of the screen we don't need to show the caret
Any suggestions?

foot:
Hi again! Sorry for the delay.

We use this `+n`[0] notation around the place for expanding out more information which could work in this case.

> when to show/hide the caret itself becomes a difficult

Yes! I don't think you can pull this off in pure CSS.. But, we store the window's width/height in the global store so you can grab that in the component by adding something like `viewportWidth: state.getIn(['viewport', 'width'])` into `mapStateToProps`. 

Then... you'd unfortunately still have to add a ref/dom measuring stuff into topoOptions.

rough pseudo js:
```js

getNumberOptionsHidden(viewportWidth, containerRef) {
  const { width, height } = containerRef.getBoundingClientRect();
  // maybe only use 50% of screen width?
  if (width < viewportWidth * 0.5) {
     return 0;
  }
  
  // calculate cumulative sum of width of DOM children and
  // takeWhile cumWidth < viewportWidth * 0.5
  ...
  return numberOfChildrenThatFitIn50Percent;
}

class TopologyOptions {

...

componentWillReceiveProps(nextProps) {
   const widthHasChanged = nextprops.viewportWidth !== this.props.viewportWidth;
   if (widthHasChanged) {
      this.setState({ 
        numberOptionsHidden: 
          getNumberOptionsHidden(nextprops.viewportWidth, this.containerRef)
      });
   }
}

render() {
  return <div ref={r => this.containerRef = r} ...
   ...
   {take(this.props.options, this.props.options.length - this.state.numberOptionsHidden)}
   ...
   {(this.state.numberOptionsHidden > 0)
     && <ShowMoreHorizontal count={this.state.numberOptionsHidden} />}
 </div>
}

...

}
```

`<ShowMore />` could be copied and adapted a bit info a `<ShowMoreHorizontal />` like thing.

WDYT?

[0] Example of `+n` in the details panel: 
screen shot 2018-04-09 at 11 04 56

2opremio:
@ycao56 did you get so see @foot 's proposal? Any chance of getting it implemented anytime soon?

2opremio:
@ycao56 Closing (I am cleaning up the lingering PRs). Please feel free to create another PR if you want this merged (we are still interested).

2opremio:
Reopening instead of

fbarl:
ping @foot - how does this PR look like - did you want to review it? :)

frederiksf:
When this PR will be checked? We currently cannot use the product because this isnt fix yet

foot:
Hi @frederiksf thanks for the feedback! You are also suffering from too many namespaces?

This PR has lost a bit of momentum but priority climbs if we know if effects more users. We'll bump it up the list.

frederiksf:
@foot Yes we do. For each customer app we create a separate NS. It would be great if this will be fixed!

foot:
@ycao56 I think this is still an improvement over the current implementation. We're happy to merge this into master sorry about the delay.

Do you want to rebase this on master resolving the conflict (Both your additions and the font-size addition should stay)? Otherwise I can resolve the issues here and create/merge a new PR based on this one. Cheers!

ycao56:
@foot done

AVRahul:
@ycao56  there are some improvements which could be better, the hovering effect can be just restricted to
the last-child class instead applying in overall sidebar.

foot:
@ycao56 thanks again for the PR!

@AVRahul nice thought, I'll open another issue for how we could improve this even further.

Issue #3112, opened 2018-03-19
Topology filtering based on k8s label


ycao56:
Currently, UI could only filter k8s resources by namespaces. It would be more useful to be able to filter by label. In k8s would, labels enable users to map their own organizational structures onto system objects in a loosely coupled fashion. Users will be benefit from this to customize their own topology view by using label filter.

ycao56:
Create a PR to implement this https://github.com/weaveworks/scope/pull/3118

PR #3118, opened 2018-03-22
Support topology filtering by k8s label

ycao56:
Resolves https://github.com/weaveworks/scope/issues/3112

- Added label filters to `PODS` `CONTROLLERS` `SERVICES`

image

Note: k8s labels are usual a lot and UI doesn't handle it well. So I created another PR to address it https://github.com/weaveworks/scope/pull/3117

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -60,6 +60,17 @@ func namespaceFilters(namespaces []string, noneLabel string) APITopologyOptionGr
 	return options
 }
 
+// labelFilters generates a label selector option group based on the given labels
+func labelFilters(labels []string, noneLabel string) APITopologyOptionGroup {
+	options := APITopologyOptionGroup{ID: "label", Default: "", SelectType: "union", NoneLabel: noneLabel}
+	for _, label := range labels {
+		options.Options = append(options.Options, APITopologyOption{
+			Value: label, Label: label, filter: render.IsLabel(label), filterPseudo: false,
+		})
+	}
+	return options
+}
+
 // updateFilters updates the available filters based on the current report.
 func updateFilters(rpt report.Report, topologies []APITopologyDesc) []APITopologyDesc {
 	topologies = updateKubeFilters(rpt, topologies)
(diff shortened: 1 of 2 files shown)


rade:
Is this really better than using the existing search/filter functionality? Seems quite unwieldy once there are lots of labels. @foot what do you think?

ycao56:
https://github.com/weaveworks/scope/pull/3117 should address the problem of large number of labels. User is going to hit it with namespaces too.
If the number is really a problem, I think the label with occurence equals 1 could be filtered. Ideas?

rade:
Why is this better than the existing search/filter functionality?

dlespiau:
A couple of random drive-by thoughts:
- A list of labels adds an exploration side search doesn't currently have: "What's the label I am supposed to search for again?" vs "Let's click on that label=value".
- Having the full carnality of of labels (ie all (key,value) pairs) seems quite noisy. It does allow an easy way to do conjunction though.
- Alternative ideas:
  - present the list of label names as a first level selection, then the list of values for that label. May be hard to express conjunction.
  - improve the search/filter to be able to filter with a label, possibly with autocomplete to cater with the label name/value discovery point.

rade:
> improve the search/filter to be able to filter with a label, possibly with autocomplete to cater with the label name/value discovery point.

I had the same idea, and I like it ;)

bboreham:
As a bystander it's unclear to me what the way forward is for this PR.  There seems to be a proposal and some discussion; can we narrow it down to a specific request to make the PR acceptable for merge?

foot:
Yeah for sure! I don't think a dropdown would work perfectly. The related PR was proposing a hiding solution that was quite nice with a couple of tweaks.

I can have a quick look over it again.

2opremio:
> I can have a quick look over it again.

Thanks, please feel free to re-open it and finish it up if you have the time.

2opremio:
Closing in favor of https://github.com/weaveworks/scope/pull/3117

Issue #3120, opened 2018-03-25
update details panel immediately on shortcut reports


rade:
When shortcut reports were introduced, [it was noted](https://github.com/weaveworks/scope/pull/640#issuecomment-155393004) that they don't trigger updates of the detail panel, so changes in the detail panel data are only reflected after the next periodic update. At the time these updates were done every 1s, which wasn't so bad, but these days they are at 5s.

The most visible user impact of this is that when pressing control buttons, such as 'pause', the button state can take up to 5s to update.

Note that, by contrast, shortcut reports update the main topology display instantly.

PR #3121, opened 2018-03-25
make shortcut reports work (again)

### LINKED ISSUE #3113: Shortcut property is not copied by report.Report.Copy method
Shortcut property is not copied by report.Report.Copy method

Method Copy ignores Shortcut property: https://github.com/weaveworks/scope/blob/3de06b5a0963549308bd28a296063ec5c5e963d7/report/report.go#L258

As a result, report is published with Shortcut property always set to false for docker.reporter
It happens because probe.drainAndPublish function makes report copy calling rpt.BackwardCompatible()
https://github.com/weaveworks/scope/blob/7b0f0cb158d3f8e468913ee5658ddbfae933b8b7/probe/probe.go#L203

rade:
To test this, run
```
docker run --name=foo --rm -ti alpine sleep 10
```
On master it can take several seconds for the container to appear on the topology view, and then several seconds for it to disappear when it stops.

After the first commit, it disappears immediately when stopped, but still takes time to appear. The latter is cured by the 2nd commit.

Fixes #3113.

In the process I discovered #3120.

### DIFF
--- a/probe/docker/registry.go
+++ b/probe/docker/registry.go
@@ -373,11 +373,9 @@ func (r *registry) updateContainerState(containerID string, intendedState *strin
 	}
 
 	// Trigger anyone watching for updates
-	if err != nil {
-		node := c.GetNode()
-		for _, f := range r.watchers {
-			f(node)
-		}
+	node := c.GetNode()
+	for _, f := range r.watchers {
+		f(node)
 	}
 
 	// And finally, ensure we gather stats for it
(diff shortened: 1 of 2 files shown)


rade:
Both these changes do have a performance impact.

1. The fix to `Report.Copy()` means shortcut reports now trigger delivery via the special shortcut path, and a render in the app (as they should).
2. The fix to `updateContainerState` results in more shortcut reports being delivered - previously they'd only be sent for containers when they were stopped, now they are sent on all state changes (as they should). This results in more report processing in general.

If we deem these too costly then we should ditch shortcut reports altogether, since in their current state they are pointless.

PR #3128, opened 2018-04-05
documentation fix: project for openshift installation

doodlesbykumbi:
Current installation instructions for OpenShift fail because they introduce policy changes to a different project than the one where weave scope pods and resources are created.
`https://cloud.weave.works/k8s/scope.yaml` explicitly uses the `weave` namespace - this ought to match the OpenShift project.

### DIFF
--- a/site/installing.md
+++ b/site/installing.md
@@ -176,7 +176,7 @@ To install Weave Scope on OpenShift, you first need to login as `system:admin` u
 
 Next, create a dedicated project for Weave Scope then apply policy changes needed to run Weave Scope:
 
-    oc new-project weave-scope
+    oc new-project weave
     # Scope probe pods need full access to Kubernetes API via 'weave-scope' service account
     oc adm policy add-cluster-role-to-user cluster-admin -z weave-scope
     # Scope probe pods also need to run as priviliaged containers, so grant 'priviliged' Security Context Constrains (SCC) for 'weave-scope' service account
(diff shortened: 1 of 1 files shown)


site/installing.md
```diff
@@ -176,7 +176,7 @@ To install Weave Scope on OpenShift, you first need to login as `system:admin` u
 
 Next, create a dedicated project for Weave Scope then apply policy changes needed to run Weave Scope:
 
-    oc new-project weave-scope
+    oc new-project weave
     # Scope probe pods need full access to Kubernetes API via 'weave-scope' service account
     oc adm policy add-cluster-role-to-user cluster-admin -z weave-scope
```
rade: Excuse my ignorance about Openshift, but...

What does the `-z` option do here? I cannot find it in the documentation. The comment above refers to it as a "service account". Is that synonymous with namespace? If so, shouldn't it be changed too?

doodlesbykumbi: No worries. 

A service account provides an identity for workloads and generally belongs to a namespace in the same way that a pod does.

The `oc` CLI help states that `-z` flag is indeed used to specify a service account.
```
Usage:
  oc adm policy add-cluster-role-to-user <role> <user | -z serviceaccount>
[user]... [options]

Options:
  -z, --serviceaccount=[]: service account in the current namespace to use as a
user
```

`https://cloud.weave.works/k8s/scope.yaml` creates and uses a service account called `weave-scope` so we don't need to change that.

doodlesbykumbi: No worries. 

A service account provides an identity for workloads and generally belongs to a namespace in the same way that a pod does.

The `oc` CLI help states that `-z` flag is indeed used to specify a service account.
```
Usage:
  oc adm policy add-cluster-role-to-user <role> <user | -z serviceaccount>
[user]... [options]

Options:
  -z, --serviceaccount=[]: service account in the current namespace to use as a
user
```

`https://cloud.weave.works/k8s/scope.yaml` creates and uses a service account called `weave-scope` so we don't need to change that.

rade: > https://cloud.weave.works/k8s/scope.yaml creates and uses a service account called `weave-scope`

Ah, I missed that bit. Thanks for the explanation.

Issue #3130, opened 2018-04-05
Add PV and PVC support to weave scope 


satyamz:
Add support for Kubernetes Storage components such as `Persistent Volume` (PV) and `Persistent Volume Claim` (PVC) . Scope UI can have new sub-topology `Persistent Volumes` under `PODS` which gives view and linking of `PV` and `PVC`.

Preparing PR for the same.

PR #3132, opened 2018-04-05
Add pv and pvc support in weave scope

satyamz:
As a part of this, a new sub-topology called `Persistent Volumes` which contain the graph of `Persistent Volume` and `Persistent Volume Claim`. 
Edges are drawn on the basis of Volume Claim Name in PV and PVC.

Fixes: #3130 

pv-pvc-support


rade:
This is great! Good effort!

Is there anything showing which pods use which claims? Would that be useful? Which makes me wonder whether persistent volumes and claims should be shown in the existing k8s views rather than getting their own view. Or, as a stepping stone, showing lists of volumes on the pod/controller detail panels, and linking to the PV node details and vice versa.

Why do the PVCs in your example have such cryptic names?

render/persistentvolume.go
```diff
@@ -0,0 +1,94 @@
...
+	MakeReduce(
+		ConnectionStorageJoin(
+			Map2PVCName,
+			report.PersistentVolume,
+			SelectPersistentVolumeClaim,
+		),
+		MapEndpoints(
+			Map2SC,
+			report.PersistentVolumeClaim,
+		)))
```
rade: Why the `MapEndpoints`? Endpoints don't have persistent volume claims.

render/render.go
```diff
@@ -251,6 +251,26 @@ func (ret *joinResults) rewriteAdjacency(outID string, adjacency report.IDList)
...
+		outID, ok := ret.mapped[n.ID]
+		if !ok {
+			continue
+		}
+		// Since PV and PVC will have only single adjacency
+		ret.storageAdjacency(outID, n.ID)
+	}
+	return Nodes{Nodes: ret.nodes}
+}
+
```
rade: Why isn't the existing `result()` method sufficient?

test/fixture/report_fixture.go
```diff
@@ -354,12 +358,48 @@ var (
 		},
 		Window: 2 * time.Second,
 	}
+	PVReport = report.Report{
```
rade: This really should be part of `Report`.

rade:
Couldn't the edges from PVCs to PVs be added by the probes, thus obviating the need for any special rendering code?

probe/kubernetes/reporter.go
```diff
@@ -98,6 +105,24 @@ var (
...
+		Message:          {ID: Message, Label: "Message", From: report.FromLatest, Priority: 7},
+	}
+
+	PersistentVolumeClaimMetadataTemplates = report.MetadataTemplates{
+		NodeType:    {ID: NodeType, Label: "Type", From: report.FromLatest, Priority: 1},
+		Namespace:   {ID: Namespace, Label: "Namespace", From: report.FromLatest, Priority: 2},
+		Status:      {ID: Status, Label: "Status", From: report.FromLatest, Priority: 3},
+		VolumeName:  {ID: VolumeName, Label: "Volume", From: report.FromLatest, Priority: 4},
+		AccessModes: {ID: AccessModes, Label: "Access Modes", From: report.FromLatest, Priority: 5},
+	}
```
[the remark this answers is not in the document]
satyamz: Yes! Name is retrieved by `MetaNode`. We have `Meta` interface which has method `Name()` to retrieve the Name of the K8s resource. Hence, no need to include Name here!

satyamz:
> Is there anything showing which pods use which claims? Would that be useful? Which makes me wonder whether persistent volumes and claims should be shown in the existing k8s views rather than getting their own view. Or, as a stepping stone, showing lists of volumes on the pod/controller detail panels, and linking to the PV node details and vice versa.

Pods has `VolumeClaim` in the Spec. We are using the same to retrieve PVC for the given Pod. However, this change is not included in this PR. We're sending another PR for that as soon as this PR is merged (just to reduce the size of PR we've split it in two).

Regarding keeping PV, PVC along with pods, We want to show complete kubernetes storage support eventually. Another reason for this can be, for storage relations we do not need any networking based linking. Hence, We have our own storage specific and simple renderer for PV, PVC objects. 
This is how final view looks:

`Cylinder` : PV
`DottedCylinder`: PVC
`Sheet` : Storage Class (SC)

  
storage-all



> Why do the PVCs in your example have such cryptic names?

These are not PVCs. These are PVs. Names for the PV is generated by K8s storage controller. You can check the same https://github.com/kubernetes-incubator/external-storage/blob/e3789b84521b5e75f161a6885010d2b7b2f50033/lib/controller/controller.go#L1160:34

> Could you post a screen shot of the table mode, please?

table-view

> Does the namespace selector apply to PVCs?

Yes! PVCs can have `Namespace` but PVs, Storage Classes don't have `Namespace`. We're still not clear whether to use `Namespace` filter for PV sub-topology. Also, We're not able to remove the Namespace filter for PV sub-topology as Namespace filter is invoked on the basis of topology and not sub-topology. I would like to know your opinion on this.


> Does searching/filtering provide reasonable results?

Yes! Searching works properly. However, we're not able to apply `Namespace` filters. 
status-searching

render/detailed/summary.go
```diff
@@ -371,6 +375,16 @@ func weaveNodeSummary(base BasicNodeSummary, n report.Node) BasicNodeSummary {
...
+func persistentVolumeNodeSummary(base BasicNodeSummary, n report.Node) BasicNodeSummary {
+	base = addKubernetesLabelAndRank(base, n)
+	return base
+}
+
+func persistentVolumeClaimNodeSummary(base BasicNodeSummary, n report.Node) BasicNodeSummary {
+	base = addKubernetesLabelAndRank(base, n)
+	return base
+}
+
```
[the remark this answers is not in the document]
satyamz: No! Apart from type of resource, nothing seem to be useful to show in `base.LabelMinor`.

rade:
Ah, I thought the dotted cylinders were PVs, and the solid ones PVCs, since the latter have "pvc-..." names in your screenshots.

IMO the pvc-pv edge has the wrong polarity. The PVC references the PV, so the arrow should go from the former to the latter. Then, by extension, we get pod -> pvc -> pv.

Regarding having a separate view... views with pods get crowded very quickly since that is the granularity at which things get scaled in k8s. Surely a user would want to see what PVs their _workloads_ (i.e. groups of pods managed by a controller) are accessing. So IMO PVs and PVCs should go in the Pod and Controller views, possible with a selector that lets users decide whether they want to see them.

satyamz:
> IMO the pvc-pv edge has the wrong polarity. The PVC references the PV, so the arrow should go from the former to the latter. Then, by extension, we get pod -> pvc -> pv.

Fixing this and other suggested changes. Will push all soon.  

> Regarding having a separate view... views with pods get crowded very quickly since that is the granularity at which things get scaled in k8s. Surely a user would want to see what PVs their workloads (i.e. groups of pods managed by a controller) are accessing. So IMO PVs and PVCs should go in the Pod and Controller views, possible with a selector that lets users decide whether they want to see them.
Merge state




Yes, I agree with you on the part where user would want to have separate storage view. Also, we already have PVC controllers and replicas in the PODS & CONTROLLERS section. As you are suggesting, we can add storage filter option there which will give mapping of stateful pods and PVC controller replicas only (no other pods). Indeed this is good suggestion, I will raise issue for this and will start work on this.

Having different storage view can help better storage management for K8s stateful workloads. Can we have new sub-topology under `Hosts` as `Storage` (below `weave net`) to show all the K8s storage components? As you know, we're finding storage relationship as `pod->pvc->pv`. In this phase, `PV` is the last components but in the `Storage` section we're extending this view to show which Disks are providing `PV` along with `storage pool` with opetations (in our scope term controls) such as snapshots, clone, backup etc. Hence, We're adding `Persistent Volumes` as a new sub-topology. We want to move this `Persistent Volumes` to `Storage` under Hosts once we've disks support. In that case, we can show disk and host relationship too. `Storage` topology will be big plus for weave scope wrt storage management. We are in the middle of development of `Node Disk Manager` ([NDM](https://docs.google.com/presentation/d/11GLg21x7G-nMTNw8aNIOhhjW_-eK19zSI9Xm-0jYHKs/edit?usp=sharing)) which we'll push to the Kubernetes. NDM will help us getting all the disk related info and operations handy.  
If you're fine with this, we can move `Persistent Volumes` to the `Storage` section in this PR itself. 

Thanks!

2opremio:
@satyamz any chance of getting fixes for the suggested changes anytime soon?

satyamz:
@2opremio I am sending changes today.

satyamz:
Hi @rade , @bboreham , @2opremio ,
Latest changes will remove `Persistent Volumes` sub topology from `PODS` topology and add K8s storage components in the storage filter in PODS topology as per suggestion from @rade 

Storage Filter On:

weave-scope-k8s-storage

Storage Filter Off:

weave-scope-k8s-pods






> Why isn't the existing result() method sufficient?

Currently, if we want to connect pod1 with pod2 based on network connection then we've network endpoints to do that. endpoints are working as a reference node & they have all the adjacencies. Further, we're passing endpoints nodes with adjacencies to `result()`. In storage we don't have any common reference node such as endpoint. Storage components are connected together directly through one to one mapping. Hence, we're introducing `storageResult()` to handle resource(PVC) to resource (PV) mapping.  

Thanks

bboreham:
@satyamz apologies I have been away from this for some time.  I finally got my own version to work - see https://github.com/weaveworks/scope/compare/master...k8s-volumes

It uses lots of code from this PR, but the key thing I wanted to show is that it has virtually no changes on the rendering side.  No new methods on `joinResult` and none of the code you used to feed things into it.  I simply create the adjacency I want in the probe.  I don't include `probeID` in the storage node; can't see any need for it.

I left in that branch a previous commit where the information goes via the `parent` field; this seemed plausible but in the end unnecessary.

What I suggest is you look at my branch, then simplify this PR based on those ideas.  I left out all your tests, also the most recent changes, because all I wanted to do was understand the rendering/adjacency side.

satyamz:
Hi @bboreham Thanks for the all the work and reply. I must say, you opened up my mind with these changes. I will reflect these changes in this PR and will push commits here soon.

Thanks a lot :+1:

satyamz:
Hi @bboreham, 
I tried the solution given you. It works fine for PV to PVC adjacency. However, this method will not work for adding adjacency from pod to pvc. If we send PVC as a adjacency for pod from probe then that'll be deleted by the call `connectionJoin()` [here](https://github.com/weaveworks/scope/blob/ca3dacdd70f981eed9876e61feaafef1b3ae7b56/render/pod.go#L62) and [here](https://github.com/weaveworks/scope/blob/ca3dacdd70f981eed9876e61feaafef1b3ae7b56/render/render.go#L156).

If we want to add adjacency for pod -> pvc with the method proposed by you then we'll have to consider `ClaimName` in the pod spec to create PVC nodeID because in pod spec `UID` for PVC is not available, `ClaimName` is unique. Similarly for `PVC` to `StorageClass` adjacency too.

bboreham:
We had a more detailed conversation in Slack.
I suggested writing a self-contained `Render()` function that shows Pods related to Storage, then you can add that into the existing Pod screen by appending to the `MakeReduce` list.

2opremio:
Awesome! Looking forward to the changes!

satyamz:
Hi @bboreham please review this approach https://github.com/openebs/scope/compare/add-pv-pvc-support...satyamz:refactor-render

satyamz:
Hi @bboreham @2opremio 
I updated this PR as per suggestions given by Bryan. 
Thanks

probe/kubernetes/persistentvolume.go
```diff
@@ -0,0 +1,35 @@
...
+func NewPersistentVolume(p *apiv1.PersistentVolume) PersistentVolume {
+	return &persistentVolume{PersistentVolume: p, Meta: meta{p.ObjectMeta}}
+}
+
+// GetNode returns Persistent Volume as Node
+func (p *persistentVolume) GetNode(probeID string) report.Node {
+	return p.MetaNode(report.MakePersistentVolumeNodeID(p.UID())).WithLatests(map[string]string{
+		report.ControlProbeID: probeID,
+		NodeType:              "Persistent Volume",
+		VolumeClaim:           p.Spec.ClaimRef.Name,
```
bboreham: can we assume `ClaimRef` is non-nil?

probe/kubernetes/storageclass.go
```diff
@@ -0,0 +1,33 @@
...
+
+// NewStorageClass returns new Storage Class type
+func NewStorageClass(p *storagev1.StorageClass) StorageClass {
+	return &storageClass{StorageClass: p, Meta: meta{p.ObjectMeta}}
+}
+
+// GetNode returns StorageClass as Node
+func (p *storageClass) GetNode(probeID string) report.Node {
+	return p.MetaNode(report.MakeStorageClassNodeID(p.UID())).WithLatests(map[string]string{
+		report.ControlProbeID: probeID,
```
bboreham: why do we need this?

probe/kubernetes/persistentvolume.go
```diff
@@ -0,0 +1,35 @@
...
+
+// NewPersistentVolume returns new persistentVolume type
+func NewPersistentVolume(p *apiv1.PersistentVolume) PersistentVolume {
+	return &persistentVolume{PersistentVolume: p, Meta: meta{p.ObjectMeta}}
+}
+
+// GetNode returns Persistent Volume as Node
+func (p *persistentVolume) GetNode(probeID string) report.Node {
+	return p.MetaNode(report.MakePersistentVolumeNodeID(p.UID())).WithLatests(map[string]string{
+		report.ControlProbeID: probeID,
```
bboreham: Why do we want this?  Won't the same volumes be reported by all probes?

probe/kubernetes/persistentvolumeclaim.go
```diff
@@ -0,0 +1,68 @@
...
+		storageClassName = *p.Spec.StorageClassName
+	}
+
+	return storageClassName
+}
+
+// GetNode returns Persistent Volume Claim as Node
+func (p *persistentVolumeClaim) GetNode(probeID string) report.Node {
+	return p.MetaNode(report.MakePersistentVolumeClaimNodeID(p.UID())).WithLatests(map[string]string{
+		report.ControlProbeID: probeID,
```
bboreham: Why do we want this?

probe/kubernetes/pod.go
```diff
@@ -75,12 +75,24 @@ func (p *pod) RestartCount() uint {
...
+	return claimName
+}
+
 func (p *pod) GetNode(probeID string) report.Node {
 	latests := map[string]string{
 		State: p.State(),
 		IP:    p.Status.PodIP,
 		report.ControlProbeID: probeID,
 		RestartCount:          strconv.FormatUint(uint64(p.RestartCount()), 10),
+		VolumeClaim:           p.VolumeClaimName(),
```
bboreham: Why maximum one VolumeClaim?
Why set it if there isn't one?

render/persistentvolume.go
```diff
@@ -0,0 +1,90 @@
...
+
+//PodToVolumeRenderer renders Pod and PVC resources
+func PodToVolumeRenderer() Renderer {
+	return podToVolumesRenderer{}
+}
+
+// VolumesRenderer is the renderer to render volumes
+type podToVolumesRenderer struct{}
+
+// Render renders the nodes
```
bboreham: This comment is not very helpful - "nodes" is a very generic term.  Say a little about what is really being done.

render/persistentvolume.go
```diff
@@ -0,0 +1,90 @@
...
+
+// VolumesRenderer returns renderer
+func VolumesRenderer() Renderer {
+	return volumesRenderer{}
+}
+
+// volumesRenderer is the renderer to render PVC & PV
+type volumesRenderer struct{}
+
+// Render renders the nodes
```
bboreham: This comment is not very helpful - "nodes" is a very generic term. Say a little about what is really being done.

render/persistentvolume.go
```diff
@@ -0,0 +1,90 @@
...
+			}
+			nodes[pvNodeID] = p
+		}
+		nodes[id] = n
+	}
+	return Nodes{Nodes: nodes}
+}
+
+//PodToVolumeRenderer renders Pod and PVC resources
+func PodToVolumeRenderer() Renderer {
```
bboreham: What is this function adding?

render/persistentvolume.go
```diff
@@ -0,0 +1,90 @@
...
+			}
+		}
+		nodes[podID] = podNode
+	}
+
+	return Nodes{Nodes: nodes}
+}
+
+// PVCToStorageClassRenderer renders PVC and Storage class objects.
+func PVCToStorageClassRenderer() Renderer {
```
bboreham: What is this function adding?

render/render.go
```diff
@@ -251,6 +251,29 @@ func (ret *joinResults) rewriteAdjacency(outID string, adjacency report.IDList)
 	ret.nodes[outID] = out
 }
 
+// storageAdjacency sets adjacency for the given node ID
+func (ret *joinResults) storageAdjacency(outID string, adjacency string) {
```
bboreham: can't see where this is used.

report/report.go
```diff
@@ -151,6 +160,18 @@ type Report struct {
...
+	// Metadata is limited for now, more to come later.
+	PersistentVolume Topology
+
+	// Persistent Volume Claim nodes represent all Kubernetes Persistent Volume Claims running on hosts running probes.
+	// Metadata is limited for now, more to come later.
+	PersistentVolumeClaim Topology
+
+	// Storage Class represent all kubernetes Storage Classes on hosts running probes.
+	// Metadata is limited for now, more to come later.
+	StorageClass Topology
```
bboreham: Do we really need three topologies to achieve the aim?  Could we just have one "storage" topology that included three kinds of nodes?

render/render.go
```diff
@@ -251,6 +251,29 @@ func (ret *joinResults) rewriteAdjacency(outID string, adjacency report.IDList)
...
+	out := ret.nodes[outID]
+	out.Adjacency = out.Adjacency.Add(adjacency)
+	ret.nodes[outID] = out
+}
+
+// storageResult returns Nodes for after adding adjacencies
+func (ret *joinResults) storageResult(input Nodes) Nodes {
```
bboreham: can't see where this is used.

report/report.go
satyamz: I am trying to bring everything related to the storage under `storage` topology.  I am giving it a shot.

satyamz: I think there'll be a problem with `shapes` of these components if we put them in the same topology.

bboreham: I haven't looked at that code, but I see other views use multiple shapes in the same view.  What actually goes wrong?

client/app/scripts/charts/edge.js
```diff
@@ -5,6 +5,20 @@ import classNames from 'classnames';
...
 
+
+function getAdjacencyClass(id) {
+  const topologyId = id.split('---');
+  const from = topologyId[0].split(';');
+  const to = topologyId[1].split(';');
+  if (from[1] !== undefined && to[1] !== undefined) {
+    from[1] = from[1].slice(1, -1);
+    to[1] = to[1].slice(1, -1);
+    if ((from[1] === 'persistent_volume' || from[1] === 'storage_class' || from[1] === 'persistent_volume_claim') || (to[1] === 'persistent_volume' || to[1] === 'storage_class' || to[1] === 'persistent_volume_claim')) {
```
foot: - This conditional is a bit long 
- assigning back into the array is confusing to read.
- Splitting into two fns might make it easier to parse!

```js
function isStorageNode(nodeId) {
  ...
}

function getAdjacencyClass(id) {
  const [ from, to ] = id.split('---');
  if (isStorageNode(from) || isStorageNode(to)) {
    ...
  }
  ...
}
```

satyamz:
@foot Please review the changes. Thanks

client/app/scripts/charts/edge.js
```diff
@@ -5,6 +5,30 @@ import classNames from 'classnames';
 import { enterEdge, leaveEdge } from '../actions/app-actions';
 import { encodeIdAttribute, decodeIdAttribute } from '../utils/dom-utils';
 
+function IsStorageComponent(id) {
```
foot: This should be `isStorageComponent `. function names are camelCase and start with a lowercase letter.

client/app/scripts/charts/edge.js
```diff
@@ -5,6 +5,30 @@ import classNames from 'classnames';
...
+// getAdjacencyClass takes id which contains information about edge as a topology
+// of parent and child node.
+
+function getAdjacencyClass(id) {
+  const topologyId = id.split('---');
+  const from = topologyId[0].split(';');
+  const to = topologyId[1].split(';');
+  if (from[1] !== undefined && to[1] !== undefined) {
+    from[1] = from[1].slice(1, -1);
+    to[1] = to[1].slice(1, -1);
```
foot: Assigning into the array here is still very confusing to me.

It would be clearer to introduce some new variables instead, something like:
```js
const fromStorageType = from[1].slice(1, -1);
const toStorageType = to[1].slice(1, -1);
```

satyamz:
I will squash commits in this PR once it's approved by all stakeholders.

client/app/scripts/charts/edge.js
```diff
@@ -5,6 +5,30 @@ import classNames from 'classnames';
...
 
+function IsStorageComponent(id) {
+  if (id === 'persitent_volume' || id === 'storage_class' || id === 'persistent_volume_claim') {
+    return true;
+  }
+  return false;
+}
+
+// getAdjacencyClass takes id which contains information about edge as a topology
+// of parent and child node.
```
foot: Nice. But please also add an example to this comment. e.g. 

`The ids take the form: "nodeA;storage_class---nodeB;persistent_volume_claim"`

satyamz: Cool! I will use this example as it is!

foot: Go for it! Is it accurate though? There is some `slice(1, -1)` logic going on too so I just wasn't sure what the edgeIds look like exactly.

satyamz: Yes Its right!

satyamz: @foot I removed the `slice` part and updated the example. Please review!

bboreham:
I just built this branch and tried it out.  Working, but I think one part of @rade's comment got misunderstood:

> Regarding having a separate view... views with pods get crowded very quickly since that is the granularity at which things get scaled in k8s. Surely a user would want to see what PVs their workloads (i.e. groups of pods managed by a controller) are accessing. So IMO PVs and PVCs should go in the Pod and Controller views, possible with a selector that lets users decide whether they want to see them.

It's in the Pod view, but with a switch that gives "only pods" or "only storage", which is basically a separate view.  If I make the following patch to `app/api_topologies.go` I get storage interspersed with pods, which is what I expect:

```
-                       {Value: "show", Label: "Show Storage", filter: render.IsVolumesComponent, filterPseudo: false},
+                       {Value: "show", Label: "Show Storage", filter: nil, filterPseudo: false},
```

Controller view can come later.

satyamz:
@bboreham If I remove the constraint on the only storage view then I get something like this:

image

And
image

Do we want it to be like this? I am bit confused here because this will make pods view visually more complicated.

umamukkara:
@satyamz , my understanding also is that "show storage" will show every pod + graph nodes related to storage. It could become bit cluttered, but with the search filter, the resulting nodes could be thinned down.

umamukkara:
May be as we add more storage related features to scope, we might need a special filter called "Show ONLY storage", but that is for future discussion I suppose

client/app/scripts/charts/edge.js
```diff
@@ -5,6 +5,28 @@ import classNames from 'classnames';
...
+// For example: id is of form "nodeA;<storage_class>---nodeB;<persistent_volume_claim>"
+function getAdjacencyClass(id) {
+  const topologyId = id.split('---');
+  const fromNode = topologyId[0].split(';');
+  const toNode = topologyId[1].split(';');
+  if (fromNode[1] !== undefined && toNode[1] !== undefined) {
```
foot: Just checking that this logic is correct:
- If we see an edge like this: `nodeA;<storage_class>---nodeB` we do _not_ want to show a line?

Otherwise I think we're good.

rndstr:
@satyamz we recently switched to sentenced cased text, would you mind changing all the strings (e.g., from "Show Storage" to "Show storage"). Thanks!

client/app/scripts/charts/edge.js
satyamz: @foot 
> If we see an edge like this: nodeA;<storage_class>---nodeB we do not want to show a line?

We're adding one more wrapper over normal line only for storage components. If it's not storage component then we're sending whatever normal line was already there. Here, if `nodeB` is Internet node then it doesn't have any topology name i.e. `<topology-name>` then we'll not add any wrapper around it i.e `link-none` for the wrapper.

satyamz:
@rndstr I made changes suggested by you on sentence cased.

probe/kubernetes/persistentvolume.go
```diff
@@ -22,16 +24,34 @@ func NewPersistentVolume(p *apiv1.PersistentVolume) PersistentVolume {
...
-		report.ControlProbeID: probeID,
-		NodeType:              "Persistent Volume",
-		VolumeClaim:           p.Spec.ClaimRef.Name,
-		StorageClassName:      p.Spec.StorageClassName,
-		Status:                string(p.Status.Phase),
-		AccessModes:           string(p.Spec.AccessModes[0]),
-		ReclaimPolicy:         string(p.Spec.PersistentVolumeReclaimPolicy),
-		Message:               p.Status.Message,
+		NodeType:         "Persistent Volume",
+		VolumeClaim:      p.GetVolume(),
```
bboreham: This can be blank; I think it would be better if we just didn't set this field in that case.

satyamz: @bboreham sure! I will fix this in subsequent PR.

tallaxes:
Note that at this time PVs and PVCs won't show up out of the box (if one follows the default install instructions) because `ClusterRole/weave-scope` does not grant access to these. `persistentvolumes`, `persistentvolumeclaims` and `storageclasses` need to be added:

```yaml
      - apiGroups:
          - ''
        resources:
          - pods/log
          - services
          - nodes
          - namespaces
          - persistentvolumes
          - persistentvolumeclaims
          - storageclasses
        verbs:
          - get
          - list
          - watch
```

satyamz:
Hi @tallaxes , I am sending PR for the same. Thanks for pointing this out. :+1:

PR #3137, opened 2018-04-11
WIP: Make Node.WithLatests() take a slice instead of a map

bboreham:
Then do the additions as a merge rather than repeated adds, to avoid memory allocations.

It appears that `WithLatests` is not something we do a lot in the benchmarks, but probably more in the probes. The follow-up change to `propagateLatest` gives us a healthy boost when rendering.

~Also a drive-by fix to `processWithContainerNameRenderer`.~

Benchmarks:
```
benchmark                            old ns/op      new ns/op      delta
BenchmarkRenderList-2                665307761      527805245      -20.67%
BenchmarkRenderHosts-2               413642784      408523260      -1.24%
BenchmarkRenderControllers-2         364129636      344651406      -5.35%
BenchmarkRenderPods-2                341522570      314960182      -7.78%
BenchmarkRenderContainers-2          204275436      193586083      -5.23%
BenchmarkRenderProcesses-2           129523723      122403160      -5.50%
BenchmarkRenderProcessNames-2        143726808      138044775      -3.95%

benchmark                            old allocs     new allocs     delta
BenchmarkRenderList-2                801654         803633         +0.25%
BenchmarkRenderHosts-2               605887         605869         -0.00%
BenchmarkRenderControllers-2         429942         429789         -0.04%
BenchmarkRenderPods-2                383866         385131         +0.33%
BenchmarkRenderContainers-2          253898         255118         +0.48%
BenchmarkRenderProcesses-2           119585         120188         +0.50%
BenchmarkRenderProcessNames-2        163417         163872         +0.28%

benchmark                            old bytes     new bytes     delta
BenchmarkRenderList-2                91623684      88391116      -3.53%
BenchmarkRenderHosts-2               65873664      62455445      -5.19%
BenchmarkRenderControllers-2         47176061      43729286      -7.31%
BenchmarkRenderPods-2                41874317      38607350      -7.80%
BenchmarkRenderContainers-2          28799258      25524452      -11.37%
BenchmarkRenderProcesses-2           13984811      14002312      +0.13%
BenchmarkRenderProcessNames-2        17930731      17981694      +0.28%
```

### DIFF
--- a/report/map_helpers.go
+++ b/report/map_helpers.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"sort"
+	"time"
 
 	"github.com/ugorji/go/codec"
 	"github.com/weaveworks/ps"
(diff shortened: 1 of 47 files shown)


report/map_helpers.go
```diff
@@ -138,3 +139,58 @@ func mapWrite(m ps.Map, encoder *codec.Encoder, encodeValue func(*codec.Encoder,
...
+		switch {
+		case j >= len(n) || m[i].key < n[j]:
+			out = append(out, m[i])
+			i++
+		case m[i].key == n[j]:
+			if m[i].Timestamp.Before(t) {
+				out = append(out, stringLatestEntry{key: n[j], Value: n[j+1], Timestamp: t})
+			} else {
+				out = append(out, m[i])
+			}
```
rade: I suggest inverting that condition. That way all the `m` appends come before the `n` appends in the code.

render/process.go
```diff
@@ -46,7 +46,7 @@ func (r processWithContainerNameRenderer) Render(rpt report.Report) Nodes {
 		if !ok {
 			continue
 		}
-		propagateLatests(container, p, docker.ContainerName)
+		p = propagateLatests(container, p, docker.ContainerName)
```
rade: That was a straight bug. How does it manifest?

bboreham:
We don't know how small the change to probe performance is; we don't have any benchmarks.
We could try it out...

The readability change comes from gofmt aligning map entries and not parameters, right?

rade:
> We don't know how small the change to probe performance is; we don't have any benchmarks.

True.

> We could try it out...

If there is a significant performance improvement, I would be swayed.

> The readability change comes from gofmt aligning map entries and not parameters, right?

Yes. And also from the fact that the representation of an association as a flat list (there is a name for that, which I've forgotten) is rather non-idiomatic golang.

2opremio:
@bboreham Are you planning to finish up this work anytime soon? Otherwise I would like to close it (I am cleaning up the lingering PRs)

render/process.go
bboreham: The absence of any ContainerName on Process nodes, I think.

bboreham:
Based on profiling some real probes, I wrote a benchmark to mimic typical operation of this code when reporting on Kubernetes objects.

Results before (best of 5):
```
BenchmarkReporter 	     500	   2671920 ns/op	  841966 B/op	    9290 allocs/op
```

After:
```
BenchmarkReporter 	     500	   2260789 ns/op	  516960 B/op	    9142 allocs/op
```

bboreham:
Replaced by #3268

Issue #3131, opened 2018-04-05
eBPF tracking causes panic on Ubuntu kernel 4.4.0-119


leth:
Full kernel version: `4.4.0-119-generic #143-Ubuntu SMP`
We were previously running `4.4.0-116-generic #140-Ubuntu SMP` with no problems

## Workaround:
Disable eBPF connection tracking with `--probe.ebpf.connections=false`

## Panic details
[panic.txt](https://github.com/weaveworks/scope/files/1879867/panic.txt)

dlespiau:
It'd be interesting to also have to known to work version so one can look at the list of commits that found their way in.

rade:
@alban @iaguis any clues what might be going wrong here?

alban:
Summary of the stack:
```
- bpf_map_lookup_elem
- bpf_prog_run
- fd_install
- do_sys_open
```

So the BPF program tries to read a eBPF map because of a open() syscall in the `scope` process:
https://github.com/weaveworks/tcptracer-bpf/blob/fecaba5/tcptracer-bpf.c#L873
```
	exists = bpf_map_lookup_elem(&fdinstall_pids, &tgid);
```

The message "BUG: unable to handle kernel paging request at 0000000063222428" might be an invalid pointer in the `bpf_map_lookup_elem()` function:
https://github.com/torvalds/linux/blob/v4.4/kernel/bpf/helpers.c#L41
```
	value = map->ops->map_lookup_elem(map, key);
```

Maybe `map` or `map->ops` is an invalid pointer? No idea why that would happen right now... is it 100% reproducible?

Given the timestamps, it seems to happen very soon after the machine boots, so probably during the initialization of Scope.

rade:
> Given the timestamps, it seems to happen very soon after the machine boots, so probably during the initialization of Scope.

Yes. We've had a couple of machines in a k8s cluster in reboot loops since updating to the `4.4.0-119-generic #143-Ubuntu SMP` kernel.

alban:
I see one [patch](https://launchpadlibrarian.net/363331633/linux-lts-xenial_4.4.0-116.140~14.04.1_4.4.0-119.143~14.04.1.diff.gz) on bpf maps in that new kernel (found from [here](https://www.ubuntuupdates.org/package/core/trusty/main/updates/linux-image-4.4.0-119-generic)):
https://pastebin.ubuntu.com/p/y58KDXwTMn/

[...]

PR #3141, opened 2018-04-13
Workaround Ubuntu kernel oops

schu:
Workaround for https://github.com/weaveworks/scope/issues/3131 until the issue is fixed upstream.

Ubuntu bug report: https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1763454

### DIFF
--- a/probe/endpoint/ebpf.go
+++ b/probe/endpoint/ebpf.go
@@ -53,16 +53,28 @@ type EbpfTracker struct {
 	closedDuringInit  map[fourTuple]struct{}
 }
 
-var releaseRegex = regexp.MustCompile(`^(\d+)\.(\d+).*$`)
+// releaseRegex should match all possible variations of a common Linux
+// version string:
+//   - 4.1
+//   - 4.22-foo
+//   - 4.1.2-foo
+//   - 4.1.2-33.44+bar
+//   - etc.
+// For example, on a Ubuntu system the vendor specific release part
+// (after the first `-`) could look like:
+// '<ABI number>.<upload number>-<flavour>' or
+// '<ABI number>-<flavour>'
+// See https://wiki.ubuntu.com/Kernel/FAQ
+var releaseRegex = regexp.MustCompile(`^(\d+)\.(\d+)\.?(\d*)-?(\d*)(.*)$`)
 
 func isKernelSupported() error {
-	release, _, err := host.GetKernelReleaseAndVersion()
+	release, version, err := host.GetKernelReleaseAndVersion()
 	if err != nil {
 		return err
 	}
 
 	releaseParts := releaseRegex.FindStringSubmatch(release)
-	if len(releaseParts) != 3 {
+	if len(releaseParts) != 6 {
 		return fmt.Errorf("got invalid release version %q (expected format '4.4[.2-1]')", release)
 	}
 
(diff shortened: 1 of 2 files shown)


probe/endpoint/ebpf.go
```diff
@@ -84,6 +84,31 @@ func isKernelSupported() error {
...
+		// cut away upload number
+		abiNumberStr := strings.Split(vendorStr, ".")[0]
+		abiNumber, err := strconv.Atoi(abiNumberStr)
+		if err != nil {
+			return err
+		}
+		// TODO: give the check an upper limit once the bug is fixed
+		if abiNumber >= 119 {
+			// https://github.com/weaveworks/scope/issues/3131
+			// https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1763454
```
alban: I'm guessing the fix in the Ubuntu kernel is to be released soon. Should we wait until it is released, so we know the exact abiNumber range where it's broken?

probe/endpoint/ebpf.go
```diff
@@ -84,6 +84,31 @@ func isKernelSupported() error {
...
+
+		// The vendor specific release part could look like:
+		// '<ABI number>.<upload number>-<flavour>' or
+		// '<ABI number>-<flavour>'
+		// See https://wiki.ubuntu.com/Kernel/FAQ
+
+		// cut away flavour
+		vendorStr := strings.Split(releaseParts[4], "-")[0]
+		// cut away upload number
+		abiNumberStr := strings.Split(vendorStr, ".")[0]
```
bboreham: I had to read the docs for `Split` to see what it did if the separator isn't found.
Why not just include the pattern for the ABI number in `releaseRegex`, i.e. replace `(.*)$` with `(\d*)(.*)$` ?

probe/endpoint/ebpf_test.go
```diff
@@ -233,3 +235,59 @@ func TestInvalidTimeStampDead(t *testing.T) {
...
+		{
+			"4.4.0-119-generic",
+			"#143-Ubuntu SMP Mon Apr 2 16:08:24 UTC 2018",
+			false,
+		},
+		{
+			"4.4.0-116-generic",
+			"#140-Ubuntu SMP Mon Feb 12 21:23:04 UTC 2018",
+			true,
+		},
```
bboreham: The tests are all for older Ubuntus; can I suggest adding some others? For example:
* The newest Ubuntu in the list on GCP - `4.13.0-38-generic #43-Ubuntu SMP Wed Mar 14 15:20:44 UTC 2018`
* A newer Ubuntu with a 119 abi - `4.13.0-119-generic #43-Ubuntu SMP Wed Apr 1 00:00:00 UTC 2018`
* Default Debian on GCP - `4.9.0-6-amd64 #1 SMP Debian 4.9.82-1+deb9u3 (2018-03-02)`

probe/endpoint/ebpf.go
rade: > I'm guessing the fix in the Ubuntu kernel is to be released soon.

Any idea how long that might take?

Frankly, if there is going to be a new LTS security-updates kernel release with the fix in the next few days, I am not sure it's even worth bothering having any check here at all.

probe/endpoint/ebpf_test.go
schu: Added, plus tightened the check in `isKernelSupported` to only match minor version `4`. Thanks @bboreham

probe/endpoint/ebpf.go
schu: >> I'm guessing the fix in the Ubuntu kernel is to be released soon.

> Any idea how long that might take?

We can't tell and have asked Seth (Ubuntu) if he can give an ETA.

> Frankly, if there is going to be a new LTS security-updates kernel release with the fix in the next few days, I am not sure it's even worth bothering having any check here at all.

I agree.

schu: Update: approximately around 4 weeks for the update to be published, https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1763454/comments/6

probe/endpoint/ebpf.go
```diff
@@ -84,6 +96,22 @@ func isKernelSupported() error {
 		return fmt.Errorf("got kernel %s but need kernel >=4.4", release)
 	}
 
+	if strings.Contains(version, "Ubuntu") {
+		// Check for specific Ubuntu kernel versions with
+		// known issues.
+
+		abiNumber, err := strconv.Atoi(releaseParts[4])
+		if err != nil {
+			return err
```
bboreham: I think this should return nil, on the basis it's at least 4.4 and we don't know that it is 119-ish, so just allow it.

schu: Updated, thanks.

Issue #3153, opened 2018-05-02
scope: k8s deployment: clusterrole: reducing permissions 


tasdikrahman:
the initial yaml borrowed from https://cloud.weave.works/k8s/scope.yaml?k8s-version=1.9.3

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: weave-scope
  labels:
    name: weave-scope
  namespace: weave
rules:
  - apiGroups:
      - '*'
    resources:
      - '*'
    verbs:
      - '*'
  - nonResourceURLs:
      - '*'
    verbs:
      - '*'
```

as discussed with @bboreham on slack [here](https://weave-community.slack.com/archives/C985DLV6F/p1523435369000321), we can reduce the number of perms given to it. 

should I add the yaml file with the changes I made to get it work here itself? If yes, please suggest the directory structuring I should follow. I was thinking putting it in 

```
scope/examples/k8s/
```

 Thanks! :)

bboreham:
I think `scope/examples/k8s/` is fine.

See also #2149 where a more restrictive config has been proposed.

PR #3155, opened 2018-05-02
examples: k8s: added manifest for scope deployment

tasdikrahman:
- closes https://github.com/weaveworks/scope/issues/3153


tasdikrahman:
@bboreham gentle poke, let me know if any changes are required on this PR :)

bboreham:
Sorry for delay.

I'm going to merge this, as it is better than we had, but it looks like the permissions are all read-only so various buttons in Scope will not function (like 'scale up').  The settings generated in  https://cloud.weave.works/k8s/scope.yaml now cover that.

PR #3156, opened 2018-05-03
Add null check to fix querier panic

leth:
I'm not sure what could cause this right now.


bboreham:
Bug was introduced in #3143

PR #3157, opened 2018-05-03
Lower SQS RPC timeout from 1m to 25s

leth:


bboreham:
The description of this PR should include why it was done, so it can be understood without reference to a private repo.

PR #3165, opened 2018-05-09
 Use the default value for a TopologyOption if omitted

bboreham:
Without this, we see unmanaged, uncontained or not-running containers in the graph until the front-end requests them to go away.

There was a shouty comment saying not to use the default, but I can't see any reason to take that stance. There is only one place where `""` is a valid value, which is namespaces, and it is also the default there.

To be safe, I changed the code to distinguish a lack of value from `""` so you can have both.


foot:
This seems like a reasonable change in behaviour, but the scope-ui always fills in the defaults provided by `/api/topologies` before requesting the nodes so I don't think this will fix the linked issue.

bboreham:
Where “always” does not cover 80% of the times I open Scope.

Issue #3163, opened 2018-05-09
Make navigation more clear


LiliC:
As discussed offline, we should try to make it more clear and intuitive that the sorting below are not subcategories of containers, processes and hosts (e.g.: I thought that I could sort "by name" of containers as well):
<img width="492" alt="screen shot 2018-05-09 at 11 36 12 am" src="https://user-images.githubusercontent.com/6232346/39807746-6211e988-537d-11e8-8767-0d868b6eecb0.png">

(Also weave net, should be one level above.)

Really liked @bia mockups:
![screen shot 2018-05-09 at 11 34 56](https://user-images.githubusercontent.com/6232346/39807961-f8884416-537d-11e8-9495-3dc3d102ce32.png)
![screen shot 2018-05-09 at 11 34 47](https://user-images.githubusercontent.com/6232346/39807962-f8a3d0e6-537d-11e8-80e0-3adc77f06910.png)

cc @fbarl

PR #3166, opened 2018-05-10
Sentence cased text everywhere

rndstr:
Follows Weave Cloud's direction of sentence case on most things.

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -43,8 +43,8 @@ var (
 		ID:      "pseudo",
 		Default: "hide",
 		Options: []APITopologyOption{
-			{Value: "show", Label: "Show Unmanaged", filter: nil, filterPseudo: false},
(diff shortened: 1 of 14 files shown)


bowenli:
I think keeping the "by foo" makes sense since it describes the grouping.

bia:
LmostlyGTM
Some spacing issues become even more prominent in sentence case.
screen shot 2018-05-11 at 12 04 53
I talked with @fbarl about it - the simplest solution would be to add about 5px more css padding around that text.

I am unable to `make` scope, so I'm not able to preview this properly. Will see if I can fix that today.

>I think keeping the "by foo" makes sense since it describes the grouping.

We're addressing this in #3163

rndstr:
@bia just fixed the alignment and adjusted the padding.
scope-table-header

bia:
Cool, anyways that looks better. Maybe a little more space between the arrow and 'Processes'?

PR #3173, opened 2018-05-15
Show image tag in node details

### LINKED ISSUE #3161: Easily see tag and with that the version of an image
Currently to see a tag of an image I would need to hover over the image name or copy and paste it somewhere. Would be nice to see it separately below the image name, so maybe something like:

```
image name: k8s.gcr.io/k8s-dns-sidecar-amd64
image tag: 1.14.5
```

Currently this is how it looks like and you would need to hover to see the tag:
screen shot 2018-05-08 at 6 28 07 pm

cc @fbarl

fbarl:
Resolves #3161.

image

image

### DIFF
--- a/probe/docker/registry.go
+++ b/probe/docker/registry.go
@@ -472,13 +472,21 @@ func (r *registry) WalkNetworks(f func(docker_client.Network)) {
 	}
 }
 
-// ImageNameWithoutVersion splits the image name apart, returning the name
-// without the version, if possible
-func ImageNameWithoutVersion(name string) string {
+func ImageNameParts(name string) []string {
 	parts := strings.SplitN(name, "/", 3)
 	if len(parts) == 3 {
 		name = fmt.Sprintf("%s/%s", parts[1], parts[2])
 	}
-	parts = strings.SplitN(name, ":", 2)
-	return parts[0]
+	return strings.SplitN(name, ":", 2)
+}
+
+// ImageNameWithoutVersion splits the image name apart, returning the name
+// without the version, if possible
+func ImageNameWithoutVersion(name string) string {
+	return ImageNameParts(name)[0]
+}
+
+// ImageNameVersion splits the image name apart, returning the version, if possible
+func ImageNameVersion(name string) string {
+	return ImageNameParts(name)[1]
 }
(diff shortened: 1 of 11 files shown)


probe/docker/registry.go
```diff
@@ -472,13 +472,26 @@ func (r *registry) WalkNetworks(f func(docker_client.Network)) {
 	}
 }
 
-// ImageNameWithoutVersion splits the image name apart, returning the name
-// without the version, if possible
-func ImageNameWithoutVersion(name string) string {
+// ImageNameParts returns parts of the full image name (image name, image tag).
```
LiliC: nit: I understand that these functions were here before, but it makes me think if `container.go` would not be a better place for them, WDYT?

probe/docker/registry.go
```diff
@@ -472,13 +472,26 @@ func (r *registry) WalkNetworks(f func(docker_client.Network)) {
...
 		name = fmt.Sprintf("%s/%s", parts[1], parts[2])
 	}
-	parts = strings.SplitN(name, ":", 2)
-	return parts[0]
+	return strings.SplitN(name, ":", 2)
+}
+
+// ImageNameWithoutVersion splits the image name apart, returning the name
+// without the version, if possible
+func ImageNameWithoutVersion(name string) string {
```
LiliC: Also I would keep the naming to be always "Tag" rather then Version, to avoid confusion. So `ImageName` and `ImageNameNoTag` for example, WDYT?

fbarl:
@LiliC PTAL.

probe/docker/reporter.go
```diff
@@ -56,8 +58,10 @@ var (
 			Label: "Image",
 			Type:  report.PropertyListType,
 			FixedRows: map[string]string{
-				ImageID:          "ID",
-				ImageName:        "Name",
+				// Prepend spaces as a hack to keep at the top when sorted.
+				ImageID:          " ID",
+				ImageName:        " Name",
```
LiliC: Not sure if per the OCI spec, it might be Image repo instead of name? Didn't have a chance to have a look.

dlespiau: I'd go for name, even if we can have $repo/$name sometimes here. There are also special cases for well known images (`busybox`, `golang`, ...).

PR #3177, opened 2018-05-17
Enforce theme fonts

fbarl:
Scope part of https://github.com/weaveworks/ui-components/issues/212.

Uses _Proxima Nova_ as a default font instead of _Roboto_.

Not to be merged before https://github.com/weaveworks/service-ui/pull/2551.


fbarl:
> Terminal is still using courier-new

I added a commit that changes that :) Unfortunately, overriding the default font family didn't really work so I needed to hack the css styles..

> Do we already have a license for proxima-nova?

I think @bowenli is looking into that right now - in any case, it's a problem for a future PR.

PR #3179, opened 2018-05-18
Adjusted terminal character width/height estimation

### LINKED ISSUE #3178: Terminal window width wrong
The character window is about half the width of the UI window, and a little less tall.

Maybe because it's a Retina display?  But why wouldn't it be the same wrong in both dimensions?

image

fbarl:
Fixes #3178.

### DIFF
--- a/client/app/scripts/components/terminal.js
+++ b/client/app/scripts/components/terminal.js
@@ -5,6 +5,7 @@ import { connect } from 'react-redux';
 import classNames from 'classnames';
 import { debounce } from 'lodash';
 import { Terminal as Term } from 'xterm';
+import * as fit from 'xterm/lib/addons/fit/fit';
 
 import { closeTerminal } from '../actions/app-actions';
 import { getNeutralColor } from '../utils/color-utils';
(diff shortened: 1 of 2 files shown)


dlespiau:
`character{Width,Height}` are only used if `wrapperNode.querySelector('.terminal .xterm-rows div');` doesn't return anything? does the selector need changing?

fbarl:
@dlespiau Good catch! Since `v3`, `xtermj.js` is doing canvas rendering so we can't probe into DOM anymore (or at least not the same way). I suggest we simply get rid of that selector and keep only the constants.

p.s. in #3177 I'll try to enforce `Roboto Mono` font in terminal for all users so those constants won't be changing anymore with future updates.

dlespiau:
Given the variety of screen DPIs and that those constants seem to be in pixels, the terminal size won't be correct almost everytime, no?

Instead of fixing the terminal width when creating the object, it seems like we my be able to leave the terminal occupy its natural allocation on the page and listen to the `resize` event, `term.on('resize', ....)` and send the resize commands on the `resize` callback.

dlespiau:
Oh! there's a `fit` addon that would help here: https://xtermjs.org/docs/api/addons/fit/

Fit the terminal to a parent element and then query its size with `cols()` and `rows()`.

fbarl:
Thanks for looking into it @dlespiau. I implemented your `fit` addon suggestion, feel free to take a look at the code and test it now :)

The code there is still pretty hacky, but hopefully now it's a bit more precise.

Issue #2914, opened 2017-10-30
No data when clicking on a service CPU/Mem graph and the service name is not the container name


dlespiau:
How to reproduce
- Select the `scope` service in the `kube-system` ns (WC)
- Click on memory or CPU usage to open the query in Monitor
- Graph show "No data"

Query:
```
namespace_label_name:container_cpu_usage_seconds_total:sum_rate{label_name="scope",namespace="kube-system"}
```

Recording rule:
```
    namespace_label_name:container_cpu_usage_seconds_total:sum_rate =
       sum by (namespace, label_name) (
          label_replace(sum(rate(container_cpu_usage_seconds_total{image!=""}[5m])) by (pod_name, namespace), "pod", "$1", "pod_name", "(.*)")
        * on (pod) group_left(label_name)
          kube_pod_labels{job="monitoring/kube-state-metrics"}
       )
```

The service name is `scope`. `kube_pod_labels` has `scope-app` as `label_name`, which is the name of the deployment and the container(s). Guessing `kube_pod_label` has the container name, not the service name. There's another interesting metric with the service name: `kube_service_labels`.

dlespiau:
Tested it on sock shop and the resulting promQL query is: `sum(rate(container_cpu_usage_seconds_total{image!="",namespace="weave",_weave_pod_name="weave-flux-memcached",job="cadvisor",container_name!="POD"}[5m]))`, which displays data. Closing!

PR #3184, opened 2018-05-21
render: The cAdvisor job is called kubernetes-cadvisor

dlespiau:
That's the job name we ship to our customers and we recently updated weave
cloud to use the same job name.

Fixes: #2914


dlespiau:
Ah, no, I got the it wrong was around. The correct job name is cadvisor, we need to renamed it in our WC manifests: https://github.com/weaveworks/launch-generator/blob/4835e062b0cc60686bd1547bfa94f681b48bc6f2/src/cortex.js#L280

So #2914 is fixed, it just doesn't work for our Weave Cloud instance. Will fix it there.

PR #3189, opened 2018-05-24
Fixes sending the default topologyOptions to the /api

foot:
- Which fails sometimes as we overwrite them whatever we find in the
  url state. Url state is not complete anymore so this doesn't work

Test with:
```
cd $WEAVE/scope/client
mkdir tmp
yarn bundle
cd $WEAVE/service-ui/client
yarn add $WEAVE/scope/client/tmp/weave-scope.tgz
SERVICE_HOST=frontend.dev.weave.works yarn start
```

### DIFF
--- a/client/app/scripts/reducers/root.js
+++ b/client/app/scripts/reducers/root.js
@@ -688,7 +688,8 @@ export function rootReducer(state = initialState, action) {
         pinnedMetricType: action.state.pinnedMetricType,
       });
       if (action.state.topologyOptions) {
-        state = state.set('topologyOptions', fromJS(action.state.topologyOptions));
+        const options = getDefaultTopologyOptions(state).mergeDeep(action.state.topologyOptions);
+        state = state.set('topologyOptions', options);
       }
       if (action.state.topologyViewMode) {
         state = state.set('topologyViewMode', action.state.topologyViewMode);


rade:
Is this a better fix to #1999 than #3165?

client/app/scripts/reducers/root.js
```diff
@@ -688,7 +688,7 @@ export function rootReducer(state = initialState, action) {
         pinnedMetricType: action.state.pinnedMetricType,
       });
       if (action.state.topologyOptions) {
-        state = state.set('topologyOptions', fromJS(action.state.topologyOptions));
+        state = state.mergeDeep(fromJS({ topologyOptions: action.state.topologyOptions }));
```
fbarl: As @foot and I discussed, it would be safer to also merge `action.state.topologyOptions` with `getDefaultTopologyOptions(state)` in here to make sure everything works fine if we switch from non-default to default topology options (otherwise `action.state.topologyOptions` would be empty and it wouldn't overwrite the current state).

p.s. Turns out the real issue here was not taking the default options into account in the first place, once they are merged with the URL state, they provide a complete description of topology options state, so it shouldn't matter whether we use `set` or `mergeDeep` :)

foot: Yeah, turns out that action is only fired onScopeLoad rather than any subsequent ui actions, which is sad. But I will update as discussed anyway in case that changes in the future! 👍

PR #3201, opened 2018-05-29
Don't return blank addresses for pods

bboreham:
A small performance improvement - instead of creating an entry in a map which doesn't match anything, just skip a pod with a blank IP.

This was sitting as part of #3089 which we're not going to merge.

### DIFF
--- a/render/pod.go
+++ b/render/pod.go
@@ -116,7 +116,7 @@ func MapPod2IP(m report.Node) []string {
 	}
 
 	ip, ok := m.Latest.Lookup(kubernetes.IP)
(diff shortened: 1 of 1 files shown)


2opremio:
@bboreham we now have a commit format policy borrowed from CNI, can you please try to comply with it? https://github.com/weaveworks/scope/blob/master/CONTRIBUTING.md#format-of-the-commit-message

We can discuss whether that's valuable or not. I think it's good to have some consistency.

bboreham:
@2opremio amended the commit message

PR #3204, opened 2018-05-31
Add option to print probe reports to stdout, for debugging

bboreham:
It's fairly common that, while debugging, I need to check whether the probe sent the wrong thing or the app rendered the wrong thing.  This option gives a crude way to view the raw probe data as json in the container logs.

It wasn't immediately obvious how to add it, so I removed two layers of cruft from the app client, and then it was easy.

A small fix-up to the `fixprobe` utility was also needed.


prog/main.go
```diff
@@ -274,6 +275,7 @@ func setupFlags(flags *flags) {
 	flag.Bool("app-only", false, "Only run the app.")
 
 	// Probe flags
+	flag.BoolVar(&flags.probe.printOnStdout, "probe.print-on-stdout", false, "Print reports on stdout instead of sending to app, for debugging")
```
rade: `probe.publish.stdout`? (we already have a `probe.publish.interval`).

prog/probe.go
```diff
@@ -86,6 +87,14 @@ func maybeExportProfileData(flags probeFlags) {
...
 }
 
+type stdoutPublisher struct{}
+
+func (stdoutPublisher) Publish(rep report.Report) error {
+	handle := &codec.JsonHandle{Indent: 2}
+	handle.Canonical = true
+	return codec.NewEncoder(os.Stdout, handle).Encode(rep)
+}
+
```
rade: this really should live in `report/marshal.go`

prog/probe.go
```diff
@@ -148,7 +157,14 @@ func probeMain(flags probeFlags, targets []appclient.Target) {
 	}
 	defer resolver.Stop()
 
-	p := probe.New(flags.spyInterval, flags.publishInterval, clients, flags.noControls)
+	var publisher probe.ReportPublisher
+	if flags.printOnStdout {
+		publisher = new(stdoutPublisher)
+	} else {
+		publisher = clients
+	}
```
rade: This is quite hacky... the fact that we are still creating the client earlier on, even though we aren't publishing to it. I cannot think of an easy way around that though since various reporters need a `controls.PipeClient`. Perhaps we could feed in a noop implementation?

prog/main.go
rade: also, perhaps the flags processing code should warn/fail when this is set and the computed `targets` is non-empty.

probe/appclient/multi_client.go
```diff
@@ -164,26 +156,25 @@ func (c *multiClient) Stop() {
...
 
-	if len(c.clients) <= 1 { // optimisation
-		for _, c := range c.clients {
-			return c.Publish(r, shortcut)
-		}
-		return nil
-	}
-
-	buf, err := ioutil.ReadAll(r)
+	buf, err := serializeReport(r)
```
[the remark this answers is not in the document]
bboreham: Previously we always serialised into a temporary buffer then sometimes read back that buffer into another buffer in order to write to multiple recipients.
We have an interface mismatch in that `http.Request` needs an `io.Reader`, whereas the encoders work with `io.Writer`, which is solved by the buffer in between. Maybe we could avoid the buffer by running the writer on a different goroutine, but it seems over-complicated.

extras/fixprobe/main.go
```diff
@@ -60,8 +62,12 @@ func main() {
 		log.Fatal(err)
 	}
 
-	rp := appclient.NewReportPublisher(client, false)
+	buf := &bytes.Buffer{}
+	err = fixedReport.WriteBinary(buf, gzip.DefaultCompression)
```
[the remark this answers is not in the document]
bboreham: Outside of tests we never call `WriteBinary` with anything other than `gzip.DefaultCompression`, so it can just be removed.

bboreham:
I have made all the suggested changes, with the exception of changing the commit message to clarify the removed "optimisation".  I should squash commits so can fix the message then.

bboreham:
Squashed and rebased

PR #3207, opened 2018-06-02
Fetch container IPs directly from the namespace instead of calling 'weave ps'

### LINKED ISSUE #2586: no connections between containers when `probe.processes=false`

bboreham:
Instead of shelling out to `weave ps`, we do what it was doing - enter the container's namespace and report IPs found there.  And whereas `weave ps` took great pains to report only Weave Net addresses, in Scope we want all addresses, so the code is simpler.

Fixes #2586
Will probably improve #3082

We can't remove the `weave` script or `weaveutil` program from the container as the App is still using them to do `weave expose`.


probe/docker/container.go
```diff
@@ -277,6 +286,18 @@ func (c *container) NetworkInfo(localAddrs []net.IP) report.Sets {
...
 	}
 
+	// Fetch IP addresses from the container's namespace
+	cidrs, err := namespaceIPAddresses(c.container.State.Pid)
+	if err != nil {
+		log.Debugf("container %s: failed to get addresses: %s", c.container.ID, err)
+	}
+	for _, cidr := range cidrs {
+		ip := cidr.IP.String()
+		if !contains(ips, ip) {
```
rade: we don't need this because we are turning `ips` into a set later on, which will get rid of the duplicates.

PR #3212, opened 2018-06-04
Render: add ContainerName into output node when rendering processes

bboreham:
This was a bug - we spend eight lines fetching the name, then didn't assign the new node into the variable.

Extracted from #3137

### DIFF
--- a/render/process.go
+++ b/render/process.go
@@ -44,7 +44,7 @@ func (r processWithContainerNameRenderer) Render(rpt report.Report) Nodes {
 		if !ok {
 			continue
 		}
-		propagateLatest(docker.ContainerName, container, p)
+		p = propagateLatest(docker.ContainerName, container, p)
 		outputs[id] = p
 	}
 	return Nodes{Nodes: outputs, Filtered: processes.Filtered}


rade:
What user-visible change does this entail? A 'Container Name' item in the 'Info' section of a process detail panel? Seems kinda superfluous - we already have a link to the container in the parents section. Also, we don't have an equivalent field for the container-to-pod relationship.

bboreham:
The feature was added in response to a user request - #331.
I noticed there is also a parent relationship, but I haven't looked into how that gets there.

I guess since it's been broken for five months we could just delete the label (and rename `ProcessWithContainerNameRenderer`).

rade:
I see; the container name is incorporated in the minor label. Usually it will be hidden and only visible on hover, since the host name, which precedes it in the label, takes up all the space. In fact it looks like there is a rendering bug and the minor label isn't rendered fully even on hover. Anyway, I'd be in favour of ditching the container name from the metadata and hence label.

2opremio:
Closing in favor of https://github.com/weaveworks/scope/pull/3263

PR #3215, opened 2018-06-05
Probe: remove backwards-compatibility code when publishing reports

bboreham:
Removed to reduce CPU and memory usage in probes.

This code was added [in August 2016](https://github.com/weaveworks/scope/pull/1682) so that newer probes could be used with older apps. Since then we have adopted the stance that new apps will accept reports from old probes but not vice-versa, on a version change.

### DIFF
--- a/probe/probe.go
+++ b/probe/probe.go
@@ -200,7 +200,7 @@ ForLoop:
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


umamukkara:
LGTM

bboreham:
Thanks for the review, @umamukkara

PR #3222, opened 2018-06-12
Show deployments in Time Travel

fbarl:
Resolves https://github.com/weaveworks/service-ui/issues/2614 (affects only Scope in Weave Cloud).

Reuses a lot of code from `<TimeTravelWrapper />` in Service UI - @foot and I had a discussion about converging those two, but that's a separate issue.

##### Other changes

* Bump `ui-components` `v0.4.82` -> `v0.4.95`
* Use `z-index: $layer-front` on the whole header

### DIFF
--- a/client/app/scripts/components/time-travel-wrapper.js
+++ b/client/app/scripts/components/time-travel-wrapper.js
@@ -2,25 +2,14 @@ import React from 'react';
 import moment from 'moment';
 import { connect } from 'react-redux';
 import { TimeTravel } from 'weaveworks-ui-components';
+import { get, orderBy } from 'lodash';
 
 import { trackAnalyticsEvent } from '../utils/tracking-utils';
 import { jumpToTime, resumeTime, pauseTimeAtNow } from '../actions/app-actions';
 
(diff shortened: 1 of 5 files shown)


client/app/scripts/components/time-travel-wrapper.js
```diff
@@ -91,27 +135,36 @@ class TimeTravelWrapper extends React.Component {
 }
 
 function mapStateToProps(state, { params }) {
-  const scopeState = state.scope || state;
```
foot: This is not important anymore?

fbarl: Nope, `<TimeTravelWrapper />` is now only used in Weave Cloud context, so it's always `state.scope`.

fbarl:
> It doesn't filter by namespace...

Related to https://github.com/weaveworks/service-ui/issues/1358.

> ...and the deployment ticks don't match the color of the deployment they represent

Not sure how to address that at the moment, but it should probably go into a separate issue :)

PR #3227, opened 2018-06-14
Use new theme gray colors

fbarl:
Scope part of https://github.com/weaveworks/ui-components/issues/260.


bia:
Table contrast colour is too light : 
screen shot 2018-06-14 at 15 08 38

Perhaps it could just be matched to the bg of 'Processes'
screen shot 2018-06-14 at 15 08 19

PR #3231, opened 2018-06-20
[Power Support] Enable weaveworks scope on Power.

meghalidhoble:
Observed this particular test  was failing on ppc64le architecture. Added the test failure log below, 
  walker_linux_test.go:97:
```                --- want
                +++ have
                @@ -32,7 +32,7 @@
                   Cmdline: (string) (len=15) "curl google.com",
                   Threads: (int) 1,
                   Jiffies: (uint64) 0,
                -  RSSBytes: (uint64) 8192,
                +  RSSBytes: (uint64) 131072,
                   RSSBytesLimit: (uint64) 2048,
                   OpenFilesCount: (int) 3,
                   OpenFilesLimit: (uint64) 32768,
```
Hence modified the value for RSSBytes to accept variable value which is computed based on pageSize instead of hard-coded values, as suggested by Adam Harrison over slack channel. 

I have verified this patch on x86 setup as well and all the tests pass there.  Please provide your comments/suggestions on this. 
Thanks,
Meghali

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,14 +1,30 @@
-FROM golang:1.10.2-stretch
+FROM golang:1.10.2
 ENV SCOPE_SKIP_UI_ASSETS true
-RUN apt-get update && \
-	apt-get install -y libpcap-dev time file shellcheck git gcc-arm-linux-gnueabihf curl build-essential python-pip && \
-	rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+RUN set -eux; \
+   export arch_val="$(dpkg --print-architecture)"; \
+   apt-get update && \
+   if [ "$arch_val" = "amd64" ]; then \
+     apt-get install -y libpcap-dev python-requests time file shellcheck git gcc-arm-linux-gnueabihf curl build-essential python-pip; \
+   else \
+     apt-get install -y libpcap-dev python-requests time file shellcheck git curl build-essential python-pip; \
+   fi; \
+   \
+   rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
 RUN go clean -i net && \
 	go install -tags netgo std && \
-	go install -race -tags netgo std
-RUN curl -fsSL -o shfmt https://github.com/mvdan/sh/releases/download/v1.3.0/shfmt_v1.3.0_linux_amd64 && \
-    chmod +x shfmt && \
-    mv shfmt /usr/bin
+   export arch_val="$(dpkg --print-architecture)"; \
+   if [ "$arch_val" != "ppc64el" ]; then \
+	go install -race -tags netgo std; \
+   fi;
+RUN export arch_val="$(dpkg --print-architecture)"; \
+   if [ "$arch_val" = "amd64" ]; then \
+     curl -fsSL -o shfmt https://github.com/mvdan/sh/releases/download/v1.3.0/shfmt_v1.3.0_linux_amd64 && \
+     chmod +x shfmt && \
+     mv shfmt /usr/bin; \
+   fi;
+   # Skipped installing shfmt, as the version v1.3.0 isn't supported for ppc64le
+   # and the later version of shfmt doesn't work with the application well
 RUN go get -tags netgo \
 		github.com/fzipp/gocyclo \
 		github.com/golang/lint/golint \
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.4.0
+FROM node:8.11
 WORKDIR /home/weave
 COPY package.json yarn.lock /home/weave/
 ENV NPM_CONFIG_LOGLEVEL=warn NPM_CONFIG_PROGRESS=false
--- a/probe/endpoint/dns_snooper.go
+++ b/probe/endpoint/dns_snooper.go
@@ -1,3 +1,7 @@
+// +build linux,amd64 linux,ppc64le
+
+// Build constraint to use this file for amd64 & ppc64le on Linux
+
 package endpoint
 
 import (
--- a/docker/Dockerfile.cloud-agent
+++ b/docker/Dockerfile.cloud-agent
@@ -1,4 +1,4 @@
-FROM alpine:3.5
+FROM alpine:3.7
 WORKDIR /home/weave
 RUN apk add --update bash conntrack-tools iproute2 util-linux curl && \
 	rm -rf /var/cache/apk/*
--- a/probe/process/walker_linux_test.go
+++ b/probe/process/walker_linux_test.go
@@ -1,6 +1,7 @@
 package process_test
 
 import (
+	"os"
 	"reflect"
 	"testing"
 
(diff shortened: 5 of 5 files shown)


bboreham:
Note there is a lint failure in this PR.

meghalidhoble:
Yes, @bboreham , I am looking into this. I am expecting my build-changes should get this resolved.

satyamz:
@meghalidhoble Please run `make lint` locally to verify linting issues.

client/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8.4.0
+FROM node:8.11
```
satyamz: Why are you upgrading versions of most of the components in `Dockerfile` ? Why it is required?

probe/endpoint/dns_snooper_linux_ppc64le.go
```diff
@@ -0,0 +1,284 @@
+package endpoint
```
satyamz: Why this file is required? This file is exact copy of `github.com/weaveworks/scope/probe/endpoint/dns_snooper_linux_amd64.go` Please explain this in more details.

client/Dockerfile
meghalidhoble: I have upgraded the versions, as these versions are now mulit-arch supported and will help building scope on many platforms easily. For me it helped to build on ppc64le.

probe/endpoint/dns_snooper_linux_ppc64le.go
meghalidhoble: @satyamz , that's right. The new file is exact copy of dns_snooper_linux_amd64.go. I had to add this file to fix the build error which says "undefined: DNSSnooper". Also observed that there is already a file dns_snooper_others.go, however the implementation was missing for the class "DNSSnooper", hence thought of reusing the amd64 file. Please suggest if this can be implemented in different way.

client/Dockerfile
satyamz: ok!  We need to make sure that these upgrades don't break anything.

bboreham:
Please run `gofmt -s -w probe/process/walker_linux_test.go` then update the PR.  This should clear the lint error.

meghalidhoble:
Thanks @bboreham , your suggestion helped to resolve the failure.

probe/endpoint/dns_snooper_linux_ppc64le.go
satyamz: Can I have build failure log?

bboreham: The reason it was undefined is the `others` file has a build constraint saying it only builds on `darwin arm`.
Given the stated reason for this PR, the simple way forward would be to add your architecture to that list.

If you can test that the amd64 file works correctly on ppc64le then renaming it and adding a build constraint for those two platforms would be fine.

Copying the whole file is not ok.

backend/Dockerfile
```diff
@@ -1,14 +1,32 @@
...
-	go install -race -tags netgo std
-RUN curl -fsSL -o shfmt https://github.com/mvdan/sh/releases/download/v1.3.0/shfmt_v1.3.0_linux_amd64 && \
-    chmod +x shfmt && \
-    mv shfmt /usr/bin
+   export arch_val="$(dpkg --print-architecture)"; \
+   if [ "$arch_val" != "ppc64el" ]; then \
+	go install -race -tags netgo std; \
+   fi
+RUN export arch_val="$(dpkg --print-architecture)"; \
+   if [ "$arch_val" != "ppc64el" ]; then \
```
bboreham: surely if the thing we are trying to download is `amd64` this test should be `== amd64` ?

backend/Dockerfile
```diff
@@ -1,14 +1,32 @@
-FROM golang:1.10.2-stretch
+FROM golang:1.10.2
```
bboreham: please explain this change

bboreham:
Please put into each commit message a description of the changes.  
Some tips [here](https://github.com/weaveworks/scope/blob/master/CONTRIBUTING.md#format-of-the-commit-message)
"code changes made" is not a useful description.

backend/Dockerfile
meghalidhoble: "golang:1.10.2" is a multi-arch image, which means it will build for amd64 & ppc64le as well. Hence the change. Sorry for not adding the description at the time of commit. I will make sure I will do that now on wards.

probe/endpoint/dns_snooper_linux_ppc64le.go
meghalidhoble: Hi @bboreham , @satyamz ,
I could add the build-constraint instead of having new file. For this I had to rename the existing file "dns_snooper_linux_amd64.go" to "dns_snooper.go" as the build-tags won't work with file-suffixes. 
Then added a line at the start of the file "dns_snooper.go" as below: 
  ```// +build linux,amd64 ppc64le ``` 

I have verified this patch works on ppc as well as on x86.
Please suggest if this approach looks good, I will commit the changes accordingly.

abdasgupta:
Hi @bboreham, @satyamz ,
Can we please merge this PR if no more modification is needed?

backend/Dockerfile
```diff
@@ -1,14 +1,32 @@
...
+   fi;
+
 RUN go clean -i net && \
 	go install -tags netgo std && \
-	go install -race -tags netgo std
-RUN curl -fsSL -o shfmt https://github.com/mvdan/sh/releases/download/v1.3.0/shfmt_v1.3.0_linux_amd64 && \
-    chmod +x shfmt && \
-    mv shfmt /usr/bin
+   export arch_val="$(dpkg --print-architecture)"; \
+   if [ "$arch_val" != "ppc64el" ]; then \
```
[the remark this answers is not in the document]
meghalidhoble: I had to exclude as the race-detector libs are not supported for power.

probe/endpoint/dns_snooper.go
```diff
@@ -1,3 +1,7 @@
+// +build linux,amd64 ppc64le
```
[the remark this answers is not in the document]
meghalidhoble: The existing file name indicated Linux and amd64, hence the build constraint also indicates the same. However for ppc64le , I want "Linux" to be the OS. I miss-read the separators. Sorry for that, I will correct it.

probe/endpoint/dns_snooper.go
```diff
@@ -1,4 +1,5 @@
-// +build linux,amd64 ppc64le
+// +build linux
+// +build amd64 ppc64le
```
meghalidhoble: Not sure why this change is causing a test-failure, as the command "COVERDIR=./coverage make RM= CODECGEN_UID=23 tests" working fine for me on my setup. 
Any pointers would really help.

meghalidhoble: Could resolve this failure, by latest commit.

meghalidhoble:
Hi @bboreham ,
I have addressed most of your comments and have pushed a few new code changes, Could you please take a look and let me know if any comments/suggestions?

meghalidhoble:
Hi @bboreham , squashed all the commits to a single commit, with all the implementation details for each file being modified. Please take a look.

bboreham:
Sadly, it is inappropriate to directly modify anything in the `tools` directory because it is a vendored copy of https://github.com/weaveworks/build-tools.  Need to modify it upstream and then bring that commit in.

I'm sure this is maddening.

I might just change it to do nothing if it can't find `shfmt`, as is done for `shellcheck`, rather than have an architecture-specific test.

meghalidhoble:
Hi @bboreham ,
I have addressed the concern raised for tools directory changes, hence I have removed the architecture specific changes. Sorry for multiple commits, I will squash those commits and will combine to a single commit. Let me know if these changes look ok.

bboreham:
Multiple commits is fine as long as they all form a progression and are all self-explanatory.

However the issue is that to change a file under the `tools` directory in this repo you need to change it in https://github.com/weaveworks/build-tools first and then do a `git subtree` to update the directory in this repo, as described in `tools/README.md`.

meghalidhoble:
Hi @bboreham , I have raised a pull request at build-tools repo with the lint changes, once those changes are approved, I will run git subtree to update the directory in this repo.

bboreham:
Thanks!

Issue #3219, opened 2018-06-08
[Crash] fatal error: concurrent map iteration and map write


panzerdev:
Hello! 

I deployed the application to a Microsoft AKS cluster version 1.9.6 with the official kubectl command from the docs and haven't changed anything about it https://www.weave.works/docs/scope/latest/installing/#k8s

The application crashes with the exception multiple times (here are the logs)
[logs-from-app-in-weave-scope-app-57566f67f6-dflmb.txt](https://github.com/weaveworks/scope/files/2083893/logs-from-app-in-weave-scope-app-57566f67f6-dflmb.txt)

I couldn't find an issue about this so I am sorry if it turnes out to be a duplicate :blush:

satyamz:
@rade I was not able to reproduce this issue. I am not sure how I can run race detector on scope binary, I could use some help here.
However, There's definite need to sync two go routines reading `Renderer`'s. This should be done by using `mutex`. 
 
```diff
diff --git a/render/render.go b/render/render.go
index eb3b6c8a..62caf2aa 100644
--- a/render/render.go
+++ b/render/render.go
@@ -1,9 +1,13 @@
 package render
 
 import (
+       "sync"
+
        "github.com/weaveworks/scope/report"
 )
 
+var mutex sync.RWMutex
+
 // MapFunc is anything which can take an arbitrary Node and
 // return another Node.
 //
@@ -72,12 +76,16 @@ func (r Reduce) Render(rpt report.Report) Nodes {
        for _, renderer := range r {
                renderer := renderer // Pike!!
                go func() {
+                       mutex.Lock()
+                       defer mutex.Unlock()
                        c <- renderer.Render(rpt)
                }()
        }
        for ; l > 1; l-- {
                left, right := <-c, <-c
                go func() {
+                       mutex.Lock()
+                       defer mutex.Unlock()
                        c <- left.Merge(right)
                }()
        }
```

rade:
Synchronising here would completely remove the parallelism, which is the whole point of spinning up multiple go routines in the first place.

What is the actual race?

bboreham:
Crash is `fatal error: concurrent map iteration and map write`

The iteration is over `range nodes.Nodes` at `render/filters.go:67`

so I would surmise that structure is not being copied some place where it should be.

bboreham:
@satyamz maybe start by running the unit tests with `-race` ?

satyamz:
@bboreham I ran tests using `-race` but nothing showed up.

```
[...]

PR #3232, opened 2018-06-20
Stop filterUnconnected() modifying its input

bboreham:
Helper function `filterInternetAdjacencies()` would sometimes copy and sometimes not copy the input, which can lead to concurrent reads and writes as seen in #3219.

Changed to always copy, before calling.

I found this by eyeballing; this is the only case I could see where the input `.Nodes` can be modified.


bboreham:
@satyamz would you like to review?

PR #3236, opened 2018-06-22
Faster report merging through mutating objects

bboreham:
When we know we have the only reference to a `Report` or `Node` object we can avoid copying the data to change it. Add "Unsafe" variants of various `Merge` operations which mutate the receiver, and a new `Merger` which takes advantage of them.

Benchmarks:
```
BenchmarkReportMerge-2 	       2	 608175373 ns/op	257994508 B/op	 1403665 allocs/op
BenchmarkReportFastMerge-2     3	 368993238 ns/op	110116594 B/op	  887828 allocs/op
BenchmarkSmartMerger-2 	      50	  28800741 ns/op	12794115 B/op	   58118 allocs/op
BenchmarkDumbMerger-2  	      20	 172918072 ns/op	54793384 B/op	  208674 allocs/op
BenchmarkFastMerger-2  	     500	   3997132 ns/op	  740686 B/op	   21410 allocs/op
```

Further refinements can be made by adding `UnsafeMerge` on `MetricTemplates`, etc., though this should be guided by profiling.

### DIFF
--- a/report/topology.go
+++ b/report/topology.go
@@ -163,6 +163,21 @@ func (t Topology) Merge(other Topology) Topology {
 	}
 }
 
+// UnsafeMerge merges the other object into this one, modifying the original.
+func (t *Topology) UnsafeMerge(other Topology) {
+	if t.Shape == "" {
+		t.Shape = other.Shape
+	}
+	if t.Label == "" {
+		t.Label, t.LabelPlural = other.Label, other.LabelPlural
+	}
+	t.Nodes.UnsafeMerge(other.Nodes)
+	t.Controls = t.Controls.Merge(other.Controls)
+	t.MetadataTemplates = t.MetadataTemplates.Merge(other.MetadataTemplates)
+	t.MetricTemplates = t.MetricTemplates.Merge(other.MetricTemplates)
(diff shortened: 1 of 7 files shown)


report/topology.go
```diff
@@ -186,14 +201,19 @@ func (n Nodes) Merge(other Nodes) Nodes {
...
+// UnsafeMerge merges the other object into this one, modifying the original.
+func (n *Nodes) UnsafeMerge(other Nodes) {
 	for k, v := range other {
-		if n, ok := cp[k]; ok { // don't overwrite
-			cp[k] = v.Merge(n)
+		if existing, ok := (*n)[k]; ok { // don't overwrite
+			(*n)[k] = v.Merge(existing)
 		} else {
-			cp[k] = v
+			(*n)[k] = v
```
rade: I reckon the pointering here hinders rather than helps comprehension - I am reading this and wondering "why is this necessary?".

bboreham: I made the receiver a pointer to stress that it modifies `n`.
Since `n` is really a `map` this isn't necessary.  I'm uncertain which is better.

PR #3244, opened 2018-06-29
Support rendering node details extras

### LINKED ISSUE #3243: Support pluggable node details attachments through app component props
In line with #3238, would make Scope more independent of Weave Cloud:

* `NodeDetailsImageStatus` component code could be pulled out of Scope code
* Instead it would be passed by Weave Cloud as a render function component prop
* Scope would not need have access to Service UI `api` and `actions` data anymore

Part of weaveworks/service-ui#2759.

fbarl:
Resolves #3243.

Not to be merged before https://github.com/weaveworks/service-ui/pull/2821.

Test like #3239.


foot:
Could this table be implemented using the scope-details-data schema? I guess not completely, but it would be another interesting way to support client plugins/customization. If there was a way to fiddle w/ the details response structure and add on another table.

Though perhaps more fragile if it was to change. Would have to document it better.

fbarl:
> Could this table be implemented using the scope-details-data schema?

@foot That's an interesting idea that might become more relevant in the future - feel free to open an issue!

PR #3253, opened 2018-07-04
Optimise merge where one side is a subset of the other

bboreham:
This is beneficial because Scope very often merges multiple versions of the same node over time that have exactly the same labels, tags, etc., so we detect this and avoid copying.
This PR covers just three data structures - `StringSet`, `StringLatestMap` and `NodeControlDataLatestMap` - which benefit the most in profiles.

All three now use the same loop structure to merge. The loop is coded twice: once as far as we can get without changing the result, and then again after allocating space for a new return value.  If we got all the way to the end without changing anything, then return what we already have.

`StringSet` is mainly used in `Sets` where it benefits to avoid assigning to its map, so add an `unchanged` return.

Before (best of five):
```
BenchmarkReportMerge-2   	       3	 415559420 ns/op	110109765 B/op	  887819 allocs/op
BenchmarkReportUpgrade-2   	      20	  74962687 ns/op	 4983769 B/op	   22949 allocs/op
```

After:
```
BenchmarkReportMerge-2   	       5	 285569184 ns/op	63896876 B/op	  475350 allocs/op
BenchmarkReportUpgrade-2   	      20	  74528873 ns/op	 4856011 B/op	   18995 allocs/op
```

### DIFF
--- a/report/latest_map_generated.go
+++ b/report/latest_map_generated.go
@@ -64,14 +64,16 @@ func (m StringLatestMap) Merge(n StringLatestMap) StringLatestMap {
 loop:
 	for i < len(m) {
 		switch {
-		case j >= len(n) || m[i].key < n[j].key:
-			i++
+		case j >= len(n):
+			return m
 		case m[i].key == n[j].key:
 			if m[i].Timestamp.Before(n[j].Timestamp) {
 				break loop
 			}
 			i++
 			j++
+		case m[i].key < n[j].key:
+			i++
 		default:
 			break loop
 		}
@@ -85,9 +87,9 @@ loop:
 
 	for i < len(m) {
 		switch {
-		case j >= len(n) || m[i].key < n[j].key:
-			out = append(out, m[i])
-			i++
+		case j >= len(n):
+			out = append(out, m[i:]...)
+			return out
 		case m[i].key == n[j].key:
 			if m[i].Timestamp.Before(n[j].Timestamp) {
 				out = append(out, n[j])
@@ -96,6 +98,9 @@ loop:
 			}
 			i++
 			j++
+		case m[i].key < n[j].key:
+			out = append(out, m[i])
+			i++
 		default:
 			out = append(out, n[j])
 			j++
@@ -311,14 +316,16 @@ func (m NodeControlDataLatestMap) Merge(n NodeControlDataLatestMap) NodeControlD
 loop:
 	for i < len(m) {
 		switch {
-		case j >= len(n) || m[i].key < n[j].key:
-			i++
+		case j >= len(n):
+			return m
 		case m[i].key == n[j].key:
 			if m[i].Timestamp.Before(n[j].Timestamp) {
 				break loop
 			}
 			i++
 			j++
+		case m[i].key < n[j].key:
+			i++
 		default:
 			break loop
 		}
@@ -332,9 +339,9 @@ loop:
 
 	for i < len(m) {
 		switch {
-		case j >= len(n) || m[i].key < n[j].key:
-			out = append(out, m[i])
-			i++
+		case j >= len(n):
+			out = append(out, m[i:]...)
+			return out
 		case m[i].key == n[j].key:
 			if m[i].Timestamp.Before(n[j].Timestamp) {
 				out = append(out, n[j])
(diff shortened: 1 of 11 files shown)


report/latest_map_generated.go
```diff
@@ -42,23 +42,45 @@ func (m StringLatestMap) Size() int {
...
 
 	i, j := 0, 0
-	for i < len(m) {
+loop:
+	for i < lenm {
+		switch {
+		case j >= len(n) || m[i].key < n[j].key:
+			i++
+		case m[i].key == n[j].key:
+			if m[i].Timestamp.Before(n[j].Timestamp) {
```
rade: Looks like this makes Merge non-commutative as far as performance is concerned, e.g. when merging two maps m,n with identical keys & values, where all entries in m are older than n, then the operation will copy, whereas it will simply return n if the args are swapped.

One way to avoid that would be to check the timestamp of the first element of equal-length args, and swap the args if necessary.

report/string_set.go
```diff
@@ -81,32 +81,49 @@ func (s StringSet) Add(strs ...string) StringSet {
...
+	if i >= len(s) && j >= len(other) {
+		return s, true
+	}
+
+	result := make(StringSet, i, len(s)+len(other))
+	copy(result, s[:i])
+
+	for i < len(s) {
+		switch {
+		case j >= len(other) || s[i] < other[j]:
```
rade: why not split these two cases, as we used to, which gets us a bulk append and early loop termination for the first case?

rade:
I suggest cherry-picking de47a0eb0b93d1 (Less verbose string dump) onto master, since it's a bit of a distraction in this PR.

report/latest_map_generated.go
bboreham: I would tend to ask for profiling evidence that that case is important.

report/string_set.go
bboreham: The simple answer is that I copy-pasted the code from `LatestMap`.  It seems the same improvement would work there.

report/latest_map_generated.go
rade: The stated purpose of this PR is to optimise the case where Scope
> merges multiple versions of the same node over time that have exactly the same labels, tags, etc., so we detect this and avoid copying.

If all the above is true, but the timestamps differ then the new code still copies. And I am quite sure the timestamps will nearly _always_ differ. I am also quite sure that in the common case the timestamps will differ systematically s.t. all timestamps in one arg are before that in the other. So we should optimise that case. Yes, it would help to check these "quite sure"s against actual data.

It is possible that the Merge benchmark, or indeed the Merge code, ensures that in most cases the Merge is always invoked as new.Merge(old). But that is not something I'd want to rely on.

rade: Also, non-commutativity on performance of an otherwise commutative operation is quite distasteful.

bboreham: Aha, I hadn't previously noticed that the probe, which is caching everything, re-generates most `Node` objects from scratch on each report.

I would expect the merge benchmark is merging from oldest to newest, since the filenames begin with a timestamp, Go sorts filenames, and there is no other sort operation. Bears examination.

bboreham: Maybe it's just evidence that allocating memory is so much slower than everything else that a few extra copies doesn't show up.

bboreham: Examination reveals that `Nodes.UnsafeMerge()` is merging left into right, which reverses the sort order from the merge benchmark. I added a `Shuffle()` to the test to defeat this.

report/latest_map_generated.go
```diff
@@ -42,27 +42,53 @@ func (m StringLatestMap) Size() int {
...
 		return m
 	}
-	l := len(m)
-	if len(n) > l {
-		l = len(n)
+	lenm := len(m)
+	if len(n) > lenm {
+		m, n, lenm = n, m, len(n) //swap so m is always at least as long as n
+	} else if len(n) == lenm && m[0].Timestamp.Before(n[0].Timestamp) {
+		m, n = n, m // swap equal-length arrays so first element of m is newer
```
rade: The mix of `lenm` and `len(n)` is quite jarring. More of that below, where we additionally get `len(m)`. I suggest getting rid of `lenm`, hence always calling `len(..)`.

rade: also, it would be helpful to expand that last comment to explain that a) this is an optimisation, and b) why it's useful.

bboreham: Removed `lenm` and extended comment.

Issue #3258, opened 2018-07-09
Add storage driver name for the given Persistent Volume


satyamz:
It would be helpful to know which storage provider was used to create a Kubernetes PV.
These details can be fetched from  [PersistentVolumeSource](https://github.com/kubernetes/api/blob/3c2a58f9923aeb5d27fa4d91249e45a1460ca3bd/core/v1/types.go#L371).

PR #3260, opened 2018-07-09
Add storage driver name to Persistent Volume

Akash4927:
Every PV is backed by storage driver.
This will show the storage driver in the PV metadata.

fixes: https://github.com/weaveworks/scope/issues/3258

### DIFF
--- a/probe/kubernetes/persistentvolume.go
+++ b/probe/kubernetes/persistentvolume.go
@@ -1,6 +1,8 @@
 package kubernetes
 
 import (
+	"reflect"
+
 	"github.com/weaveworks/scope/report"
 	apiv1 "k8s.io/api/core/v1"
 )
@@ -11,6 +13,7 @@ type PersistentVolume interface {
 	GetNode() report.Node
 	GetAccessMode() string
 	GetVolume() string
+	GetStorageDriver() string
 }
 
 // persistentVolume represents kubernetes persistent volume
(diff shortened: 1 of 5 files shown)


satyamz:
@Akash4927 Can you please attach screenshot for this change? How this will appear on the UI?

Akash4927:
Sure @satyamz

Akash4927:
screenshot from 2018-07-09 15-28-41
screenshot from 2018-07-09 15-28-30

@satyamz I tried it with `OpenEBS` and `PortWorx` to make sure that nothing is specific to a single vendor.

probe/kubernetes/persistentvolume.go
```diff
@@ -45,13 +48,31 @@ func (p *persistentVolume) GetVolume() string {
 	return volume
 }
 
+// GetStorageDriver returns the backing driver of Persistent Volume
+func (p *persistentVolume) GetStorageDriver() string {
```
2opremio: Are you sure that the k8s library doesn't supply a way to get the textual representation of the driver?

2opremio: (if not, then that's OK, we could use something like https://github.com/fatih/structs but it's completely overkill for just this)

Akash4927: @2opremio AFAIK, k8s library doesn't provide the textual representation of the driver directly, 
and yes, we can use https://github.com/fatih/structs but I don't want to introduce any dependency for such a small change.

probe/kubernetes/persistentvolume.go
```diff
@@ -45,13 +48,31 @@ func (p *persistentVolume) GetVolume() string {
...
+	persistentVolumeSource := reflect.ValueOf(p.Spec.PersistentVolumeSource)
+
+	for i := 0; i < persistentVolumeSource.NumField(); i++ {
```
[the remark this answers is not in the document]
Akash4927: Ok, @bboreham I'll add the comment.

Akash4927:
@bboreham Can you please review this?

PR #3262, opened 2018-07-11
Use GraphNode component

### LINKED ISSUE #3261: Use GraphNode component for graph nodes
Depends on https://github.com/weaveworks/ui-components/pull/291.

fbarl:
Resolves #3261.

The behaviour should have stayed pretty much the same. Some differences include:

* Not fading out the node shapes when metric is selected
* The shapes being rendered a tiny bit differently
* Removed the borders on `MatchedText` items

Also, the `ui-components` have been bumped to the latest version.


client/app/scripts/charts/nodes-chart-elements.js
```diff
@@ -157,18 +157,17 @@ class NodesChartElements extends React.Component {
...
         label={node.get('label')}
         labelMinor={node.get('labelMinor')}
         pseudo={node.get('pseudo')}
         rank={node.get('rank')}
-        dx={node.get('x')}
-        dy={node.get('y')}
-        scale={node.get('scale')}
+        x={node.get('x')}
+        y={node.get('y')}
+        size={100 * node.get('scale')}
```
guyfedwards: just commenting to check this is intentional as `100 * x` is quite a jump from `x` 👌

fbarl: Oh yes, the scale is a relative scale actually (normalized at `1`) and `100` corresponds to node base size, so I should replace it with the actual existing `NODE_BASE_SIZE` constant. Good catch! :)

PR #3266, opened 2018-07-11
Remove First and Last data members from Metrics structs

bboreham:
They can be trivially computed when required.
Smaller structs = less garbage-collection = better performance.

This is a breaking change in the wire protocol; will need a version bump.  New app will work with older probes, just ignoring the data.

I checked the JavaScript and can only find one use of the field, and that already falls back to the alternative.

Benchmarks (unmarshalling older reports which do contain the fields)

Before:
```
BenchmarkReportUnmarshal-2   	       1	1090111087 ns/op	184140088 B/op	 2497821 allocs/op
BenchmarkReportMerge-2   	       3	 437993382 ns/op	110116245 B/op	  887843 allocs/op
```

After:
```
BenchmarkReportUnmarshal-2   	       1	1089826095 ns/op	172437016 B/op	 2497790 allocs/op
BenchmarkReportMerge-2   	       3	 397821741 ns/op	102478133 B/op	  887843 allocs/op
```

The impact on `Merge()` is more noticeable when combined with #3253:
```
BenchmarkReportMerge-2   	       5	 303941149 ns/op	56328974 B/op	  476558 allocs/op
```


report/metrics.go
```diff
@@ -48,11 +48,17 @@ func (m Metrics) Copy() Metrics {
...
 // Add method to add values.  Metrics are immutable.
 type Metric struct {
-	Samples     []Sample
-	Min, Max    float64
-	First, Last time.Time
+	Samples  []Sample
+	Min, Max float64
 }
 
+// Following two functions are exported for testing only: make sure there are samples before calling.
```
rade: Perhaps ditch those tests instead. They don't seem all that useful.

PR #3268, opened 2018-07-17
Optimise Node.WithLatests()

bboreham:
This is a different take on solving the same problem as in #3137.  It's not quite as fast but has lower memory usage, and it doesn't have the same readability issue.

Here we make one memory allocation to resize the array, then add the contents of the map, sort to regain the `LatestMap` invariant, then remove any keys which are now duplicated (in normal usage this doesn't happen, but better to be safe. It does happen in at least one test).

Based on profiling some real probes, I wrote a benchmark to mimic typical operation of this code.

Results before (best of 5):
```
BenchmarkReporter 	     500	   2671920 ns/op	  841966 B/op	    9290 allocs/op
```

After:
```
BenchmarkReporter 	     500	   2364309 ns/op	  472526 B/op	    8479 allocs/op
```

~Note the optimisation to `propagateLatests()` in #3137 is not included in this PR.~

### DIFF
--- a/report/map_helpers.go
+++ b/report/map_helpers.go
@@ -173,3 +173,16 @@ func (m StringLatestMap) addMapEntries(ts time.Time, n map[string]string) String
 	out.fixup()
 	return out
 }
+
+// Propagate a set of latest values from one set to another.
+func (m StringLatestMap) Propagate(from StringLatestMap, keys ...string) StringLatestMap {
+	out := make(StringLatestMap, len(m), len(m)+len(keys))
+	copy(out, m)
+	for _, k := range keys {
+		if v, ts, ok := from.LookupEntry(k); ok {
+			out = append(out, stringLatestEntry{key: k, Value: v, Timestamp: ts})
+		}
+	}
+	out.fixup()
+	return out
+}
--- a/report/map_helpers.go
+++ b/report/map_helpers.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"sort"
+	"time"
 
 	"github.com/ugorji/go/codec"
 	"github.com/weaveworks/ps"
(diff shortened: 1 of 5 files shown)


bboreham:
I added in the optimisation to `propagateLatests()` from #3137, re-done slightly differently.  I can squash the refactor after review.

Before:
```
BenchmarkRenderContainers-2   	      10	 149366804 ns/op	18293102 B/op	  120577 allocs/op
```
After
```
BenchmarkRenderContainers-2   	      10	 147485047 ns/op	14980076 B/op	  119862 allocs/op
```

report/map_helpers.go
```diff
@@ -138,3 +139,31 @@ func mapWrite(m ps.Map, encoder *codec.Encoder, encodeValue func(*codec.Encoder,
...
+	for i := 1; i < len(out); { // shuffle down any duplicates
+		if out[i-1].key == out[i].key {
+			if out[i-1].Timestamp.Before(out[i].Timestamp) {
+				out = append(out[:i-1], out[i:]...)
+			} else {
+				out = append(out[:i], out[i+1:]...)
+			}
+			continue
+		}
+		i++
```
rade: nitpick: ditching the `continue` and instead having an `else` clause would be neater, I think.

report/map_helpers.go
```diff
@@ -138,3 +139,48 @@ func mapWrite(m ps.Map, encoder *codec.Encoder, encodeValue func(*codec.Encoder,
...
+			if m[i-1].Timestamp.Before(m[i].Timestamp) {
+				m = append(m[:i-1], m[i:]...)
+			} else {
+				m = append(m[:i], m[i+1:]...)
+			}
+		} else {
+			i++
+		}
+	}
+}
```
rade: nitpick: I'd move this method up so it comes before the two methods that invoke it.

bboreham: moved.

Issue #3242, opened 2018-06-28
Stop talking to the kubelet unsecured port


bboreham:
When running under Kubernetes, the probe will try to talk to kubelet on 10255, unsecured, if it doesn't know the node name. We should not assume unsecured access.

There is a secured port at 10250; probably we have access to credentials.  Or we could just remove that code and insist on knowing the node name.

Kubeadm removes the unsecured port by default: https://github.com/kubernetes/kubernetes/pull/64187

This is possibly the underlying issue for #3104 (OpenShift)

yuriprym:
I was wonder how can I assist in you solving your problem?

bboreham:
@yuriprym I'm not clear what kind of advice you need.
Given other developments it's probably best to remove the code that talks to kubelet and rely on filtering by node name (if per-node reporting is still enabled).  So the main work is in checking the system still works, and considering how to warn the user if they don't supply a node name.

PR #3274, opened 2018-07-23
Enable Kubernetes objects to be reported just once in a cluster

bboreham:
This set of changes allows you to configure the probe on each node with Kubernetes probing disabled, and run one extra probe with Kubernetes enabled and processes, containers, etc., disabled. This is quite simple to arrange with one DaemonSet and one Deployment.

Benefits:
 * less impact on the Kubernetes api-server
 * less work on each node gathering and reporting the data in each probe
 * less work in the app to merge N copies of identical nodes
 * less network traffic to send them to the app.  
Disappointingly the CPU-usage benefit wasn't huge when I tried it in our staging cluster, but I didn't spend long looking at why.

It has one disruptive change: pods never get tagged with a host-id.  The rendering code is changed to find this on a container node. When reporting Kubernetes on just one node in the cluster, it doesn't know what hostID has been given to any other nodes.

Alternatives considered to the above change:
 * Add the Kubernetes node-name to each host node, then have the renderer match them up.
 * use Kubernetes node-name instead of OS-supplied hostname as the host ID. This seems like a better idea, but is a more disruptive change if people are used to seeing a certain format on the screen. Maybe we can separate the internal ID from the on-screen presentation?

### DIFF
--- a/probe/kubernetes/reporter.go
+++ b/probe/kubernetes/reporter.go
@@ -238,8 +238,15 @@ func isPauseContainer(n report.Node, rpt report.Report) bool {
 	return false
 }
 
+// Tagger adds pod parents to container nodes.
+type Tagger struct {
+}
+
+// Name of this tagger, for metrics gathering
+func (Tagger) Name() string { return "K8s" }
+
 // Tag adds pod parents to container nodes.
-func (r *Reporter) Tag(rpt report.Report) (report.Report, error) {
+func (r *Tagger) Tag(rpt report.Report) (report.Report, error) {
 	for id, n := range rpt.Container.Nodes {
 		uid, ok := n.Latest.Lookup(docker.LabelPrefix + "io.kubernetes.pod.uid")
 		if !ok {
--- a/probe/kubernetes/reporter.go
+++ b/probe/kubernetes/reporter.go
@@ -561,7 +561,7 @@ func (r *Reporter) podTopology(services []Service, deployments []Deployment, dae
 	}
 
 	var localPodUIDs map[string]struct{}
-	if r.nodeName == "" {
+	if r.nodeName == "" && r.kubeletPort != 0 {
 		// We don't know the node name: fall back to obtaining the local pods from kubelet
 		var err error
 		localPodUIDs, err = GetLocalPodUIDs(fmt.Sprintf("127.0.0.1:%d", r.kubeletPort))
--- a/prog/main.go
+++ b/prog/main.go
@@ -330,7 +330,7 @@ func setupFlags(flags *flags) {
 	flag.StringVar(&flags.probe.kubernetesClientConfig.User, "probe.kubernetes.user", "", "The name of the kubeconfig user to use")
 	flag.StringVar(&flags.probe.kubernetesClientConfig.Username, "probe.kubernetes.username", "", "Username for basic authentication to the API server")
 	flag.StringVar(&flags.probe.kubernetesNodeName, "probe.kubernetes.node-name", "", "Name of this node, for filtering pods")
-	flag.UintVar(&flags.probe.kubernetesKubeletPort, "probe.kubernetes.kubelet-port", 10255, "Node-local TCP port for contacting kubelet")
+	flag.UintVar(&flags.probe.kubernetesKubeletPort, "probe.kubernetes.kubelet-port", 10255, "Node-local TCP port for contacting kubelet (zero to disable)")
 
 	// AWS ECS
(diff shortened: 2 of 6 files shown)


2opremio:
>  The rendering code is changed to find this on a container node

I think this is acceptable

> use Kubernetes node-name instead of OS-supplied hostname as the host ID

Wouldn't this require hosts would also provide their Kubernetes node-name? Otherwise how would you  map pods to hosts?

2opremio:
A fre comments:

1. Shouldn't this PR remove https://github.com/weaveworks/scope/blob/052ff39bf1b1e92f3127aeb4cfbe9cf554d105b9/probe/kubernetes/kubelet.go (and its test file)? 

2. I think t's important to make sure that the apps in Weave Cloud are updated ASAP (as I understand it, k8s rendering will break for probes using this code until the counterpart app is updated).

3. Did you test backwards compatibility with probes using the hostID?

prog/main.go
```diff
@@ -320,7 +322,7 @@ func setupFlags(flags *flags) {
 	flag.StringVar(&flags.probe.kubernetesClientConfig.User, "probe.kubernetes.user", "", "The name of the kubeconfig user to use")
 	flag.StringVar(&flags.probe.kubernetesClientConfig.Username, "probe.kubernetes.username", "", "Username for basic authentication to the API server")
 	flag.StringVar(&flags.probe.kubernetesNodeName, "probe.kubernetes.node-name", "", "Name of this node, for filtering pods")
-	flag.UintVar(&flags.probe.kubernetesKubeletPort, "probe.kubernetes.kubelet-port", 10255, "Node-local TCP port for contacting kubelet")
+	flag.UintVar(&flags.probe.kubernetesKubeletPort, "probe.kubernetes.kubelet-port", 10255, "Node-local TCP port for contacting kubelet (zero to disable)")
```
2opremio: Can't we simply disable it completely, even when not using a separate deployment for gathering k8s information?

2opremio:
If I understood properly, after this PR:

1. The normal probes (daemon sets) will run with `--probe.kubernetes-tag=true`
2. The deployment gathering k8s information will run with `--probe.kubernetes =true`

If that's correct I think it would be more intuitive to something like:

1. Run the daemon set with `--probe.kubernetes=true``--probe.kubernetes.role=host-agent`
2. Run the deployment with `--probe.kubernetes=true` `--probe.kubernetes.role=api-server-agent`

(we could optionally drop the `--probe.kubernetes=true` since it's redundant.)


I think that exposing whether it tags or not is an implementation detail which may be confusing for the end user. Indicating what the probe is used for would more useful.

probe/kubernetes/reporter.go
```diff
@@ -552,7 +559,7 @@ func (r *Reporter) podTopology(services []Service, deployments []Deployment, dae
 	}
 
 	var localPodUIDs map[string]struct{}
-	if r.nodeName == "" {
+	if r.nodeName == "" && r.kubeletPort != 0 {
 		// We don't know the node name: fall back to obtaining the local pods from kubelet
 		var err error
 		localPodUIDs, err = GetLocalPodUIDs(fmt.Sprintf("127.0.0.1:%d", r.kubeletPort))
```
2opremio: I don't think you want to call `GetLocalPodUIDs` in the separate deployment since it will only report pods running in the host where Kubernetes decide it to run it.

prog/main.go
bboreham: That is the suggestion in #3242.

bboreham:
I have rebased and updated the flag settings in line with what @2opremio suggested.  Now it is:

1. Run the daemon set with `--probe.kubernetes.role=host`
2. Run the deployment with `--probe.kubernetes.role=cluster`

Talking directly to kubelet is now disabled in both modes, although the code remains - removing it would be a separate PR.

bboreham:
I tested this again in our staging cluster; backwards-compatibility is fine.
However I hadn't quite got the host parent propagation right, so have fixed that.

Impact on our staging cluster was about a 10% reduction in CPU usage by probes.  In bigger clusters the net impact should be better.

I think this is good to go now.

2opremio:
Are there followup issues/PRs for:
* Updating https://cloud.weave.works/k8s/ ? (so that we only talk to the api server from a deployement)
* Removing the kubelet code ?

bboreham:
#3242 is the latter.  Yes we are very keen to update the cloud.weave.works config, but need to do a Scope release first, which was waiting on a review here.

PR #3275, opened 2018-07-23
Add initial CRI support

LiliC:
This adds a CRI probe which you can use with different container runtimes, such as cri-o in the below example. The probe defaults to the dockershim endpoint, but can be modified by passing the `probe.cri.endpoint`. To use you need to expose those sockets in your deployment file as well. I tested with cri-o and dockershim so it should work fine with containerd as well as it's the same API anyways.

In the followup work we need to have a way of how to gather various other information that we cannot get via CRI, such as network information, hostname etc. Will open follow up issue/issues for that!

As for the examples, there is an image tag in the deployment manifest files but since we don't know when this will be merged in the release was not sure what to add there? Maybe should leave that out for now and add it in the following PR when this is merged in the release and we have an image?

This is an example of the results of containers view in a local minikube cluster started with cri-o:
screen shot 2018-07-23 at 12 05 44


prog/main.go
```diff
@@ -304,6 +307,10 @@ func setupFlags(flags *flags) {
 	flag.DurationVar(&flags.probe.dockerInterval, "probe.docker.interval", 10*time.Second, "how often to update Docker attributes")
 	flag.StringVar(&flags.probe.dockerBridge, "probe.docker.bridge", "docker0", "the docker bridge name")
 
+	// CRI
+	flag.BoolVar(&flags.probe.criEnabled, "probe.cri", true, "collect CRIr-related attributes for processes")
```
marccarre: Typo? `s/CRIr-related/CRI-related/`

Makefile
```diff
@@ -45,6 +45,15 @@ IMAGE_TAG=$(shell ./tools/image-tag)
 
 all: $(SCOPE_EXPORT)
 
+update-cri:
+	curl https://raw.githubusercontent.com/kubernetes/kubernetes/master/pkg/kubelet/apis/cri/runtime/v1alpha2/api.proto > runtime/api.proto
+
+protoc-gen-gofast:
+	@go get -u -v github.com/gogo/protobuf/protoc-gen-gofast
+
+cri: update-cri protoc-gen-gofast
```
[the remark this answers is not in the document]
LiliC: We should only run `make cri` when we want to update the files, at least that was the intention of it, as we don't need to update it every time but when the upstream will get updated, hence the checked in files to guarantee the API does not change each time. That includes the actual `cri` make target which just regenerates the newly downloaded files. Open to suggestions of course!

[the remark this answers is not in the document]
marccarre: Perfect then!
It can be guessed given the file is checked in, but you may nevertheless want to add a comment next to the `cri` target to clarify this?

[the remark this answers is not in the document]
LiliC: Makes sense, done! Let me know if it needs rephrasing to understand it better?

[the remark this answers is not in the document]
marccarre: No, perfect, merge at will! 👍

runtime/api.proto
```diff
@@ -0,0 +1,1227 @@
...
+
+option (gogoproto.goproto_stringer_all) = false;
+option (gogoproto.stringer_all) =  true;
+option (gogoproto.goproto_getters_all) = true;
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.sizer_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (gogoproto.goproto_unrecognized_all) = false;
+
+// Runtime service defines the public APIs for remote container runtimes
```
leth: Should the path to this file include something something about CRI or Kubernetes?

Makefile
```diff
@@ -45,6 +45,16 @@ IMAGE_TAG=$(shell ./tools/image-tag)
 
 all: $(SCOPE_EXPORT)
 
+update-cri:
+	curl https://raw.githubusercontent.com/kubernetes/kubernetes/master/pkg/kubelet/apis/cri/runtime/v1alpha2/api.proto > runtime/api.proto
```
leth: Should this also add a header to that file?

runtime/api.proto
LiliC: You mean to change the `runtime` directory name?

Makefile
LiliC: A header? Not sure I understand what you meant here.

runtime/api.proto
leth: yeah! Otherwise it looks like this is the scope runtime module or something.

Makefile
leth: Do we want it to add a big warning to the top to say something like "// DO NOT EDIT: This file is imported from kubernetes, update it with make blabla"

jml:
> In the followup work we need to have a way of how to gather various other information that we cannot get via CRI, such as network information, hostname etc. Will open follow up issue/issues for that!

Are these in https://github.com/orgs/weaveworks/projects/39?

LiliC:
@jml Added the open issues to the board and updated the progress. Sorry about that!

PR #3281, opened 2018-07-25
Fix WithLatests() fixup on duplicate keys

bboreham:
The `fixup()` method was modifying a copy of its input, so you could get duplicate keys in the output.

Thankfully this is rare: in most cases `WithLatests()` is called with fields that are not duplicates of existing ones.

Added a test that would have caught it.

I made this PR against the release-1.9 branch so we can make a clean patch release.

### DIFF
--- a/report/map_helpers.go
+++ b/report/map_helpers.go
@@ -147,8 +147,8 @@ func (m StringLatestMap) Len() int           { return len(m) }
 func (m StringLatestMap) Swap(i, j int)      { m[i], m[j] = m[j], m[i] }
 func (m StringLatestMap) Less(i, j int) bool { return m[i].key < m[j].key }
 
-// sort entries and shuffle down any duplicates
-func (m StringLatestMap) fixup() {
+// sort entries and shuffle down any duplicates. NOTE: may modify contents of m.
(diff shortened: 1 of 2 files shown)


report/map_helpers.go
```diff
@@ -148,19 +148,21 @@ func (m StringLatestMap) Swap(i, j int)      { m[i], m[j] = m[j], m[i] }
 func (m StringLatestMap) Less(i, j int) bool { return m[i].key < m[j].key }
 
 // sort entries and shuffle down any duplicates
-func (m StringLatestMap) fixup() {
-	sort.Sort(m)
-	for i := 1; i < len(m); {
-		if m[i-1].key == m[i].key {
-			if m[i-1].Timestamp.Before(m[i].Timestamp) {
-				m = append(m[:i-1], m[i:]...)
+func (m *StringLatestMap) fixup() {
```
rade: hmm. maybe make it return a slice instead. Would simplify the call sites too: `return out.fixup()`.

Downside is that it might give the appearance of being a pure function.

bboreham: See if a different verb case helps.

PR #3282, opened 2018-07-25
Bump ui-components version

fbarl:
##### Changes

* Reduced border radius on a lot of elements (https://github.com/weaveworks/ui-components/pull/320)
* Fixed the top border radius of node details panel
* Add `react-router` dep for `ui-components` compatilibity

### DIFF
--- a/client/app/scripts/charts/nodes-grid.js
+++ b/client/app/scripts/charts/nodes-grid.js
@@ -1,5 +1,6 @@
 /* eslint react/jsx-no-bind: "off", no-multi-comp: "off" */
 import React from 'react';
+import styled from 'styled-components';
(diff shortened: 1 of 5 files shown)


bia:
Looks good.

These color squares should also have a 2px roundrect.
screen shot 2018-07-26 at 12 23 22

The blocks containing the titles 'Containers' and 'Hosts' should also be rounded.
screen shot 2018-07-26 at 12 23 37

fbarl:
> Oh ok, I didn't intend for everything to be rounded...

Yeah, I thought it looks better if everything in that view is rounded actually, since otherwise the metric boxes would be one of the rare rectangle elements I know (or perhaps the only one?) with sharp corners.

PR #3294, opened 2018-07-31
Make header semitransparent

### LINKED ISSUE #3285: Background color not opaque enough on tab selectors
It's difficult to read titles when graph nodes are shifted underneath.

screen shot 2018-07-26 at 12 22 58

fbarl:
Resolves #3285.

I initially tried to make only the children text elements (topology labels, etc..) semitransparent but didn't look that good so I thought it might be better to simply make the whole header strip more opaque to keep a stronger separation between the navigation elements and the content:

image


fbarl:
@bia PTAL. I added the following changes:

* Increased the filters' background opacity
* Made all the active items in selectors more purple
* Put Node Details panel one level on the top - I'm mainly asking for your approval on this part as things look a bit different now with the node details panel in focus (introduced by the first commit):

image

bia:
>Made all the active items in selectors more purple

more purple helps a lot.

Why did you make it more opaque? Now it's barely noticeable that it's transparent. I would go back to how it was before.

Node details looks good!

Issue #3209, opened 2018-06-04
flaky test failure in RegistryDelete


rade:
I ran into this yesterday in a local build - re-running "fixed it" - and @bboreham [encountered it in CI today](https://circleci.com/gh/weaveworks/scope/8868#tests/containers/1)
```
--- FAIL: TestRegistryDelete (0.01s)
	registry_test.go:511: Didn't get right container updates: 
		--- want
		+++ have
		@@ -1,4 +1,322 @@
		-([]report.Node) (len=1 cap=1) {
		+([]report.Node) (len=2 cap=2) {
                ...
```
Presumably a race condition in the test rather than a bug in the code.

satyamz:
I am also facing this issue on local `make tests` with `-slow` flag.

bboreham:
Looks the same here: https://circleci.com/gh/weaveworks/scope/8977

I suspect the 100ms timeout.

bboreham:
> I suspect the 100ms timeout.

except I just raised it to 10 seconds and it still fails.

bboreham:
The race is:
 - `newMockClient()` sets up a fake Docker with one container, container1.
 - `TestRegistryDelete()` calls `testRegistry()` calls `NewRegistry()` which starts a goroutine that calls `listenForEvents()`
- `TestRegistryDelete()` calls `registry.WatchContainerUpdates()`
 - If `listenForEvents()` runs after `registry.WatchContainerUpdates()`, then the callback is fired with info about container1.
 - If `listenForEvents()` runs before `registry.WatchContainerUpdates()`, then the callback is not fired.

The test fails in the first case because it has two updates instead of the required one.

bboreham:
Reopening: more `TestRegistryDelete` test failures recently:
https://circleci.com/gh/weaveworks/scope/12848
https://circleci.com/gh/weaveworks/scope/12880
https://circleci.com/gh/weaveworks/scope/12884
https://circleci.com/gh/weaveworks/scope/12885

PR #3305, opened 2018-08-08
Disable CRI probe by default

LiliC:
This was enabled by default but we want it to be disabled instead, just
like with the docker probe.

### DIFF
--- a/prog/main.go
+++ b/prog/main.go
@@ -308,7 +308,7 @@ func setupFlags(flags *flags) {
 	flag.StringVar(&flags.probe.dockerBridge, "probe.docker.bridge", "docker0", "the docker bridge name")
 
 	// CRI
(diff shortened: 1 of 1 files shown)


bboreham:
CI failed on #3209 - I hit the re-run button

bboreham:
Should we write some code to try to figure out which type of container runtime is in use, without configuration flags?

LiliC:
> Should we write some code to try to figure out which type of container runtime is in use, without configuration flags?

Problem is we need to also mount the container runtime endpoint in the deployment manifest file anyways when deploying scope so the user needs to know which endpoint to use before starting scope anyways. We do that in the launcher for the weave cloud scope deployment -> https://github.com/weaveworks/launcher/pull/252

PR #3307, opened 2018-08-08
Add Jaeger distributed tracing support

leth:
Allows trace reporting to be configured via environment variables e.g. `JAEGER_AGENT_HOST`, `JAEGER_SAMPLER_TYPE`.


bboreham:
List of changes pulled in from `weaveworks/common`:
 - Don't use full URL as metric label on unmatched path (weaveworks/common#108)
 - Raise default server timeouts from 5s to 30s (weaveworks/common#109)
 - Return errors when http/grpc servers exit. (weaveworks/common#92)
 - Expose jaeger metrics (weaveworks/common#114)
 - Allow the AWS backend to get credentials from its environment. (weaveworks/common#111)

PR #3317, opened 2018-08-14
vendor: update gopkg.in/yaml.v2 to latest upstream

### LINKED ISSUE #3283: Please update vendored copy of yaml.v2
yaml.v2 was relicensed from LGPL3 to Apache 2.0 in https://github.com/go-yaml/yaml/commit/e4d366fc3c7938e2958e662b4258c7a89e1f0e3e#diff-9879d6db96fd29134fc802214163b95a

Scope is licensed under Apache 2.0 as well, so it'd be good to use code under the same license.

dholbach:
yaml.v2 was relicensed under Apache 2.0 in 2016, as Scope is Apache 2.0, we want this too. A lot of other bugs were fixed in the code since then as well.

  fixes #3283


bboreham:
As far as I can see in the code, Scope never calls this library; it is vendor'd as a dependency of various other libraries.

PR #3325, opened 2018-08-23
Add middleware to detect jaeger trace context over HTTP

leth:
Previously scope would not honor http headers requesting that a request be traced.

Updates weaveworks/common to d442d08d89b51712ca61de3f7c14e2e218a739d7
- Default diff-printer to hide details (weaveworks/common#103)
- Improve logging of http errors (weaveworks/common#115)
- Update httpgrpc to match weaveworks/cortex#910 (weaveworks/common#117)
- Expose the HTTP server from the server struct. (weaveworks/common#118)
- Add HTTP tracing middleware (weaveworks/common#119)


bboreham:
Why wouldn't this go in `weaveworks/common/middleware` ?

leth:
> Why wouldn't this go in weaveworks/common/middleware ?

I didn't think this was possible initially; I've worked out how to do it now!

PR #3333, opened 2018-08-25
Move to CircleCI 2.0

### LINKED ISSUE #3217: Migrate Scope off CircleCI 1.0
Because it is claimed to be shutting down https://circleci.com/blog/sunsetting-1-0/

Easiest would be CircleCI 2.0, but other CI systems are available.

Prior art at https://github.com/weaveworks/cortex/pull/799, https://github.com/weaveworks/weave/pull/3270

Note there are a bunch of dependent packages that also build at CircleCI 1.0, e.g. github.com/weaveworks/tcptracer-bpf

bboreham:
Fixes #3217 

Once the changes are merged to master I will update the `image:` lines to a master tag, the same tag for both images.


bboreham:
@guyfedwards I took your starting-point, adapted it to multi-step in the same pattern as weaveworks/weave, then got it to build and pass tests.

I haven't tried the deployment step yet; it's set to run on master branch only.

.circleci/config.yml
```diff
@@ -0,0 +1,186 @@
...
+      IMAGES: scope cloud-agent
+    steps:
+      - checkout
+      - setup_remote_docker
+      - attach_workspace:
+          at: .
+      # PATH must be set here as circle currently does not support string interpolation
+      # for environment
+      # https://discuss.circleci.com/t/bin-sh-mkdir-command-not-found/8710
+      - run: export PATH="$PATH:$HOME/.local/bin"
```
leth: I think this `export` might not work: https://circleci.com/docs/2.0/env-vars/#using-bash_env-to-set-environment-variables

.circleci/config.yml
```diff
@@ -0,0 +1,186 @@
...
+          mv client/app client/test /home/weave/
+          cd /home/weave; yarn run lint
+          cd /home/weave; yarn test
+
+  xplatform-build:
+    <<: *defaults
+    steps:
+      - checkout
+      - run: GOARCH=arm  make BUILD_IN_CONTAINER=false GO_BUILD_INSTALL_DEPS= prog/scope
+      - run: GOOS=darwin make BUILD_IN_CONTAINER=false GO_BUILD_INSTALL_DEPS= prog/scope
```
leth: Does this make any artefacts that we need to keep?

.circleci/config.yml
```diff
@@ -0,0 +1,186 @@
...
+
+defaults: &defaults
+  working_directory: /go/src/github.com/weaveworks/scope
+  docker:
+    - image: weaveworks/scope-backend-build:circle-2.0-4ad2f14
+
+client-defaults: &client-defaults
+  working_directory: /home/weave/scope
+  docker:
+    - image: weaveworks/scope-ui-build:circle-2.0-2885645
```
leth: These two are on different versions, which seems odd?

.circleci/config.yml
```diff
@@ -0,0 +1,186 @@
...
+          requires:
+            - client-test
+            - integration-tests
+
+jobs:
+  lint:
+    <<: *defaults
+    steps:
+      - checkout
+      - run: make BUILD_IN_CONTAINER=false lint
```
leth: Would it make sense to move `BUILD_IN_CONTAINER=false` up to the job defaults?

.circleci/config.yml
```diff
@@ -0,0 +1,186 @@
...
+
+  unit-test:
+    <<: *defaults
+    parallelism: 1
```
leth: Does `parallelism: 1` do anything? As far as I can tell from the docs, I don't think so?

.circleci/config.yml
bboreham: no

.circleci/config.yml
bboreham: Once the changes are merged to master I will update to a master tag, the same tag for both images.

.circleci/config.yml
bboreham: I'm not sure that environment variables override settings inside `Makefile` in the same way this does.

.circleci/config.yml
bboreham: No, it's there because I edited it down from 2 to 1 to optimise CircleCI resource usage, and I left the line in place in case I wanted to put it back up again.

.circleci/config.yml
bboreham: Removed the line and comment.

bboreham:
I tested the `deploy` step and addressed comments, so I think this is good to go now.

bboreham:
Just realised I included the commit from #3334 in here to stop tests failing, so I won't merge this until #3334 is approved.

rade:
Yeah!

PR #3335, opened 2018-08-27
Simplify fetch of IP addresses in a namespace

bboreham:
Code was unnecessarily fetching all links then fetching all addresses filtered by link, when we can just get the addresses without any filter.


bboreham:
I haven't seen that test failure before.  It passed on re-try.

PR #3337, opened 2018-09-03
Use new Search component

### LINKED ISSUE #3314: Use new search bar component
We should replace the current search bar:

image

with the new component from the `ui-components` repo:

http://weaveworks-ui-components.s3-website-us-west-2.amazonaws.com/#/components/search

fbarl:
Resolves #3314.

Transferring search context between Weave Cloud aspects is to be done in separate PRs.


client/app/scripts/components/search.js
```diff
@@ -1,27 +1,61 @@
...
+  width: 100%;
+`;
+
+const SearchHint = styled.div`
+  font-size: ${props => props.theme.fontSizes.tiny};
+  color: ${props => props.theme.colors.purple400};
+  transition: transform 0.3s 0s ease-in-out, opacity 0.3s 0s ease-in-out;
+  text-align: left;
+  padding: 0 1em;
+  opacity: 0;
```
foot: maybe a margin-top:2px?

fbarl:
> Only issue: new default behaviour of cmp/Search is to pin items onBlur. I think this feels quite strange when you: filter > click a node > close > topo is filtered w/ pin and everything has disappeared.

That's a good point - as discussed, we can resolve this with https://github.com/weaveworks/ui-components/issues/365.

PR #3353, opened 2018-09-24
Rework UI build to improve caching and fix packaging issue

### LINKED ISSUE #3351: package.json content in weave-scope.tgz not always up to date
For example, commit https://github.com/weaveworks/scope/commit/3f66348ad78dba8f000ec5a9cb5c314f66d63417 bumps the `weaveworks-ui-components` version to `0.12.2`, but looking into the corresponding package build https://s3.amazonaws.com/weaveworks-js-modules/weave-scope/3f66348a/weave-scope.tgz, there the `package.json` file still says `"weaveworks-ui-components": "0.11.14"`.

Looks like a caching issue with the UI part of our build pipeline, possibly related to #3333.

cc @bboreham @guyfedwards @foot

leth:
Stop baking the toolchain and dependencies into the build image. Instead, run the install step each time, but use volume mounts or CircleCI caching to keep the happy path fast.

Previously, some parts of the client (UI) directory were baked into the build image, and some parts were mounted or copied into the build environment. As a result, files baked into the build image require a two-step update for changes to take effect in CI.

Now, for dockerised builds, we pre-install very little into the build image and mount the whole directory into the build environment. However, we do overlay a volume on the node_modules folder to allow the standard build toolchain to be separate from the host build toolchain.

Non-dockerised builds (e.g. CI) are now more similar to the dockerised versions.

Fixes #3351

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,11 +62,19 @@ jobs:
     <<: *client-defaults
     steps:
       - checkout
-      # Convoluted set of steps here to mimic Makefile actions of bind-mounting different dirs into container
-      - run: mv client/app /home/weave/
-      - run: cd /home/weave; mkdir build         ; yarn run build         ; mv build          scope/client
-      - run: cd /home/weave; mkdir build-external; yarn run build-external; mv build-external scope/client
-      - run: cd /home/weave; mkdir tmp           ; yarn run bundle        ; mv tmp            scope
+      - restore_cache:
+          keys:
+          - yarn-cache-{{ checksum "client/yarn.lock" }}
+      - run: cd client; yarn install
+      - save_cache:
+          key: yarn-cache-{{ checksum "client/yarn.lock" }}
+          paths:
+            - "/home/weave/scope/.cache/yarn"
+      - run: |
+          cd client
+          yarn run build
+          yarn run build-external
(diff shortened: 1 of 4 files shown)


bboreham:
Can you outline what you intended to do, so I don't have to reverse the intent out of the code?

leth:
Sorry @bboreham! You're right, it was missing any explanation. Hopefully it makes sense now!

One improvement I'll make in a few minutes:

- [ ] add a Makefile target for the build toolchain

bboreham:
In the CI build the cache is basically empty - do you think it will work better if we merge it, or still more work required?

leth:
I think I need to update the tag for the UI build image in the CI config, to pull in the `XDG_CACHE_HOME` var, before things will work properly.

The cache dirs work in the local build; I dont think I've messed anything up, so it should work in CI too 🤞

fbarl:
I think it would be great if we could merge the changes that fix the packaging issue, which is a blocker for having latest Scope in Weave Cloud (and https://github.com/weaveworks/service-ui/pull/3202 in particular) and then try to further improve the caching in a separate PR :)

Does that make sense @leth?

leth:
I'm just doing some tests to check that the caching stuff works; It should be ready to go after lunch!

PR #3360, opened 2018-09-28
Update CI UI build container to master-c0b60a16

leth:
This is a follow up to #3353. It should make the cache directory be properly used.


bboreham:
How about you change the cache identifier, e.g. to `yarn-cacheb-...`, then it will stop saying "I've already got one with that identifier".

PR #3369, opened 2018-10-02
Remove weaveutil and weave from Dockerfile.cloud-agent

### LINKED ISSUE #3341: Remove weaveutil and weave from cloud-agent container
The cloud-agent container (built from `scope/docker/Dockerfile.cloud-agent`) only runs the probe, which no longer calls `weave`, therefore we can remove these files from that container and move them to the `scope` container.

bhavin192:
- Moves the `weave` and `weaveutil` from `Dockerfile.cloud-agent` to
  `Dockerfile.scope`
- Fixes #3341

### DIFF
--- a/docker/Dockerfile.cloud-agent
+++ b/docker/Dockerfile.cloud-agent
@@ -2,7 +2,6 @@ FROM alpine:3.7
 WORKDIR /home/weave
 RUN apk add --update bash conntrack-tools iproute2 util-linux curl && \
 	rm -rf /var/cache/apk/*
-ADD ./weave ./weaveutil /usr/bin/
 COPY ./scope /home/weave/
 ENTRYPOINT ["/home/weave/scope", "--mode=probe", "--no-app", "--probe.docker=true"]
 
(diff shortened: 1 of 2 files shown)


bboreham:
Thanks!  I've kicked off full integration tests.

Have you done any analysis to check there are no dependencies left, or are you relying on the statement in #3341

bhavin192:
@bboreham right now I'm relying on the references from the issue. But can you help me to test this out? I was not able to find instructions to build this images and run them.

dholbach:
If you build scope locally (you might have to merge from master to pull in #3382 for the build to pass), does running `./scope launch` and pointing the browser at it work for you? I'm getting an empty browser window.

bboreham:
I tried out the cloud-agent image in our test cluster, and it worked fine.

I also did the following analysis:

In `weaveworks/scope` directory:

    grep -nH -w weave -r . --include=*.go --exclude-dir=vendor --exclude=*test.go

This gives us 76 candidate matches, of which many are in tests and comments.

After eliminating all non-matches, the only use of `weave` is in `common/weave/client.go`:
  * `client.Expose()` calls `weave ps` and `weave expose`
  * `client.Expose()` is called by `updateDNS()`.
  * `updateDNS()` is called by `NewWeavePublisher()`
  * `NewWeavePublisher()` is called from `prog/app.go`

so I conclude that it is indeed only used in the app.

Thanks!

PR #3370, opened 2018-10-02
Report installed docker version

### LINKED ISSUE #1376: We should report what Docker version is installed

teisena:
Report docker version found on the system after it was obtained during version check.
    
Fixes #1376

### DIFF
--- a/scope
+++ b/scope
@@ -92,6 +92,8 @@ check_docker_version() {
         exit 1
     fi
 
+    echo "Docker version installed: $DOCKER_VERSION"
+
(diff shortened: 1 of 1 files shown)


bboreham:
Thanks for the contribution!

However, I think what #1376 is referring to is sending the information from the Scope probe to the app (back-end).  The information-gathering process in the probe is called "reporting".
 
So we cannot accept this change as-is.

teisena:
Ohh, understood, thanks for the explanation.

PR #3382, opened 2018-10-10
Fix UI toolchain dir permissions error

leth:
Create the UI toolchain dir from the Makefile to prevent docker from creating it with root as the owner


bboreham:
I had to chown two directories:

    .cache/build_node_modules/
    client/node_modules/

so is this one change enough?

dholbach:
Yes, I'm seeing the same problem as Bryan.

leth:
Ah yes, the extra one is also created by docker, thanks!

bboreham:
I tried this and it still didn't work.  There are two directories and only one change to the makefile.

PR #3383, opened 2018-10-11
Add more Opentracing detail to the app

bboreham:
This lets you see much more clearly which parts of the processing are taking time.
Unfortunately we have to add a thousand `context.Context` parameters, but that can't be helped - this is the Go way.

I used reflection to get the name of `Map` and `Reduce` objects, to avoid typing in specific span calls in each of them.

The overall result can look like this (using Jaeger):

image

### DIFF
--- a/app/api_topologies.go
+++ b/app/api_topologies.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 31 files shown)


2opremio:
Can you fix the conflicts? Otherwise LGTM

PR #3384, opened 2018-10-11
In multitenant app, drop all nodes for big topologies

bboreham:
As a protection against overloading the app server, drop topologies that have really large node counts. In practice we only see this with runaway numbers of zombie processes.

I've only wired this in to the AWS collector, since that is where we felt the pain.

Also some refactoring of the AWS collector parameters where the duplication started to get annoying.

### DIFF
--- a/app/multitenant/aws_collector.go
+++ b/app/multitenant/aws_collector.go
@@ -107,14 +107,10 @@ type AWSCollectorConfig struct {
 }
 
 type awsCollector struct {
-	userIDer  UserIDer
+	cfg       AWSCollectorConfig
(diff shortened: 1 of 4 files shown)


satyamz:
@bboreham I am looking into this, 
How do I run `multi tenant app `? I mean apart from options `-app.collector` & `-app.collector.s3`, do I need to setup anything?
I am setting up `dynamodb` & `s3` in my aws account.

bboreham:
@satyamz thanks for looking.
I think you also need `--app.userid.header=X-Scope-OrgID` to set how it finds the tenant ID.

PR #3385, opened 2018-10-11
[WIP] Add the message and reason fields for pods

### LINKED ISSUE #2989: Show more information for failed pods
This is what I get from Scope:
image

compare this from `kubectl`:
```
$ kubectl describe -n sock-shop pod carts-db-549516398-5dnvn
Name:		carts-db-549516398-5dnvn
Namespace:	sock-shop
Node:		gke-sock-shop-default-pool-9652982b-7pbs/
Start Time:	Thu, 30 Nov 2017 22:10:49 +0000
Labels:		name=carts-db
		pod-template-hash=549516398
Annotations:	kubernetes.io/created-by={"kind":"SerializedReference","apiVersion":"v1","reference":{"kind":"ReplicaSet","namespace":"sock-shop","name":"carts-db-549516398","uid":"ef4cef2b-cde6-11e7-91c4-42010a9a008...
Status:		Failed
Reason:		Evicted
Message:	Pod The node was low on resource: [MemoryPressure].
...
```

zkry:
Fixes #2989
I think I've got a viable PR for this issue. 🙂

The only problem is that the new reason and message fields always show, even if they are blank. I was trying to figure out how to optionally show these if the status was an unhealthy or blank, but there didn't seem to be an easy way. A possible idea that came to my mind was to add another field to the report.MetadataTemplate struct to indicate to the frontend to hide if it is the zero value.

What do you think about this?

### DIFF
--- a/client/app/scripts/charts/nodes-grid.js
+++ b/client/app/scripts/charts/nodes-grid.js
@@ -18,7 +18,7 @@ import { getNodeColor } from '../utils/color-utils';
 
 
 const IGNORED_COLUMNS = ['docker_container_ports', 'docker_container_id', 'docker_image_id',
-  'docker_container_command', 'docker_container_networks'];
+  'docker_container_command', 'docker_container_networks', 'kubernetes_reason', 'kubernetes_message'];
 
 
 const Icon = styled.span`
--- a/client/app/scripts/components/node-details/node-details-info.js
+++ b/client/app/scripts/components/node-details/node-details-info.js
@@ -41,6 +41,9 @@ class NodeDetailsInfo extends React.Component {
       <div className="node-details-info">
         {rows.map((field) => {
           const { value, title } = formatDataType(field, timestamp);
+          if (field.hideIfEmpty && value === '') {
+            return null;
+          }
           return (
             <div className="node-details-info-field" key={field.id}>
               <div className="node-details-info-field-label truncate" title={field.label}>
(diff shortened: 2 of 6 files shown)


bboreham:
Thanks for this @zkry !

I might suggest looking at the pod status conditions - they will say whether it is ready, running, terminated, etc., and so you can fill in the details accordingly.

I would also say it's worth bringing some of that info out to the UI.

zkry:
That part makes sense. Since a lot of those fields can be irrelevant (reason, message) should we sill display them in the UI even if they are blank?

bboreham:
@zkry feel free to experiment.  Try to think about how you, or another user, would use this information on the screen. I suggest it's better to be concise - if you know some fields are useless then don't take up space with them.

But we have people on the project who are much more skilled in visual design than I am - if you put up some screen shots I'm sure you'll get comments.

zkry:
Perfect. I'll look into the best way to do this then. 👍

zkry:
I think I am much closer if not done. So what I decided to do was to display the Pods 'Reason' and 'Message' information on the details card only if it is not empty. If the reason and/or message is empty, it will be hidden from the card altogether.

I was having a really hard time testing this in a Kubernetes environment as I couldn't get a pod to be listed as failed with a reason and message (as in the picture of issue #2989).

bboreham:
@zkry you could create pods via a `Job` which have invalid config, or which run `/bin/false` to indicate failure.

zkry:
Thanks 🙂 Yeah, I actually kind of got stuck. It seemed like that no matter what I thew at the Kubernetes configuration, I couldn't get a Failed pod to stay on the console. The kubectl tool might show otherwise but the scope ui would never show them.

bboreham:
Closing this to tidy up.  Happy to re-open if you come back to it.

PR #3386, opened 2018-10-13
Probe: Rate-limit report publishing

### LINKED ISSUE #3150: Lots of 'Dropping report' logs under Kubernetes
I observe at startup the Scope probe logs a lot of these:

```
<probe> WARN: 2018/04/23 14:59:16.547846 Dropping report to frontend.dev.weave.works.
<probe> WARN: 2018/04/23 14:59:16.603809 Dropping report to frontend.dev.weave.works.
<probe> WARN: 2018/04/23 14:59:16.632496 Dropping report to frontend.dev.weave.works.
<probe> WARN: 2018/04/23 14:59:16.865601 Dropping report to frontend.dev.weave.works.
<probe> WARN: 2018/04/23 14:59:16.865617 Dropping report to scope.weave.svc.cluster.local.
<probe> WARN: 2018/04/23 14:59:16.877569 Dropping report to frontend.dev.weave.works.
```

Further investigation shows this is triggered by `reporter.podEvent` being called for every pod in the cluster at startup. This amount of reporting is unnecessary.

Further, we don't filter these shortcut reports by node name, as we do in `podTopology`

bboreham:
Fixes #3150 

Rate-limit report publishing so that, if many shortcut reports are produced in quick succession, they will tend to get merged together before sending.

Also expand the queue size for shortcut reports, to avoid holding up the producer so much.

The behaviour of the Go rate-limiter with a burst size of 1 is that it will let 1 report go through immediately, then the next one will wait until the time specified (300ms by default).  If nothing happens for more than 300ms the next report will go immediately, and so on.

### DIFF
--- a/probe/probe.go
+++ b/probe/probe.go
@@ -83,7 +83,7 @@ func New(
 		spyInterval:     spyInterval,
 		publishInterval: publishInterval,
 		publisher:       publisher,
-		rateLimiter:     rate.NewLimiter(rate.Every(publishInterval)*10, 1),
+		rateLimiter:     rate.NewLimiter(rate.Every(publishInterval/100), 1),
 		noControls:      noControls,
 		quit:            make(chan struct{}),
 		spiedReports:    make(chan report.Report, spiedReportBufferSize),
--- a/probe/probe.go
+++ b/probe/probe.go
@@ -1,17 +1,20 @@
 package probe
 
 import (
+	"context"
 	"sync"
 	"time"
 
 	"github.com/armon/go-metrics"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/time/rate"
 
 	"github.com/weaveworks/scope/report"
 )
 
 const (
-	reportBufferSize = 16
+	spiedReportBufferSize    = 16
+	shortcutReportBufferSize = 1024
 )
 
 // ReportPublisher publishes reports, probably to a remote collector.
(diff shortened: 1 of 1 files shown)


rade:
I don't understand what the rate limiting achieves here. According to #3150 the objective of this PR is to "allow the pod-reporting code to run at full speed while merging those reports and not dropping so many on the way to the app." Why wouldn't increasing the chan buffer size alone achieve that? And that wouldn't have the rather undesirable side-effect of delaying short-cut reports by up to 300ms.

bboreham:
Currently, shortcut reports are pulled out of the chan as fast as possible, so they never queue up. They are the passed immediately to the app client which either sends them or drops them. 

And the rate-limiter doesn’t delay the first shortcut report, only subsequent ones in rapid succession. I could have it go faster - 300ms was just a first guess.

rade:
> And the rate-limiter doesn’t delay the first shortcut report, only subsequent ones in rapid succession.

I get that, but the whole point of shortcut reports - as the name suggests - is to send them quickly. Introducing _any_ delay undermines that. OTOH _dropping_ a report is the ultimate delay.

The issue focuses on the deluge of shortcut reports sent on startup. Does the dropping of these actually cause any problems?

rade:
> How about if the delay was 10ms? Well below human reaction times.

Fine by me.

Issue #2511, opened 2017-05-09
Graph icon is confusing


bowenli:
This:
<img width="49" alt="screen shot 2017-05-09 at 14 34 29" src="https://cloud.githubusercontent.com/assets/915421/25873775/a1795fbc-34c4-11e7-9eef-cc7c7c354771.png">
is the share icon.
Confusing when used in a web app:
![share-icon-hi](https://cloud.githubusercontent.com/assets/915421/25873789/b45a5da2-34c4-11e7-8d44-dfec82ce6d1f.png)

fbarl:
Yeah, my thoughts were the same when I was making that selector responsive.

PR #3387, opened 2018-10-13
scripts: replace share icon with sitemap (graph)

hexmind:
#2511 
replace share icon with sitemap (graph)
sitemap


bboreham:
Thanks!

PR #3402, opened 2018-10-22
Slack Status link is broken

muthumalla:


muthumalla:
Removed that line altogether as suggested. Thank you @rndstr

Issue #3400, opened 2018-10-22
Error validating examples/k8s/cluster-role.yaml


dholbach:
```
[daniel@reef site (⎈ |minikube:default)]$ kubectl apply -f ../examples/k8s/
clusterrolebinding.rbac.authorization.k8s.io/weave-scope created
namespace/weave created
podsecuritypolicy.extensions/weave-scope created
serviceaccount/weave-scope created
service/weave-scope-app created
error validating "../examples/k8s/cluster-role.yaml": error validating data: [ValidationError(ClusterRole.rules[5]): unknown field "apigroups" in io.k8s.api.rbac.v1.PolicyRule, ValidationError(ClusterRole.rules[5]): unknown field "resourcenames" in io.k8s.api.rbac.v1.PolicyRule]; if you choose to ignore these errors, turn validation off with --validate=false
error validating "../examples/k8s/deploy.yaml": error validating data: ValidationError(Deployment.spec): missing required field "selector" in io.k8s.api.apps.v1.DeploymentSpec; if you choose to ignore these errors, turn validation off with --validate=false
error validating "../examples/k8s/ds.yaml": error validating data: ValidationError(DaemonSet.spec): missing required field "selector" in io.k8s.api.apps.v1.DaemonSetSpec; if you choose to ignore these errors, turn validation off with --validate=false
[daniel@reef site (⎈ |minikube:default)]$ 
```

PR #3403, opened 2018-10-23
Fix issues in k8s yaml files

Akash4927:
- Change `apigroups` to `apiGroups` in cluster-role.yaml
- Change `resourcenames` to `resourceNames` in cluster-role.yaml
- Add `selector` in deploy.yaml and ds.yaml

fixes: #3400

### DIFF
--- a/examples/k8s/cluster-role.yaml
+++ b/examples/k8s/cluster-role.yaml
@@ -51,9 +51,9 @@ rules:
   verbs:
   - list
   - watch
-- apigroups:
+- apiGroups:
   - extensions
-  resourcenames:
+  resourceNames:
   - weave-scope
   resources:
   - podsecuritypolicies
--- a/examples/k8s/deploy.yaml
+++ b/examples/k8s/deploy.yaml
@@ -11,6 +11,9 @@ metadata:
   namespace: weave
 spec:
   replicas: 1
+  selector:
+    matchLabels:
+      app: weave-scope
   revisionHistoryLimit: 2
   template:
     metadata:
--- a/examples/k8s/ds.yaml
+++ b/examples/k8s/ds.yaml
@@ -11,6 +11,9 @@ metadata:
   namespace: weave
 spec:
   minReadySeconds: 5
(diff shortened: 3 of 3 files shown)


dholbach:
Looks like the namespace 'weave' needs to be created still:
```
daniel@reef examples (⎈ |minikube:default)]$ kubectl apply -f k8s/
clusterrolebinding.rbac.authorization.k8s.io/weave-scope created
clusterrole.rbac.authorization.k8s.io/weave-scope created
namespace/weave created
podsecuritypolicy.extensions/weave-scope created
serviceaccount/weave-scope created
service/weave-scope-app created
Error from server (NotFound): error when creating "k8s/deploy.yaml": namespaces "weave" not found
Error from server (NotFound): error when creating "k8s/ds.yaml": namespaces "weave" not found
[daniel@reef examples (⎈ |minikube:default)]$
```

dholbach:
Only slightly related to this PR: is this the best way to spin up Scope in a kubernetes environment (asking for the contributor in 3149 for example).

Shall we document this somewhere? E.g. 
 - spin up minikube
 - apply example/k8s
 - port-forward to access weave-scope-app

Or is there a better way?

(And thanks Akash for looking into this issue!)

satyamz:
@dholbach steps look good to me. Minor feedback on the above steps,
 
- We need to include installation of Docker.
- Scope app service can be exposed as a `NodePort` on the local setup. (Sometimes, namespace & other filters doesn't work behind `kube-proxy` and `port-forward`)

It'd be good if you can file an issue containing this info. I will have someone to send PR for the same.

Thanks.

Akash4927:
I think for the local setup, `NodePort` is a good idea to access the Scope.

dholbach:
I filed https://github.com/weaveworks/scope/issues/3406 for this. Thanks a lot.

dholbach:
Are we happy with this PR now or do we want to change something here?

satyamz:
Thanks!

Issue #3408, opened 2018-10-26
client: namespaces not wrapped


joelmukuthu:
When you have many namespaces, they are not all visible from the widget at the bottom:

![weave-namespaces-nowrap](https://user-images.githubusercontent.com/374638/47567686-8b62d300-d92f-11e8-9257-0c797e9d2759.png)

## What you expected to happen?

To see all namespaces listed:

![weave-namespaces-wrapped](https://user-images.githubusercontent.com/374638/47567864-e5fc2f00-d92f-11e8-9d77-515901bc76b2.png)

## What happened?

The list of namespaces was cut off at the right.

## How to reproduce it?

Have many namespaces, such that they don't all fit in one row.

## Versions:

Version 1.9.1 on weave-scope-app-2064699473-389
Plugins:n/a

## Anything else we need to know?

I had to hide the `div.footer` element to view the full list of namespaces as well (and to take the screenshots). Would be nice to fix that too.

```
$ scope
$ docker version
$ uname -a
$ kubectl version
```

rade:
We already have #3096 for this. And there's a PR too - #3117.

PR #3409, opened 2018-10-26
fix(client): allow wrapping items in namespace-list

joelmukuthu:
Not sure if this is the right fix, but this attempts to fix https://github.com/weaveworks/scope/issues/3408


joelmukuthu:
I'd have also liked to fix the issue with the footer floating on top of the namespace-list, but that seems more of a UX issue rather than UI

joelmukuthu:
Closed with deference to https://github.com/weaveworks/scope/pull/3117

PR #3413, opened 2018-10-28
Don't count page cache as part of the process memory

sharanp98:
The container_memory_usage_bytes metric contains the page cache allocations
done by the kernel when the process does some I/O to speed subsequent accesses
eg. when reading 32 bytes a file, the kernel I/O scheduler will optimistically
load more data into that page cache to speed subsequent read operations.

This page cache is an implementation detail of disk I/O and the process have
roughly no control over it (not exactly true, but a good enough approximation).

RSS is a truer measure of memory consumption ie. the actual number of system
memory pages used by the process eg. when allocating 10MB on the heap with
malloc() but not accessing it, that memory isn't actually pinned in RAM and not
accounted for in RSS.


satyamz:
This is invalid PR. Closing this.

PR #3419, opened 2018-11-05
Disable host and endpoint reporting in Kubernetes cluster mode

bboreham:
When the probe is run with `--probe.kubernetes.role=cluster`, disable host reporting and endpoint reporting since it duplicates what is being done by the per-host probe.

This was missed from #3274

### DIFF
--- a/prog/probe.go
+++ b/prog/probe.go
@@ -213,40 +213,42 @@ func probeMain(flags probeFlags, targets []appclient.Target) {
 	}
 
 	p := probe.New(flags.spyInterval, flags.publishInterval, clients, flags.noControls)
-
-	hostReporter := host.NewReporter(hostID, hostName, probeID, version, clients, handlerRegistry)
-	defer hostReporter.Stop()
-	p.AddReporter(hostReporter)
-	p.AddTagger(probe.NewTopologyTagger(), host.NewTagger(hostID))
-
 	var processCache *process.CachingWalker
-	if flags.procEnabled {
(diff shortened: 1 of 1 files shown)


prog/probe.go
```diff
@@ -213,40 +213,42 @@ func probeMain(flags probeFlags, targets []appclient.Target) {
...
-	})
-	defer endpointReporter.Stop()
-	p.AddReporter(endpointReporter)
+		endpointReporter := endpoint.NewReporter(endpoint.ReporterConfig{
+			HostID:       hostID,
+			HostName:     hostName,
+			SpyProcs:     flags.spyProcs,
+			UseConntrack: flags.useConntrack,
+			WalkProc:     flags.procEnabled,
+			UseEbpfConn:  flags.useEbpfConn,
```
rade: Since this now only executes in the non-k8s-cluster-role case, you can get rid of the code that clears all these flags in https://github.com/weaveworks/scope/blob/71972c302ad9612b243a47fadfeb9489800e7157/prog/probe.go#L115.

bboreham: Yeah, I thought it was better left as-is than to cherry-pick those that are used now. It’s clearer.

rade: but _none_ of the flags set earlier on are used anymore, except for the two that are clearly kubernetes related, i.e.
```
		flags.kubernetesKubeletPort = 0
		flags.kubernetesEnabled = true
```

bboreham: I guess the fact that we are having this debate proves my point, but still, I see `spyProcs` cleared on line 118 and used on line 126.

PR #3426, opened 2018-11-13
Upgrade to font-awesome 5

foot:
- Switch to fa5 icons and tweak font-sizes a bit
- Load up fa5 instead of 4
- Include fa4 shim so that old probes don't lose their icons
- Switch the probe over to use the new icons in controls too.


client/app/styles/_base.scss
```diff
@@ -737,6 +739,8 @@ a {
         color: $text-secondary-color;
         font-size: $font-size-small;
 
+        // TODO: remove this in the future you can't count on .fa but I don't
+        // know where label .fa comes from.
         .fa {
           margin-left: 0.5em;
         }
```
fbarl: I think it's safe to remove this chunk :)

PR #3427, opened 2018-11-13
Update aws-sdk-go library to latest release

bboreham:
From version 1.0.5 to version 1.15.74.
Simply because the old version was really old - December 2015.


fbarl:
@bboreham Is this still a WIP or it's ready for review?

I could take a look if I get more context about which parts of Scope (if any) should be affected :)

bboreham:
It is done as-is; nothing should be affected because it's just a version update.
But since there is no real driver I don't know if we should do this at all.

PR #3428, opened 2018-11-14
Check if host record has uptime in billing emitter

bboreham:
This is to address an inconsistency introduced by #3274.
#3419 wasn't enough to fix it, because the Kubernetes reporter [still injects a host record.](https://github.com/weaveworks/scope/blob/4ac9d1f946efc530ee686ffff380076a7fd7cb30/probe/kubernetes/reporter.go#L372-L400).

I could just return early if there are no hosts reported. Advice welcome.

### DIFF
--- a/app/multitenant/billing_emitter.go
+++ b/app/multitenant/billing_emitter.go
@@ -15,6 +15,10 @@ import (
 	"github.com/weaveworks/scope/report"
 )
 
+const (
+	uptime = "uptime"
+)
+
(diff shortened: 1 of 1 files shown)


rade:
This is too horrific.

Suggestion:... move the horrible hack to the app; don't send a fake host record. The app-side hack would need the service IPs, which we are sending already as part of the service records.

bboreham:
Replaced by #3432

PR #3432, opened 2018-11-19
Move "hideous hack" for Kubernetes service network from probe to app

bboreham:
This is to address an inconsistency introduced by #3274, reporting Kubernetes info from a cluster-level probe, and removes a lie from the probe data.

Suggested at https://github.com/weaveworks/scope/pull/3428#issuecomment-439064091

Rather than do the work every time the probe sends a report, we do it every time the app renders endpoints, which should be once per merged report set.

Replaces #3428

### DIFF
--- a/probe/kubernetes/reporter.go
+++ b/probe/kubernetes/reporter.go
@@ -2,7 +2,6 @@ package kubernetes
 
 import (
 	"fmt"
-	"net"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


bboreham:
> If the app receives old reports, nothing breaks, it just duplicates some work, right?

Right, if the probe has already put the service route in then this should re-add it which is a no-op.

PR #3437, opened 2018-11-22
Release 1.10

bboreham:
See changelog for contents of release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,143 @@
+## Release 1.10.0
+
+Highlights:
+
+- Add Kubernetes Persistent Volume snapshot and clone operations
+	[#3355](https://github.com/weaveworks/scope/pull/3355)
+
+- Kubernetes objects can be reported just once in a cluster, instead
+  of reporting the same data from every node.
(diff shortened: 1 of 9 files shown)


satyamz:
Hi @bboreham This looks good! 
It'd be good if you can include these two files:
- https://github.com/weaveworks/scope/blob/master/examples/k8s/deploy.yaml
- https://github.com/weaveworks/scope/blob/master/examples/k8s/ds.yaml

bboreham:
How do you mean "include" ?  If they're on master they will be in the release.

satyamz:
I meant, if you can bump up the version in these files too.

bboreham:
Oh, right.  Release process needs updating https://github.com/weaveworks/scope/wiki/Release-Process

Really we should add the single-probe-per-cluster config too, but the release doesn't need to wait for that.

satyamz:
By this `single-probe-per-cluster`, you mean k8s probe?

bboreham:
I mean the feature added at #3274 whereby we can stop all the per-node probes reporting Kubernetes info and just report it once for the whole cluster.

satyamz:
Ok! This would be good addition. We can include that later in the example files. Let's go ahead with current changes.

Issue #3438, opened 2018-11-23
Error fetching app details: Get http://127.0.0.1:4040/api: dial tcp 127.0.0.1:4040: connect: connection refused


zhwill:
i can't access the website：http://192.168.92.6:4040/，when i accesss ,it is black.
![weavescopeerror](https://user-images.githubusercontent.com/33590311/48929153-dff27180-ef20-11e8-8381-d732162e9e32.png)

i use the follow command to install scope
```
curl -L git.io/scope -o /usr/local/bin/scope
chmod a+x /usr/local/bin/scope
scope launch
```

and the OS version is:
```
[root@localhost ~]# cat /etc/redhat-release 
CentOS Linux release 7.5.1804 (Core) 
[root@localhost ~]# uname -sr
Linux 4.16.3-1.el7.elrepo.x86_64
[root@localhost ~]# 
```

docker container is running:
```
[root@localhost ~]# docker ps
CONTAINER ID        IMAGE                     COMMAND                  CREATED              STATUS              PORTS               NAMES
8b098a695874        weaveworks/scope:1.10.0   "/home/weave/entrypo…"   About a minute ago   Up About a minute                       weavescope
[root@localhost ~]# 
```
docker info:
```
[root@localhost ~]# docker info
Containers: 1
 Running: 1
 Paused: 0
 Stopped: 0
Images: 32
Server Version: 18.09.0
Storage Driver: overlay
 Backing Filesystem: xfs
 Supports d_type: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: c4446665cb9c30056f4998ed953e6d4ff22c7c39
runc version: 4fc53a81fb7c994640722ac585fa9ca548971871
init version: fec3683
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.16.3-1.el7.elrepo.x86_64
Operating System: CentOS Linux 7 (Core)
OSType: linux
Architecture: x86_64
CPUs: 2
Total Memory: 1.925GiB
Name: localhost.localdomain
ID: FHGI:SFWY:IVIZ:FEBC:CKZL:WUUB:LWAS:IP7Y:OKYD:VW6G:4P3N:N2JD
[...]

PR #3439, opened 2018-11-23
UI Build: stop deleting static ui files when building external ui

bboreham:
This can cause an empty static UI if things build in an unfortunate order.

Fixes #3438

PR #3440, opened 2018-11-23
Add a check that the top-level UI is not empty

bboreham:
Should detect the symptom at #3438


bboreham:
Test failed as desired on bad build: https://circleci.com/gh/weaveworks/scope/10998

bboreham:
Test succeeded after including the fix in #3439 https://circleci.com/gh/weaveworks/scope/11023

PR #3451, opened 2018-11-28
[WIP] Probe: Better mapping of NATted connections

bboreham:
Now that we get NAT information via netlink we have a flag to distinguish `SRC_NAT` from `DST_NAT`, so we can do a better job of mapping connections. See extensive comment in `nat.go` for explanation of the approach.

One example how we get a better result is that incoming connections to a Kubernetes NodePort are visible with their remote IP address - previously they did not get mapped.  Unfortunately, when using an AWS ELB the source endpoint is excluded because its address is on the same network as the host where the probe runs.

Also stop special-casing the Docker bridge in Kubernetes mode.  Most often, Kubernetes uses a bridge named `cbr0` so the 'docker0' bridge should be locally-scoped, and exceptionally users can turn that off with `--probe.docker.bridge=""`

### DIFF
--- a/probe/endpoint/nat.go
+++ b/probe/endpoint/nat.go
@@ -38,61 +38,66 @@ func endpointNodeID(scope string, ip net.IP, port uint16) string {
 
 Some examples of connections with NAT:
 
+Here 10.32.0.X are pod addresses; 172.31.X.X are node addresses; 10.10X.X.X are service virtual addresses.
+
 Pod to pod via Kubernetes service
   picked up by ebpf as 10.32.0.16:47600->10.105.173.176:5432 and 10.32.0.6:5432 (??)
   NAT IPS_DST_NAT orig: 10.32.0.16:47600->10.105.173.176:5432, reply: 10.32.0.6:5432->10.32.0.16:47600
   We want: 10.32.0.16:47600->10.32.0.6:5432
-   - replace the destination (== NAT orig dst) with the NAT reply source
+   - replace the destination (== NAT orig dst) with the NAT reply source (A)
 
 Incoming from outside the cluster to a NodePort:
   picked up by ebpf as 10.32.0.1:13488->10.32.0.7:80
   NAT: IPS_SRC_NAT IPS_DST_NAT orig: 37.157.33.76:13488->172.31.2.17:30081, reply: 10.32.0.7:80->10.32.0.1:13488
   We want: 37.157.33.76:13488->10.32.0.7:80
-   - replace the source (== NAT reply dst) with the NAT original source
+   - replace the source (== NAT reply dst) with the NAT original source (B)
+  To match another probe with the other side of this connection, also want 37.157.33.76:13488->172.31.2.17:30081
+   - add NAT original dst as a copy of nat reply source (C)
 
 Outgoing from a pod:
   picked up by ebpf as 10.32.0.7:36078->18.221.99.178:443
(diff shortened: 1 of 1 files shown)


prog/probe.go
```diff
@@ -266,9 +266,7 @@ func probeMain(flags probeFlags, targets []appclient.Target) {
 	}
 
 	if flags.dockerEnabled {
-		// Don't add the bridge in Kubernetes since container IPs are global and
-		// shouldn't be scoped
-		if flags.dockerBridge != "" && !flags.kubernetesEnabled {
+		if flags.dockerBridge != "" {
```
rade: This is an improvement for when we have "unmanaged" containers running on a k8s host. However, it will also result in

1. spurious error messages when docker isn't running on a k8s host (as will happen by default on some future k8s releases)

2. incorrect treatment of the docker bridge as local if it is in fact the bridge used by k8s

We could avoid the first one by only erroring when the bridge name has been specified on the command line (rather than left to the default 'docker0' value).

Is the 2nd problem real?

probe/endpoint/nat.go
```diff
@@ -30,45 +30,102 @@ func makeNATMapper(fw flowWalker) natMapper {
...
+				if !ok {
+					replySrcNode = report.MakeNode(replySrcID)
+				}
+				newNode := replySrcNode.WithID(toID).WithLatests(map[string]string{
+					CopyOf: replySrcID,
+				})
+				rpt.Endpoint.AddNode(newNode)
+			}
+
+		}
```
rade: The difference in structure between the two 'big ifs' has me scratching my head. Could we regularise this to
```golang
if (f.Status & conntrack.IPS_SRC_NAT) != 0 {
    replyDstID := endpointNodeID(scope, f.Reply.Dst, f.Reply.DstPort)
    origSrcID := endpointNodeID(scope, f.Orig.Src, f.Orig.SrcPort)
    if replyDstID != origSrcID { ... }
}
if (f.Status & conntrack.IPS_DST_NAT) != 0 {
    replySrcID := endpointNodeID(scope, f.Reply.Src, f.Reply.SrcPort)
    origDstID := endpointNodeID(scope, f.Orig.Dst, f.Orig.DstPort)
    if replySrcID != origDstID { ... }
}
```
?

probe/endpoint/nat.go
```diff
@@ -30,45 +30,102 @@ func makeNATMapper(fw flowWalker) natMapper {
...
+			if !ok {
+				return
+			}
+			toID := endpointNodeID(scope, f.Orig.Dst, f.Orig.DstPort)
+
+			// replace destination with reply source
+			replySrcID := endpointNodeID(scope, f.Reply.Src, f.Reply.SrcPort)
+			if replySrcID != toID {
+				fromNode.Adjacency = fromNode.Adjacency.Minus(toID)
+				fromNode = fromNode.WithAdjacent(replySrcID)
```
rade: Don't you need to store `fromNode` in `rpt.Endpoint` after this?

prog/probe.go
bboreham: 2. I don't think it is practical to use the `docker0` bridge as the Kubernetes bridge nowadays except on a single-node cluster.
I think Flannel used to connect one machine's Docker bridge to another, but nowadays that is driven via CNI with its own bridge.  And Docker's own overlay networking will create a different bridge for each network.

fbarl:
@bboreham What's the status of this PR? Is there something blocking you from proceeding / merging it?

PR #3452, opened 2018-11-28
[WIP] Probe: report fewer closed connections

bboreham:
Specifically, only one for each (source address, destination address, destination port) triple.

It's fairly common for programs to open and close a lot of TCP connections to the same service, and we don't need to report all of them for Scope to paint the picture.

### DIFF
--- a/probe/endpoint/ebpf.go
+++ b/probe/endpoint/ebpf.go
@@ -51,7 +51,7 @@ type EbpfTracker struct {
 	debugBPF bool
 
 	openConnections   map[fourTuple]ebpfConnection
-	closedConnections []ebpfConnection
+	closedConnections map[fourTuple]ebpfConnection
 	closedDuringInit  map[fourTuple]struct{}
 }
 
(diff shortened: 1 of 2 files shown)


probe/endpoint/ebpf.go
```diff
@@ -288,7 +288,13 @@ func (t *EbpfTracker) handleConnection(ev tracer.EventType, tuple fourTuple, pid
...
 		if deadConn, ok := t.openConnections[tuple]; ok {
 			delete(t.openConnections, tuple)
-			t.closedConnections = append(t.closedConnections, deadConn)
+			// mask the source port on closed connections so we only report one for each destination
+			if deadConn.incoming {
+				tuple.fromPort = 0
+			} else {
+				tuple.toPort = 0
+			}
+			t.closedConnections[tuple] = deadConn
```
rade: The effect of this when connections are short-lived is that we'll mostly only report one connection per destination per host, thus losing the ability to differentiate between different processes (and hence containers and services) connecting to a particular destination. This will also make views unstable.

Perhaps include the pid in the key?

rade: > Perhaps include the pid in the key?

Except that would only work for outbound connections.

Another concern I have is that we won't be able to "match up" inbound and outbound connections, though I don't remember exactly how that code works.

bboreham: Thinking about it, there's probably a deeper problem of matching up short-lived connections from two different nodes.

Maybe if I take care to report the lowest-numbered connection (per pid) it would work better.

bboreham: > we'll mostly only report one connection per destination per host, thus losing the ability to differentiate between different processes (and hence containers and services) 

The ebpf reporter sees connections inside the container network namespace, so it will report different containers as different IP addresses.

bboreham:
Closing this in view of the review comments; hope to come back with a better solution.

PR #3530, opened 2018-12-08
Add adjacency between Pod and PVC when claim name and namespace are same

Akash4927:
- Name of the PVC can be the same in different namespaces, in this
scenario an application pod is connected to multiple PVC.
screenshot from 2018-12-07 22-46-05

- Add adjacency between Pod and PVC when `claim name` and `namespace` are same.


fbarl:
@bboreham would you mind reviewing this one?

PR #3534, opened 2018-12-14
Rename :orgId to :instanceId in metrics graph url

rndstr:
- Renames placeholders `:orgId` to `:instanceId` in
  `--app.metrics-graph`
- Adjusts frontend client to use the provided instanceId by its host
  application (Weave Cloud)

### DIFF
--- a/client/app/scripts/components/cloud-link.js
+++ b/client/app/scripts/components/cloud-link.js
@@ -9,7 +9,7 @@ import CloudFeature from './cloud-feature';
  * that is comprised of Weave Cloud related pieces.
  *
  * We support here relative links with a leading `/` that rewrite
- * the browser url as well as cloud-related placeholders (:orgId).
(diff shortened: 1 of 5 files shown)


bboreham:
What happens if an old client meets a new server?

rndstr:
> What happens if an old client meets a new server?

what exactly is client/server in this case? the probe has no play here, the app is given the format as config, and the host (weave cloud for us) of the frontend is passing in `instanceId=<value>`.

bboreham:
OK, so all the bits that know the name will get reloaded at the same time into the browser, so they will all change at once?

rade:
IIRC we traditionally haven't worried about compatibility between f/e and b/e. If something looks broken, a reload should fix it.

rndstr:
@bboreham the thing to note here is that it's currently broken in any case. so there is no need to manage a seamingless migration path.

PR #3540, opened 2018-12-21
Reverted ruby version to 2.3.0

Rich-Core:


Rich-Core:
@dholbach requested I revert his change because it has been reverted on the Weave.works website repo. The original change to 2.3.6 was made because I was looking for ways to increase build speed, and the warning - which you can see at the top of that Netlify build there - was a good start. It was reverted because it caused problems with local builds.

bboreham:
If the intention here is to improve Netlify builds, I'm not going to merge it until the Netlify build passes.

I can't see anything in the Netlify log that explains why it reports an error.

Issue #3550, opened 2019-01-08
One-to-many connections between a pod and many PVCs


Akash4927:
## What you expected to happen?

One pod can mount multiple volumes and those volumes can be represented as PVCs. Hence, one pod can have multiple PVCs. However, Scope shows only one mapping. 

It should show all the mappings.

PR #3553, opened 2019-01-09
Fix one pod to multiple PVC connection

Akash4927:
Read all the PVC source from a pod spec
and add adjacency between the pod and its PVCs.

screenshot from 2019-01-09 16-25-07

Signed-off-by: Akash Srivastava <akashsrivastava4927@gmail.com>

fixes: #3550

### DIFF
--- a/probe/kubernetes/pod.go
+++ b/probe/kubernetes/pod.go
@@ -2,6 +2,7 @@ package kubernetes
 
 import (
 	"strconv"
+	"strings"
 
 	"github.com/weaveworks/scope/report"
 
@@ -29,6 +30,7 @@ type Pod interface {
 	GetNode(probeID string) report.Node
 	RestartCount() uint
 	ContainerNames() []string
+	VolumeClaimNames() []string
 }
 
 type pod struct {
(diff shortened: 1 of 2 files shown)


probe/kubernetes/pod.go
```diff
@@ -94,8 +95,8 @@ func (p *pod) GetNode(probeID string) report.Node {
 		RestartCount:          strconv.FormatUint(uint64(p.RestartCount()), 10),
 	}
 
-	if p.VolumeClaimName() != "" {
-		latests[VolumeClaim] = p.VolumeClaimName()
+	if len(p.VolumeClaimNames()) > 0 {
+		latests[VolumeClaim] = strings.Join(p.VolumeClaimNames(), "~p$")
```
bboreham: `"~p$"` should be a constant.
And is that necessary and/or sufficient?  What are the rules on volume claim names?

Akash4927: PVC name must consist of lower case alphanumeric characters, `-` or `.`, and must start and end with an alphanumeric character.
We can use a single space or any special character to join them.

bboreham: OK, can you please add a comment about the rules and put the separator in a `const`.
I would probably just use `,` as the separator, if that works, but don't really mind.

Akash4927: Sure @bboreham, `,` will work.
I'll update the PR soon.

PR #3557, opened 2019-01-15
update xterm.js dependency (CVE-2019-0542)

dholbach:
Updating xterm.js because of https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-0542

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -45,7 +45,7 @@
     "styled-components": "3.4.10",
     "weaveworks-ui-components": "0.15.2",
     "whatwg-fetch": "2.0.3",
-    "xterm": "3.3.0"
+    "xterm": "3.10.1"
   },
   "devDependencies": {
     "@fortawesome/fontawesome-free": "^5.5.0",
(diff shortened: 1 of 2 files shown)


dholbach:
Hum. This is probably not as straight-forward as I thought.

Now that I retried it in a local build and running `scope launch`, it works only the second time (clicking on host to start a shell there), the first time the screen is blank.

dholbach:
Tried it on `minikube` (built locally, pushed image to dockerhub, pointed [k8s deployment](https://github.com/weaveworks/scope/blob/master/site/faq.md#running-scope-in-a-kubernetes-setting) to those images). The first launch of a shell in a container came up blank (as in #3559), this is with the current PR (using xterm.js 3.8.1). I can't say I'm confident in this change now.

Help wanted, setting back to WIP.

dholbach:
Let's go with 3.10.1 then. It fixes a number of bugs and is faster. Originally I wanted to just update to the closest, unaffected version, but since the behaviour is the same between all versions, let's go with the most recent.

PR #3562, opened 2019-01-18
Fixed the sidebar namespaces hover effect at last-child

### LINKED ISSUE #3555: Improve namespace-filter expansion UX.
https://github.com/weaveworks/scope/pull/3117 made having a lot of namespaces more manageable in the UI.

There were some ideas to improve this even further:
- @AVRahul:
  > there are some improvements which could be better, the hovering effect can be just restricted to the last-child class instead applying in overall sidebar.
  
- Provide an explicit "show more" button to expand and collapse the namespace list.

AVRahul:
Fixes #3555 
Applied the hover effect to the last child to fix the other above sidebar nodes to remain constant and to improve experience through all the browsers. 

Signed-off-by: A V RAHUL  a.v.rahul11@gmail.com

### DIFF
--- a/client/app/styles/_base.scss
+++ b/client/app/styles/_base.scss
@@ -1329,17 +1329,20 @@ a {
     overflow: hidden;
     max-height: 27px;
     transition: max-height 0.5s 0s $base-ease;
-    &:hover {
-      height: auto;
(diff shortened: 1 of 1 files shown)


foot:
Ah hah this stops the filters other than namespace from subtlety jumping around?

Issue #2192, opened 2017-02-03
Run integration tests on external pull requests


2opremio:
Right now we don't run integration tests on external pull requests because there is no way to securely use the credentials to handle the VMs.

This leads to the increasingly frequent external PRs to break master. For instance https://github.com/weaveworks/scope/issues/2175

2opremio:
https://github.com/weaveworks/scope/pull/2135 is a great example of a PR in which we need integration tests for better quality assurance.

2opremio:
CC @marccarre

alban:
[Jenkins' GitHub plugin](https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin) has a bot that asks on external PRs if it is "ok to test". When someone whitelisted by the config comments "ok to test" on the PR (after testing that circle.yaml is not modified to print the credentials), Jenkins runs the test.

Could CircleCI have something like that?

2opremio:
Related: https://github.com/weaveworks/scope/issues/2193 and  https://github.com/weaveworks/scope/issues/2270

2opremio:
As a remark, it's possible to mimic the same setup in a fork (e.g. https://github.com/kinvolk/scope have done it), it requires setting up your own infrastructure, which is hard and and creates a lot of friction.

bboreham:
Why do we have "Build forked pull requests" turned off at CircleCI when there is a separate setting "Pass secrets to builds from forked pull requests" that meets the requirement stated at the top?

dholbach:
@2opremio You said you had some ideas to fix this. Do you think you could summarise them here?

PR #3566, opened 2019-01-28
Fix dnssnooper probe for multiple CNAMEs.

tiriplicamihai:
Hello,
Playing with scope I noticed a bug where the dns snooper does not identify some external hosts. Digging a bit deeper I found the faulty pattern - if there are multiple CNAMEs in a record but only one of it has a corresponding A address then the match will not be done. Here's an example:

```
100.x.x.x.domain > 100.y.y.y.54637: 34745 3/0/0 graph.facebook.com. CNAME api.facebook.com., api.facebook.com. CNAME star.c10r.facebook.com., star.c10r.facebook.com. A 31.13.66.4 (174)
```

For this frame the outbound calls from my service to `graph.facebook.com.` where never caught.

The fix I considered goes through all the CNAME records and builds a list of aliases. Let me know what you think.

P.S. Scope is awesome, thanks for building it!

### DIFF
--- a/probe/endpoint/dns_snooper.go
+++ b/probe/endpoint/dns_snooper.go
@@ -250,15 +250,14 @@ func (s *DNSSnooper) processDNSMessage(dns *layers.DNS) {
 		domainQueried = question.Name
 		records       = append(dns.Answers, dns.Additionals...)
 		ips           = map[string]struct{}{}
-		alias         []byte
+		aliases       = [][]byte{}
 	)
 
-	// Traverse records for a CNAME first since the DNS RFCs don't seem to guarantee it
-	// appearing before its A-records
+	// Traverse all the CNAME records and the get the aliases. There are cases when the A record is for only one of the
+	// aliases. We traverse CNAME records first because there is no guarantee that the A records will be the first ones
 	for _, record := range records {
-		if record.Type == layers.DNSTypeCNAME && record.Class == layers.DNSClassIN && bytes.Equal(domainQueried, record.Name) {
-			alias = record.CNAME
-			break
+		if record.Type == layers.DNSTypeCNAME && record.Class == layers.DNSClassIN {
+			aliases = append(aliases, record.CNAME)
 		}
 	}
 
@@ -267,8 +266,15 @@ func (s *DNSSnooper) processDNSMessage(dns *layers.DNS) {
 		if record.Type != layers.DNSTypeA || record.Class != layers.DNSClassIN {
 			continue
 		}
-		if bytes.Equal(domainQueried, record.Name) || (alias != nil && bytes.Equal(alias, record.Name)) {
+		if bytes.Equal(domainQueried, record.Name) {
 			ips[record.IP.String()] = struct{}{}
+			continue
+		}
+		for _, alias := range aliases {
+			if bytes.Equal(alias, record.Name) {
+				ips[record.IP.String()] = struct{}{}
+				continue
+			}
 		}
 	}
 
--- a/probe/endpoint/dns_snooper.go
+++ b/probe/endpoint/dns_snooper.go
@@ -273,7 +273,7 @@ func (s *DNSSnooper) processDNSMessage(dns *layers.DNS) {
 		for _, alias := range aliases {
 			if bytes.Equal(alias, record.Name) {
 				ips[record.IP.String()] = struct{}{}
-				continue
+				break
 			}
 		}
 	}
--- a/probe/endpoint/dns_snooper_test.go
+++ b/probe/endpoint/dns_snooper_test.go
@@ -20,19 +20,19 @@ func TestProcessDNSMessageMultipleCNAME(t *testing.T) {
 
 	ipAddressCNAME := "127.0.0.1"
 	answers := []layers.DNSResourceRecord{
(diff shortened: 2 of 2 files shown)


tiriplicamihai:
I don't really get why the integration tests failed. The error doesn't seem related to my code:

```
/tmp/go-build425036814/github.com/weaveworks/scope/probe/docker/_obj/container.go:328:17: undefined: namespaceIPAddresses
# github.com/weaveworks/scope/vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf
In file included from vendor/github.com/weaveworks/tcptracer-bpf/vendor/github.com/iovisor/gobpf/elf/elf.go:52:0:
./include/bpf.h:12:30: fatal error: linux/bpf_common.h: No such file or directory
 #include <linux/bpf_common.h>
                              ^
compilation terminated.
# github.com/weaveworks/scope/vendor/github.com/google/gopacket/pcap
vendor/github.com/google/gopacket/pcap/pcap.go:22:18: fatal error: pcap.h: No such file or directory
 #include <pcap.h>
                  ^
compilation terminated.
FAIL	github.com/weaveworks/scope/app [build failed]
```

bboreham:
@tiriplicamihai thanks for the PR!

The integration test error messages arise because some of the set-up is disabled because we don't want to run unchecked code with full permissions to our test environment. Discussed at #2192

probe/endpoint/dns_snooper.go
```diff
@@ -250,15 +250,14 @@ func (s *DNSSnooper) processDNSMessage(dns *layers.DNS) {
...
 
-	// Traverse records for a CNAME first since the DNS RFCs don't seem to guarantee it
-	// appearing before its A-records
+	// Traverse all the CNAME records and the get the aliases. There are cases when the A record is for only one of the
+	// aliases. We traverse CNAME records first because there is no guarantee that the A records will be the first ones
 	for _, record := range records {
-		if record.Type == layers.DNSTypeCNAME && record.Class == layers.DNSClassIN && bytes.Equal(domainQueried, record.Name) {
-			alias = record.CNAME
-			break
+		if record.Type == layers.DNSTypeCNAME && record.Class == layers.DNSClassIN {
```
bboreham: Why don't we need the `bytes.Equal(domainQueried, record.Name)` here now?

probe/endpoint/dns_snooper.go
```diff
@@ -250,15 +250,14 @@ func (s *DNSSnooper) processDNSMessage(dns *layers.DNS) {
 		domainQueried = question.Name
 		records       = append(dns.Answers, dns.Additionals...)
 		ips           = map[string]struct{}{}
-		alias         []byte
+		aliases       = [][]byte{}
```
bboreham: Would this work better as a `map` ? I suspect it depends on how many aliases we get.

tiriplicamihai: I can rewrite it as a map, no problem.

probe/endpoint/dns_snooper.go
tiriplicamihai: In my experiments I found that the record name for the `star.c10r.facebook.com.` was not `graph.facebook.com`. It looked like a "cascading" CNAME where the record name is the one of the previous CNAME.

I do have a follow up question though - do you see any problems with this approach? I don't see why we would get CNAME answers for a different hostname in the same message.

tiriplicamihai: In my experiments it seemed that the CNAME was not for the queried domain but for another name in that record (`star.c10r.facebook.com.` record name was other than "graph.facebook.com.".

tiriplicamihai:
@bboreham Thanks for the quick review! I will update my pr based on your recommendations from my 2 questions.

probe/endpoint/dns_snooper.go
bboreham: `[]byte` is not allowed as a map key - it would have to be `string`.
Yes, the relative expense of map vs linear search depends on how many elements you have.
From [Rob Pike's 5 Rules of Programming](https://users.ece.utexas.edu/~adnan/pike.html) "n is usually small"
However I was thinking more about code readability than performance.  It's not a big deal.

probe/endpoint/dns_snooper.go
bboreham: I don't know.  I'm not familiar with the code or the problem space.
Would you consider encoding two or three of the cases you saw as tests for this code, so at least we know in future what you were working off?

tiriplicamihai:
@bboreham Thanks for the response! I will try to address the changes next weekend.

bboreham:
@tiriplicamihai did you manage to get back to this?

dholbach:
@2opremio Do you think you can help us nudge this forward?

tiriplicamihai:
Ah, sorry! I completely forgot about this. I can write a unit test by the end of the week. I have a week off so it was the perfect reminder.

tiriplicamihai:
I'm working right now on a test. Sorry again for missing this! I should get better at notification management.

tiriplicamihai:
@dholbach @bboreham can you please take another look? The linting failure doesn't seem to be related with my changes.

PR #3572, opened 2019-02-25
Add a confirmation dialog for deleting a pod.

### LINKED ISSUE #3548: Confirmation dialog for deleting a pod
## What you expected to happen? 
There should be a dialog or pop up about whether user really wants to delete the Pod/Container. Sometimes, Pod is deleted when mistakenly clicked on the `delete` control. Deleting a resource should not be that simple. 
 
## What happened?

User should be asked explicitly while deleting a resource.

Have discussed this with @fbarl already!

fbarl:
Resolves #3548.

On the UI side I used the native `window.confirm` dialog which is generally not encouraged, but is the quickest way of solving the issue at hand. If this becomes a more common pattern, we should put more thought in what sort of component we want to replace it with.

### DIFF
--- a/client/app/scripts/components/node-details/node-details-control-button.js
+++ b/client/app/scripts/components/node-details/node-details-control-button.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { connect } from 'react-redux';
+import { isEmpty } from 'lodash';
 import classNames from 'classnames';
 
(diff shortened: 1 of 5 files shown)


satyamz:
Hi @fbarl ! This is great. Can you attach a screenshot for preview?

fbarl:
Sure @satyamz, here it is:

image

PR #3574, opened 2019-02-27
Upgrade ui-cmp and some of similar deps

foot:
- So everyone is on the same styled-components
- Only visual diff found: terminal font is a bit heavier


foot:
@fbarl any conflicts / concerns come to mind? Everything seems to work.

PR #3580, opened 2019-03-06
Upgrade to Webpack 4

fbarl:
Some of our node dependencies were quite behind, especially Webpack which was blocking bumping other deps and thus resolving some of the latest security issues. Hopefully after this PR, we'll both be able to resolve the remaining security issues and also get some of the Webpack optimizations for free.

#### Changes

* Bumped a couple of deps, most notably Webpack 4 and Babel 7
* Bumped node `v8.4.0` -> `v8.9.0` and locked it in `.nvmrc` file
* Got rid of the `ContrastStyleCompiler` webpack plugin in favor of a pure SCSS solution, so the contrast mode should be working also in Weave Cloud again
* Fixed the `schemeCategory10` import issue

#### Notes

* The production build of `app.js` entry point file grew `1.36MB` -> `1.8MB` - could be for a couple of reasons, but maybe we could address it in a separate issue


dholbach:
Thanks for your work on this! It seems to work great.

A few things I noted: Upon startup I see `@babel/polyfill is loaded more than once on this page. This is probably not desirable/intended and may have consequences if different versions of the polyfills are applied sequentially. If you do need to load the polyfill more than once, use @babel/polyfill/noConflict instead to bypass the warning.` in Firefox console.

`yarn build` lists these two:
```
  terminal-app-25212d40fcac7d911c2d.js   1010 KiB       1  [emitted]  [big]  terminal-app
           app-06be00839d294e8cf164.js   1.82 MiB       2  [emitted]  [big]  app
```

dholbach:
Also `yarn audit` seems to be a lot happier.

@foot what do you think?

PR #3587, opened 2019-03-26
Fix build-pkg script

fbarl:
Follow-up to #3580.

This change makes babel in `build-pkg` script fully transform the JS files to make them more compatible when imported from Weave Cloud.


foot:
> Setting this to false will not transform modules.

foot:
https://babeljs.io/docs/en/babel-preset-env#modules

PR #3591, opened 2019-04-02
app: delete duplicates in report

### LINKED ISSUE #3433: Remove docker_container_ips from reports as it duplicates information
The same addresses are added under `docker_container_ips` and `docker_container_ips_with_scopes`, with slightly different decoration.

The only place `docker_container_ips` is used in the app is via `ContainerMetadataTemplates` for the "IPs" column, so if we teach the app how to remove the decoration we can save sending the addresses twice.

najeal:
this deletes docker_container_ips from reports and keeps only docker_container_ips_with_scopes. IPs are extracted from docker_container_ips_with_scopes when needed.

Fixes #3433

### DIFF
--- a/probe/docker/container.go
+++ b/probe/docker/container.go
@@ -11,7 +11,6 @@ import (
 
 	docker "github.com/fsouza/go-dockerclient"
 	log "github.com/sirupsen/logrus"
-
 	"github.com/weaveworks/common/mtime"
 	"github.com/weaveworks/scope/report"
 )
@@ -322,9 +321,6 @@ func (c *container) NetworkInfo(localAddrs []net.IP) report.Sets {
 	if len(c.container.NetworkSettings.Ports) > 0 {
 		s = s.Add(ContainerPorts, c.ports(localAddrs))
 	}
-	if len(ipv4s) > 0 {
-		s = s.Add(ContainerIPs, report.MakeStringSet(ipv4s...))
-	}
 	if len(ipsWithScopes) > 0 {
 		s = s.Add(ContainerIPsWithScopes, report.MakeStringSet(ipsWithScopes...))
 	}
@@ -465,7 +461,7 @@ func (c *container) GetNode() report.Node {
 
 // ExtractContainerIPs returns the list of container IPs given a Node from the Container topology.
 func ExtractContainerIPs(nmd report.Node) []string {
-	v, _ := nmd.Sets.Lookup(ContainerIPs)
+	v, _ := nmd.Sets.LookupMask(ContainerIPsWithScopes, `[.]?;(.*)$`, 1)
 	return []string(v)
 }
 
--- a/probe/docker/container_test.go
+++ b/probe/docker/container_test.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	client "github.com/fsouza/go-dockerclient"
-
 	"github.com/weaveworks/common/mtime"
 	"github.com/weaveworks/scope/probe/docker"
 	"github.com/weaveworks/scope/report"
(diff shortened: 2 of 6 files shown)


satyamz:
Hi @najeal Thanks for the PR.

report/metadata_template.go
```diff
@@ -75,6 +78,11 @@ func fromCounters(n Node, key string) (string, bool) {
 	return strconv.Itoa(val), ok
 }
 
+func fromSetsIPScope(n Node, key string) (string, bool) {
```
bboreham: This file is fairly generic; is it possible to move the "IP scopes" functionality somewhere more specific to IP addresses or containers?

report/sets.go
```diff
@@ -84,6 +86,33 @@ func (s Sets) Lookup(key string) (StringSet, bool) {
 	return MakeStringSet(), false
 }
 
+// LookupMask returns the sets stored under key with using mask and index identifying match.
+func (s Sets) LookupMask(key string, mask string, matchSelect int) (StringSet, bool) {
```
bboreham: I haven't looked very deeply, but I suspect this can be achieved using `ParseAddressNodeID()` rather than a regex; it seems overkill to embed knowledge of the string format _and_ use a regex.
This function is only called in one place (apart from tests), so it doesn't need to be generic.

report/metadata_template.go
najeal: You entier code is fairly generic 😛 
It seems that **ContainerMetadataTemplates** only allows to gather data from one source but there is no option to modify the data. Hence, I cannot extract **IPs** from **IPsWithScope**
MetadataTemplate looks like:
```
type MetadataTemplate struct {
	ID       string  `json:"id"`
	Label    string  `json:"label,omitempty"`    // Human-readable descriptor for this row
	Truncate int     `json:"truncate,omitempty"` // If > 0, truncate the value to this length.
	Datatype string  `json:"dataType,omitempty"`
	Priority float64 `json:"priority,omitempty"`
	From     string  `json:"from,omitempty"` // Defines how to get the value from a report node
}
```
I could add a **To** optional field function, to ask for specific changes on the found data.

`func fromDefault(n Node, key string) (string, bool)` **From** action returns always a string.
My optional **To** field could be a function with `func(data string) string` signature. Then it could be called immediately after the **From** function if it exists.

report/sets.go
satyamz: `ParseAddressNodeID()` seems a good approach to me too.

probe/docker/container.go
```diff
@@ -465,7 +461,7 @@ func (c *container) GetNode() report.Node {
 
 // ExtractContainerIPs returns the list of container IPs given a Node from the Container topology.
 func ExtractContainerIPs(nmd report.Node) []string {
-	v, _ := nmd.Sets.Lookup(ContainerIPs)
+	v, _ := nmd.Sets.LookupMask(ContainerIPsWithScopes, `[.]?;(.*)$`, 1)
```
satyamz: Can we have a regex ``[.]?;(.*)$``  as a constant variable?

report/metadata_template.go
bboreham: You're right, it is all a bit generic. The original vision was that the app was completely dumb, and all information was supplied by probes.  But this vision didn't really persist so we have a sort of half-and-half situation where some things are heavily coded into the app.

Unfortunately I think the metadata template is passed from the probe, so a function pointer wouldn't work.  Maybe it has to remain the way you had it.

probe/docker/container_test.go
```diff
@@ -108,12 +107,11 @@ func TestContainer(t *testing.T) {
...
 			Add("docker_container_ports", report.MakeStringSet("1.2.3.4:80->80/tcp", "81/tcp")).
 			Add("docker_container_networks", nil).
-			Add("docker_container_ips", report.MakeStringSet("1.2.3.4")).
-			Add("docker_container_ips", report.MakeStringSet("5.6.7.8")).
 			Add("docker_container_ips_with_scopes", report.MakeStringSet(";1.2.3.4")).
 			Add("docker_container_ips_with_scopes", report.MakeStringSet(";5.6.7.8")).
 			Add("docker_container_networks", report.MakeStringSet("network1"))
-
+			// Add("docker_container_ips", report.MakeStringSet("1.2.3.4")).
+			// Add("docker_container_ips", report.MakeStringSet("5.6.7.8")).
```
fbarl: I think these commented lines could be removed now given https://github.com/weaveworks/scope/pull/3591/files#diff-51f61e163768377668ac60822adf1f67L229.

fbarl:
@najeal Thanks for your contribution!

We discussed this PR in the last Scope community meeting and thought we should probably open a new `tech-debt` issue for https://github.com/weaveworks/scope/pull/3591#discussion_r273414109 that could be addressed afterwards.

However it would be nice to give https://github.com/weaveworks/scope/pull/3591#discussion_r273414695 a try before we merge the PR - otherwise things look good! :)

fbarl:
I'm closing this PR as there has been no activity here for about 5 months.

@najeal If you're planning to pick it up again, please feel free to reopen the PR and ping us again if you need any help or feedback.

Issue #3597, opened 2019-04-30
Missing pod view all of a sudden


paulmorabito:
Hi,

I'm using Weave Scope 1.10.2 on Kubernetes 1.13.5. This is on a bare metal cluster running Debian.

Previously, when accessing Weave Scope I would see a Pod alongside Processes, Containers and Hosts. This has now suddenly vanished. 

If I log into the Kubernetes Dashboard or access via kubectl, I can see my pods by namespace but it is only not displaying within Weave Scope. 

I don't recall doing anything to the UI view and nothing in the cluster has changed recently.

Any ideas how this can be restored?

qiell:
Can you please share the logs of `weave-scope-agent` and `weave-scope-app` pods?

paulmorabito:
I am using kubetail to get the logs and only this line appears in
weave-scope-agent:

Error from server: Get
https://192.168.1.113:10250/containerLogs/weave/weave-scope-agent-cpxq9/scope-agent?follow=true&sinceSeconds=10:
dial tcp 192.168.1.113:10250: connect: no route to host

And this is because one of my nodes is down at the moment. It's a manager
node but the other two are up and running.

There is nothing appearing in the logs for weave-scope-app


Is there any way I can grab the logs or query for them?

paulmorabito:
This the logs for one of the weave-scope-agent pods. 

```
<probe> INFO: 2019/04/29 00:23:01.741729 Pipe pipe-3566376904579192063 connection to weave-scope-app.weave.svc.cluster.local starting
<probe> INFO: 2019/04/29 00:23:04.464767 Pipe pipe-3566376904579192063 connection to weave-scope-app.weave.svc.cluster.local exiting
<probe> INFO: 2019/04/29 00:23:12.114149 Pipe pipe-2906881728982743550 connection to weave-scope-app.weave.svc.cluster.local starting
<probe> INFO: 2019/04/29 00:23:13.505390 Pipe pipe-2906881728982743550 connection to weave-scope-app.weave.svc.cluster.local exiting
<probe> INFO: 2019/04/29 00:25:49.424499 Pipe pipe-5906913349546651901 connection to weave-scope-app.weave.svc.cluster.local starting
<probe> INFO: 2019/04/29 00:25:58.973341 Pipe pipe-5906913349546651901 connection to weave-scope-app.weave.svc.cluster.local exiting
[...]

PR #3598, opened 2019-05-10
Add OpenTracing span for report.ReadBinary()

bboreham:
So we can see the timing and size in Jaeger.


bboreham:
> Was this in any way motivated by #3597 (comment)?

No, that's much better addressed by the probe metrics I found (and am changing in #3600).

This span is on the read side, i.e. in the app or collector.

PR #3599, opened 2019-05-10
Add metrics for report size and count per tenant

bboreham:
In a multitenant system it is useful to be able to drill into which tenants have the most or biggest reports.

### DIFF
--- a/app/multitenant/aws_collector.go
+++ b/app/multitenant/aws_collector.go
@@ -71,7 +71,7 @@ var (
 	reportsPerUser = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "scope",
 		Name:      "reports_stored_total",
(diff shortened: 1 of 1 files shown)


app/multitenant/aws_collector.go
```diff
@@ -68,6 +68,16 @@ var (
 		Help:      "Distribution of memcache report sizes",
 		Buckets:   prometheus.ExponentialBuckets(4096, 2.0, 10),
 	})
+	reportsPerUser = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "scope",
+		Name:      "reports_stored_total",
+		Help:      "Total stored reports per user.",
```
fbarl: Just to make it more consistent with the rest -> _Total **count of** stored..._

PR #3606, opened 2019-05-15
Update google/gopacket library

bboreham:
To pick up a fix https://github.com/google/gopacket/pull/589 for Go 1.12


dholbach:
I retried the integration-tests, but looks like they're broken for some reason.

fbarl:
@bboreham Was there an issue manifestation in Scope that motivated this PR?

bboreham:
> Was there an issue manifestation in Scope that motivated this PR?

Scope's build is still using an older Go, but on my dev machine with Go 1.12 this change is needed to compile.

bboreham:
I found the unlikely reason for the test failure.

PR #3623, opened 2019-05-24
Always delete container info when we get a 'destroy' event

### LINKED ISSUE #3576: Report sizes seem to increase over days after a restart
Noticed in data from a lot of probes - this is the average report size over the last 30 days:

image

The two drops are at points where all probes were restarted.
Pattern seems to suggest the probes are accumulating stuff in data structures which does not come back on restart, and therefore probably unnecessary.

bboreham:
Fixes #3576 

Previously it would only delete if Docker said the container didn't exist, which is a race between Docker sending the event and Docker removing the info from its own records. If Scope lost the race it would never check again, so the set of containers grows and grows.

Extract a couple of functions to make the action clearer.

### DIFF
--- a/probe/docker/registry.go
+++ b/probe/docker/registry.go
@@ -262,7 +262,7 @@ func (r *registry) updateContainers() error {
 	}
 
 	for _, apiContainer := range apiContainers {
-		r.updateContainerState(apiContainer.ID, nil)
+		r.updateContainerState(apiContainer.ID)
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


fbarl:
> Fixes #3576

How confident are you about this @bboreham? :) Maybe we should wait a week or two after merging to see if this PR really fixes it?

bboreham:
> > Fixes #3576

> How confident are you about this

I expect this is 80-90% of the problem.  Probably other bugs after this one.

PR #3627, opened 2019-06-04
fix(probe): Use a buffered chan to reduce the chance of losing events

### LINKED ISSUE #3576: Report sizes seem to increase over days after a restart

bboreham:
Fixes #3576 (I know, we've been here before...)

Event notifications from Docker have been seen to get dropped; using a buffered chan seems to reduce the chance of this happening.

### DIFF
--- a/probe/docker/registry.go
+++ b/probe/docker/registry.go
@@ -176,7 +176,9 @@ func (r *registry) listenForEvents() bool {
 	// Next, start listening for events.  We do this before fetching
 	// the list of containers so we don't miss containers created
 	// after listing but before listening for events.
-	events := make(chan *docker_client.APIEvents)
+	// Use a buffered chan so the client library can run ahead of the listener
+	// - Docker will drop an event if it is not collected quickly enough.
+	events := make(chan *docker_client.APIEvents, 1024)
 	if err := r.client.AddEventListener(events); err != nil {
 		log.Errorf("docker registry: %s", err)
 		return true


probe/docker/registry.go
```diff
@@ -176,7 +176,8 @@ func (r *registry) listenForEvents() bool {
 	// Next, start listening for events.  We do this before fetching
 	// the list of containers so we don't miss containers created
 	// after listing but before listening for events.
-	events := make(chan *docker_client.APIEvents)
+	// Use a buffered chan to reduce the chance of losing events
+	events := make(chan *docker_client.APIEvents, 32)
```
rade: Where do the events get dropped? Is the sender using a non-blocking send?

bboreham: Unclear. Not in `fsouza/go-dockerclient`.
Maybe the other side of the socket.

bboreham: Here you go: https://github.com/moby/moby/blob/a6e1502575858cd063b0e75d400e04b6f334e86d/daemon/events/events.go#L114

bboreham: Also `// Each listener has 100 milliseconds to receive the event or it will be skipped.`

rade: Is there any downside to making the buffer a lot larger (but not as large as to become a memory hog)?

bboreham: Like 1024?  I don't see any issue.
Really we need to do a full sync periodically, so I see this change as an improvement but not a final fix.

rade: 1024 sounds reasonable.

Pls file an issue for the periodic sync.

bboreham: Filed https://github.com/weaveworks/scope/issues/3629

PR #3633, opened 2019-06-07
Security updates

dholbach:
I updated a bunch of Javascript dependencies to eliminate some known security issues in Scope. Probably not all of them are fixed in this round, it also brings us closer to upstream's latest and greatest.


client/package.json
```diff
@@ -108,7 +108,7 @@
     "start-production": "NODE_ENV=production node server.js",
     "test": "jest",
     "coveralls": "cat coverage/lcov.info | coveralls",
-    "lint": "eslint app server.js && stylelint src && sass-lint -v",
+    "lint": "eslint app server.js && stylelint --allow-empty-input src && sass-lint -v",
```
dholbach: @foot @fbarl Please see if we want to change in the future. I mostly just wanted to help us fix some of the security issues before we get out another release. (10.0.0 of stylelint errored out on empty input, it looks like we were not feeding it the right list of files for some time?)

dholbach: ccing @bboreham too.

client/package.json
```diff
@@ -108,7 +108,7 @@
     "start-production": "NODE_ENV=production node server.js",
     "test": "jest",
     "coveralls": "cat coverage/lcov.info | coveralls",
-    "lint": "eslint app server.js && stylelint --allow-empty-input src && sass-lint -v",
+    "lint": "eslint app server.js && stylelint app/scripts && sass-lint -v",
```
dholbach: I played around with stylelint-scss and other things, to no avail. As somebody not knowledgeable, this takes a bit much time to investigate. Happy to file an issue about this to track it further.

bboreham:
Clarification: they are not "known security issues in Scope", they are "known security issues in the dependencies, which are not believed to affect Scope but best to be safe".

dholbach:
Thanks for the clarification, yes.

client/package.json
dholbach: I retract my last comment. Apparently we run `sass-lint` over the files in this directory. The current fix should be good. A look over it from Simon or Filip wouldn't hurt though.

bboreham: There is no directory `src` in the same place as `app/scripts`, so this line wasn't checking anything before?

dholbach: Yes. I didn't do any git archaeology to find out if there ever was a `src` directory, but at least this is checking the scripts now and as a I said, `sass-lint` takes care of the rest.

PR #3642, opened 2019-06-24
Add CODEOWNERS file

dholbach:
Follow https://github.blog/2017-07-06-introducing-code-owners/ to allow us to use Pull reminders.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,7 +2,7 @@
 # Each line is a file pattern followed by one or more owners.
 
 # These owners will be the default owners for everything in the repo.
-*       @bboreham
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,8 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in the repo.
+*       @bboreham
```
fbarl: There seems to be an extra space here :)

dholbach: Thanks. Fixed.

Issue #3624, opened 2019-05-27
today,I find a Mining procedure on my computer, the path show the /scope


youyouvi:
## Versions:

```
$ scope version
$ docker version
$ uname -a
$ kubectl version
```

## Logs:
```
$ docker logs weavescope
```
or, if using Kubernetes:
```
$ kubectl logs <weave-scope-pod> -n <namespace> 
```

bboreham:
Can you give more detail?  That’s not really enough information to do anything with. 

Those bits at the end, like `version` and `logs`, you are supposed to run the command and paste the output here. 

Use the address security@weave.works if it is confidential or you are reporting a vulnerability.

youyouvi:
This morning, I found the mining program on the server I managed, disguised as a MySQL user.
So I used the "ls/proc/pid/cwd" instruction and found that the path pointed to / scope.
So I executed rm-rf/usr/local/bin/scope and rm-rf/run/scope.
Weave is installed according to the instructions on GitHub。
Sorry，my english is poor





------------------ 原始邮件 ------------------
发件人: "Bryan Boreham"<notifications@github.com>;
发送时间: 2019年5月27日(星期一) 下午3:29
收件人: "weaveworks/scope"<scope@noreply.github.com>;
抄送: "yo.vi"<yo.vi@foxmail.com>;"Author"<author@noreply.github.com>;
主题: Re: [weaveworks/scope] today,I find a Mining procedure on mycomputer, the path show the /scope (#3624)




Can you give more detail?  That’s not really enough information to do anything with.
 
Those bits at the end, like version and logs, you are supposed to run the command and paste the output here.
 
Use the address security@weave.works if it is confidential or you are reporting a vulnerability.

bboreham:
Is your server open to the Internet?  In the installation instructions we warn " the Scope app endpoint (port 4040) should not be made accessible on the Internet"

youyouvi:
Yes,  my server open to the Internet 


 
---Original---

bboreham:
OK, suggest you use a firewall to block port 4040 from public access (indeed, block all ports you don't explicitly need to open).
We do try to warn about this [in the installation instructions](https://github.com/weaveworks/scope/blob/master/site/installing.md#installing-on-docker) - I've created #3650 to cover another place where users might arrive first.

Issue #2814, opened 2017-08-15
Help users avoid accidentally opening Scope to the whole Internet


bboreham:
We do [warn](https://github.com/weaveworks/scope/blob/12028fb950daddd18ed2ceccacf9af59254d1d59/site/installing.md#L50) users not to run the Scope app exposed to the Internet, but apparently some people carry on.

Perhaps Scope should default to a simple basic-auth scheme with a randomly-generated password so it is a little safer?

dholbach:
Maybe as a first step we could document how to do that in a very basic way?

bboreham:
Do you mean "we could expand the instructions on how to use a reverse proxy"?

Since this issue is about changing the code to behave less open, we wouldn't document that specifically.

ycao56:
I have some code working to add basic auth support to protect weave. I can create a pr if people are interested.

ycao56:
https://github.com/weaveworks/scope/pull/3393

PR #3650, opened 2019-07-09
Docs: improve the warning about opening Scope to the Internet

bboreham:
Trying to improve the situation at #3624 and #2814 

Also drive-by removal of `boot2docker` since it's ancient history and complicates the instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ If you would like to see your name added, let us know on Slack, or send a PR ple
 
 ## <a name="getting-started"></a>Getting Started
 
+Ensure your computer is behind a firewall that blocks port 4040 then,
+
 ```console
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -42,15 +42,16 @@ If you would like to see your name added, let us know on Slack, or send a PR ple
 
 ## <a name="getting-started"></a>Getting Started
 
+Ensure your computer is behind a firewall that blocks port 4040 then,
```
satyamz: Can we add a small note like, **If you're using Scope in Kubernetes then do not expose it as a NodePort service** ? 

Also, it'd be good to have these instructions in **bold** format. :smile:

bboreham: done

PR #3653, opened 2019-07-17
fix(probe): Avoid ebpf events out of order by adding offset to time

### LINKED ISSUE #2827: eBPF tracker bounces due to late events
Follow-up to #2650... as of #2735 the eBPF tracker restarts on error, but we should still try to eliminate the underlying cause at some point.

bboreham:
Fixes #2827

Needs the [upstream PR](https://github.com/weaveworks/tcptracer-bpf/pull/67) to be merged then vendor that change here.

The idea is described at https://github.com/iovisor/gobpf/issues/42#issuecomment-508500065 but no response in that repo so I'm making the change in a weaveworks-controlled repo.

### DIFF
--- a/vendor/github.com/weaveworks/tcptracer-bpf/pkg/tracer/event.go
+++ b/vendor/github.com/weaveworks/tcptracer-bpf/pkg/tracer/event.go
@@ -1,3 +1,5 @@
+// +build linux
+
 package tracer
 
 import (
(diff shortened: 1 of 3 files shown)


bboreham:
I've been running this branch in our staging and prod environments for 12 days now, with zero "Stopping the eBPF tracker" messages.  Complete success.

Issue #3256, opened 2018-07-06
Limit the number of report decodings that we do in parallel, in multitenant code


bboreham:
Currently we have this call sequence:
```
awsCollector.Report()
  awsCollector.getReports()
    S3Store/MemcacheClient.FetchReports()
      // fetch and decode all reports in parallel
  Merger.Merge()
```

When you have a lot of reports this is maximally bad for memory consumption.

Since `Merge()` is now serial we could return compressed blobs from `FetchReports()` and decode just before they are merged.  It's probably more effective to have some parallelism - fetch, decode and merge subsets in parallel, then merge the outputs of those subsets.

bboreham:
This has been fixed by #3671.  It is still unlimited in the sense that if the app receives a large number of requests in parallel it will start to handle them all, but that is a different feature.

PR #3671, opened 2019-09-09
Quantise report cache in query side of aws-collector

bboreham:
Cache merged groups of reports, to reduce the number we handle in parallel.

Previously we would merge all reports in a 15-second window. Now we use a 'quantum' of 3 seconds, similar to the single-user app.

E.g. a 30-node cluster will have 150 individual reports over 15 seconds, but the new code will merge 5 pre-merged reports plus 20-ish very recent individual ones.

This reduces the max heap size used for deserialising, since we only do 3 seconds at once per instance. This improves #3256, but I'm not marking it as a fix since the number done in parallel is still unlimited.

Individual reports are still put into the cache, but should get displaced by the pre-merged ones under LRU.

### DIFF
--- a/app/multitenant/aws_collector.go
+++ b/app/multitenant/aws_collector.go
@@ -31,10 +31,10 @@ const (
 	reportField = "report"
 	natsTimeout = 10 * time.Second
 
-	reportQuantisationInterval = 3000000000 // 3 seconds in nanoseconds
+	reportQuantisationInterval = 3 * time.Second
 	// Grace period allows for some gap between the timestamp on reports
 	// (assigned when they arrive at collector) and them appearing in DynamoDB query
-	gracePeriod = 500000000 // 1/2 second in nanoseconds
+	gracePeriod = 500 * time.Millisecond
 )
 
 var (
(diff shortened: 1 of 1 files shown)


app/multitenant/aws_collector.go
```diff
@@ -30,6 +30,11 @@ const (
 	tsField     = "ts"
 	reportField = "report"
 	natsTimeout = 10 * time.Second
+
+	reportQuantisationInterval = 3000000000 // 3 seconds in nanoseconds
+	// Grace period allows for some gap between the timestamp on reports
+	// (assigned when they arrive at collector) and them appearing in DynamoDB query
+	gracePeriod = 500000000 // 1/2 second in nanoseconds
```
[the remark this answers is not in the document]
bboreham: I did "make these constants `time.Duration` and `call .Nanoseconds()` on every use"

app/multitenant/aws_collector.go
```diff
@@ -348,25 +354,89 @@ func (c *awsCollector) getReports(ctx context.Context, reportKeys []string) ([]r
...
+	// Fetch a merged report for each time quantum covering the window
+	startTS, endTS := start.UnixNano(), end.UnixNano()
+	ts := startTS - (startTS % reportQuantisationInterval.Nanoseconds())
+	for ; ts+(reportQuantisationInterval+gracePeriod).Nanoseconds() < endTS; ts += reportQuantisationInterval.Nanoseconds() {
+		quantumReport, err := c.reportForQuantum(ctx, userid, reportKeys, ts)
+		if err != nil {
+			return report.MakeReport(), err
+		}
+		reports = append(reports, quantumReport)
+	}
```
rade: I was wondering whether that entire loop could operate on `time.Time`, converting to unix nanos only in the call to (or even inside) `reportForQuantum`.

bboreham: It’s the modulus that scares me. 
I’ve made one bug that way already this year.

PR #3678, opened 2019-09-15
perf(probe): add 'omitempty' tag to Topology.Nodes

bboreham:
So we save space writing out empty topologies, which we see more often after #3677.  Very small impact overall.

Need to fix up `app_client_internal_test.go` to use Scope's `test/reflect` package that understands empty==nil, so now it doesn't need a previous workaround.

Remove a similar workaround in `probe_internal_test.go` that isn't necessary since it's already using that package.

### DIFF
--- a/probe/appclient/app_client_internal_test.go
+++ b/probe/appclient/app_client_internal_test.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
-	"reflect"
 	"strings"
 	"testing"
 	"time"
@@ -16,6 +15,7 @@ import (
 	"github.com/weaveworks/common/test"
 	"github.com/weaveworks/scope/common/xfer"
 	"github.com/weaveworks/scope/report"
+	"github.com/weaveworks/scope/test/reflect"
 )
 
 func dummyServer(t *testing.T, expectedToken, expectedID string, expectedVersion string, expectedReport report.Report, done chan struct{}) *httptest.Server {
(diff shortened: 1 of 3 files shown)


report/topology.go
```diff
@@ -13,7 +13,7 @@ type Topology struct {
 	Tag               string            `json:"tag,omitempty"`
 	Label             string            `json:"label,omitempty"`
 	LabelPlural       string            `json:"label_plural,omitempty"`
-	Nodes             Nodes             `json:"nodes"`
+	Nodes             Nodes             `json:"nodes,omitempty" deepequal:"nil==empty"`
```
fbarl: I want to check how `deepequal` here works but don't know what exactly to Google.. Do you have links/references at hand?

bboreham: that feature was added here: https://github.com/weaveworks/scope/commit/f44f8806cd7f20036362f73f6c749294fecc61b2#diff-1aaeb558d4801562ecccd7b7352209dbR149

probe/appclient/app_client_internal_test.go
```diff
@@ -16,6 +15,7 @@ import (
 	"github.com/weaveworks/common/test"
 	"github.com/weaveworks/scope/common/xfer"
 	"github.com/weaveworks/scope/report"
+	"github.com/weaveworks/scope/test/reflect"
```
fbarl: Did this `"reflect"` -> `"github.com/weaveworks/scope/test/reflect"` happen by mistake?

PR #3679, opened 2019-09-15
perf(probe): reduce copying of nodes

bboreham:
Where we know we are merging several reports into the one, we can call `UnsafeMerge()` and skip the copy that `Merge()` will do.

Small impact; most of the effort goes into collecting the data not publishing it.

### DIFF
--- a/probe/probe.go
+++ b/probe/probe.go
@@ -183,7 +183,7 @@ func (p *Probe) report() report.Report {
 
 	result := report.MakeReport()
 	for i := 0; i < cap(reports); i++ {
-		result = result.Merge(<-reports)
+		result.UnsafeMerge(<-reports)
 	}
 	return result
 }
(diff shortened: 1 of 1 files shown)


probe/probe.go
```diff
@@ -210,11 +210,12 @@ func (p *Probe) tag(r report.Report) report.Report {
 
 func (p *Probe) drainAndPublish(rpt report.Report, rs chan report.Report) {
 	p.rateLimiter.Wait(context.Background())
+	rpt = rpt.Copy()
```
fbarl: Nit: Does the `rpt report.Report` in the function arg only do a shallow copy? If it's a deep copy, then we don't need this line, if it's shallow, then why not only pass the reference there `rpt *report.Report`?

bboreham: Since Report holds some slices and maps, copying it is shallow so we need the deep copy.
However in Go it is idiomatic to pass a reference only where you intend to modify the data, or when profiling shows the reference to be way more efficient.

PR #3682, opened 2019-09-15
Websocket tracing spans

bboreham:
Since the http call lasts forever, the spans for it don't appear.
Instead, start a new root span every time round the loop (which is typically every 5 seconds).

For the multitenant case it's useful to have a tag with the userid.

### DIFF
--- a/app/api_topology.go
+++ b/app/api_topology.go
@@ -7,6 +7,7 @@ import (
 	"context"
 
 	"github.com/gorilla/mux"
+	ot "github.com/opentracing/opentracing-go"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/weaveworks/scope/common/xfer"
@@ -133,6 +134,8 @@ func handleWebsocket(
 	defer rep.UnWait(ctx, wait)
 
 	for {
+		span := ot.StartSpan("ws.Render", ot.Tag{"topology", topologyID})
+		ctx := ot.ContextWithSpan(ctx, span)
 		// We measure how much time has passed since the channel was opened
 		// and add it to the initial report timestamp to get the timestamp
 		// of the snapshot we want to report right now.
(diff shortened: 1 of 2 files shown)


app/api_topology.go
```diff
@@ -144,11 +147,13 @@ func handleWebsocket(
 		re, err := rep.Report(ctx, reportTimestamp)
 		if err != nil {
 			log.Errorf("Error generating report: %v", err)
+			span.Finish()
```
fbarl: Can't we defer `span.Finish()` at the top instead of adding it before every `return`?

app/api_topology.go
```diff
@@ -133,6 +134,8 @@ func handleWebsocket(
 	defer rep.UnWait(ctx, wait)
 
 	for {
+		span := ot.StartSpan("ws.Render", ot.Tag{"topology", topologyID})
```
fbarl: I wonder if we'd also want to add `channelOpenedAt` and `reportTimestamp` to the span tags.

fbarl: Also, `websocket.Render` for more clarity?

app/api_topology.go
bboreham: I actually coded that first time, but it doesn't work.  `defer` runs things at the end of the function, and we want to run at the end of the loop.
Possibly the whole thing would be neater if the loop body was written as a separate function, which could then use `defer` as you suggest.

app/api_topology.go
bboreham: Tags are indexed for lookup; it's rather unlikely someone would want to do that so I don't feel the overhead is warranted.
I could add them in a `Log` line since the data is interesting.

app/api_topology.go
fbarl: > Possibly the whole thing would be neater if the loop body was written as a separate function, which could then use defer as you suggest.

Not sure if that would make it better or worse, your call :)

app/api_topology.go
fbarl: > I could add them in a Log line since the data is interesting.

yeah, maybe you can do that

bboreham: Done.

app/api_topology.go
bboreham: I refactored the code to move several variables into a struct and the loop body into a member function on that struct.  See what you think.

fbarl:
Thanks for the changes, feel free to merge now!

PR #3685, opened 2019-09-16
Update javascript libraries in various places

dholbach:
Update javascript libraries in various places, needs newer node.

```console
[daniel@reef scope ]$ cat client/Dockerfile 
# Changes to this file will not take effect in CI
# until the image version in the CI config is updated. See
# https://github.com/weaveworks/scope/blob/master/.circleci/config.yml#L11
FROM node:8.12
[...]
```


dholbach:
@fbarl Can you advise on the update for the CI related bits?

fbarl:
> Can you advise on the update for the CI related bits?

So, the UI image that CircleCI currently uses is `weaveworks/scope-ui-build:master-c0b60a16` which seems to be about [a year old](https://cloud.docker.com/u/weaveworks/repository/docker/weaveworks/scope-ui-build). The last node version bump happened in https://github.com/weaveworks/scope/pull/3231 also about a year ago, so that seems right.

I'm not sure this is the best way to proceed but what we could do for sure is:

1) Merge a separate node version bump PR into master first (would be a forced merge as the CircleCI would fail with the same errors as here)
2) I can build that master locally and push the new `weaveworks/scope-ui-build` image to Dockerhub manually
3) We update `.circleci/config.yml#L11` here with the new Dockerhub image and fix the CircleCI errors

This doesn't sound like the optimal approach though and I wonder if @bboreham knows of any better way to proceed. If not, we could follow those 3 steps above and get it out of the way.

bboreham:
The change to use specific tags for CircleCI build image was in #3333, but I can't see that I explained it very well.
The build image has always been hand-pushed from someone's machine.
I suggest building and pushing from this branch, then add a commit to this PR to make `config.yml` use that image, then after it is merged we can update with a tag from master.

fbarl:
Checklist (suggested by @bboreham):

* [ ] Built and pushed the image from this branch to Dockerhub
* [ ] Add a commit that updates `config.yml` with the new image tag + comment about reverting
* [ ] Open a new PR which switches back to the new master tag

fbarl:
Going to merge this PR after https://github.com/weaveworks/scope/pull/3689 is merged.

PR #3686, opened 2019-09-17
feature(app): Add a debugging summary function, exposed via http

bboreham:
When someone says they have a performance problem, the first thing I want to know is whether Scope is reporting 26,000 processes on a node, or similar.  It's awkward to do that via the raw reports, so I wrote this summarising function you can call via URL `/admin/summary`

Example output:
```
pod:183, service:81, deployment:87, daemon_set:8, stateful_set:2, cron_job:5, namespace:14, persistent_volume:3, persistent_volume_claim:4, storage_class:1
ip-172-20-2-197;<host>:endpoint:992, process:369, container:101, container_image:41, host:1, overlay:1
ip-172-20-1-59;<host>:endpoint:266, process:250, container:43, container_image:15, host:1, overlay:1
ip-172-20-3-247;<host>:endpoint:891, process:383, container:107, container_image:43, host:1, overlay:1
```

### DIFF
--- a/app/multitenant/aws_collector.go
+++ b/app/multitenant/aws_collector.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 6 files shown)


app/multitenant/aws_collector.go
```diff
@@ -444,6 +445,34 @@ func (c *awsCollector) HasHistoricReports() bool {
...
+	if err != nil {
+		return "", err
+	}
+	reports, err := c.reportsForKeysInRange(ctx, reportKeys, start.UnixNano(), end.UnixNano())
+	if err != nil {
+		return "", err
+	}
+	var b strings.Builder
+	for i := range reports {
+		// TODO: print the key - note reports may be in a different order from reportKeys
```
fbarl: Is this a non-trivial thing to do in this PR?

bboreham: Yes, it requires creating some new data structure to hold both the key and the report.

bboreham:
> Update the Github issue template to paste the output of /admin/summary whenever filing a new bug

I think I will leave this until experience tells us we need the info very often.

PR #3687, opened 2019-09-17
Refactor report reading

bboreham:
There was a lot of duplication and confusingly-similar functions.


bboreham:
Given review comments, I replaced the last commit (which removed `MakeFromBinary()`) in favour of one removing `ReadBinary()`.

fbarl:
> Given review comments, I replaced the last commit (which removed MakeFromBinary()) in favour of one removing ReadBinary().

Thanks @bboreham, LGTM!

PR #3691, opened 2019-09-21
Move to Kubernetes 'v1' APIs

bboreham:
The latest Kubernetes release, 1.16, no longer serves the v1beta1 APIs by default.

Scope will no longer work with Kubernetes 1.8 and below.

### DIFF
--- a/probe/kubernetes/client.go
+++ b/probe/kubernetes/client.go
@@ -17,12 +17,10 @@ import (
 	"github.com/pborman/uuid"
 	log "github.com/sirupsen/logrus"
 	apiappsv1 "k8s.io/api/apps/v1"
-	apiappsv1beta1 "k8s.io/api/apps/v1beta1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	apibatchv1 "k8s.io/api/batch/v1"
 	apibatchv1beta1 "k8s.io/api/batch/v1beta1"
-	apibatchv2alpha1 "k8s.io/api/batch/v2alpha1"
 	apiv1 "k8s.io/api/core/v1"
-	apiextensionsv1beta1 "k8s.io/api/extensions/v1beta1"
 	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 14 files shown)


probe/kubernetes/client.go
```diff
@@ -242,28 +240,19 @@ func (c *client) clientAndType(resource string) (rest.Interface, interface{}, er
 	case "storageclasses":
 		return c.client.StorageV1().RESTClient(), &storagev1.StorageClass{}, nil
 	case "deployments":
-		return c.client.ExtensionsV1beta1().RESTClient(), &apiextensionsv1beta1.Deployment{}, nil
+		return c.client.AppsV1().RESTClient(), &apiappsv1.Deployment{}, nil
```
satyamz: New Scope versions will not work with `kubernetes < 1.16` ? 
Does it makes sense to have checks for `kubernetes < 1.16` too?

probe/kubernetes/client.go
```diff
@@ -242,28 +240,19 @@ func (c *client) clientAndType(resource string) (rest.Interface, interface{}, er
...
-		if err != nil {
-			return nil, nil, err
-		}
-		if ok {
-			// kubernetes >= 1.8
-			return c.client.BatchV1beta1().RESTClient(), &apibatchv1beta1.CronJob{}, nil
-		}
-		// kubernetes < 1.8
-		return c.client.BatchV2alpha1().RESTClient(), &apibatchv2alpha1.CronJob{}, nil
+		return c.client.BatchV1beta1().RESTClient(), &apibatchv1beta1.CronJob{}, nil
```
satyamz: This would deprecate Scope for kubernetes < `1.8` (which is ok, I believe).

probe/kubernetes/client.go
```diff
@@ -609,19 +599,19 @@ func (c *client) DeleteVolumeSnapshot(namespaceID, volumeSnapshotID string) erro
 }
 
 func (c *client) ScaleUp(namespaceID, id string) error {
-	return c.modifyScale(namespaceID, id, func(scale *apiextensionsv1beta1.Scale) {
+	return c.modifyScale(namespaceID, id, func(scale *autoscalingv1.Scale) {
```
satyamz: I understand that we would not be able to achieve both implementations since method signatures are changing like this. So it looks like blocker for supporting older versions.

bboreham:
It will work with versions from 1.9 to 1.16.

PR #3700, opened 2019-10-02
Get rid of SASS files

### LINKED ISSUE #3577: Replace use of sasslint-webpack-plugin
sasslint-webpack-plugin is dead.

fbarl:
Fixes #3577.

Turned out to be more involved than I expected, quite a few variables needed to be renamed and quite a few SASS concepts replaced even when moving everything just into `createGlobalStyle`.

This PR gets rid of all SASS dependencies, including `sass-lint` which should make our build times faster, styling more modular and consistent and also shrink down our security vulnerabilities list. This piece of work was long due :)

There should be no user visible changes in this PR, which is also the best way to test it - check the UI and make sure no styles are broken.


dholbach:
great work!

PR #3713, opened 2019-10-14
Fix accidental rename of DNS to 'nodes'

bboreham:
In commit 951629af292a (part of #3677) the `DNS` field was accidentally renamed in serialised data to `nodes`.  Put it back, and also add a field to fix up data coming from a probe with that fault.

🤦‍♂

### DIFF
--- a/report/report.go
+++ b/report/report.go
@@ -188,7 +188,9 @@ type Report struct {
 	// Job represent all Kubernetes Job on hosts running probes.
 	Job Topology
 
-	DNS DNSRecords `json:"nodes,omitempty" deepequal:"nil==empty"`
+	DNS DNSRecords `json:"DNS,omitempty" deepequal:"nil==empty"`
+	// For release 1.11.6, probes accidentally sent DNS records labeled "nodes".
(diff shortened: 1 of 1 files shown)


report/report.go
```diff
@@ -188,7 +188,9 @@ type Report struct {
 	// Job represent all Kubernetes Job on hosts running probes.
 	Job Topology
 
-	DNS DNSRecords `json:"nodes,omitempty" deepequal:"nil==empty"`
+	DNS DNSRecords `json:"DNS,omitempty" deepequal:"nil==empty"`
+	// For release 1.11.6, probes accidentally sent DNS records labeled "nodes".
```
fbarl: Could you please link https://github.com/weaveworks/scope/commit/951629af292ad2964a73766840d8d2e3ac983e99 in the comment? Just so that it's easier to track how much time has passed.

Related to that, how long do you think we should keep `BugDNS` field around? Hopefully not forever :O

bboreham: Not forever, but If 1.11.6 is the last version that supports pre-1.9 Kubernetes we probably need to keep compatibility for a year or two.

PR #3714, opened 2019-10-17
performance: send active controls as a single string per node

bboreham:
Scope "controls" are buttons to do something like restart a container.
These controls can be turned on and off by the probe, e.g. depending on whether the container is running.

Previously this information was conveyed by a complicated data structure mapping each control name to a `Dead` boolean which had a last-modified timestamp.  All of which takes effort to send and merge.
This PR changes the representation to a single string saying which controls are active, e.g. `"pause;restart;stop"`. String values have a timestamp, so the latest value prevails in a merge.

I guess we have to update the plugins docs but haven't really looked.

This is a protocol change, requires a version bump.


probe/awsecs/reporter.go
```diff
@@ -150,17 +150,18 @@ func (r Reporter) Tag(rpt report.Report) (report.Report, error) {
 		// Create all the services first
 		for serviceName, service := range ecsInfo.Services {
 			serviceID := report.MakeECSServiceNodeID(cluster, serviceName)
+			activeControls := []string{ScaleUp}
+			// Disable ScaleDown when only 1 task is desired, since
+			// scaling down to 0 would cause the service to disappear (#2085)
+			if service.DesiredCount < 1 {
```
fbarl: Why did you change the condition? :thinking: 

```suggestion
			if service.DesiredCount <= 1 {
```

probe/docker/container.go
```diff
@@ -389,20 +389,18 @@ func (c *container) getBaseNode() report.Node {
...
-		AttachContainer:  {Dead: !running},
-		ExecContainer:    {Dead: !running},
-		StartContainer:   {Dead: !stopped},
-		RemoveContainer:  {Dead: !stopped},
+	switch {
+	case paused:
+		return []string{UnpauseContainer}
+	case c.container.State.Running:
+		return []string{RestartContainer, StopContainer, PauseContainer, AttachContainer, ExecContainer}
+	case !c.container.State.Running:
```
fbarl: Nit: why not like this?

```suggestion
	default:
```

If the reason is readability, I feel it's a bit odd to single out the Running condition from above as the real condition that applies here is `!paused && !running`.

probe/docker/container.go
```diff
@@ -389,20 +389,18 @@ func (c *container) getBaseNode() report.Node {
...
-		UnpauseContainer: {Dead: !paused},
-		RestartContainer: {Dead: !running},
-		StopContainer:    {Dead: !running},
-		PauseContainer:   {Dead: !running},
-		AttachContainer:  {Dead: !running},
-		ExecContainer:    {Dead: !running},
-		StartContainer:   {Dead: !stopped},
-		RemoveContainer:  {Dead: !stopped},
+	switch {
+	case paused:
```
fbarl: Nit: I guess I do have a little bit of OCD:

```suggestion
	case c.container.State.Paused:
```

report/node.go
```diff
@@ -118,28 +117,16 @@ func (n Node) WithAdjacent(a ...string) Node {
...
-	return n
+	return n.WithLatest(NodeActiveControls, mtime.Now(), strings.Join(cs, ScopeDelim))
 }
 
-// WithLatestControl produces a new Node with control added to it
-func (n Node) WithLatestControl(control string, ts time.Time, data NodeControlData) Node {
-	n.LatestControls = n.LatestControls.Set(control, ts, data)
-	return n
+// ActiveControls returns a string slice with the names of active controls.
+func (n Node) ActiveControls(cs ...string) []string {
```
fbarl: `cs` doesn't seem to be used in the function or am I missing something? Weird how linter didn't complain:

```suggestion
func (n Node) ActiveControls() []string {
```

report/backcompat.go
```diff
@@ -0,0 +1,153 @@
...
+	*n = in.Node
+	if len(in.LatestControls) > 0 {
+		// Convert the map into a delimited string
+		cs := make([]string, 0, len(in.LatestControls))
+		var ts time.Time
+		for name, v := range in.LatestControls {
+			if !v.Value.Dead {
+				cs = append(cs, name)
+				// Pull out the newest timestamp to use for the whole set
+				if ts.Before(v.Timestamp) {
```
fbarl: Is this always true before `ts` is set? Because otherwise we should make sure it doesn't end up being `nil` I guess?

report/backcompat.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"strings"
+	"time"
+
+	"github.com/ugorji/go/codec"
+)
+
+// For backwards-compatibility with probes that sent a map of latestControls data
+type bcNode struct {
```
fbarl:
```suggestion
type nodeWithLatestControls struct {
```

Just for greater clarity and also to be open for the time when "backwards-compatible nodes" might diverge along more than just one property.

fbarl: Actually never mind, I misinterpreted the intention here and forgot that this lives in `backcompat.go` :)

report/marshal_test.go
```diff
@@ -71,3 +73,51 @@ func TestBiggerRoundtrip(t *testing.T) {
...
+        "latestControls": {
+          "docker_attach_container": {
+            "timestamp": "2019-10-14T14:36:01Z",
+            "value": {"dead": true}
+          },
+          "docker_remove_container": {
+            "timestamp": "2019-10-14T14:36:01Z",
+            "value": {"dead": false}
+          }
+        },
```
fbarl: Could you add a couple of more non-dead entries here? e.g. some with the same control name but different timestamps and some other with different control names for a more complete test?

report/marshal_test.go
```diff
@@ -72,3 +73,51 @@ func TestBiggerRoundtrip(t *testing.T) {
...
+        "latestControls": {
+          "docker_attach_container": {
+            "timestamp": "2019-10-14T14:36:01Z",
+            "value": {"dead": true}
+          },
+          "docker_remove_container": {
+            "timestamp": "2019-10-14T14:36:01Z",
+            "value": {"dead": false}
+          }
```
fbarl: Could you add a couple more entries here? e.g. some with the same control name but different timestamps and some with differing control names for a more complete test?

report/backcompat.go
bboreham: `ts` is initialized to the "zero value" for `time.Time` which compares as before any real time.
`ts` cannot be `nil` as it is a value not a pointer type.

report/node.go
bboreham: Right, it was bad copy/paste from the method above, thanks.
Go doesn't complain about unused parameters inside the function because they may be needed to conform to an interface.  And it's ok to supply an empty list when calling a `...` function.

report/marshal_test.go
bboreham: I added a little bit more but not sure if I understand you.

PR #3716, opened 2019-10-17
performance(aws_collector): don't persist shortcut reports

bboreham:
Shortcut reports are sent to update the UI quickly, on events like a container starting. 

We don't need to persist them in the time-travel data since the same information will be covered by a regular report a few seconds later.

### DIFF
--- a/app/multitenant/aws_collector.go
+++ b/app/multitenant/aws_collector.go
@@ -544,13 +544,17 @@ func (c *awsCollector) Add(ctx context.Context, rep report.Report, buf []byte) e
 		return err
 	}
 
-	reportSize, err := c.cfg.S3Store.StoreReportBytes(ctx, reportKey, buf)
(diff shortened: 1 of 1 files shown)


bboreham:
S3 charges per write (and read) as well as for monthly storage. 

I believe shortcut reports are sent immediately to the UI.

rade:
> I believe shortcut reports are sent immediately to the UI.

Even when time-travelling?

bboreham:
> I was wondering whether this make the UI less fluid when time-travelling

Sorry I didn't read that last bit.

In time-travel mode, if you inch the time forward by one second you would be able to tell the difference.
I'm not sure I would call that "fluid" since it's quite hard to achieve such fine control, and a moderately-sized report will take a couple of seconds to render.
I don't think this is worth commenting.

PR #3720, opened 2019-10-23
Adding the "user-agent" Header

### LINKED ISSUE #3715: Scope probe should set the user-agent string
Right now it shows up in access logs as "Go-http-client/1.1", which is very dull.

We could copy Prometheus, which sets it to something like "Prometheus/2.12.0".

DarthSett:
Added the "user-agent" header with the value: "Scope/1.11.6" and added a test for it.

Fixes #3715.

### DIFF
--- a/probe/appclient/app_client_internal_test.go
+++ b/probe/appclient/app_client_internal_test.go
@@ -33,7 +33,7 @@ func dummyServer(t *testing.T, expectedToken, expectedID string, expectedVersion
 			t.Errorf("want %q, have %q", expectedID, have)
 		}
 
-		if have := r.Header.Get("user-agent"); "Scope/1.11.6" != have {
+		if have := r.Header.Get("user-agent");  have != "Scope/1.11.6"  {
 			t.Errorf("want %q, have %q", app.Version, have)
 		}
 		var have report.Report
--- a/probe/appclient/app_client_internal_test.go
+++ b/probe/appclient/app_client_internal_test.go
@@ -32,6 +32,10 @@ func dummyServer(t *testing.T, expectedToken, expectedID string, expectedVersion
 			t.Errorf("want %q, have %q", expectedVersion, have)
 		}
 
+		if have := r.Header.Get("user-agent"); "Scope_Probe/"+expectedVersion!=have {
+			t.Errorf("want %q, have %q","Scope_Probe/"+expectedVersion,have)
+		}
+
 		var have report.Report
 
 		reader := r.Body
--- a/probe/appclient/app_client_internal_test.go
+++ b/probe/appclient/app_client_internal_test.go
@@ -3,7 +3,6 @@ package appclient
 import (
 	"compress/gzip"
 	"fmt"
-	"github.com/weaveworks/scope/app"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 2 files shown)


probe/appclient/probe_config.go
```diff
@@ -46,6 +46,7 @@ func (pc ProbeConfig) authorizeHeaders(headers http.Header) {
 	}
 	headers.Set(xfer.ScopeProbeIDHeader, pc.ProbeID)
 	headers.Set(xfer.ScopeProbeVersionHeader, pc.ProbeVersion)
+	headers.Set("user-agent","Scope/1.11.6" )
```
bboreham: Could you have this use the `ProbeVersion` string instead of hard-coding 1.11.6?

bboreham:
Thanks for the PR!

probe/appclient/probe_config.go
bboreham: It should be the probe version.  You could prepend "Scope_Probe/".
The app is the thing it is talking to.  They could be at different versions.

DarthSett:
I changed it to "Probe_Scope/probeversion". Please review the changes. I have committed and pushed them to my fork.

probe/appclient/probe_config.go
```diff
@@ -46,6 +46,9 @@ func (pc ProbeConfig) authorizeHeaders(headers http.Header) {
 	}
 	headers.Set(xfer.ScopeProbeIDHeader, pc.ProbeID)
 	headers.Set(xfer.ScopeProbeVersionHeader, pc.ProbeVersion)
+	headers.Set("user-agent","Probe_Scope/"+pc.ProbeVersion )
```
fbarl: Could you please rename like this as @bboreham suggested?

```suggestion
	headers.Set("user-agent","Scope_Probe/"+pc.ProbeVersion )
```

It's also more consistent with the rest of the code.

probe/appclient/app_client_internal_test.go
```diff
@@ -29,7 +29,7 @@ func dummyServer(t *testing.T, expectedToken, expectedID string, expectedVersion
 		}
 
 		if have := r.Header.Get(xfer.ScopeProbeVersionHeader); expectedVersion != have {
-			t.Errorf("want %q, have %q", expectedID, have)
+			t.Errorf("want %q, have %q", expectedVersion, have)
```
fbarl: Good catch, I wonder why the tests were passing before :thinking:

fbarl: Could you please also add a similar check here for the `user-agent` field while you're at it?

DarthSett: Oh yea. Sorry forgot to mention this change in the pr

probe/appclient/probe_config.go
DarthSett: Aaah. I thought i got it the right way the first time itself. Will commit this change.

probe/appclient/app_client_internal_test.go
bboreham: The tests are passing because this is just the bit that prints out what it found - the check is on the line before.

DarthSett:
I have committed the changes mentioned

fbarl:
Thanks @DarthSett, `user-agent` header looks good to me now.

Before we merge, could you please add a test for it? See https://github.com/weaveworks/scope/pull/3720#discussion_r338989917.

> ... added a test for it.

So far you've only fixed the existing test for `xfer.ScopeProbeVersionHeader`, but it would be nice to do the same check for `user-agent` ;)

DarthSett:
@fbarl Added the check for user-agent and pushed it to the fork. Please review if it looks good or not

PR #3721, opened 2019-10-24
build(client): deduplicate all yarn dependencies

bboreham:
Using https://github.com/atlassian/yarn-deduplicate

Where there were multiple versions of same dependency that all satisfied semver requirements, the highest version was used.

I see a lot of lines beginning ` integrity sha512-` have been removed - is that OK?

Following #3719 which was the first tentative step.

The image size hasn't gone down, so the duplicates don't appear to have been taking up any space.

### DIFF
--- a/client/yarn.lock
+++ b/client/yarn.lock
@@ -5,7 +5,6 @@
 "@babel/cli@^7.6.0":
   version "7.6.0"
   resolved "https://registry.yarnpkg.com/@babel/cli/-/cli-7.6.0.tgz#1470a04394eaf37862989ea4912adf440fa6ff8d"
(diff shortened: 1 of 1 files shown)


fbarl:
> I see a lot of lines beginning `integrity sha512`- have been removed - is that OK?

You probably have an old version of `yarn` installed - https://gitlab.com/gitlab-org/gitlab-foss/merge_requests/22027 (I'm on `1.19.0`). Try upgrading and then running `yarn-deduplicate` from `master` again, then I will take another look :)

bboreham:
I ran it inside the build container - it claims to be 1.19.4.
I can take another look...

fbarl:
Closing in favor of #3733.

PR #3733, opened 2019-12-05
Dedupe yarn.lock

fbarl:
Does the same #3721 did with two additions:

* I ran it on a current fresh master
* The `integrity sha512-` bits weren't removed

I gave it a smoke test with `yarn start` and it seems to be idempotent under `yarn install` now so things look good :+1:


fbarl:
> seems legit

Thanks for the review @bboreham!

PR #3735, opened 2019-12-12
Add more rbac permissions in the 'apps' group

bboreham:
Apparently `daemonset` was missed in the update from `v1beta1` to `v1`; also we need `get` and `update` on `deployments/scale`.

(I am a bit confused why nobody has pointed this out before)


bboreham:
I noticed in logs that getting the initial list of daemonsets was blocked.

PR #3740, opened 2020-01-13
Update Ruby version for the benefit of Netlify build

bboreham:
Scope does not use Ruby, but we (apparently) need this file to make the Netlify website build go.

And Netlify builds have been failing recently saying the version had to be at least 2.5.2.


bboreham:
It's still failing, possibly because a contentstack token is needed.
I know nothing about how the website is set up.

fbarl:
> Missing Contentstack environment variable: CONTENTSTACK_STACK_TOKEN

Hmm, it's weird there's nothing about this on the internet :thinking: 

The Ruby version bump still feels like a step forward though, as at passes through the step where it was failing before.

Maybe we could similarly set `.nvmrc` to something never, e.g. `v12.9.1`, the currently cached one `v6.17.1` is really old!

dholbach:
I don't know much about this either, but the Website repo, scope and weave-net will all have to have the same ruby version AFAIR.

Rich-Core:
I have gone ahead and added the missing Contentstack env variable to Netlify. Could we re-run this build test, please?

PR #3742, opened 2020-01-14
Convert to Go modules

bboreham:
Done by running `go mod init` then fixing compilation errors then `go mod vendor`.
A few C files needed to be manually added.

There are a lot of files moved or deleted, because `go mod vendor` flattens dependencies at the top level and the previous tool, `gvt`, did not.


Makefile
```diff
@@ -23,9 +23,8 @@ RM=--rm
 RUN_FLAGS=-ti
 BUILD_IN_CONTAINER=true
 GO_ENV=GOGC=off
-GO_BUILD_INSTALL_DEPS=-i
```
fbarl: Do we want to drop `-i` from the instruction `echo "    sudo go clean -i net";` down below as well?

fbarl:
Thanks for taking your time with this @bboreham!

Good catch with the last commit as well, as you mentioned `bpf.h` and `bpf_map.h` are waiting for https://github.com/iovisor/gobpf/pull/206, but what about `vendor/github.com/weaveworks/tcptracer-bpf/tcptracer-bpf.h`? Shall we open a PR that fixes that one?

Overall, the code looks good, wanted to look into the build and test the behavior quickly! :)

Makefile
bboreham: `-i` is actually correct there, as the intention was to install a different build of the standard library, but the whole bit is irrelevant now that Go caches builds for different tags in different places, so I'll remove it.

bboreham:
I made https://github.com/weaveworks/tcptracer-bpf/pull/69 to avoid the manual copy for one file.

PR #3743, opened 2020-01-15
kubernetes: detect more 'pause' containers

bboreham:
Dockershim has added a label `io.kubernetes.docker.type` for at least four years, where the pause container is of type `podsandbox`.  This should be more reliable than trying to keep up with everyone's name for the pause container.

I don't like the way we do this in two places (one probe, one app), but it seemed a bit complicated to unwind that right now.

I left in the old name-based code just in case; I don't think it is necessary any more.


bboreham:
I should really do a unit test for this.

PR #3746, opened 2020-01-28
probe: avoid reporting the container ID as its hostname

bboreham:
If hostname isn't set on a container it defaults to a random hex number, which we don't want to show in the UI.


probe/docker/container.go
```diff
@@ -108,6 +108,11 @@ func (c *container) PID() int {
 
 func (c *container) Hostname() string {
 	if c.container.Config.Domainname == "" {
+		// If hostname isn't set on a container it defaults to a random hex
+		// number which we don't want to show in the UI.
```
fbarl:
```suggestion
		// number starting with the container ID which we don't want to show in the UI.
```

bboreham:
Example: you can see that the hostname of the container is the first 12 characters of the container ID.

image

bboreham:
However, with this PR the second container disappears from the "containers by DNS name" view. 
I'm not sure that is an improvement.

PR #3747, opened 2020-01-28
Simplify some renderers

bboreham:
`ContainerImageRenderer`: Crunch a `Map` and `Reduce` into a single `Renderer` that has the same function.

`ContainerHostnameRenderer`: The second half of the render added all the same nodes, except for
those with no Docker image information, so we could show a figure for those filtered out on that basis. This just isn't worth the effort.

And remove unused parameterisation of `containerWithImageNameRenderer`

### DIFF
--- a/render/container.go
+++ b/render/container.go
@@ -116,14 +116,12 @@ func HasChildren(topology string) FilterFunc {
 	}
 }
 
-type containerWithImageNameRenderer struct {
-	Renderer
-}
+type containerWithImageNameRenderer struct{}
 
 // Render produces a container graph where the the latest metadata contains the
 // container image name, if found.
 func (r containerWithImageNameRenderer) Render(ctx context.Context, rpt report.Report) Nodes {
-	containers := r.Renderer.Render(ctx, rpt)
+	containers := ContainerRenderer.Render(ctx, rpt)
 	images := SelectContainerImage.Render(ctx, rpt)
 
 	outputs := make(report.Nodes, len(containers.Nodes))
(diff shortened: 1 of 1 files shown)


render/container.go
```diff
@@ -177,20 +169,9 @@ var ContainerImageRenderer = Memoise(FilterEmpty(report.Container,
 //
 // not memoised
 var ContainerHostnameRenderer = FilterEmpty(report.Container,
-	MakeReduce(
-		MakeMap(
-			MapContainer2Hostname,
-			ContainerWithImageNameRenderer,
-		),
-		// Grab *all* the hostnames, so we can count the number which were empty
-		// for accurate stats.
```
fbarl: Where would these stats be displayed?

render/container.go
```diff
@@ -177,20 +169,9 @@ var ContainerImageRenderer = Memoise(FilterEmpty(report.Container,
...
-		MakeMap(
-			MapToEmpty,
```
fbarl: After removing it from here, this method is not used anymore so feel free to remove it from `container.go`.

render/container.go
```diff
@@ -155,20 +153,14 @@ func (r containerWithImageNameRenderer) Render(ctx context.Context, rpt report.R
 
 // ContainerWithImageNameRenderer is a Renderer which produces a container
 // graph where the ranks are the image names, not their IDs
-var ContainerWithImageNameRenderer = Memoise(containerWithImageNameRenderer{ContainerRenderer})
+var ContainerWithImageNameRenderer = Memoise(containerWithImageNameRenderer{})
 
-// ContainerImageRenderer is a Renderer which produces a renderable container
-// image graph by merging the container graph and the container image topology.
+// ContainerImageRenderer produces a graph where each node is a container image
+// with the original containers as children
```
fbarl: :+1: Thanks for clarity!

render/container.go
bboreham: Bottom-left of the screen, like "21 nodes (119 filtered)"

bboreham:
I am yak-shaving towards a significant performance improvement; this one piece doesn't do much but it helps to simplify the problem.

render/container.go
fbarl: Ok cool. That information's been misleading and incorrect at times for container topologies so I'm ok seeing it removed.

Looking at the UI code, it seems it's already robust enough to not display filtered nodes count if it's not provided so probably nothing will need to be updated there :+1:

PR #3748, opened 2020-01-31
Move Counters into Latest

bboreham:
The only place `Counters` are used is in rendering, for the number of nodes under a topology, so the overhead of holding a unique data structure in every `Node` is unwarranted.

Counters are not set in the probe, so we don't need any backwards-compatibility in report decoding. Similarly they are not set until after all nodes are merged, so we don't need that logic.
(It was a bug if we did use it - merges are supposed to be idempotent).


report/node.go
```diff
@@ -80,10 +79,26 @@ func (n Node) WithLatest(k string, ts time.Time, v string) Node {
...
+}
+
+// WithCounter returns a fresh copy of n, with Counter c added in.
+// (counters are stored as strings, to keep the data structure simple)
+func (n Node) WithCounter(k string, value int) Node {
+	name := CounterPrefix + k
+	if prevStr, found := n.Latest.Lookup(name); found {
+		prevValue, _ := strconv.Atoi(prevStr)
+		value += prevValue
+	}
```
fbarl: Why not use `LookupCounter` here to further simplify it?

report/node_test.go
```diff
@@ -146,33 +146,34 @@ func TestMergeNodes(t *testing.T) {
...
-			},
-		},
 	} {
 		if have := c.a.Merge(c.b); !reflect.DeepEqual(c.want, have) {
 			t.Errorf("%s: %s", name, test.Diff(c.want, have))
 		}
 	}
 }
+
+func TestCounters(t *testing.T) {
```
fbarl: > Counters are not set in the probe, so we don't need any backwards-compatibility in report decoding. Similarly they are not set until after all nodes are merged, so we don't need that logic.
(It was a bug if we did use it - merges are supposed to be idempotent).

Could you please add (some of) this is a comment to the test, just to justify that merges are not suppose to work anymore.

report/node.go
```diff
@@ -80,10 +79,26 @@ func (n Node) WithLatest(k string, ts time.Time, v string) Node {
...
+	var str string
+	if str, found = n.Latest.Lookup(name); found {
+		value, _ = strconv.Atoi(str)
+	}
+	return value, found
+}
+
+// WithCounter returns a fresh copy of n, with Counter c added in.
+// (counters are stored as strings, to keep the data structure simple)
+func (n Node) WithCounter(k string, value int) Node {
```
fbarl:
```suggestion
func (n Node) AddCounter(k string, value int) Node {
```

I suggest this change for two reasons:

1. `AddCounter` suggests more explicitly that if the new value will be _added_ to the old counter value if it existed, rather than overwritting it
2. `Counters.Add` -> `AddCounter` is a smoother change, also consistent with `Counters.Lookup` -> `LookupCounter`

WDYT?

report/node_test.go
bboreham: They _work_ ; it's just that there is no special logic to merge counters now they are just values.  I'll add a note that we don't check this.

report/node.go
bboreham: Possibly because I thought computing `name := CounterPrefix + k` twice would slow it down, but this doesn't appear to be the case. Maybe the compiler can inline `LookupCounter`.
So done.

report/node_test.go
bboreham: done

bboreham:
It turned out this needed more work because `ContainerHostnameRenderer` was relying on the additive merge behaviour, also this wasn't being checked in unit tests.
So I extended the unit tests to check counts and to cover the case where the count is not 1.
Then re-did `ContainerHostnameRenderer` as a `Renderer` so it didn't need additive merge.

PR #3754, opened 2020-03-10
Removed kubelet port flag. Node name now always need from env/flag

CiMaol:
Fix to issue #3242. Kubelet flag is removed from the spec and tests relying on mocked versions of the getlocalpods method have been changed to no longer rely on this method. Kubelet API files, which were using the unsecure methods, have been removed.


CiMaol:
Any pointers on why the integration tests aren't going through? Seems it's not able to resolve some files in vendor and it's not able to resolve a symbol in the docker package "namespaceIPAddress"

bboreham:
Thanks for the PR!  Integration tests are turned off on forks; see #2192.
It is unfortunate that when this happens the reason is not made clear.

I have pushed your fork to the weaveworks/scope repo so the tests will now re-run.

bboreham:
In the issue #3242 I noted:
> "the main work is in checking the system still works, and considering how to warn the user if they don't supply a node name"

Could you say what you did on those points?  What does happen if they don't supply a node name?

There are now two modes to run the Scope probe with Kubernetes: the default every-node-captures-everything, or via `-probe.kubernetes.role` designate just one to connect to the api-server.  I think this PR is most relevant to the default mode, but it should do something reasonable in both cases.

CiMaol:
The nodeName flag can be provided manually and if not scope will attempt to pull it from the KUBERNETES_NODENAME env variable. I think right now the reporter that calls the podTopology method will only get added if the kubernetes role is cluster.

probe/kubernetes/reporter.go
```diff
@@ -658,26 +654,13 @@ func (r *Reporter) podTopology(services []Service, deployments []Deployment, dae
...
-	var localPodUIDs map[string]struct{}
-	if r.nodeName == "" && r.kubeletPort != 0 {
-		// We don't know the node name: fall back to obtaining the local pods from kubelet
-		var err error
-		localPodUIDs, err = GetLocalPodUIDs(fmt.Sprintf("127.0.0.1:%d", r.kubeletPort))
-		if err != nil {
-			log.Warnf("No node name and cannot obtain local pods, reporting all (which may impact performance): %v", err)
-		}
+	// filter out non-local pods: we only want to report local ones for performance reasons.
+	if r.nodeName == "" {
```
CiMaol: I think, on rereading this, that the nodeName=="" check should be inside rather than outside the WalkPods function call. right now this is not allowing a cluster agent to get the node for any pod. If inside it will move on to the pods.AddNode() call which will allow it to attach a node to a pod.

CiMaol: Changed the order of this check in latest commit.

bboreham:
> I think right now the reporter that calls the podTopology method will only get added if the kubernetes role is cluster.

It checks that the mode is not host, but there are three possibilities: host, cluster or neither (for backwards-compatibility, since it requires changing the way you deploy the probe).

CiMaol:
@bboreham I'm assuming that neither is the out-of-cluster deployment? Or is there a separate deployment mode for Kubernetes? 

If it's the out of cluster mode then this check for kubelet won't have any impact. The code as is will only fire for kubernetesRole=Cluster, and when it does it will add a node name for each pod in the list.

bboreham:
The three possibilities are:
- `kubernetesRole` is `""`.  This is how one would deploy Scope prior to #3274; each probe talks to Kubernetes and tries to report its own pods.
- `kubernetesRole` is `"cluster"`; only this probe talks to Kubernetes and should report all pods.
- `kubernetesRole` is `"host"`; this probe should not talk to Kubernetes.

I believe the change you have made will impact the first two.

Still, we don't need to argue about theoreticals; my question is which of these have you tested and what does it do in the case that the node name is not supplied?

CiMaol:
I've manually tested the named kubernetesRoles. Like you said in the host role case this method isn't called and has no effect.  In the cluster role case the WalkPods will run the addNode method for every pod if no nodeName is passed to it.

CiMaol:
I'm not certain of how to approach the first Role - i.e. no role given. Is it a case that this is supported as a run-mode even with new changes to the code such as this one? In the other thread there seem to be explicit calls for removing the kubelet code. 

If this is still supposed to support per-node reporting maybe we can do a check for the flags and a warning at set up time? Do you think this is a reasonable way to signal to the user - or should there be something harder than a log.Warn()?

bboreham:
Thanks for doing those tests.

I think we have to expect people are still running Scope as a single DaemonSet with blank role.  Currently it works, without calling kubelet, so long as a node name is supplied. 
We could decide not to support blank role after this PR, given some justification.

Logging a warning would be fine, together with either reporting no pods or all pods.  The main thing I want to ensure is that it doesn't crash.

bboreham:
I've pushed a commit to your branch adding a warning.

PR #3755, opened 2020-03-10
Update babel, jest, webpack bits

dholbach:
This should fix the handlebars and serialize-javascript security alerts on GH at least.

It also updates the node requirement (I think through the jest update this was necessary) - somebody should check if we can actually make the jump.


dholbach:
https://circleci.com/gh/weaveworks/scope/14642 says

> [1/5] Validating package.json...
> error weave-scope@0.0.0: The engine "node" is incompatible with this module. Expected version "^10.19.0".
> error Found incompatible module

I guess something like https://github.com/weaveworks/scope/commit/56f2fac4874fcdad9463eb3c0eb72024be05fb38 is needed here?

bboreham:
I applaud your efforts, but really have no idea what changes are required when you move two major versions of Node.

I think you should have permissions to push a new build image; let me know if not.

dholbach:
> I think you should have permissions to push a new build image; let me know if not.

Looks like I don't. Can you help me out? My user is `dholbach` on Dockerhub too.

```
denied: requested access to the resource is denied
```

PR #3757, opened 2020-03-11
Switch UI build container back to master, and update more JS deps

dholbach:
Revert https://github.com/weaveworks/scope/commit/795eaab7f52c5de6ae393741aed403c025851b3a and match deps with service-ui/ui-components.


dholbach:
Follow up to #3755.

PR #3763, opened 2020-03-19
security: update JS dependencies which have notified vulnerabilities

bboreham:
I updated five listed on https://github.com/weaveworks/scope/network/alerts:
minimist, acorn, set-value, mixin-deep, sshpk

I did not update clean-css because I believe it's only used at build time.

All changes are point versions only.

I ran `yarn-deduplicate` as in #3733 prior to starting, and after each change.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -43,6 +43,7 @@
     "reqwest": "2.0.5",
     "reselect": "4.0.0",
     "reselect-map": "1.0.4",
+    "sshpk": "^1.13.2",
     "styled-components": "4.3.2",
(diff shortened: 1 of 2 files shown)


fbarl:
I think this is a good way to go @bboreham!

However, once you updated `yarn.lock`, you could have removed all these entries you added to `package.json` and reran `yarn install` - I just did that locally and `yarn.lock` stayed the same.

That means we maintain the JS security updates from this PR while still keeping `package.json` relatively clean; let me try making a small follow-up PR for this.

PR #3767, opened 2020-03-21
Security ypgrade to Alpine:1.11.0

dnascimento:
Container scanning tools are reporting 6 CVEs due to an older version of Alpine. 
 
| Vulnerability | Package | Fix |
|-|-|-|
| CVE-2019-14697 | musl-utils-1.1.18-r3   | 1.1.18-r4 |
| CVE-2019-14697 | musl-1.1.18-r3         | 1.1.18-r4 |
| CVE-2019-5482  | libcurl-7.61.1-r2      | 7.61.1-r3 |
| CVE-2019-5481  | libcurl-7.61.1-r2      | 7.61.1-r3 |
| CVE-2019-5482  | curl-7.61.1-r2         | 7.61.1-r3 |
| CVE-2019-5481  | curl-7.61.1-r2         | 7.61.1-r3 |

PR #3768, opened 2020-03-23
Update alpine base image to 3.11.3

bboreham:
Note I'm pinning the minor version here, otherwise it can vary depending on what's on your machine when you build.

Replaces #3767


dnascimento:
Thanks @bboreham ! Nice work 👍

PR #3773, opened 2020-03-26
fix (multitenant collector): Use consistent report timestamp

bboreham:
Previously the code called `time.Now()` in two different places so the timestamps didn't match. 
Now we use the timestamp of the report itself.

Add the collector's local time to the report if it didn't have one.


bboreham:
It turns out this change is problematic in the case that the probe's idea of the time is different to the app's: the newest reports may fall outside of the "display window" and hence nothing appears in view.

PR #3776, opened 2020-03-30
security(js-dependency) update html-webpack-plugin to most recent stable

dholbach:


client/package.json
```diff
@@ -72,7 +72,7 @@
     "eslint-plugin-react": "7.14.3",
     "express": "4.16.4",
     "file-loader": "^2.0.0",
-    "html-webpack-plugin": "3.2.0",
+    "html-webpack-plugin": "4.0.3",
```
fbarl: Couldn't find the docs if there were any breaking changes introduced with this major bump, but I checked the options docs https://github.com/jantimon/html-webpack-plugin#options and we use them the same way. Also ran Scope with `yarn start` locally and it looks fine :+1:

dholbach: I ran it locally too, just to be sure.

PR #3780, opened 2020-04-13
Multitenant: merge incoming reports in collector

bboreham:
This means we store fewer, bigger, reports, which reduces cost of storage and time to render when data is viewed.

### DIFF
--- a/app/multitenant/aws_collector.go
+++ b/app/multitenant/aws_collector.go
@@ -215,6 +215,27 @@ func (c *awsCollector) flushLoop() {
 // Range over all users (instances) that have pending reports and send to store
 func (c *awsCollector) flushPending(ctx context.Context) {
 	instrument.CollectedRequest(ctx, "FlushPending", flushDuration, nil, func(ctx context.Context) error {
+		type queueEntry struct {
+			userid string
+			buf    []byte
+		}
+		queue := make(chan queueEntry)
+		const numParallel = 10
+		var group sync.WaitGroup
+		group.Add(numParallel)
+		// Run n parallel goroutines fetching reports from the queue and flushing them
+		for i := 0; i < numParallel; i++ {
+			go func() {
+				for entry := range queue {
+					rowKey, colKey, reportKey := calculateReportKeys(entry.userid, time.Now())
+					err := c.persistReport(ctx, entry.userid, rowKey, colKey, reportKey, entry.buf)
+					if err != nil {
+						log.Errorf("Could not persist combined report: %v", err)
+					}
+				}
+				group.Done()
+			}()
+		}
 		c.pending.Range(func(key, value interface{}) bool {
 			userid := key.(string)
 			entry := value.(*pendingEntry)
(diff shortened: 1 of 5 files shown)


app/multitenant/aws_collector.go
```diff
@@ -169,14 +190,82 @@ func NewAWSCollector(config AWSCollectorConfig) (AWSCollector, error) {
...
+
+// Range over all users (instances) that have pending reports and send to store
+func (c *awsCollector) flushPending(ctx context.Context) {
+	instrument.CollectedRequest(ctx, "FlushPending", flushDuration, nil, func(ctx context.Context) error {
+		type queueEntry struct {
+			userid string
+			buf    []byte
+		}
+		queue := make(chan queueEntry)
+		const numParallel = 10
```
fbarl, line 223: How did you decide on the value here? Do you think it would useful to make it an arg param?

app/multitenant/aws_collector.go
```diff
@@ -537,63 +659,40 @@ func (c *awsCollector) Add(ctx context.Context, rep report.Report, buf []byte) e
...
-		reportsPerUser.WithLabelValues(userid).Inc()
-	}
-
-	// third, put it in memcache
-	if c.cfg.MemcacheClient != nil {
-		_, err = c.cfg.MemcacheClient.StoreReportBytes(ctx, reportKey, buf)
-		if err != nil {
-			// NOTE: We don't abort here because failing to store in memcache
-			// doesn't actually break anything else -- it's just an
-			// optimization.
```
fbarl, line 565: Why not keep this comment?

fbarl: Ah I see, now we actually do abort here - any reason why that `return nil` was added?

app/multitenant/aws_collector.go
```diff
@@ -169,14 +190,82 @@ func NewAWSCollector(config AWSCollectorConfig) (AWSCollector, error) {
...
+			userid := key.(string)
+			entry := value.(*pendingEntry)
+
+			entry.Lock()
+			rpt, count := entry.report, entry.count
+			entry.report, entry.count = report.MakeReport(), 0
+			entry.Unlock()
+
+			if count > 0 {
+				// serialise reports on one goroutine to limit CPU usage
```
fbarl, line 249: Can you think of cases (hardware configurations) where it might be desirable to branch out here instead? I'm thinking whether there would be any value in making this an optional arg as well.

fbarl:
Conceptually LGTM! I only added a few minor comments concerning the config robustness.

I just looked at the code though (didn't go as far as trying to test) - how well do you think the current unit tests cover the logic modified here?

app/multitenant/aws_collector.go
bboreham: Yes it would be more flexible, but the way flags are plumbed through from `main()` makes it annoying to add more.
There is a balancing act to get all serialisation plus IO completed within the tick interval, and I don't really know what might need adjusting.  I'm happy changing the code as I learn from running it; more than adding all possible configuration parameters then never using most of them.

app/multitenant/aws_collector.go
bboreham: It's possible, but the collector can be run across multiple processes (pods) to get more CPUs working on the task.
See comments above about the pain of adding more options.  Plus it raises the testing burden.

bboreham:
The current unit tests don't cover any of the DynamoDB/S3 code.
I think if I was going to add tests I'd do it as integration tests, running mock DynamoDB and S3 containers.

app/multitenant/aws_collector.go
bboreham: I have added it back, also another comment on the case where a shortcut report can't be stored.

PR #3797, opened 2020-06-09
Update build tools

dholbach:


bboreham:
One of the changes from the `build-tools` repo was https://github.com/weaveworks/build-tools/pull/163 to fix the lint script which was doing nothing. Now it is actually checking files, we need to fix all the places where it complains (and ignore a few that we aren't going to fix).

bboreham:
@dholbach can you either let me know what more needs to be done, or take off the 'draft' marker.

dholbach:
This looks good to me.

Thanks for your work on this Bryan!

Issue #3805, opened 2020-06-29
weavescope installation on k3s not working


milindchawre:
**BUG REPORT**

## What you expected to happen?
Expecting weavescope to work on k3s.

## What happened?
- I installed weavescope on my k3s cluster using https://github.com/weaveworks/scope/tree/master/examples/k8s
- Weavescope is configured and can see the UI, but containers are not visible all the k8s and docker probes where not working.
- Found error logs in `weave-scope-cluster-agent` pod.
```
root@kube-master-ce86:/# kubectl logs pod/weave-scope-cluster-agent-669d84579b-qm6dx -n weave
time="2020-06-29T17:22:35Z" level=info msg="publishing to: weave-scope-app.weave.svc.cluster.local.:80"
<probe> INFO: 2020/06/29 17:22:35.936466 Basic authentication disabled
<probe> INFO: 2020/06/29 17:22:35.950455 command line args: --mode=probe --probe-only=true --probe.cri.endpoint=unix///var/run/k3s/containerd/containerd.sock --probe.kubernetes.role=cluster weave-scope-app.weave.svc.cluster.local.:80
<probe> INFO: 2020/06/29 17:22:35.950581 probe starting, version 1.13.1, ID 37b7bd86d15ce42f
<probe> ERRO: 2020/06/29 17:22:35.950669 Error getting docker bridge ip: route ip+net: no such network interface
<probe> INFO: 2020/06/29 17:22:35.950974 kubernetes: targeting api server https://192.168.128.1:443
<probe> WARN: 2020/06/29 17:22:35.997726 Error collecting weave status, backing off 10s: Get http://127.0.0.1:6784/report: dial tcp 127.0.0.1:6784: connect: connection refused. If you are not running Weave Net, you may wish to suppress this warning by launching scope with the `--weave=false` option.
<probe> INFO: 2020/06/29 17:22:36.070069 volumesnapshots are not supported by this Kubernetes version
<probe> INFO: 2020/06/29 17:22:36.074895 Control connection to weave-scope-app.weave.svc.cluster.local. starting
<probe> INFO: 2020/06/29 17:22:36.101873 Publish loop for weave-scope-app.weave.svc.cluster.local. starting
[...]

PR #3806, opened 2020-07-01
Updating cluster-role.yaml to fix rbac

milindchawre:
This PR is intended to fix the issue mentioned in [this](https://github.com/weaveworks/scope/issues/3805#issuecomment-652585371) comment.

**Issue:** The serviceaccount - weave-scope don't have permission to list resource "namespaces" in API group "" at the cluster scope.

**Fix:** Have updated the cluster role to include listing of `namespaces` resource.

### DIFF
--- a/examples/cri/cluster-role.yaml
+++ b/examples/cri/cluster-role.yaml
@@ -15,6 +15,7 @@ rules:
   - replicationcontrollers
   - services
   - nodes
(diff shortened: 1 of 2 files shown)


bboreham (approved):
Thanks!  I guess it used to be possible to list namespaces without explicit permission.

PR #3833, opened 2021-02-18
Update build tools including Go 1.16.2

bboreham:
- Stop creating a NodeJS build image, just use `node:10.19` directly.
 - Update to Go 1.16.2.
 - Remove some unused tools.
 - Fix some build issues on MacOS.
 - Add 'osusergo' tag to remove some link-time warnings.
 - Remove 'demo.json' file, not used in Scope.
 - Miscelaneous other build improvements.


backend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.14.2
+FROM golang:1.16.2
```
dholbach, line 1: Will `go.mod` need an update too?

dholbach:
Great work - this looks much cleaner now!

backend/Dockerfile
bboreham: You mean the line in `go.mod` saying `go 1.13` ?
I read up on that https://golang.org/ref/mod#go-mod-file-go
It seems there would be some benefit from updating it.

If you mean updating the other Go package dependencies, then I'm happy to leave that to another PR.

dholbach: :+1:

Issue #3832, opened 2021-02-18
make command is failing with printer dependency


srajmane:
## What you expected to happen?
To pass make command

## What happened?

While building weaveworks/scope-backend-build docker image
make command is giving below issue on master and v1.13.1 :
```
/src/github.com/weaveworks/scope$ make
sudo -E docker build -t weaveworks/scope-backend-build backend
Sending build context to Docker daemon  6.144kB
Step 1/12 : FROM golang:1.13.9
...
...
Step 4/12 : RUN go clean -i net &&      go install -tags netgo std &&
..
package github.com/hashicorp/hcl/hcl/printer: cannot find package "github.com/hashicorp/hcl/hcl/printer" in any of:
        /usr/local/go/src/github.com/hashicorp/hcl/hcl/printer (from $GOROOT)
        /go/src/github.com/hashicorp/hcl/hcl/printer (from $GOPATH)
```

## How to reproduce it?

To reproduce this error just clone scope and run make command
```
git clone https://github.com/weaveworks/scope.git $GOPATH/src/github.com/weaveworks/scope
cd $GOPATH/src/github.com/weaveworks/scope
make
```

## Versions:

```
$ scope version : v1.13.1 and master
$ docker version: 19.03.6
$ uname -a : Linux  4.15.0-135-generic #139-Ubuntu x86_64 x86_64 x86_64 GNU/Linux
$ kubectl version
```

bboreham:
> ..

You appear to have elided the part that would show what it was doing at the time it failed.

However it looks like the upstream dependency made a breaking change. Probably something else needs to be updated to match (or removed; there is some old cruft in there)

srajmane:
@bboreham To avoid too many lines in the log I have added `...`.
Please check the below-mentioned log.
```
~/src/github.com/weaveworks/scope$ make
sudo -E docker build -t weaveworks/scope-backend-build backend
Sending build context to Docker daemon  6.144kB
Step 1/12 : FROM golang:1.14.2
 ---> 2421885b04da
Step 2/12 : ENV SCOPE_SKIP_UI_ASSETS true
 ---> Using cache
 ---> e2b73c3a8738
[...]

PR #3834, opened 2021-03-03
Remove obsolete dependency github.com/fatih/hclfmt

Kielek:
github.com/fatih/hclfmt was archived. Technically there is a possibility to fork (or used another forked version) to executle hclfmt.

But it leads to another issues - problems with importing https://github.com/hashicorp/hcl/hcl/printers. HCL was released in version 2 so that the file structure was changes.
I do not see any way to install it in changed Makefile. Go installs does not support pining to given tags so that always the newest version is installed.

For now, the impact on build process should be minimal. it is used only in ./tools/lint/lint() method. I was able to execute it locally without executing hclfmt.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -21,7 +21,6 @@ RUN go clean -i net && \
 		github.com/fzipp/gocyclo \
 		golang.org/x/lint/golint \
 		github.com/kisielk/errcheck \
-		github.com/fatih/hclfmt \
 		github.com/mjibson/esc \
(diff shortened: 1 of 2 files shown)


Kielek:
Should solbe #3832

Kielek:
@bboreham, could you please check?

bboreham:
Sorry, been doing other things.
Please feel free to ping me again in a few days if I still haven't looked.

Kielek:
@bboreham?

bboreham (approved):
Thanks!

PR #3836, opened 2021-03-08
Update faq.md - fix read-only mode gotcha

### LINKED ISSUE #3835: Allow disabling container operations
I have setup weavescope on kubernetes. I would like to expose it to developers behind a vpn. We do not want to allow developers to directly pause, stop or restart individual containers.

I have tried restricting access in the cluster role but that doesn't seem to be working. I am able to restrict terminating pods but not containers. 

I may have missed something fairly obvious here. Please do let me know if its possible to disable container operations.

kingdonb:
Someone's text editor has made this `U+2014 : EM DASH` instead of a double-dash. Closes #3835

### DIFF
--- a/site/faq.md
+++ b/site/faq.md
@@ -15,7 +15,7 @@ the YAML manifests from `cloud.weave.works`.
 
 ## Disabling Scope Write Access
 
-Can be done by using the `probe.no-controls` option and set it to false for the scope agents. This can be done in the scope deployment manifest under the `weave-scope-agent`'s argument section with `—probe.no-controls=true`.
+Can be done by using the `probe.no-controls` option and set it to true for the scope agents. This can be done in the scope agents' manifests under the `weave-scope-cluster-agent` deployment and `weave-scope-agent` daemonset argument section with `--probe.no-controls=true`.
 
 ## RBAC and Weave Scope OSS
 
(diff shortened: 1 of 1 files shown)


kingdonb:
I think the following section has a similar error. The dashes are proper ASCII hyphen-minus, but they are single dashes and should be double dashes. Pushing up another commit to address this while I'm here.

site/faq.md
```diff
@@ -15,7 +15,7 @@ the YAML manifests from `cloud.weave.works`.
 
 ## Disabling Scope Write Access
 
-Can be done by using the `probe.no-controls` option and set it to false for the scope agents. This can be done in the scope deployment manifest under the `weave-scope-agent`'s argument section with `—probe.no-controls=true`.
+Can be done by using the `probe.no-controls` option and set it to false for the scope agents. This can be done in the scope daemonset manifest under the `weave-scope-agent`'s argument section with `--probe.no-controls=true`.
```
bboreham, line 18: Nowadays we have both a Deployment (which handles Kubernetes data for the whole cluster) and a DaemonSet (which handles everything else on a per-host basis).

bboreham (commented):
Thanks; it turns out single-hyphen arguments are valid, though I'm ok to change for consistency.
https://golang.org/pkg/flag/#hdr-Command_line_flag_syntax

site/faq.md
kingdonb: Ah. Does it need to be passed on to both then? I only noticed one (my cluster is also a single node)

My update went to the daemonset and it seemed to do the trick

kingdonb:
That makes sense. I was just going by what I saw was already in use in existing args in the daemonset.

site/faq.md
kingdonb: I read once more carefully and it looks like there was another error in this paragraph. I think it reads correctly now.

kingdonb:
This was confirmed in #3835 👍

bboreham (approved):

PR #3838, opened 2021-03-11
Update alpine base image to 3.13

gaby:
Update alpine 3.11 to latest patch version.

It would make more sense at this point to update to 3.13, let me know if you want make to make the change. 

Change log for 3.11.8: https://alpinelinux.org/posts/Alpine-3.10.6-3.11.8-3.12.4-released.html

### DIFF
--- a/docker/Dockerfile.cloud-agent
+++ b/docker/Dockerfile.cloud-agent
@@ -1,4 +1,4 @@
-FROM alpine:3.11.3
+FROM alpine:3.11.8
 WORKDIR /home/weave
 RUN apk add --no-cache bash conntrack-tools iproute2 util-linux curl
 COPY ./scope /home/weave/
--- a/docker/Dockerfile.cloud-agent
+++ b/docker/Dockerfile.cloud-agent
@@ -1,6 +1,6 @@
-FROM alpine:3.13
+FROM alpine:3.13.2
 WORKDIR /home/weave
-RUN apk add --no-cache bash conntrack-tools iproute2 util-linux curl
+RUN apk add --no-cache bash conntrack-tools iproute2 util-linux curl libpcap-dev
 COPY ./scope /home/weave/
 ENTRYPOINT ["/home/weave/scope", "--mode=probe", "--no-app", "--probe.docker=true"]
 
--- a/docker/Dockerfile.cloud-agent
+++ b/docker/Dockerfile.cloud-agent
@@ -1,6 +1,6 @@
 FROM alpine:3.13.2
 WORKDIR /home/weave
-RUN apk add --no-cache bash conntrack-tools iproute2 util-linux curl libpcap-dev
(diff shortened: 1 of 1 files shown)


bboreham:
Happy to make a bigger jump.

I don't believe Alpine flaws are exploitable in a normal Scope situation, but I recognize that folks run tools which highlight them.

gaby:
@bboreham Great, changes have been made. Alpine now points to 3.13

kingdonb:
Looks like the build fails now, you might try adding `libpcap-dev`?

Also, the old Dockerfile was specific about `MAJOR`.`MINOR`.`Patch`, let's keep that property... would you change the form of the FROM line from `3.13` back to `3.13.2` that mentions the more specific latest release available at this time?

gaby:
Test still failing, I will take a look again tonight. The integration test for alpine 3.11.8 also failed, so it must be something else missing. 

See https://github.com/google/gopacket/issues/734

gaby:
I tried debugging this, but i can't figure out why it fails. PR #3834 is also failing with the same error.

bboreham:
The integration tests basically don't run at all for 3rd-party forks, due to mining. #2192

I have pushed your branch as 1st-party, so now the tests will re-run and hopefully succeed.

I apoologise the error message is very hard to understand; it would be better to fail the job straight off if it doesn't have secrets.

docker/Dockerfile.cloud-agent
```diff
@@ -1,6 +1,6 @@
-FROM alpine:3.11.3
+FROM alpine:3.13.2
 WORKDIR /home/weave
-RUN apk add --no-cache bash conntrack-tools iproute2 util-linux curl
+RUN apk add --no-cache bash conntrack-tools iproute2 util-linux curl bcc-tools libpcap-dev
```
bboreham, line 3: We don't want things relating to 'dev' in the final release container; we have a build container for those.

I backed out this change at https://github.com/weaveworks/scope/commit/5331034059ab9d5ed49afb58f854a8715e87c05e
so we can check the CI run at https://app.circleci.com/pipelines/github/weaveworks/scope/357/workflows/a74f43bf-9696-4bff-bb60-a969f6a02b46

gaby: @bboreham Thank you for looking at this, tests look good.

bboreham: Ok, do you want to update your branch so I can merge this PR, or shall I merge that branch above?

gaby: @bboreham Done, I just removed the dev dependencies from my branch.

Kielek:
@bboreham, do you have any plans to official 1.13.2 release with the latest changes? Including new alpine?

Kielek:
@bboreham, any change to official 1.13.2 release with the latest changes? Including

PR #3845, opened 2021-04-13
Add s390x support

vibhutisawant:
This PR adds support for s390x architecture. 

Changes implemented:
1) Skipped `go install -race -tags netgo std;` in backend/Dockerfile as ` -race ` flag is not supported on s390x.
2) Added docker binary link for s390x arch in backend/Dockerfile
3) Added s390x in `probe/endpoint/dns_snooper_others.go`  as the make command failed for s390x arch with `undefined: DNSSnooper` error.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -14,7 +14,7 @@ RUN set -eux; \
 RUN go clean -i net && \
 	go install -tags netgo std && \
    export arch_val="$(dpkg --print-architecture)"; \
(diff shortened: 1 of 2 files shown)


bboreham (approved):
Thanks for the PR! LGTM.

I'm just running the integration tests to check it hasn't broken anything.

vibhutisawant:
Hi @bboreham 
Looks like integration tests have passed. Can this PR be merged?

PR #3846, opened 2021-04-18
Multitenant mode: fetch live data from collectors

### LINKED ISSUE #3783: Idea for more efficient multitenant rendering
Scope has an optional multitenant deployment model:
 - 1 or more "collector" processes save reports to store (e.g. S3+DynamoDB).
 - Probes send reports to a load-balancer front-end which distributes incoming reports across collectors.
 - 1 or more "query" processes handle requests from browsers, fetch reports from store, merge and render them.

Currently, the collectors save all reports to store in near-real-time, and queriers do all rendering by querying the store as-at the time required. Live rendering is done by repeatedly querying as-at "now".

New proposal, which makes the store optional for live rendering:
  - Each collector process holds recent reports in memory, for the length of the rendering window (e.g. 15 seconds)
   - Need to maintain a sliding window of data; can't just store 15 seconds then flush and start from scratch.
 - When a query process needs to render "live", it fetches the data held by all collectors.
   - Query process needs to be able to find all collectors (i.e. some sort of service discovery).

We can still save historical data, but at a greatly reduced rate:
 - Collectors save a report to store once per window period (e.g. once per 15 seconds).

Fixes #2297

bboreham:
Fixes #3783 

Collectors hold recent reports in memory; when querier needs 'live' data, it fetches it from collectors instead of from the long-term store.

This means queriers get more up-to-date data, by the store-frequency collector was using (e.g. 2 seconds), and we can slow down that frequency to spend less on S3/DynamoDB.

Included in this change: multitenant collector now always saves async - removed support for saving all reports immediately to store.

### DIFF
--- a/app/api_report.go
+++ b/app/api_report.go
@@ -19,7 +19,7 @@ func makeRawReportHandler(rep Reporter) CtxHandlerFunc {
 			return
 		}
 		censorCfg := report.GetCensorConfigFromRequest(r)
-		respondWith(ctx, w, http.StatusOK, report.CensorRawReport(rawReport, censorCfg))
+		respondWithReport(ctx, w, r, report.CensorRawReport(rawReport, censorCfg))
 	}
 }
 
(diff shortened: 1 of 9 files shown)


bboreham:
This is in production at Weave Cloud now.
One glitch: if you forget to set `--app.collector-addr` that should be a fatal error so the administrator can spot the problem earlier.

PR #3847, opened 2021-04-20
Release docker image for arm64

odidev:
The following files have been modified and added:

- In **.circleci/config.yml** added a new job **publish** in order to use buildx to build and push images for both the platforms to dockerhub.

- Added **docker/Dockerfile.cloud-agent_multiarch** file for building cloud-agent image which uses multi-stage build that builds the scope binary in stage 1 and in stage 2 follows the same steps as written in Dockerfile.cloud-agent.

- Added **docker/Dockerfile.scope_multiarch** file for building scope image which uses multi-stage that builds all the binaries in stage 1 and in stage 2 copies all the binaries same as Dockerfile.scope.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,10 @@ workflows:
           requires:
             - client-test
             - integration-tests
+      - publish:
+          filters:
+            branches:
+              only: master
 
 jobs:
   lint:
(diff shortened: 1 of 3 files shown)


odidev:
Can you please review this PR and let me know if any changes are required?

kingdonb:
I appreciate the work that has been done here, thanks very much for contributing this!

I don't quite have access to see what's wrong with the failing integration tests, but I suspect it has something to do with the recent CircleCI deprecations. Maybe there's more to it than that.

If there is someone who has the desire to perform the maintainer role, it will help a lot, as I'm able to facilitate but cannot take on the role myself. There are too many other responsibilities, primarily around Flux and getting to GA/CNCF graduation.

PR #3851, opened 2021-04-23
allow to read password from file

Kielek:
it enables option to utilize docker swarm secret storage for password instead of env variables or app switches

### DIFF
--- a/prog/main.go
+++ b/prog/main.go
@@ -4,6 +4,7 @@ import (
 	"compress/gzip"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
 	"regexp"
@@ -97,6 +98,7 @@ type probeFlags struct {
 	basicAuth              bool
 	username               string
 	password               string
+	passwordFilename       string
 	token                  string
 	httpListen             string
 	publishInterval        time.Duration
@@ -152,9 +154,10 @@ type appFlags struct {
 	logHTTP        bool
 	logHTTPHeaders bool
 
-	basicAuth bool
-	username  string
-	password  string
+	basicAuth        bool
+	username         string
+	password         string
+	passwordFilename string
 
 	weaveEnabled   bool
 	weaveAddr      string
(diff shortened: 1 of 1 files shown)


Kielek:
@bboreham, if you are fine with these changes, it will be great to have new _offical release_. 1.13.3?

site/faq.md
```diff
@@ -33,10 +33,14 @@ OSS Scope has no user concept, this is only available in Weave Cloud. To limit t
         Password for basic authentication (default "admin")
   --app.basicAuth.username string
         Username for basic authentication (default "admin")
+  --probe.basicAuth.password.filename string
+        Password filename for basic authentication. It overwrites app.basicAuth.password
```
bboreham, line 37: Is this meant to be `--app...` ?

site/faq.md
```diff
@@ -23,7 +23,7 @@ OSS Scope has no user concept, this is only available in Weave Cloud. To limit t
 
 - setup a reverse proxy with auth and block access to non admin users,
 - capture the calls with something like Chrome network console to get the endpoints to know which requests to authenticate in the proxy server.
-- you can use Basic HTTP Auth since Scope 1.10.0 - just use these command line
+- you can use Basic HTTP Auth since Scope 1.10.0 (support for loading password from file sice Scope 1.13.3) - just use these command line
```
bboreham, line 26: Typo s/sice/since/

Put this in a separate PR so we don't publish incorrect info for a time.

prog/main.go
```diff
@@ -504,3 +518,15 @@ func main() {
...
 }
+
+func getPassword(password string, passwordFilename string) string {
+	if passwordFilename == "" {
+		return password
+	}
+
+	b, err := ioutil.ReadFile(passwordFilename)
+	if err != nil {
+		log.Errorf("Cannor read password from: %v", passwordFilename)
```
bboreham, line 529: Typo s/Cannor/Cannot/

bboreham, line 529: Please include `err` in the message.

bboreham (changes requested):
Thanks for the PR.  The concept is fine; I made some small points inline.

Bigger point though: what happens if the file is not readable?  Looks to me like the system will then be open to all callers, which is probably not what the administrator would expect if they configured a password file.

prog/main.go
Kielek: fixed

Kielek: fixed

site/faq.md
Kielek: fixed, faq moved to #3852

site/faq.md
Kielek: fixed, faq moved to #3852

Kielek:
> Bigger point though: what happens if the file is not readable? Looks to me like the system will then be open to all callers, which is probably not what the administrator would expect if they configured a password file.

You are right. I have change log to Fatal, it should close the process and I think, it is expected solution.

Kielek:
@bboreham, any chance to move forward with this PR (and prepare next official release)?

Kielek:
@bboreham, is there any chance that you check this PR?

bboreham (approved):
Thanks! Lgtm; just checking if integration tests pass

PR #3865, opened 2021-05-30
Fix pod topology for jobs when cronJobs not exist.

### LINKED ISSUE #3864: Weave Scope does not record the pod topology when there is only job(not cronjob) in kubernetes
## What you expected to happen?

When there is only a job in my kubernetes cluster, Weave Scope does not show the topology of the pod created by this job.

I have debugged the code and found that Scope only calculates the topology when cronJob exists. And looks like when  there are more than one cronJob, job selector will register many times.

https://github.com/weaveworks/scope/blob/master/probe/kubernetes/reporter.go#L664

## Versions:

```
$ scope version
$ docker version
$ uname -a
$ kubectl version
```

## Logs:
```
$ docker logs weavescope
```
or, if using Kubernetes:
```
$ kubectl logs <weave-scope-pod> -n <namespace> 
```

shabicheng:
Fixes #3864.

### DIFF
--- a/probe/kubernetes/reporter.go
+++ b/probe/kubernetes/reporter.go
@@ -661,18 +661,18 @@ func (r *Reporter) podTopology(services []Service, deployments []Deployment, dae
 				report.MakeCronJobNodeID(cronJob.UID()),
 			))
 		}
-		for _, job := range jobs {
-			selector, err := job.Selector()
-			if err != nil {
-				return pods, err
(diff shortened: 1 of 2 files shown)


bboreham:
Thanks for the PR - great spot! 
Apologies for long delay in looking at it.  I've kicked off the tests now so it should run the integration tests.

Issue #3880, opened 2021-07-14
Weave Scope not support containerd in GKS


draconiun:
Hi, I currently have a kubernetes cluster deployed on GCP. The version is 1.19 and as an image type I use containerd.
I have deployed weave scope version 1.13 and it does not show me the containers in the panel. And it is important to be able to visualize the CPU / Memory of my pods. 

![image](https://user-images.githubusercontent.com/22685421/125706349-83bcb7d5-c081-437b-a82d-540aeca16fb2.png)

Pool in GKS:

![image](https://user-images.githubusercontent.com/22685421/125706396-dcc28c92-9211-498f-94ec-a460e442ba2a.png)

I have checked this Weave Scope repository for CRI I tried to implement it but it still doesn't work.
https://github.com/weaveworks/scope/blob/master/examples/cri/ds.yaml

I will be grateful for your responses. Thank you.

bboreham:
I had a go at this today.  (Sorry it is a long time since you asked!)

GKE, at least when I tested it today, is running containerd, and the CRI socket can be accessed at `/var/run/containerd/containerd.sock`. Therefore you need to change this argument on the agent:
```
            - '--probe.cri.endpoint=unix:///var/run/containerd/containerd.sock'
```

You need to mount this path from the host into the container, like this:

```
          volumeMounts:
            - name: cri-socket
              mountPath: /var/run/containerd/containerd.sock
      volumes:
        - name: cri-socket
          hostPath:
            path: /var/run/containerd/containerd.sock
            type: Socket
```

Once I did that, Scope knew about my containers, however it thought they were all stopped.
I can make a PR to fix that.

(Also that `ds.yaml` file is out of date in other ways; it should use `--probe.kubernetes.role=host` and run a separate probe for Kuebrnetes like in [the k8s directory](https://github.com/weaveworks/scope/blob/5c228dc898400b932ea96e0a35288d86dfb8c3ca/examples/k8s/probe-deploy.yaml))

PR #3899, opened 2022-02-28
probe: translate CRI container state

bboreham:
Use values consistent with what we use for Docker.
This avoids seeing all containers as 'stopped', as noted at #3880.

### DIFF
--- a/probe/cri/reporter.go
+++ b/probe/cri/reporter.go
@@ -56,11 +56,23 @@ func (r *Reporter) containerTopology() (report.Topology, error) {
 	return result, nil
 }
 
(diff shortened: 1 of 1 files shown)


CVilinsky:
Hi, can you elaborate on that?
I changed the **ds.yaml** as you mentioned, and now I can see the containers but they are all 'stopped'.

bboreham:
This is a proposed change to the Go code to fix the issue. Are you are able to download this branch and build your own container?

CVilinsky:
Yes, I have cloned your branch, but still I still can't see any containers.
I have also made the changes you've noted in #3880 , because my container-runtime is 'containerd'.

PR #3902, opened 2022-03-14
Update Go to 1.17.8; tidy build image

bboreham:
Need to specify `-buster` otherwise libpcap gets unresolved symbols.

Add back a missing `&&` after installing packages; it doesn't make a practical difference but looks more consistent to me.

Stop rebuilding various Go libraries since they are now cached automatically for each different set of tags, and in a different
directory so the ones inside the image aren't actually used.
Compare build on master vs this branch:
```
weaveworks/scope-backend-build      master-a774857b        af5063a2025a     1.54GB
weaveworks/scope-backend-build      go-1-17-8-87988e5a     0bec25faf9c5     1.15GB
```

I have uploaded a copy of the build image to DockerHub at `bboreham/scope-backend-build:go-1-17-8-87988e5a` since I don't have write access to `weaveworks/scope-backend-build`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 defaults: &defaults
   working_directory: /go/src/github.com/weaveworks/scope
   docker:
-    - image: bboreham/scope-backend-build:go-1-17-8-87988e5a
+    - image: weaveworks/scope-backend-build:go-1-17-8-bcdf2caa
 
 client-defaults: &client-defaults
   working_directory: /home/weave/scope
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 defaults: &defaults
   working_directory: /go/src/github.com/weaveworks/scope
   docker:
-    - image: weaveworks/scope-backend-build:build-updates-1b1c9664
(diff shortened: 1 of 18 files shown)


mnowster:
Hi @bboreham is there something that needs updating to the weaveworks build? I don't think uploading something to your personal one seems like a good idea on a weave open project. Is there something I can help with here? do you want to email me and talk about it?

bboreham:
Hi @mnowster, 

The steps are: build Weave Scope (clone the repo and run `make`); this will create a Docker image tagged as `weaveworks/scope-backend-build`; push that image to DockerHub and this PR can be changed to use that tag.

Me pushing my own image serves to confirm it will go through CI cleanly once tagged appropriately.

mnowster:
Hi @bboreham, apologies for the delay turns out I didn't have permissions either! Fixed now.

We pushed a new image `weaveworks/scope-backend-build:go-1-17-8-bcdf2caa`

Let me know if I can do anything else to help! Hope you're well :)

bboreham:
> Add back a missing `&&` after installing packages, so temp files aren't committed to the image.

For the record this didn't make the difference, it was all in the Go build, the next change in Dockerfile.

However I'm going to merge this as-is to avoid another round of pushing image updates.

